PR #20, opened 2016-08-14
Fix bad include path for OSD headers.

mont29:
Current (3.0.5) OSD build puts its headers into `<install_path>/include/opensubdiv`, not into `<install_path>/include/opensubdiv3`...

I guess this comes from dev env where you have several versions of OSD installed, but... not nice for common case. ;)

### DIFF
--- a/pxr/imaging/lib/hd/codeGen.cpp
+++ b/pxr/imaging/lib/hd/codeGen.cpp
@@ -46,7 +46,7 @@
 
 #include <sstream>
 
-#include <opensubdiv3/osd/glslPatchShaderSource.h>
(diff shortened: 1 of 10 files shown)


c64kernal:
Thanks mont29 -- we have this fixed on our end too and we'll be pushing it soon.  Incidentally, do we have a CLA from you?  Thanks again!

mont29:
Thanks for the answer @c64kernal .
And no, did not thought about the CLA, just sent it a few minutes ago now. :)

c64kernal:
Great, thanks so much!

c64kernal:
Our change was pushed through.  Closing.  Thanks again!

PR #37, opened 2016-09-01
Dev win ip

JamieKenyonFoundry:
Clean up of Arch based off Sunya's feedback.

Windows now uses forward slashes in all paths.


JamieKenyonFoundry:
Cleanup of of cmake - added PXR_HYBRID_BUILD_MODE build flag to enable symbols in release

ArchGetEnv no longer uses static results (via feedback from Sunya).

JamieKenyonFoundry:
end-to-end example can now be completed.

extras/usd/tutorials/endToEnd/tutorial_scripts/add_shadingVariants.py
```diff
@@ -93,6 +93,7 @@ def _AddShadingToBall(stage):
         with shadingVariant.GetVariantEditContext():
             whichBall = variantName.split('_')[-1]
             texPath = os.path.join(texDir, 'ball%s.tex' % whichBall)
+            texPath = texPath.replace('\\', '/')
             # in the current variant, modify the color
```
meshula: Might be an idea here to expose the path normalizer you wrote earlier via Python and call that here instead ~ then normalization will behave the same everywhere. Then it would cover drive letter upper-lower-casing etc...

pxr/imaging/lib/garch/glPlatformContextWindows.cpp
```diff
@@ -10,11 +10,19 @@
 // GarchGLWContextState
 //
 
+static HDC g_device;
```
meshula: This'll break in multi-context environments, for example when using GLEW_MX.

JamieKenyonFoundry: Indeed. Thanks Nick!

pxr/imaging/lib/garch/glPlatformContextWindows.cpp
```diff
@@ -10,19 +10,11 @@
 // GarchGLWContextState
 //
 
-static HDC g_device;
-static HGLRC g_context;
-
 GarchGLWContextState::GarchGLWContextState() :
+	device(wglGetCurrentDC()),
+	context(wglGetCurrentContext()),
```
meshula: Thanks :)

extras/usd/tutorials/endToEnd/tutorial_scripts/add_shadingVariants.py
JamieKenyonFoundry: I'm not sure I understand Nick. Are you saying the tutorial script should have a dependency on Tf?

meshula: I guess yeah. It's a tutorial so maybe it's not that important, but I was thinking that it's probably good that normalization is consistent everywhere, especially if the rules ever get more complex. For example, at ILM, we had scripts that performed much more stringent normalization including illegal character substitutions. If it all goes through one place, that's enabled universally.

ZeroCrunch: Nick, I totally agree with you but lets discuss this offline.

meshula: Sure thing! Anyway, I'm not suggesting that a merge be gated by my comment. It's only a comment ;)

PR #38, opened 2016-09-01
usdObj: Clean up build and plugInfo for proper installation into examples.

asluk:
…PXR_PLUGINPATH_NAME bootstrapping. Otherwise the user must manually edit the usdObj plugInfo.json , etc., in order for Plug to find what it needs.

### DIFF
--- a/extras/usd/examples/usdObj/CMakeLists.txt
+++ b/extras/usd/examples/usdObj/CMakeLists.txt
@@ -1,3 +1,4 @@
+set(PXR_PREFIX pxr/usd)
 set(PXR_PACKAGE usdObj)
 
 pxr_plugin(${PXR_PACKAGE}
--- a/extras/usd/examples/usdObj/plugInfo.json
+++ b/extras/usd/examples/usdObj/plugInfo.json
@@ -17,10 +17,10 @@
                     }
                 }
             },
-            "LibraryPath": "usdObj",
+            "LibraryPath": "@PLUG_INFO_LIBRARY_PATH@",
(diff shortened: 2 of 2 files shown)


extras/usd/examples/usdObj/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
+set(PXR_PREFIX pxr/usd)
 set(PXR_PACKAGE usdObj)
+unset(PXR_INSTALL_SUBDIR)
```
sunyab: I think we just need to remove this line from your change, since we don't want the usdObj plugin to build in the standard plugin location.

sunyab:
The usdObj plugin is meant to be an example plugin, but not necessarily as a fully-featured and supported .obj reader for USD. For that reason, I think we'd rather have usdObj still install to the examples area and force consumers to point to it explicitly to via PXR_PLUGINPATH_NAME.

That said, I think we still want most of what you have in your change, since the plugInfo.json file did need to get fixed up. If you fix up the code review comments, I'll merge the rest of this in.

Note that there seems to be some other rpath issue that causes the usdObj plugin not to work if left in the examples area. We're looking at that internally.

Thanks!

asluk:
Done; thanks! Should I open an issue for PXR_PLUGINPATH_NAME to work with installed examples?

asluk:
Cleaning up the change history; will reopen when my tree is back in shape.

asluk:
Tree repaired; reopening.

PR #39, opened 2016-09-01
Ignore .DS_Store and backup emacs files.

asluk:


sybrenstuvel:
Just curious: wouldn't this be environment-specific, and thus be more suitable to be placed in your own `.git/info/exclude` file or global `core.excludesfile`?

asluk:
Fair point, though I also don't see the harm in always excluding these. I'll leave it open for the Pixar folks to decide. Thanks for the tip!

sunyab:
Hey @asluk, we think this would be better off as a user-environment-specific thing so we're going to close this pull request out. Thanks!

Issue #6, opened 2016-07-27
Remove unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null) like the following.
- [GlfUVTextureStorageData](https://github.com/OpenVSP/OpenVSP/blob/73d3e0f2e4326323036a6c7a3f63300882d99f91/src/geom_api/GuiInterface.cpp#L30)
- [HdMeshTopology::SetQuadInfo](https://github.com/PixarAnimationStudios/USD/blob/2eb01f5cd4c2dae4e1ef9912ca27a93083bb6ef4/pxr/imaging/lib/hd/meshTopology.cpp#L198)

amooty:
Filed as internal issue #135159.

superfunc:
Note this code has been cleaned up since and can be closed: https://github.com/PixarAnimationStudios/USD/blob/022f7ef54ff8ce15120f34e61e078c1daf471153/pxr/imaging/lib/hdSt/meshTopology.cpp#L83

elfring:
There are [further update candidates](https://github.com/PixarAnimationStudios/USD/blob/65320d266057ae0a68626217f54a0298ac092799/pxr/imaging/lib/hd/glslProgram.cpp#L49 "Update candidate: Destructor for HdGLSLProgram") to [consider](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteProgram.xhtml "Delete a program object"), aren't there?

superfunc:
Looks like I was wrong; I hadn't realized this was pointing to a more general problem and not those specific callouts, my apologies.

elfring:
I am just curious if all possible improvements were integrated into this software for the reported search pattern.

elfring:
Would you like to [take another look also at any included software libraries](https://github.com/PixarAnimationStudios/USD/blob/022f7ef54ff8ce15120f34e61e078c1daf471153/third_party/houdini/lib/gusd/UT_StaticInit.h#L62 "Update candidate: Constructor for GusdUT_StaticValHolder")?

elfring:
[...]

PR #43, opened 2016-09-03
Remove unnecessary null pointer check from meshTopology.cpp

Geertiebear:
This is an issue which was brought up by issue #6 .


sunyab:
Hi @Geertiebear, thank you for your contribution and apologies for not getting back to you in a timely manner. The change you suggested was made recently during some code reorganization in commit 01769bad:

https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/imaging/lib/hdSt/meshTopology.cpp#L79

So, I'm closing out this pull request. Thanks again!

Issue #53, opened 2016-09-14
CMake GLEW handling on Windows is broken-ish


jcowles:
On Windows, when statically linking against Glew, GLEW_STATIC must be defined, here is the relevant part of glew.h:

```
/*
 * GLEW_STATIC is defined for static library.
 * GLEW_BUILD  is defined for building the DLL library.
 */

#ifdef GLEW_STATIC
#  define GLEWAPI extern
#else
#  ifdef GLEW_BUILD
#    define GLEWAPI extern __declspec(dllexport)
#  else
#    define GLEWAPI extern __declspec(dllimport)
#  endif
#endif
```

So on Windows when using glew32.lib (or glew32d.lib, which OpenSubdiv/USD doesn't support!) -> no switch needed.

When using glew32s.lib (glew32sd.lib) -> -DGLEW_STATIC is required.

Also, OpenSubdiv should guard against glew leaking into the public headers/templates, which I think is what I ran into -- will file a separate bug.

jcowles:
I have a fix

jtran56:
Filed as internal issue #137371.

sunyab:
We think this issue should be fixed in OpenSubdiv 3.1.1 by this commit:
PixarAnimationStudios/OpenSubdiv#905

I'm going to close this (and the associated pull request) out for now, please reopen if this is still an issue. Thanks!

PR #54, opened 2016-09-14
Fix static GLEW linkage

jcowles:
In glew.h, it is required to define GLEW_STATIC when using the static library,
however USD never sets this flag, thus previously all glew linkage was assumed
to be dynamic with "dllimport" linkage.

Now the CMake FindGlew module sets this flag when running in Win32 and the
static library is discovered (this is determined by convention of the "s" or "sd"
suffix).

Also, GLEW debug libraries are now supported (e.g. glew32[s]d.lib).

Addresses Issue #53

### DIFF
--- a/cmake/modules/FindGLEW.cmake
+++ b/cmake/modules/FindGLEW.cmake
@@ -51,7 +51,7 @@ if (WIN32)
 
     find_library(GLEW_LIBRARY
         NAMES
-            glew GLEW glew32s glew32
+            glew GLEW glew32s glew32 glew32sd glew32d
         HINTS
             "${GLEW_LOCATION}/lib"
             "$ENV{GLEW_LOCATION}/lib"
@@ -62,6 +62,12 @@ if (WIN32)
         PATH_SUFFIXES
             Release/${ARCH}
         DOC "The GLEW library")
+
+    if ("${GLEW_LIBRARY}" MATCHES "glew32s(d|)")
+        add_definitions("/DGLEW_STATIC")
+        MESSAGE(STATUS "Glew API: GLEW_STATIC " ${GLEW_LIBRARY})
+    endif()
+
 endif ()
 
 if (${CMAKE_HOST_UNIX})


jcowles:
Don't merge this yet, it needs another tweak to be correct.

cmake/modules/FindGLEW.cmake
```diff
@@ -62,6 +62,12 @@ if (WIN32)
         PATH_SUFFIXES
             Release/${ARCH}
         DOC "The GLEW library")
+
+    if ("${GLEW_LIBRARY}" MATCHES "glew32s(d|)")
+        add_definitions("/DGLEW_STATIC")
```
jcowles: I believe this should be GLEW_BUILD (not GLEW_STATIC) -- I will confirm and update the PR.

jcowles: Yeah, that's what I thought from the docs too, however Windows has a quirk where global variables declared "extern" will not be shared over library boundaries unless they are declared with dllexport declspec. Looking at the GLEW_STATIC/GLEW_BUILD preprocessor logic, the way to get dllexport is by declaring GLEW_BUILD.

So it seems this is the only way to make it work, unless you are ok with unique glew state per-library, but the USD imaging init pattern requires shared state, since Glf calls glewInit() (via GlfGlewInit); so without shared state, Glf will be the only library that gets initialized correctly.

jcowles: Oh, the unstated bit in my comment is that glew stores GL functions as global function pointer variables of the form __glewApiName -- e.g. glDrawBuffers is internally stored as a global function pointer called __glewDrawBuffers, thus hitting the scenario I described above.

meshula:
````
/*
 * GLEW_STATIC is defined for static library.
 * GLEW_BUILD  is defined for building the DLL library.
 */

#ifdef GLEW_STATIC
#  define GLEWAPI extern
#else
#  ifdef GLEW_BUILD
#    define GLEWAPI extern __declspec(dllexport)
#  else
#    define GLEWAPI extern __declspec(dllimport)
#  endif
#endif
````
If you don't have GLEW_STATIC, and you do have GLEW_BUILD, USD will attempt to re-export the symbols, so if you use GLEW yourself, and link glf, then the linker will report duplicate symbols? I agree that these macros in glew.h are bogus, extern is seldom what you want in this case. 

In Zeno, I worked around this issue by requiring Zeno libs to link the PlatformGL library, and never glew directly. Maybe this hiccough needs to be documented...

jcowles: Yeah, seems wrong -- and thinking about this more this morning, I believe the real problem is static linkage of Glew into a dynamic shared library. We should really just be linking Glew dynamically, then everything would work as expected... aside from the linker errors I get from dynamic linking when combining USD and OpenSubdiv (but that's just something that needs to be fixed).

meshula: Yeah, that's why build-club is using dynamic now. Yesterday, I did a build of Pixar's dev_win_ip with @JamieKenyonFoundry's latest dev_win_ip merged on top, and usd-build-club built GLEW and OpenSubdiv and it all worked :) What link problem are you seeing in OSD?

jcowles:
I'm getting opensubdiv related errors (missing__glesXxxxx symbols) when
linking Hd.dll

Was the build working without Jamie's changes?

meshula:
Hmmm, I had a local change. When I back it out, I get the same failure.

https://github.com/meshula/USD/commit/c698023ff9c2fdd40b5a8bac5bc61fc5a55fa090

This is complicated :\

jcowles:
What was your local change?

jcowles:
Yeah, I wondered about that too, but given the fact that Windows DLLs
won't share global variables across library boundaries without declspec of
dllexport, It seems GLEW_BUILD is required -- unless you are ok with
private glew state per library, but this breaks USD imaging assumptions.

ZeroCrunch:
Have I broken something?

ZeroCrunch:
GLEW_STATIC should not be needed to make dev_win_ip build. USD compiles and links here at The Foundry. 

Any ideas Nick?!

JamieKenyonFoundry:
I'm seeing this now too!

jcowles:
@JamieKenyonFoundry The OpenSubdiv fix is here: https://github.com/PixarAnimationStudios/OpenSubdiv/pull/905

ZeroCrunch:
Thats awesome! Thanks @jcowles!

sunyab:
Closing this pull request along with #53, please reopen if this is still an issue. Thanks!

Issue #52, opened 2016-09-13
Build fails with "'isnan' was not declared in this scope" on Ubuntu


popey:
See below for extract. I'm building on Ubuntu 16.04, 64-bit, and have got the following when building either the release, or the tip of trunk.

[ 25%] Building CXX object pxr/base/lib/gf/CMakeFiles/gf.dir/bbox3d.cpp.o
[ 25%] Building CXX object pxr/base/lib/gf/CMakeFiles/gf.dir/camera.cpp.o
[ 25%] Building CXX object pxr/base/lib/gf/CMakeFiles/gf.dir/colorRamp.cpp.o
In file included from /root/parts/usd/build/include/pxr/base/gf/colorRamp.h:28:0,
                 from /root/parts/usd/src/pxr/base/lib/gf/colorRamp.cpp:24:
/root/parts/usd/build/include/pxr/base/gf/rgb.h: In member function 'bool GfRGB::IsValid() const':
/root/parts/usd/build/include/pxr/base/gf/rgb.h:80:49: error: 'isnan' was not declared in this scope
     bool IsValid() const { return !isnan(_rgb[0]); }
                                                 ^
/root/parts/usd/build/include/pxr/base/gf/rgb.h:80:49: note: suggested alternative:
In file included from /root/parts/usd/build/include/pxr/base/arch/math.h:36:0,
                 from /root/parts/usd/build/include/pxr/base/gf/math.h:27,
                 from /root/parts/usd/build/include/pxr/base/gf/colorRamp.h:27,
                 from /root/parts/usd/src/pxr/base/lib/gf/colorRamp.cpp:24:
/usr/include/c++/5/cmath:641:5: note:   'std::isnan'
     isnan(_Tp __x)
     ^
pxr/base/lib/gf/CMakeFiles/gf.dir/build.make:386: recipe for target 'pxr/base/lib/gf/CMakeFiles/gf.dir/colorRamp.cpp.o' failed
make[2]: **\* [pxr/base/lib/gf/CMakeFiles/gf.dir/colorRamp.cpp.o] Error 1
CMakeFiles/Makefile2:1345: recipe for target 'pxr/base/lib/gf/CMakeFiles/gf.dir/all' failed
make[1]: **\* [pxr/base/lib/gf/CMakeFiles/gf.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: **\* [all] Error 2

jtran56:
Filed as internal issue #137247.

[...]

PR #55, opened 2016-09-15
Use std::isnan instead of isnan in rgb and rgba.h

doctormo:
Fix issue #52 for popey. Use std::isnan instead of just isnan, std was already used in `pxr/usd/lib/usd/timeCode.h` so it should be fine.

### DIFF
--- a/pxr/base/lib/gf/rgb.h
+++ b/pxr/base/lib/gf/rgb.h
@@ -77,7 +77,7 @@ class GfRGB {
     
     //! Returns whether or not the color is valid. By convention, a color
     //  is valid if the first color component is not NAN.
-    bool IsValid() const { return !isnan(_rgb[0]); }
+    bool IsValid() const { return !std::isnan(_rgb[0]); }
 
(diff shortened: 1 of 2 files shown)


sunyab:
Hi @doctormo, just to update you on this, our current plan is to simply remove gf/rgb.h and rgba.h from the distribution since they're not being used. I'll leave this open until that actually goes through.

Also, just wanted to ask you to fill out the CLA per the instructions below should you wind up contributing more changes in the future: http://graphics.pixar.com/usd/docs/Contributing-to-USD.html

Thanks!

doctormo:
Good to hear. Thanks for the update. No need to merge or anything.

We do a Public Domain declaration, since CLAs cause more paperwork and messing about than a normal multi-stakeholder open source project. So probably no large contributions, but the odd patch is easy enough to PD and hopefully is enough for your lawyers.

sunyab:
Commit 57d4d15 was pushed last week, closing this out per my comment above. Thanks again!

PR #56, opened 2016-09-15
Alembic support

JamieKenyonFoundry:
We can now build and run the usdAbc plugin.


cmake/modules/FindMaya.cmake
```diff
@@ -223,11 +223,21 @@ endif()
...
 
-find_package_handle_standard_args(Maya
-    REQUIRED_VARS
-        MAYA_EXECUTABLE
-        MAYA_INCLUDE_DIRS
-        MAYA_LIBRARIES
-    VERSION_VAR
-        MAYA_API_VERSION
-)
+if (${PXR_MAYA_API_ONLY} AND NOT ${PXR_MAYA_API_ONLY})
```
meshula: Is this test for ${A} AND NOT ${A} a cmake-ism or a typo?

ZeroCrunch: As I understand its the difference between

#define A

and

#define A 1

So it tests the variable exists and its not been set to true.

meshula: Ah, gotcha :) So much cmake to learn.

Issue #70, opened 2016-09-30
Cmake Alembic detection script needs to recognize newer Alembic builds.


meshula:
On Windows, Alembic no longer builds to many small libraries, instead only a single DLL and lib are generated.

That means that the detection should look something like this:

```
    find_library(ALEMBIC_LIBRARY
                 NAMES Alembic
                 PATHS ${LIBRARY_PATHS})

    set(ALEMBIC_LIBRARIES ${ALEMBIC_LIBRARY})

    get_filename_component(ALEMBIC_LIBRARY_DIR ${ALEMBIC_LIBRARY} PATH)

```

A possible solution would be to add a detection phase in the Cmake script, along the lines of

```
if (NOT ALEMBIC_ABC_LIBRARY_FOUND)
   message("Historic Alembic not detected, looking for modern Alembic")
   .... then do something like what I listed above.
endif()

```

jtran56:
Filed as internal issue #138129.

JamieKenyonFoundry:
Nick,

I believe this only applies to the latest version. But agreed, it would be nice if we are to support versions > 1.5.2.

asluk:
Yep, I think this affects versions >= 1.6.0 , per https://github.com/alembic/alembic/commit/834eca66a05a06d45699600ce4bd724fe1c581b9

I can take a look at this.

sunyab:
Commit has been merged into release v0.7.2, closing.

PR #75, opened 2016-10-01
Support detection of single-library alembic

asluk:
as shipped in alembic >= 1.6.0 . Addresses issue #70 .


asluk:
I squashed these changes and pushed, but the remote doesn't seem to be updating.

meshula:
Nice!

Issue #84, opened 2016-10-18
cmake DESTDIR support is broken


davvid:
### System Information (OS, Hardware, etc.)

Linux RHEL7
### Package Versions

N/A
### Build Flags

Set the install prefix to a non-writable directory
### Steps to Reproduce

```
$ mkdir -p build
$ cd build
$ cmake -D CMAKE_INSTALL_PREFIX=/usr/local/foo ..
$ make DESTDIR=/tmp/foo install
[....]
Install the project...
-- Install configuration: "Release"
CMake Error at pxr/cmake_install.cmake:36 (file):
  file Internal CMake error when trying to open file:
  /usr/local/foo/lib/python/pxr/__init__.py for writing.
Call Stack (most recent call first):
  cmake_install.cmake:37 (INCLUDE)
```

I'll submit a pull request shortly with a proposed solution.

jtran56:
Filed as internal issue #138888.

PR #85, opened 2016-10-18
build: fix cmake DESTDIR support

davvid:
Fix issue #84 by generating files and installing them in separate steps.


davvid:
Thanks for the careful review @sunyab .  I've updated `davvid/cmake-destdir` with your review notes, it should be correct now.  `DESTINATION` is now correctly specified as a directory and the double-escaping was removed.  I also rebased to the latest `dev` branch.

PR #112, opened 2016-11-30
Windows: Fixed path separators for usdview stylesheet

krzykli:
Replaced Windows style slashes with forward slashes so that the path can be digested correctly in the Qt stylesheet.

Also removed redundant slash at the end of `resourceDir`, since this separator is added in usdviewstyle.qss


krzykli:
Apologies, something clearly went wrong here, I'll redo the pull request.

PR #115, opened 2016-12-06
Logic clean-up for finding Maya package on Windows

krzykli:
Currently if we try to configure build with USD Maya plug-in on Windows we need to have `PXR_MAYA_API_ONLY` set, otherwise cmake will fail at line 226. Not only that, if we do have it set the statement will anyway evaluate to false, which is not very useful.

The proposed fix amends the logic so that it's optional for the user to have this variable set.

I imagine `PXR_MAYA_API_ONLY` is still present in this branch for convenience purposes only and will go away at some point.

### DIFF
--- a/cmake/modules/FindMaya.cmake
+++ b/cmake/modules/FindMaya.cmake
@@ -223,10 +223,10 @@ endif()
 # all listed variables are TRUE
 include(FindPackageHandleStandardArgs)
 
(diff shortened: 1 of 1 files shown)


sunyab:
We're in the process of getting initial (and experimental) Windows support in the dev branch and retiring dev_win_ip. As you suspected, the PXR_MAYA_API_ONLY flag was only an interim thing and is no longer needed, so I'm going to close out this pull request. Thanks!

PR #121, opened 2016-12-21
Ptex prefer static

elrond79:
### Description of Change(s)

Adds a cmake option, PTEX_PREFER_STATIC, which for linux and osx will make FindPTex.cmake use static versions of the ptex libraries preferentially to dynamic ones

### Included Commit(s)
- aa36361a1e22af427c06ac2f2cd2f01ff036845c
- ~~8885d0c5176a5a061825f910a00bbc71005c6921~~
- b0f253f8b53e3bbd4f054d3c41b0900bce568f2d
### Fixes Issue(s)
- none

### DIFF
--- a/cmake/modules/FindPTex.cmake
+++ b/cmake/modules/FindPTex.cmake
@@ -14,7 +14,7 @@
 # You may obtain a copy of the License at
 # http://www.apache.org/licenses/LICENSE-2.0
 
-option(PTEX_PREFER_STATIC "Prefer static library when linking pTex" OFF)
+option(PTEX_PREFER_STATIC "Prefer static library when linking pTex on linux or osx" OFF)
 
 if (WIN32)
     find_path( PTEX_INCLUDE_DIR
@@ -52,9 +52,14 @@ elseif (APPLE)
             "$ENV{PTEX_LOCATION}/include"
         PATHS
             DOC "The directory where Ptexture.h resides")
+    if (${PTEX_PREFER_STATIC})
+        set(PTEX_LIBRARY_NAMES libPtex.a Ptex)
+    else ()
+        set(PTEX_LIBRARY_NAMES Ptex libPtex.a)
+    endif ()
     find_library( PTEX_LIBRARY
         NAMES
-            Ptex libPtex.a
+            ${PTEX_LIBRARY_NAMES}
         PATHS
             "${PTEX_LOCATION}/lib"
             "$ENV{PTEX_LOCATION}/lib"
--- a/cmake/modules/FindPTex.cmake
+++ b/cmake/modules/FindPTex.cmake
@@ -14,6 +14,7 @@
 # You may obtain a copy of the License at
 # http://www.apache.org/licenses/LICENSE-2.0
 
+option(PTEX_PREFER_STATIC "Prefer static library when linking pTex" OFF)
 
 if (WIN32)
     find_path( PTEX_INCLUDE_DIR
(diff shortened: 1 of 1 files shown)


cmake/modules/FindPTex.cmake
```diff
@@ -27,6 +27,8 @@ if (WIN32)
             "$ENV{PROGRAMFILES}/Ptex/include"
             /usr/include
             DOC "The directory where Ptexture.h resides")
+    # TODO: figure out a way to add windows support for PTEX_PREFER_STATIC
+    # (since both static and dynamic libs can have a ".lib" suffix)
```
meshula: Maybe ask the Ptex devs to name the libs according to some convention, like boost's?

elrond79: Yeah, I suspect something upstream (on the Ptex side) may be the only way to handle this - since I'm not even sure the libraries can / do exist side-by-side, like they do for OSX/Linux.

Or maybe it's just less of an issue on Windows - perhaps there the convention is that you simply have two completely separate builds for static + dynamic, and you would simply set PTEX_LOCATION accordingly?

meshula: You can do it two ways, quarantine by directory, or by filename. filename has the advantage that the versions can co-exist in one directory and there will never ever be ambiguity.

It's a serious issue because in general with gcc/clang, if you mix and match release and debug dylibs, in general nothing terrible will happen, modulo missing inlined symbols. On Windows, debug libs and release libs have separate heaps so you can end up in the amazing world of allocating in one heap and releasing to another, yielding absolute havoc.

sunyab:
Hi @elrond79, would you mind changing the base branch for this pull request to the dev branch?
See https://github.com/blog/2224-change-the-base-branch-of-a-pull-request.

Thanks!

elrond79:
Done - didn't know github had that feature, thanks for pointing it out!

elrond79:
Removed the TODO about windows support (since it is outside the scope of this project to allow this with a simple CMake option), and changed the description of the option to clarify that it only works on linux and osx.

chadrik:
Below are some snippets from the official FindBoost.cmake for setting a static library preference.  I used a similar approach when adding this feature to OIIO.  Perhaps we should adopt it here as well?  

```cmake
# Support preference of static libs by adjusting CMAKE_FIND_LIBRARY_SUFFIXES
if( Boost_USE_STATIC_LIBS )
  set( _boost_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
  if(WIN32)
    set(CMAKE_FIND_LIBRARY_SUFFIXES .lib .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
  else()
    set(CMAKE_FIND_LIBRARY_SUFFIXES .a )
  endif()
endif()

...

# Restore the original find library ordering
if( Boost_USE_STATIC_LIBS )
  set(CMAKE_FIND_LIBRARY_SUFFIXES ${_boost_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES})
endif()
```

Also, there's this bit which is used in calls to `find_library` to narrow the search for static libraries on windows:

```cmake
# Setting some more suffixes for the library
set(Boost_LIB_PREFIX "")
if ( WIN32 AND Boost_USE_STATIC_LIBS AND NOT CYGWIN)
  set(Boost_LIB_PREFIX "lib")
endif()
```

Presumably building static boost libs on windows adds the 'lib' prefix.  Can we rely on, or enforce, that behavior for Ptex?

There's also this note about the previous behavior:

```cmake
# The previous behavior of FindBoost when Boost_USE_STATIC_LIBS was enabled
# on WIN32 was to:
#  1. Search for static libs compiled against a SHARED C++ standard runtime library (use if found)
#  2. Search for static libs compiled against a STATIC C++ standard runtime library (use if found)
# We maintain this behavior since changing it could break people's builds.
# To disable the ambiguous behavior, the user need only
# set Boost_USE_STATIC_RUNTIME either ON or OFF.
```

Good for learning from their previous mistakes.

jtran56:
Filed as internal issue #147122.

elrond79:
Going to close the PR, since there hasn't been much action, and we don't use it ourselves anymore either (since we just set PXR_ENABLE_PTEX_SUPPORT=OFF nowadays).

PR #135, opened 2017-01-05
Iterating maya underworld when exporting

sirpalee:
### Description of Change(s)
The change adds the ability to iterate objects in the Maya underworld during USD export from Maya.

### Fixes Issue(s)
Objects are not exported from the underworld. This happens if somebody writes an exporter plugin for imagePlane.

### DIFF
--- a/third_party/maya/lib/usdMaya/util.cpp
+++ b/third_party/maya/lib/usdMaya/util.cpp
@@ -1105,17 +1105,28 @@ _IsShape(const MDagPath& dagPath) {
     return (numberOfShapesDirectlyBelow == 1);
 }
 
+std::string
+PxrUsdMayaUtil::MDagPathToUsdPathString(const MDagPath& dagPath)
+{
+    std::string usdPathStr(dagPath.fullPathName().asChar());
+    // We are keeping the iterators around, to avoid extra memory allocations.
+    // Replacing characters because of underworld (transform|shapeNode->|underWorldNode).
+    // Maya inserts "->|", we can safely eliminate both the "-" and ">", and keep the rest.
+    auto itBegin = usdPathStr.begin();
+    auto itEnd = std::remove(itBegin, usdPathStr.end(), '-');
+    itEnd = std::remove(itBegin, itEnd, '>');
+    // Replacing MDagPath separators with the USD ones.
+    std::replace(itBegin, itEnd, '|', '/');
+    // We may want to have another option that allows us to drop namespace's
+    // instead of making them part of the path.
+    std::replace(itBegin, itEnd, ':', '_');
+    return std::string(itBegin, itEnd);
+}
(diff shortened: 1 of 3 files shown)


elrond79:
Squashed the three commits into one, for cleaner history

third_party/maya/lib/usdMaya/util.cpp
```diff
@@ -1095,17 +1095,25 @@ _IsShape(const MDagPath& dagPath) {
...
+PxrUsdMayaUtil::MDagPathToUsdPathString(const MDagPath& dagPath)
 {
     std::string usdPathStr(dagPath.fullPathName().asChar());
     std::replace( usdPathStr.begin(), usdPathStr.end(), '|', '/');
-
     // We may want to have another option that allows us to drop namespace's
     // instead of making them part of the path.
     std::replace( usdPathStr.begin(), usdPathStr.end(), ':', '_'); // replace namespace ":" with "_"
+    // Replacing characters because of underworld (transform|shapeNode->|underWorldNode)
+    // Because maya inserts "->|", we can safely eliminate both the "-" and ">"
```
sirpalee: No, we want to get rid of -> and keep the rest.

I cleaned up the implementation a bit, first doing the removal then the replaces, kept the iterators and added a bit of extra comment to clarify it further. What do you think, is it easier to follow?
`
std::string usdPathStr(dagPath.fullPathName().asChar());
// We are keeping the iterators around, to avoid extra memory allocations.
// Replacing characters because of underworld (transform|shapeNode->|underWorldNode).
// Maya inserts "->|", we can safely eliminate both the "-" and ">", and keep the rest.
auto itBegin = usdPathStr.begin();
auto itEnd = std::remove(itBegin, usdPathStr.end(), '-');
itEnd = std::remove(itBegin, itEnd, '>');
// Replacing MDagPath separators with the USD ones.
std::replace(itBegin, itEnd, '|', '/');
// We may want to have another option that allows us to drop namespace's
// instead of making them part of the path.
std::replace(itBegin, itEnd, ':', '_');
return std::string(itBegin, itEnd);
`

sirpalee:
@sunyab I wasn't able to reproduce the nurbs issue locally (we are not much of a nurbs users), so I'm looking forward to the example. The rest of the issues were fixed and I pushed back a change. (I'll squash them later on, but keeping the changes separate for now)

elrond79:
I rebased / collapsed our changes, and pushed a fresh / clean copy. Look here for the full history:

https://github.com/LumaPictures/USD/tree/tg/simple_features/iterating_maya_underworld
https://github.com/LumaPictures/USD/commit/5794dc39e1d3157cbf786f6e18a85471ffcf7bcf

jtran56:
Filed as internal issue #147123.

PR #139, opened 2017-01-05
Fallback to lstat if platform does not support fstatat.

marsupial:


jtran56:
Filed as internal issue #147124.

sunyab:
I'm going to close this pull request out. fstatat was introduced in MacOS 10.10, and USD uses C++11 features (primarily thread_local) that became available in Xcode 8, which requires MacOS 10.11. Thanks!

PR #140, opened 2017-01-05
Fixes for building with clang.

marsupial:
### Description of Change(s)
Use constructor/destructor attributes to add to ELF's .init .fini sections.
Enable pause instruction on Intel CPU built with clang.
Only add -Wno-unused-local-typedefs if clang version is > 3.5.

### Fixes Issue(s)
ARCH_CONSTRUCTOR/ARCH_DESTRUCTOR not being called when built with clang on Linux.
ArchThreadPause doing nothing built with clang.
Tons of warning related to -Wno-unused-local-typedefs not being understood.

### DIFF
--- a/cmake/defaults/gccclangshareddefaults.cmake
+++ b/cmake/defaults/gccclangshareddefaults.cmake
@@ -43,7 +43,13 @@ _disable_warning("deprecated")
 _disable_warning("deprecated-declarations")
 
 # Suppress unused typedef warnings emanating from boost.
-_disable_warning("unused-local-typedefs")
+if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR
+    NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.6)
+    if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR
+        NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6.1)
+            _disable_warning("unused-local-typedefs")
+    endif()
+endif()
 
 if (${PXR_MAYA_TBB_BUG_WORKAROUND})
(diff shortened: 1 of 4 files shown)


sunyab:
Hi @marsupial, we ask that all contributors sign a CLA before we can accept their changes. Once we receive a signed CLA we can have a look at merging your pull requests in. You can find the necessary info at the top of this page:

http://openusd.org/docs/Contributing-to-USD.html

Also, our convention is for the 'master' branch to contain only released changes and to have all new changes to go into the 'dev' branch. Could you change the base branch of your pull requests to the 'dev' branch instead?

https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

Thank you!

cmake/defaults/gccclangshareddefaults.cmake
```diff
@@ -42,8 +42,14 @@ endif()
 _disable_warning("deprecated")
 _disable_warning("deprecated-declarations")
 
-# Suppress unused typedef warnings emanating from boost.
-_disable_warning("unused-local-typedefs")
+# Suppress unused typedef warnings eminating from boost.
```
sunyab: Ah, and there's a typo that was introduced here -- it should be "emanating" not "eminating".

marsupial: I fixed the typo.
Perhaps I'm not understanding it right, but the other request seems a more confusing logic.

Instead of `if (!clang || clang > 3.5)`, you'd like  `if (!clang || !(clang < 3.6))`?

sunyab: So what I'm really trying to get at is I think the check ought to be if (!clang || clang >= 3.6), since no-unused-local-typedefs was introduced in 3.6. The check for (clang > 3.5) allows 3.5.x to pass, which I don't think is correct.

Unfortunately, cmake doesn't provide a VERSION_GREATER_OR_EQUAL, so "NOT ... VERSION_LESS 3.6" was the most succinct alternative.

marsupial: Changed.

PR #148, opened 2017-01-13
Arch_DebuggerIsAttachedPosix failure.

marsupial:
### Description of Change(s)
Arch_DebuggerIsAttachedPosix can return true in cases where the process is in fact not being debugged (on OS X always and when YAMA trace_scope is enabled). This will cause the process to hang after a crash as it assumes the process is being traced. Fixed on OS X by testing if debugger is attached according to Apple docs, and on Linux by making sure the forked child has permission to attach.

Also fixes a typo where variable is compared rather than assigned to.

### Fixes Issue(s)
Hanging after crash in OS X and Linux kernels with YAMA
defaultGalId not begin set properly when empty.


marsupial:
These are major defects. Any feedback?

c64kernal:
Thanks for the PR.  We still need time to go through the ARCH_CONSTRUCTOR change (we have a similar one internally that we're testing).

Thanks for the finding the Hd typo.  That is a dead code path right now, but it's a great catch nonetheless.  If you wouldn't mind breaking it out into its own pull request (it's not related to this PR), that would be great.

Thanks again!

marsupial:
After rebasing to dev I seem to have pushed the wrong commits into this PR.
It was (_and now is again_) only related to `Arch_DebuggerIsAttachedPosix` failing  on OS X and Ubuntu.

Sorry bout the confusion.

jtran56:
Filed as internal issue #147125.

sunyab:
We made a similar change to the OSX portions of this pull request in commit 3f9467c. 

I've been looking at merging the YAMA-related parts of this change, but I think there's a potential race condition here: if the child process tries to attach before the parent process has a chance to finish calling prctl, I think the attach will fail. 

It seems like we'd want the child process to wait until the parent process says it's ready before trying to attach. Perhaps we can use pipes to do this, similar to Arch_DebuggerRunUnrelatedProcessPosix.

nvmkuruc:
@pmolodo Do you think #3014 might address this same issue?

PR #150, opened 2017-01-18
add PXR_SYMLINK_HEADER_FILES cmake option

elrond79:
### Description of Change(s)

On build, CMake automatically copies the header files into a slightly different location in the cmake build directory (ie, from pxr/base/lib/tf to CMAKE_DIR/pxr/base/tf).  Unfortunately, this plays havoc with certain IDEs (ie, CLion), which find the "copied" headers instead of the "source" headers.

This change adds a  PXR_SYMLINK_HEADER_FILES  option which makes CMake symlink these headers to the cmake build dir, instead of copying, so that you can freely edit the headers in either location without having to worry which header you have.

On install, the headers are still properly copied, not symlinked.

### Included Commit(s)
- ~~c9301a7e3d78858bf81849f22bdf898ae4dca195~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/150/commits/7b45b91ffa5807231cca4210a92ccfc8e5f98c14~~
- ~~acf470c71a525eff8cca84c02d753c66e7ff5169~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/150/commits/6dd67544918dd8a8d54c0ab36cae000796ece9ff~~
- https://github.com/PixarAnimationStudios/USD/pull/150/commits/a23b785ef91a245c55e21da04d7d504a244e209d

### Fixes Issue(s)
- none

### DIFF
--- a/cmake/defaults/Options.cmake
+++ b/cmake/defaults/Options.cmake
@@ -36,6 +36,7 @@ option(PXR_ENABLE_HDF5_SUPPORT "Enable HDF5 backend in the Alembic plugin for US
 option(PXR_ENABLE_PTEX_SUPPORT "Enable Ptex support" ON)
 option(PXR_MAYA_TBB_BUG_WORKAROUND "Turn on linker flag (-Wl,-Bsymbolic) to work around a Maya TBB bug" OFF)
 option(PXR_ENABLE_NAMESPACES "Enable C++ namespaces." ON)
+option(PXR_SYMLINK_HEADER_FILES "Symlink the header files from, ie, pxr/base/lib/tf to CMAKE_DIR/pxr/base/tf, instead of copying; ensures that you may edit the header file in either location, and improves experience in IDEs which find normally the \"copied\" header, ie, CLion; has no effect on windows" OFF)
 
 # Precompiled headers are a win on Windows, not on gcc.
 set(pxr_enable_pch "OFF")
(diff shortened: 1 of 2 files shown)


sunyab:
Hi @elrond79, I just committed cff6390dc which I think resolves the IDE issues with the copied headers without needing this build option. I tried this on Xcode and MSVC and confirmed that compile errors take me to headers in the source tree instead of the build directory. I haven't tried CLion, but I'm hoping it'll work the same there. Let me know if this works for you if/when you get a chance to try it out. Thanks!

elrond79:
Cool, we'll check to see if this does the trick!

jtran56:
Filed as internal issue #147126.

elrond79:
So, finally got around to giving this a shot - alas, it looks like our IDE (clion) does not support special handling for #line directives.

Also - one nice thing about the symlink approach is it really works for pretty much everything - ie, if you use "find" or "grep -R" from the command line, then open it up in a text editor, you don't need to worry about which version of the file you have.

However - if you folks find really are against this, then we can drop the PR (though I think some others might find it handy!)

Also - out of curiosity, what IDE are you folks using for developing USD?

elrond79:
Ok, seems like this PR likely isn't going to happen (though I still think it's a good idea!), so I'm going to close the request.

PR #152, opened 2017-01-19
Refactor maya plugin to use UsdStage::IsSupportedFile

elrond79:
### Description of Change(s)

There's some code in the maya plugin that re-implements the functionality of UsdStage::IsSupportedFile - this PR standardizes on that method.  It also eliminates some repeated string constants.

### Included Commit(s)
- 63b7f8c0b09992f05f38e9ef5c14d2036e68e36e

### Fixes Issue(s)
-


jtran56:
Filed as internal issue #147127.

elrond79:
Closing this PR, as it has some negative side-effects - ie, we don't want to use this plugin to read or write .abc files by default!

PR #159, opened 2017-02-03
usdMaya: warn if falling back to pseudoRoot due to failed primPath

elrond79:
### Description of Change(s)
Currently, when the maya plugin is passed an invalid primPath (either one that has bad syntax, like "bad;primPath", or one that doesn't exist in the scene, like "/prim/doesnt/exist"), the plugin silently just converts this to a pseudoRoot ("/").  I feel that it should, at minimum, print a warning - so I've made a commit / pull request that does that.

It think it would be still more preferable to have it "abort" and return false, similar to the behavior if the stage has an invalid defaultPrim (the handling of which is a few lines below).  However, I was hesitant to change existing behavior without sanction.  If returning false would be preferable, though, it's an easy addition.

### Included Commit(s)
- https://github.com/PixarAnimationStudios/USD/commit/efaec0a98654e8d2eb001b0668f95fbff86aaa60

### Fixes Issue(s)
- None reported, but the silent conversion of an invalid input could be considered an issue...


jtran56:
Filed as internal issue #147128.

PR #161, opened 2017-02-08
GetSessionLayerForVariantSelections takes full path, not model name

elrond79:
This makes the variant-selections cache key more accurate, as the variants are tied to a particular prim-path, not a root name.

This necessitated some other changes - for instance, pxrUsdReferenceAssembly.inStageDataCached is now affected by .primPath.  Also, it allowed us to eliminate a double-setting of variants that was (potentially) occuring in usdReadJob::doIt

### Description of Change(s)

The current args to GetSessionLayerForVariantSelections seemed to make it re-used cached session layers inappropriately in some cases (as well as set variant overrides incorrectly), as the variant selections often do NOT apply to a root-level prim.  This fixes this by making it take a full SdfPath as the arg to build the key (and apply the variant overs).

### Included Commit(s)
- ~~187718eaadd37b1f42b82787a197ce736f69fbda~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/161/commits/059d7947917bb7b3c5f22437a95d0b39e97965ed~~
- https://github.com/PixarAnimationStudios/USD/pull/161/commits/318372e917d19c176df0e7e96380f180fe1a7fd7

### Fixes Issue(s)
- None filed

### DIFF
--- a/pxr/usd/lib/usdUtils/stageCache.cpp
+++ b/pxr/usd/lib/usdUtils/stageCache.cpp
@@ -62,7 +62,7 @@ UsdUtilsStageCache::Get()
 
 SdfLayerRefPtr 
 UsdUtilsStageCache::GetSessionLayerForVariantSelections(
-        const TfToken& modelName,
+        const SdfPath& primPath,
         const std::vector<std::pair<std::string, std::string> >&variantSelections)
 {
     // Sort so that the key is deterministic.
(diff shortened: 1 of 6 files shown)


elrond79:
I rebased / collapsed our changes, and pushed a fresh / clean copy. Look here for the full history:

https://github.com/LumaPictures/USD/commits/tg/simple_features/GetSessionLayerForVariantSelections_SdfPath

elrond79:
Updated with version which fixes a test failure

formerly failed testUsdReferenceAssemblyChangeRepresentations - did not error if no explicit primPath given, and the usd has no default prim defined

jtran56:
Filed as internal issue #147129.

elrond79:
Pushed a refreshed version of this PR for posterity, but going to just close this PR, due to the fact that, after discussing with Spiff, it seems unlikely to be merged, and because we no longer use this method ourselves.

Ideally would like to revisit this at some point, and refactor along lines Spiff suggested, to make it take a vector of SdfPaths + var selections, similar to katana's wrapper, but then, ideally I'd ride a dragon to work every day...

PR #164, opened 2017-02-14
make default value for readAnimData controlled by env. var

elrond79:
### Description of Change(s)

Adds an environment var, $PIXMAYA_READ_ANIM_BY_DEFAULT, for controlling the default setting for readAnimData on import.

### Included Commit(s)
- df084351aaeb4f61f4ef29b96f290f17a27a8045

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/145


spiffmon:
Hi Paul,
  0.7.3 makes readAnimData true by default. Given that, do we need this?

--SpiffiPhone

elrond79:
It does? It's still false in JobArgs (https://github.com/PixarAnimationStudios/USD/blob/master/third_party/maya/lib/usdMaya/JobArgs.cpp#L70), and in usdTranslatorImport (https://github.com/PixarAnimationStudios/USD/blob/master/third_party/maya/lib/usdMaya/usdTranslatorImport.h#L46).  And I tested the behavior on this commit, which I believe comes off the dev branch that was merged for 7.3, and it still defaulted to off.

However - if the plan is to make it true by default, then no, we don't need this...

spiffmon:
My mistake, twice over.
1) The change I was thinking of went in just after we cut 0.7.3, and it hasn't pushed out to github yet.
2) It only affects the import UI default.  

Here's the description:
Make USD import UI read anim data by default

- This only affects the translator's default UI options
- This does not change the usdExport scripting command, which assumes readAnimData=false by default unless the flag is explicitly set to true

--spiff

elrond79:
Hmm - if it doesn't affect the scripting command, then the PR would still
be handy for us.

chadrik:
@spiff, what do you think about forgoing the environment variable and just changing the default at the command level?  It's a breaking change, but now's the time to make those kinds of choices.

spiffmon:
@chadrik - sorry for the delay.  We discussed and do support changing the default to true at all levels.  Thanks!

elrond79:
Should I modify this pull request to just set the default at all levels? Or did you mean that was something you were going to do on your end?

elrond79:
(and sorry for the close / reopen spam, that was something suggested by github support deal with the fact that github though the travis-ci build hadn't finished, while travis-ci said it had...)

spiffmon:
Hi Paul,
   Go ahead and change all the defaults, please. 

--SpiffiPhone

elrond79:
Closing in favor of https://github.com/PixarAnimationStudios/USD/pull/174, which just defaults to true on all levels, with no env var

PR #165, opened 2017-02-22
Cmdline ar resolver

elrond79:
### Description of Change(s)
Fixes to allow usage of unresolved asset paths with both the command-line tools and the katana plugin

### Included Commit(s)
- ~~https://github.com/LumaPictures/USD/commit/f903a6e130787e38808ea014a7bd1b9f7c2c9551~~
- ~~https://github.com/LumaPictures/USD/commit/c575ac5035c7f005da67d357dac84ee7b2baf6b6~~
- ~~https://github.com/LumaPictures/USD/commit/55502b743275390d9ef4f3ed0b623324c6123f13~~

- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/562ec2cf0abd854e4919247c479ce1bb46322b3a~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/2a63b3628aab2e7011b56658f43b57c4f4a277c2~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/d4a74d56b94f0c9aae0ff48820368903c59de470~~

- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/eb955914b8bcf72670fd22b1e135def3e2c1573e~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/790810b2fe1d090650d73854169203045f762973~~
- ~~https://github.com/PixarAnimationStudios/USD/pull/165/commits/dd7f456d4a953b99e662d0c824b363d5e70cbce6~~

- https://github.com/PixarAnimationStudios/USD/pull/165/commits/cf84e05f45df535a687150bb617c1fb42505b232
- https://github.com/PixarAnimationStudios/USD/pull/165/commits/b9cef7bb2d72122dfeeed273eca4a31996d26245
- https://github.com/PixarAnimationStudios/USD/pull/165/commits/fdb3a78e084fea49b0cc726cdf262491bc851d54

### Fixes Issue(s)
None previously reported (that I know of...)

### DIFF
--- a/pxr/usd/bin/usddiff/usddiff.py
+++ b/pxr/usd/bin/usddiff/usddiff.py
@@ -85,12 +85,16 @@ def _findDiffTools():
     return (usdcatCmd, diffCmd)
 
 def _getFileFormat(path):
-    from pxr import Sdf
+    from pxr import Sdf, Ar
 
     # Note that python's os.path.splitext retains the '.' portion
     # when obtaining an extension, but Sdf's Fileformat API doesn't 
     # expect one. We also make sure to prune out any version specifiers.
     _, ext = os.path.splitext(path)
+    path = Ar.GetResolver().Resolve(path)
+    if path is None:
+        return None
+
     if len(ext) <= 1:
         fileFormat = Sdf.FileFormat.FindByExtension('usd')
     else:
@@ -101,8 +105,9 @@ def _getFileFormat(path):
          
         fileFormat = Sdf.FileFormat.FindByExtension(prunedExtension)
 
-    # Allow an empty file.
-    if fileFormat and (os.stat(path).st_size == 0 or fileFormat.CanRead(path)):
+    # Don't check if file exists - this should be handled by resolver (and
+    # path may not exist / have been fetched yet)
+    if fileFormat:
         return fileFormat.formatId
 
     return None
@@ -110,15 +115,19 @@ def _getFileFormat(path):
 def _convertTo(inPath, outPath, usdcatCmd, flatten=None, fmt=None):
     # Just copy empty files -- we want something to diff against but
     # the file isn't valid usd.
-    if os.stat(inPath).st_size == 0:
-        import shutil
-        try:
-            shutil.copy(inPath, outPath)
-            return 0
-        except:
-            return 1
-    else:
-        return call(_generateCatCommand(usdcatCmd, inPath, outPath, flatten, fmt))
+    try:
+        if os.stat(inPath).st_size == 0:
+            import shutil
+            try:
+                shutil.copy(inPath, outPath)
+                return 0
+            except:
+                return 1
+    except (IOError, OSError):
+        # assume it's because file doesn't exist yet, because it's an unresolved
+        # path...
+        pass
+    return call(_generateCatCommand(usdcatCmd, inPath, outPath, flatten, fmt))
 
 def _tryEdit(fileName, tempFileName, usdcatCmd, fileType, flattened):
     if flattened:
@@ -216,6 +225,9 @@ def _findFiles(args):
     baseline-only and comparison-only are lists of individual files, while
     matching is a list of corresponding pairs of files.'''
     import os
+    import stat
+    from pxr import Ar
+
     join = os.path.join
     basename = os.path.basename
     exists = os.path.exists
@@ -233,12 +245,33 @@ def _findFiles(args):
     if len(args) < 2:
         raise err
 
-    for index, exist in enumerate(map(exists, args)):
-        if not exist:
-            raise ValueError("Error: %s does not exist" % args[index]) 
+    # For speed, since filestats can be slow, stat all args once, then reuse that
+    # to determine isdir/isfile
+    resolver = Ar.GetResolver()
+    stats = []
+    for arg in args:
+        try:
+            st = os.stat(arg)
+        except (OSError, IOError):
+            if not resolver.Resolve(arg):
+                raise ValueError("Error: %s does not exist, and cannot be "
+                                 "resolved" % arg)
+            st = None
+        stats.append(st)
+
+    def isdir(st):
+        return st and stat.S_ISDIR(st.st_mode)
+
+    # if any of the directory forms are used, no paths may be unresolved assets
+    def validateFiles():
+        for i, st in enumerate(stats):
+            if st is None:
+                raise ValueError("Error: %s did not exist on disk, and using a "
+                                 "directory comparison form" % args[i])
 
     # DIR FILES...
-    if os.path.isdir(args[0]) and not any(map(os.path.isdir, args[1:])):
+    if isdir(stats[0]) and not any(map(isdir, stats[1:])):
+        validateFiles()
         dirpath = args[0]
         files = set(map(os.path.relpath, args[1:]))
         dirfiles = listFiles(dirpath)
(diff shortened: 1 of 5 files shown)


pxr/usd/bin/usddiff/usddiff.py
```diff
@@ -170,16 +172,33 @@ def listFiles(dirpath):
...
         return set(ret)
 
-    # Must have FILE FILE, DIR DIR, DIR FILES... or FILES... DIR.
+    origArgs = args
+
+    # path may not be a "real" path due to custom resolvers... so use
+    # Sdf.Layer.FindOrOpen first
+    args = []
+    for arg in origArgs:
+        layer = Sdf.Layer.FindOrOpen(arg)
```
sunyab: I think this would break usddiff in the case where you pass in directories to compare, since args (and origArgs) would contain directory names. You would wind up calling Sdf.Layer.FindOrOpen on a directory, which would fail and result in a ValueError getting raised.

pxr/usd/bin/usddiff/usddiff.py
```diff
@@ -170,16 +172,33 @@ def listFiles(dirpath):
                     for file in files]
         return set(ret)
 
-    # Must have FILE FILE, DIR DIR, DIR FILES... or FILES... DIR.
+    origArgs = args
+
+    # path may not be a "real" path due to custom resolvers... so use
+    # Sdf.Layer.FindOrOpen first
```
sunyab: I'm concerned about the performance impact of using Sdf.Layer.FindOrOpen to try to deal with custom asset paths that may not have a corresponding file on the filesystem. It seems like we don't really care about loading the layer here, we just want to know whether a layer at that path actually exists. So instead of doing this, could we try to use Ar.Resolver.Resolve somehow to give us this information?

third_party/katana/lib/usdKatana/cache.cpp
```diff
@@ -264,23 +264,21 @@ UsdStageRefPtr UsdKatanaCache::GetStage(
...
                 _FindOrCreateSessionLayer(sessionAttr, sessionRootLocation);
 
         UsdStageCache& stageCache = UsdUtilsStageCache::Get();
         UsdStageCacheContext cacheCtx(stageCache);
         const UsdStage::InitialLoadSet load = 
             (forcePopulate ? UsdStage::LoadAll : UsdStage::LoadNone);
-        UsdStageRefPtr const& stage = UsdStage::Open(rootLayer, sessionLayer, 
-                ArGetResolver().GetCurrentContext(),
+        UsdStageRefPtr const& stage = UsdStage::Open(rootLayer, sessionLayer,
+                ArGetResolver().CreateDefaultContextForAsset(fileName),
```
sunyab: What was the reason for changing this to use a new context rather than whatever context was currently bound?

third_party/maya/lib/usdMaya/usdImport.cpp
```diff
@@ -109,18 +111,28 @@ MStatus usdImport::doIt(const MArgList & args)
...
+        // First resolve using maya's resolver, to expand any environment
+        // variables, or any maya-specific tokens
+        MFileObject mayaFile;
+        mayaFile.setRawFullName(tmpVal);
+
+        // Use the usd resolver - this should also have the side effect of making
+        // absolute paths
+        std::string expandedPath = mayaFile.expandedFullName().asChar();
+        ArResolver& resolver = ArGetResolver();
+        resolver.ConfigureResolverForAsset(expandedPath);
```
sunyab: I don't think we should reconfigure the resolver here, since that's a stateful change that potentially affects all subsequent consumers of ArResolver for path resolution. Can you describe what issue you were running into here?

pxr/usd/bin/usddiff/usddiff.py
elrond79: Hmm, good point. Essentially want we want to do here is just validate using the ArResolver. I actually posted a question on the mailing list about this (and some other resolver issues): [https://groups.google.com/forum/#!topic/usd-interest/E3huSatYrLI]https://groups.google.com/forum/#!topic/usd-interest/E3huSatYrLI

Assuming that we can rely on getting back an empty string as a form of validation, I think maybe the best thing to do is to just replace the look that checks for existence with one which checks that either it resolves to non-empty, OR it's a dir (per your next note).

pxr/usd/bin/usddiff/usddiff.py
elrond79: See my response to the comment above...

third_party/maya/lib/usdMaya/usdImport.cpp
elrond79: As far as configuring the resolver, that wasn't done in response to a specific problem, but rather just confusion about best practices - again, see my first question here: https://groups.google.com/forum/#!topic/usd-interest/E3huSatYrLI

Your comment about this being a stateful change helps clarify things somewhat, though - on re-reviewing the existing code, it seems like everywhere that's calling ConfigureResolverForAsset is doing so from within "application level" code.

third_party/katana/lib/usdKatana/cache.cpp
elrond79: Hmm, just confusion about best practices - see my first question here: https://groups.google.com/forum/#!topic/usd-interest/E3huSatYrLI

Going off of some other examples (ie, usdviewq/mainWindow.py, and _createPathResolverContext), it seemed like the "best" thing to do was, if you knew the path you were about to open, to create a context for it.  Apologies if this isn't the proper thing to do - I will remove this change!

elrond79:
First - thanks for the code review, was very helpful!

Second - I'm making the changes noted... and realizing that using on-disk filepaths and unresolved asset paths interchangeably is tricky, once directories get involved.  For simplicity, I'm going to make the following assumptions:

- we first check an arg to see if it is a file/dir, and if so, assume it is NOT an unresolved asset path
- if any of the directory forms are used, then none of the other args may be unresolved asset paths

I'll add a note to the parser help to this effect. Let me know if you'd rather proceed in a different manner.

elrond79:
So, I pushed a fresh branch where I've made the changes requeted.  Sorry for the delay - we're in the middle of migrating our commits over to topgit to make juggling our various patches more manageable.  Anyway, this branch is a "clean" history - if you want to see all the incremental changes, I can push the topgit patch branches. Thanks again for the review!

elrond79:
Pushed a new version that fixed a test failure...

jtran56:
Filed as internal issue #147131.

sunyab:
Hi @elrond79, I was trying to merge this in but I still run into a test failure with testUsdDiffToolVariousFormats5:

  File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/bin/usddiff", line 340, in <module>
    main()
  File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/bin/usddiff", line 324, in main
    results.flatten, results.noeffect):
  File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/bin/usddiff", line 152, in _runDiff
    baselineFileType = _getFileFormat(baseline)
  File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/bin/usddiff", line 117, in _getFileFormat
    return fileFormat.formatId
AttributeError: 'NoneType' object has no attribute 'formatId'

Do you guys see the same thing? Or is this something on my end?

elrond79:
Oops - sorry, yes, that is a problem we saw, and fixed a while ago... but apparently forgot to push a refreshed PR.  The PR should be up-to-date now.

For your reference, here's the (non-collapsed) commit that fixed this:

https://github.com/LumaPictures/USD/commit/2f05bd706b17a8476a15404a8bbb06418a97eee2

elrond79:
Rebased the PR to resolve merge conflicts...

PR #166, opened 2017-02-22
Setting up the shading mode option menu dynamically.

sirpalee:
### Description of Change(s)
The change builds the Shading Mode list dynamically, instead of hard coding the values in the Mel scripts. The advantage of this change is adding a new shading mode export won't require manual changes in the UI scripts, and the translatorMaterial will use the right exporter at export time.

To make this work, I needed to add a new command to list the export/import Shading Modes, similar functions to the Registry, and modify the UI scripts to build the optionMenu based on the output of the command.

### Fixes Issue(s)
N/A


sirpalee:
@sunyab @spiffmon Can we get this reviewed as well?

PR #168, opened 2017-02-23
Redesigning the shading mode export interface

sirpalee:
### Description of Change(s)
This change deals with redesigning the shading mode export/import interface. As a first step, it's a take on the export interface. I tried to create a bit more flexible export process, so exporters can overwrite the whole process too if they don't want to go through the exporter context.

This is just a take on the shading mode export, so feedback can be gathered, before moving on to the importers.

### Fixes Issue(s)
- There is no way to do stateful shading mode export / import interfaces.
- There is no way to include spaces in the shading mode exporter's/importer's name, so lots of manual remapping is required.
- No easy way to link the same exporter to multiple modes.
- Exporters can't take over the whole process themselves, just by one shading node at a time.

### DIFF
--- a/third_party/maya/lib/usdMaya/shadingModeExporterContext.h
+++ b/third_party/maya/lib/usdMaya/shadingModeExporterContext.h
@@ -1,0 +1,88 @@
+//
+// Copyright 2016 Pixar
+//
+// Licensed under the Apache License, Version 2.0 (the "Apache License")
+// with the following modification; you may not use this file except in
+// compliance with the Apache License and the following modification to it:
+// Section 6. Trademarks. is deleted and replaced with:
+//
+// 6. Trademarks. This License does not grant permission to use the trade
+//    names, trademarks, service marks, or product names of the Licensor
+//    and its affiliates, except as required to comply with Section 4(c) of
+//    the License and to reproduce the content of the NOTICE file.
+//
+// You may obtain a copy of the Apache License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the Apache License with the above modification is
+// distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the Apache License for the specific
+// language governing permissions and limitations under the Apache License.
+//
+#ifndef PXRUSDMAYA_SHADINGMODEEXPORTERCONTEXT_H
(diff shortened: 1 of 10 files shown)


sirpalee:
If we combine these changes with https://github.com/PixarAnimationStudios/USD/pull/166 , we can completely remove the need for the branching logic when reading in the shading modes from the commands, because we can re-register the displayColor exporter under multiple aliases, and pxrRis under an alias that displayed properly in the Option Menu.

third_party/maya/lib/usdMaya/shadingModeExporterContext.h
```diff
@@ -0,0 +1,93 @@
...
+    MObject _shadingEngine;
+    const UsdStageRefPtr& _stage;
+    bool _mergeTransformAndShape;
+    bool _handleUsdNamespaces;
+    SdfPath _overrideRootPath;
+
+    SdfPathSet _bindableRoots;
+};
+
+typedef boost::function< void (PxrUsdMayaShadingModeExportContext*) > PxrUsdMayaShadingModeExporterContext;
```
sunyab: From what we could tell, this typedef isn't used anywhere -- could you confirm and remove if it isn't?

Otherwise, (and this is nit-picky) -- could we rename this typedef (and use a C++11 type alias), since this type isn't really a context, it's a function accepting a context?

third_party/maya/lib/usdMaya/translatorMaterial.cpp
```diff
@@ -317,21 +317,11 @@ PxrUsdMayaTranslatorMaterial::ExportShadingEngines(
...
-            PxrUsdMayaShadingModeExportContext c(
-                    shadingEngine,
-                    stage,
-                    mergeTransformAndShape,
-                    bindableRoots,
-                    overrideRootPath);
-
-            exporter(&c);
+        auto exporter = exporterCreator();
+        if (exporter != nullptr) {
```
sunyab: Since we used an inline assignment at line 320, can we make this if check also do the same for consistency? e.g.,

if (auto exporter = exporterCreator()) { ... }

sirpalee:
@sunyab Thanks for the feedback!

About the copyright notice, it was probably lost during a rebase when we switched between our branch and yours.

For future reference, what do you prefer when addressing feedback? Updating the original commit with the changes (cleaner history), or making an extra commit (simpler tracking of changes)?

sunyab:
@sirpalee Thanks for the changes! I think I'd prefer to squash and get the cleaner history for now.

sirpalee:
@sunyab Done! If you are happy with the current state, I'll go ahead and change the import interface in a similar manner.

PR #172, opened 2017-03-06
[usdMaya] Exporting instances from Maya.

sirpalee:
### Description of Change(s)
This PR focuses on export instances from Maya. Right now import is not yet supported since this is a significant change, I want to gather some feedback before implementing the import side of the modification.

A couple of notes.
- I renamed "Make Instances" to "Make References Instanceable" to avoid confusion and make the difference between instances and that flag.
- The central part of the change is creating a new class, usdWriteJobCtx, which is the parent of usdWriteJob. This ctx is meant to be passed to the writers and has common functionality that's useful both in usdWriteJob and each of the writers. The writer creation is also moved here, because in our approach to instances, ever shape is an instance and the instances created in a different scope. The ctx handles this creation process.
- The ctx also has a bunch of shared variables, like the args, and the stage, what was accessible through the prim writer earlier.
- Most of the instancing related logic is done in the MayaTransformWriter. This way, all the shapes can be instanced automatically, no need to add code to each shape.
- I decided to put the instances in a separate scope. This seemed a bit cleaner, but it also lead to using inherits instead of references. (no non-root prims with references)
- The change doesn't support instancing transforms. Instead, it instances the shapes and creates a unique shape / xform for each instance source in the scope.
- The scope is created first, then removed at saving, if it's empty. This way we could have it at the beginning of the file, this is strictly a question of what looks nicer.
- I also removed a bunch of boost code, like using std::shared_ptr for all the writers.
- The MDagPathComparator also changed, using strcmp is simpler than creating two std::strings.

### DIFF
--- a/third_party/maya/lib/usdMaya/MayaPrimWriter.h
+++ b/third_party/maya/lib/usdMaya/MayaPrimWriter.h
@@ -26,16 +26,11 @@
 
 #include "pxr/pxr.h"
 #include "usdMaya/api.h"
-#include "usdMaya/JobArgs.h"
+#include "usdMaya/usdWriteJobCtx.h"
 
 #include "pxr/usd/usd/stage.h"
 
-#include <boost/smart_ptr.hpp>
-
 PXR_NAMESPACE_OPEN_SCOPE
-
-
-using boost::shared_ptr;
 
 class UsdGeomImageable;
 class UsdTimeCode;
@@ -46,13 +41,13 @@ class MayaPrimWriter
   public:
     PXRUSDMAYA_API
     MayaPrimWriter(
-            MDagPath & iDag, 
-            UsdStageRefPtr stage, 
-            const JobExportArgs & iArgs);
+            const MDagPath& iDag,
+            const SdfPath& uPath,
+            usdWriteJobCtx& jobCtx);
     virtual ~MayaPrimWriter() {};
 
-    virtual UsdPrim write(const UsdTimeCode &usdTime) = 0;
-    virtual bool isShapeAnimated()     const = 0;
+    virtual void write(const UsdTimeCode &usdTime) = 0;
+    virtual bool isShapeAnimated() const = 0;
 
     /// Does this PrimWriter directly create one or more gprims on the UsdStage?
     ///
@@ -75,11 +70,13 @@ class MayaPrimWriter
     virtual bool shouldPruneChildren() const;
 
 public:
-    const MDagPath&        getDagPath()    const { return mDagPath;};
-    const SdfPath &        getUsdPath()    const { return mUsdPath; };
-    const UsdStageRefPtr&  getUsdStage()   const { return mStage;};
-    bool isValid()                         const { return mIsValid;};
-    const JobExportArgs&   getArgs()       const { return mArgs;};
+    const MDagPath&        getDagPath()    const { return mDagPath; }
+    const SdfPath&         getUsdPath()    const { return mUsdPath; }
+    const UsdStageRefPtr&  getUsdStage()   const { return mWriteJobCtx.getUsdStage(); }
+    bool isValid()                         const { return mIsValid; }
+    const JobExportArgs&   getArgs()       const { return mWriteJobCtx.getArgs(); }
+    const UsdPrim&         getPrim()       const { return mUsdPrim; }
+
 
 protected:
     void setValid(bool isValid) { mIsValid = isValid;};
@@ -87,19 +84,15 @@ protected:
     PXRUSDMAYA_API
     bool writePrimAttrs(const MDagPath & iDag2, const UsdTimeCode &usdTime, UsdGeomImageable &primSchema);
 
-  private:
+    UsdPrim mUsdPrim;
+    usdWriteJobCtx& mWriteJobCtx;
+
+private:
     MDagPath mDagPath;
-    UsdStageRefPtr mStage;
     SdfPath mUsdPath;
+
     bool mIsValid;
-
-    // This is just a reference to the JobExportArgs object of the UsdWriteJob
-    // that creates this prim writer. This prevents copying when the set of
-    // dagPaths can potentially be large.
-    const JobExportArgs& mArgs;
 };
-
-typedef shared_ptr < MayaPrimWriter > MayaPrimWriterPtr;
 
 PXR_NAMESPACE_CLOSE_SCOPE
 
--- a/third_party/maya/lib/usdMaya/MayaTransformWriter.h
+++ b/third_party/maya/lib/usdMaya/MayaTransformWriter.h
@@ -35,7 +35,6 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-
 class UsdGeomXformable;
 class UsdTimeCode;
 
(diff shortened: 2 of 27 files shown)


sunyab:
Hi @sirpalee, our sets team would like to review this pull request -- would you be able to re-merge dev into this branch and resolve the conflicts? I think a good chunk of the conflicts are due to commit 8b9e9a. Thanks!

sirpalee:
@sunyab sorry for the late reply, but I got the pr updated. I'm still working on some changes requested by @spiffmon , I'm hoping to be done with those tomorrow.

sirpalee:
@sunyab The extra changes were done.
- Using an over instead of a scope for the instance sources group.
- Making sure the instance sources won't be the default prim.

sunyab:
@sirpalee: Thanks! I've pulled these changes locally for testing and review.

third_party/maya/lib/usdMaya/MayaPrimWriter.h
```diff
@@ -75,32 +70,30 @@ class MayaPrimWriter
...
 
 protected:
     void setValid(bool isValid) { mIsValid = isValid;};
     void setUsdPath(const SdfPath &newPath) { mUsdPath = newPath;};
     PXRUSDMAYA_API
     bool writePrimAttrs(const MDagPath & iDag2, const UsdTimeCode &usdTime, UsdGeomImageable &primSchema);
 
-  private:
+    UsdPrim mUsdPrim;
+    usdWriteJobCtx& mJob;
```
sunyab: i'd call this ```mWriteJobCtx``` instead of ```mJob```

third_party/maya/lib/usdMaya/MayaTransformWriter.h
```diff
@@ -64,6 +63,7 @@ class MayaTransformWriter : public MayaPrimWriter
 
     PXRUSDMAYA_API
     MayaTransformWriter(MDagPath & iDag, UsdStageRefPtr stage, const JobExportArgs & iArgs);
```
sunyab: Doesn't look like the implementation for this constructor exists anymore. (remove declaration since looks like all references were switched to the new constructor?)

third_party/maya/lib/usdMaya/primWriterRegistry.h
```diff
@@ -67,9 +67,9 @@ PXR_NAMESPACE_OPEN_SCOPE
 struct PxrUsdMayaPrimWriterRegistry
 {
     typedef std::function< MayaPrimWriterPtr (
-            MDagPath &curDag,
-            UsdStageRefPtr& stage,
-            const JobExportArgs& args) > WriterFactoryFn;
+            const MDagPath&,
+            const SdfPath&, bool,
+            usdWriteJobCtx&) > WriterFactoryFn;
```
sunyab: The example code on Register (line 81-85) here should be updated to have the same signature as a WriterFactoryFn

third_party/maya/lib/usdMaya/usdWriteJob.cpp
```diff
@@ -221,45 +212,44 @@ bool usdWriteJob::beginJob(const std::string &iFileName,
...
+                        MayaTransformWriterPtr xformWriter =
+                            std::dynamic_pointer_cast<MayaTransformWriter>(primWriter);
+                        if (xformWriter) {
+                            MDagPath xformDag = xformWriter->getTransformDagPath();
+                            mDagPathToUsdPathMap[xformDag] = usdPrim.GetPath();
+                        }
                     }
                 }
 
-                mModelKindWriter.OnWritePrim(usdPrim, primWriter);
```
sunyab: Did you intend to remove the call to mModelKindWriter.OnWritePrim?

third_party/maya/lib/usdMaya/usdWriteJob.h
```diff
@@ -31,15 +31,18 @@
...
+
+#include <maya/MObjectHandle.h>
+
 #include <string>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
 
-class usdWriteJob
+class usdWriteJob : public usdWriteJobCtx
```
sunyab: What's the rationale here for inheriting? It seems like this should just have a usdWriteJobCtx object instead?

third_party/maya/lib/usdMaya/usdWriteJobCtx.h
```diff
@@ -0,0 +1,58 @@
...
+typedef std::shared_ptr<MayaPrimWriter> MayaPrimWriterPtr;
+
+class usdWriteJobCtx {
+public:
+    PXRUSDMAYA_API
+    usdWriteJobCtx(const JobExportArgs& args);
+    const JobExportArgs& getArgs() const { return mArgs; };
+    const UsdStageRefPtr& getUsdStage() const { return mStage; };
+    // Querying the master path for instancing. This also creates the mesh if it doesn't exists.
+    SdfPath getMasterPath(const MDagPath& dg);
```
sunyab: This function should have the PXRUSDMAYA_API decoration as well since it's not implemented inline

third_party/maya/lib/usdMaya/usdWriteJobCtx.h
```diff
@@ -0,0 +1,58 @@
...
+class MayaPrimWriter;
+typedef std::shared_ptr<MayaPrimWriter> MayaPrimWriterPtr;
+
+class usdWriteJobCtx {
+public:
+    PXRUSDMAYA_API
+    usdWriteJobCtx(const JobExportArgs& args);
+    const JobExportArgs& getArgs() const { return mArgs; };
+    const UsdStageRefPtr& getUsdStage() const { return mStage; };
+    // Querying the master path for instancing. This also creates the mesh if it doesn't exists.
```
sunyab: should "mesh" here be "shape"?

third_party/maya/lib/usdMaya/usdWriteJob.h
sirpalee: To avoid creating a usdWriteJobCtx member on the usdWriteJob. Note, I should have moved the usdWriteJobCtx constructor to protected

third_party/maya/lib/usdMaya/usdWriteJob.cpp
sirpalee: No, there were various rewrites while working on this change, and I forgot to add this back to the code. (it was present somewhere else for a while) Added it back.

third_party/maya/lib/usdMaya/MayaPrimWriter.h
sirpalee: I also renamed the parameters in the constructors to jobCtx.

sirpalee:
@sunyab Updated the commit with the changes.

third_party/maya/lib/usdMaya/usdWriteJob.h
sunyab: Are we expecting to derive more things from usdWriteJobCtx? If not, I think we'd prefer usdWriteJob to have a usdWriteJobCtx as a member instead and add a public accessor like:

 ```const usdWriteJobCtx& usdWriteJob::getContext() const``` 

or some such. From what we could tell, it seemed like that wouldn't be too big of a change to your code. What do you think?

Also, if we could add some simple class documentation to the top of usdWriteJobCtx that describes its purpose (e.g., serves as interface for export plugins) that'd be great.

sirpalee: @sunyab done the requested change and change the inheritance to a member variable more than a week ago... I just forgot to reply here.

jtran56:
Filed as internal issue #147132.

PR #191, opened 2017-04-03
Enabling proxy draw overrides for VP2.

sirpalee:
### Description of Change(s)
This PR enables Maya VP2 rendering of proxy shapes.

### Fixes Issue(s)
Proxy shapes were not visible in VP2 because of the missing supportedDrawAPIs.

### DIFF
--- a/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
+++ b/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
@@ -184,6 +184,16 @@ UsdMayaProxyDrawOverride::prepareForDraw(
     return userData;
 }
 
+MHWRender::DrawAPI
+UsdMayaProxyDrawOverride::supportedDrawAPIs() const
+{
+#if MAYA_API_VERSION >= 201600
+    return MHWRender::kOpenGL | MHWRender::kOpenGLCoreProfile;
+#else
+    return MHWRender::kOpenGL;
+#endif
+}
+
 void
 UsdMayaProxyDrawOverride::draw(const MHWRender::MDrawContext& context, const MUserData *data)
 {
--- a/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.h
+++ b/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.h
@@ -71,6 +71,9 @@ public:
         MUserData* oldData);
 
     PXRUSDMAYAGL_API
+    virtual MHWRender::DrawAPI supportedDrawAPIs() const;
+
+    PXRUSDMAYAGL_API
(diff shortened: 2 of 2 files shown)


mattyjams:
Hi @sirpalee,

Thanks for the PR. We actually have this exact same change stashed away internally. We've been holding off on committing it so far though because even though we need it to make the proxies draw, we're then seeing some incorrect/inconsistent picking behavior in the viewport when we use VP2.0.

We've been talking with Autodesk about the issue and it sounds like they've added some new API on MPxDrawOverride (wantUserSelection() and userSelect() callbacks, I believe) that we can use to properly support selection when rendering in VP2.0. They added that very recently though, and at the moment it's only available in pre-release versions of Maya 2017 (starting with PR76, I'm told). I've heard that it will be included as part of 2017 Update 4 when that is released. We don't have any of these versions deployed at the moment, so we're not currently able to give it a try.

We made a note about VP2.0-related issues on the [Maya USD Plugin page](http://graphics.pixar.com/usd/docs/Maya-USD-Plugins.html#MayaUSDPlugins-TheMayaViewportandOpenGL) and included a Maya environment variable we're using to work around the problem for now. We're primarily on Maya 2016 still and have really only just started testing with 2017, so I'm not certain of the state of things with 2017. The environment variable described there has been working for us with 2016 though.

Are you seeing these issues with viewport picking when you build with this change?

sirpalee:
Hey! We are using OpenGL - Core Profile (Compabitility) mode for 2016 ext 2 sp2 on CentOS 7. Selection seems to work fine. Is there anything to test that's known for failing?

I understand these concerns, and the confusion caused by having selection issue, but sometimes enabling the display, even with the selection issues, is a better approach. Maybe we could add an env variable to enable draw overrides?

mattyjams:
Hmm... We have 2016 extension 2 sp2 installed here, so I've been giving that a try this afternoon. I've set my rendering engine preference to "OpenGL - Core Profile (Compatibility)", and I have a Maya scene with one pxrUsdReferenceAssembly node in it that has a nested structure such that when I switch it to "Full", it has multiple proxy nodes underneath.

When I have the viewport set to "Legacy Default Viewport", I see everything drawing correctly, and selecting proxy nodes by picking in the viewport works correctly too. However, when I have the viewport set to "Viewport 2.0", everything draws, but selection by picking in the viewport does **not** work, or at least does not work consistently. I can occasionally get a marquee selection to select the proxies depending on my camera location/angle, but I have so far not gotten a single selection (non-marquee) to work. Selecting nodes directly in the outliner does correctly draw them with a wireframe outline in the viewport, but actually selecting **in** the viewport doesn't seem to be working.

Is there anything else about your setup I might be missing? What does the structure of the Maya scene you're testing with look like?

jtran56:
Filed as internal issue #147133.

PR #195, opened 2017-04-07
[usdMaya] merge usdTranslatorImport/Export

elrond79:
### Description of Change(s)
necessary because if there are two separate translators, when saving a reference, maya has a bug where it recorded the incorrect translator (ie, pxrUsdExport), even though pxrUsdImport was used to create the reference

### Included Commit(s)
- https://github.com/PixarAnimationStudios/USD/commit/36db2a0f62e42f62b667841fa07a1449bcd36570

### Fixes Issue(s)
- If you create a reference to a USD file, then save the scene containing the reference with it loaded, then open it again, there will be an error on attempting to reload the USD reference. This is because maya (incorrectly) attempts to the load the reference using the usdExport translator.  This is a bug on maya's part, which I've reported to them, but until it's fixed, this is the only workaround if you want to use USD references.


jtran56:
Filed as internal issue #147134.

elrond79:
Going to go ahead and close this PR, as we no longer use it ourselves...

PR #206, opened 2017-04-28
[tests] deal with env vars that have an '=' in them

elrond79:
### Description of Change(s)
Currently running the tests will error in testWrapper.py if you have an environ var (in our case, it was PYTHONPATH) that has an '=' in it. This fixes it deal properly with such vars.

### Included Commit(s)
- https://github.com/PixarAnimationStudios/USD/commit/6f9a52afdcdb09dffbc72a0e872591deec3dfd7a

### Fixes Issue(s)
- running tests when an env var has a '='

### DIFF
--- a/cmake/macros/testWrapper.py
+++ b/cmake/macros/testWrapper.py
@@ -240,7 +240,7 @@ if __name__ == '__main__':
             sys.stderr.write("Error: use --pre-path or --post-path to edit PATH.")
             sys.exit(1)
         try:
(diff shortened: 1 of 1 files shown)


chadrik:
You're probably curious why we have paths with '=' in them, but even if you're not, here's an explanation!   [rez](https://github.com/nerdvegas/rez) generates folder structures based on package version rules, which look like `usd<=0.7.5`, etc.

PR #208, opened 2017-05-04
[usdMaya] Adding a maya particle writer.

sirpalee:
### Description of Change(s)
Adding a prim writer for Maya Particles and NParticles. The logic for exporting user attributes (which ones to ignore mostly) is taken from partio4Maya.

Note, the exporter fails on Maya 2016 ext2 if you are writing out a cached NParticle. However, the same code works with Maya 2016 and Maya 2017. There seems to be a critical bug in 2016 ext2 that zeroes out all the position values, but all the other attributes work just fine. The same problem was seen with Maya To Arnold, PartIO Export, and even the built-in Maya Instancer.

### Fixes Issue(s)
No reported issue.


sirpalee:
Uh-oh... I forgot that the velocity in maya is per second, and it's most likely per frame in USD. I'm going to check that and update the pull request with scaling down the velocity if necessary.

sirpalee:
Added velocity scaling.

sirpalee:
The documentation is going to be changed for pointBased, and the requirement for the velocity is going to be per second, so I removed the velocity scaling from the writer.

jtran56:
Filed as internal issue #147136.

third_party/maya/lib/usdMaya/MayaParticleWriter.cpp
```diff
@@ -0,0 +1,362 @@
+//
+// Copyright 2016 Pixar
```
sunyab: Could you update the year in the copyright notice here to 2017?

third_party/maya/lib/usdMaya/MayaParticleWriter.cpp
```diff
@@ -0,0 +1,362 @@
...
+    // All the initial state attributes end with 0
+    bool _isInitialAttribute(const std::string& attrName) {
+        return attrName.back() == '0';
+    }
+
+    bool _isCachedAttribute(const std::string& attrName) {
+        constexpr auto _cached = "cached";
+        constexpr auto _Cache = "Cache";
+        const auto attrNameSize = attrName.size();
+        if (attrNameSize > 5 && attrName.substr(attrNameSize - 6) == _Cache) {
```
sunyab: I think this could be:
``return TfStringEndsWith(attrName, _Cache) || TfStringStartsWith(attrName, _cached);``

(need to include "pxr/base/tf/stringUtils.h")

third_party/maya/lib/usdMaya/MayaParticleWriter.cpp
```diff
@@ -0,0 +1,362 @@
...
+    writeParams(usdTime, primSchema);
+}
+
+void MayaParticleWriter::writeParams(const UsdTimeCode& usdTime, UsdGeomPoints& points) {
+    if (usdTime.IsDefault() == isShapeAnimated()) {
+        return;
+    }
+
+    const auto particleNode = getDagPath().node();
+    MFnParticleSystem PS(particleNode);
```
sunyab: Just one nitpick: for consistency with coding conventions, could we change the variable names ``PS`` and ``DPS``?  I think i would've gone with ``particleSys`` and ``deformedParticleSys`` but that is quite a lot to type..

third_party/maya/lib/usdMaya/MayaParticleWriter.h
```diff
@@ -0,0 +1,68 @@
+//
+// Copyright 2016 Pixar
```
sunyab: Like above, please update year to 2017 here

sirpalee:
@sunyab updated the PR and added a simple test.

sunyab:
I finally got around to merging this change, thanks so much for your patience. 

There were a few fixups I had to make to resolve conflicts and to get this change trued up with the current state of the world. Most notably, the particle writing code was using a VtArray<long> when writing the ids attribute, which broke on Windows and MacOS. The documentation on the ids attribute does say to use VtArray<long>, but this is a bug in our code generation we need to fix. I've changed this to use VtArray<int64_t>, which lines up with the Int64Array attribute type. The unit test you added (thank you for that!) passes with this change.

Please let me know if there's any issues with these changes. Thanks again!

PR #209, opened 2017-05-05
set $MAYA_APP_DIR before running tests, to use a default maya profile

elrond79:
### Description of Change(s)
in testWrapper.py, set the $MAYA_APP_DIR env var, so that maya tests run using a default user profile

### Included Commit(s)
- ~~https://github.com/PixarAnimationStudios/USD/commit/718780496f8cfd88492a2f54fb783b348b3be5f3~~
- https://github.com/PixarAnimationStudios/USD/pull/209/commits/f75077e557a7a5c15ccf5ea318076e52237e62e8

### Fixes Issue(s)
- We were experiencing some crashes when running the tests on maya-2017, due to some problems with the color-management settings in the user profile

### DIFF
--- a/cmake/macros/testWrapper.py
+++ b/cmake/macros/testWrapper.py
@@ -78,7 +78,10 @@ def _parseArgs():
     parser.add_argument('--expected-return-code', type=int, default=0,
             help='Expected return code of this test.')
     parser.add_argument('--env-var', dest='envVars', default=[], type=str, 
-            action='append', help='Variable to set in the test environment.')
+            action='append',
+            help=('Variable to set in the test environment, in KEY=VALUE form. '
+                  'If "<PXR_TEST_DIR>" is in the value, it is replaced with the '
+                  'absolute path of the temp directory the tests are run in'))
     parser.add_argument('--pre-path', dest='prePaths', default=[], type=str, 
             action='append', help='Path to prepend to the PATH env var.')
     parser.add_argument('--post-path', dest='postPaths', default=[], type=str, 
@@ -243,6 +246,7 @@ if __name__ == '__main__':
             sys.exit(1)
         try:
             k, v = varStr.split('=', 1)
+            v = v.replace('<PXR_TEST_DIR>', testDir)
             env[k] = v
         except IndexError:
             sys.stderr.write("Error: envvar '{0}' not of the form "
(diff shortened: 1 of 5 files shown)


jtran56:
Filed as internal issue #147137.

sunyab:
Hey @elrond79, at the moment we'd prefer to keep library or plugin-specific environment variables out of testWrapper.py. Can we add MAYA_APP_DIR to the tests in usdMaya/CMakeLists.txt and pxrUsd/CMakeLists.txt instead, just like we do for MAYA_DISABLE_CIP? It's more boilerplate (which we can maybe move to a macro or something later) but it makes it less mysterious what environment is set when the tests are run.

I guess the one tricky bit is that in testWrapper.py, you can set MAYA_APP_DIR to the temporary test directory, which you can't really do in the CMakeLists.txt. However, would it work to set MAYA_APP_DIR to a relative path, like "./maya_profile" instead?

elrond79:
Makes sense - moved it into the CMakeLists.txt files. I checked and confirmed that maya will use relative paths for MAYA_APP_DIR.  Latest commit should have these changes!

sunyab:
Just wanted to provide an update: I tried merging these changes earlier but the Maya tests on MacOS and Windows with Maya 2017 failed with the following errors:

On MacOS:
*** Fatal Error: Failed creating directory: ./maya_profile

On Windows:
*** Fatal Error: Failed creating directory: ./maya_profile/2017/prefs/icons
(It's able to create the maya_profile directory but not a subdirectory beneath it? Weird.)

On Linux, everything worked as expected.

Running the test using an absolute path for MAYA_APP_DIR seemed to work in all cases, so it seems like support for relative paths is inconsistent at best. I wonder if another possible solution is to have testWrapper.py do some kind of substitution; for example, if it gets passed a parameter like $TESTDIR/maya_profile, it'll substitute $TESTDIR with the temporary test directory. Still thinking about this one, though.

elrond79:
So, I fixed it along the lines you suggested - I altered testWrapper.py to automatically replace `<PXR_TEST_DIR>` in any --env-var args passed into it.  Decided to go with <NAME> syntax instead of $NAME because I wanted it to be distinct from env vars or CMake vars, to avoid confusion, but I don't feel strongly about that... we can use whatever syntax or var name you prefer.

PR #210, opened 2017-05-06
[usdview] add --camera cmdline arg

elrond79:
### Description of Change(s)

Allows choosing of what camera to use when opening the scene

### Included Commit(s)
- https://github.com/PixarAnimationStudios/USD/commit/f2f5e997be58d64f76a754ba703c7e0203bc0cc3

### Fixes Issue(s)
- None, perse, but there was a TODO note about not hardcoding the default to be a pixar-specific name, which this effectively does


elrond79:
Added in the ability for the command-line arg to take full prim paths, as requested.  Didn't touch the handling in the UI menu - will let you folks decide how you want to deal with that.

Lemme know if you need anything else!

jtran56:
Filed as internal issue #147138.

PR #214, opened 2017-05-24
[USD] Adding position interpolation to points.

sirpalee:
### Description of Change(s)
The PR adds a new function, ComputePointsAtTimes to UsdGeomPoints, including tests and python bindings.

The goal of the change is to have an easy to use interface to calculate velocity based interpolation for points, and in the case of no velocity fall back to the built-in interpolation mode. The change also updates the katana point-reader, so interpolated positions will are generated at shutter open and close. For example, with this change, Arnold in Katana interpolates particles nicely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,49 @@
 # Change Log
 
+## [0.8.0] - 2017-07-29
+
+### Added
+- Added option to build_usd.py for building monolithic shared library
+  and for building new sample Embree plugin.
+- Added --mask option to usdcat, matching the --mask option in usdview.
+- Added CPU-side external computation support to Hydra.
+- Added Embree backend to serve as an example of how to add renderer 
+  plugins to Hydra.
+- Added camera gates to usdview.
+- Added ability to Houdini plugin to set active state via usdactive 
+  attribute and to write static data to default time via usdwritestatictopology,
+  usdwritestaticprimvars, and usdwritestaticgeo attributes.
+- Added UsdExportAttributes hda to Houdini plugin for setting primitive and 
+  detail attributes which control various USD attributes or metadata.
+
+### Changed
+- Restored export of CMake targets from shared library builds.
+- Asset paths in .usda files may now contain any printable character.
+  (Issue #73)
+- Legacy special behavior for variant sets named "standin" has been disabled
+  by default.
+- Interpolation is now applied to time samples from clips when no sample
+  exists at a specified clip time.
+- Removed deprecated UsdShadePShader schema.
+- Enabled vertex primvar sharing in Hydra. Hydra can now use significantly 
+  less GPU memory.
+- Many internal performance improvements to Hydra.
+- usdview HUD now shows the backend renderer plugin used by Hydra
+- Wireframe drawing now ignores opacity.
+- Katana plugin properly interprets schema-supported triangle subdivision rule 
+  when meshes are read in and rendered.
+
+### Fixed
+- Fixed several issues in build_usd.py related to building dependencies.
+  (Issue #225, #230, #234)
+- Fixed bug where UsdGeomPrimvar::IsIndexed() would not work for attributes
+  with only authored time samples. (Issue #238)
+- Fixed small platform inconsistencies for Windows in Hydra.
+- Fixed crash in Katana plugin when using point instancers that did not have
+  scales or orientations specified. (Issue #239)
+- Fixed issue in Houdini plugin where the "w" attribute was not converted
+  to the "angularVelocities" attribute for point instancer prims.
+	
 ## [0.7.6] - 2017-06-30
 
 ### Added
--- a/cmake/defaults/Version.cmake
+++ b/cmake/defaults/Version.cmake
@@ -23,5 +23,5 @@
 #
 # Versioning information
 set(PXR_MAJOR_VERSION "0")
-set(PXR_MINOR_VERSION "7")
-set(PXR_PATCH_VERSION "6")
+set(PXR_MINOR_VERSION "8")
+set(PXR_PATCH_VERSION "0")


third_party/katana/lib/usdKatana/readPoints.cpp
```diff
@@ -34,6 +34,49 @@ PXR_NAMESPACE_OPEN_SCOPE
...
 static FnKat::Attribute
+_GetPositionAttr(
+        const UsdGeomPoints& points,
+        const PxrUsdKatanaUsdInPrivateData& data)
+{
+    // Because of the logic used to gather these (in PxrUsdInOp::InitUsdInArgs),
+    // we can use the size of this vector as a simple test for whether motion
+    // blur is enabled, without needing access to the cook interface.
+    const std::vector<double>& motionSampleTimes =
+        data.GetUsdInArgs()->GetMotionSampleTimes();
```
spiffmon: There is actually important logic in PxrUsdKatanaUsdInPrivateData::GetMotionSampleTimes() beyond what the UsdInArgs version does.  I think you should call it here on the "points" attr.  That said, there is some craziness regarding IsMotionBackwards() that we are trying to work through right now, so that we can get the PointInstancer and your new code doing the right thing in both forwards and backwards motion blur (we do backwards here, but want to make sure not to foist that decision on everyone).  I'll contact you again when we get it sorted, which should hopefully be soon.

pxr/usd/lib/usdGeom/points.cpp
```diff
@@ -238,4 +238,82 @@ TF_REGISTRY_FUNCTION(UsdGeomBoundable)
         _ComputeExtentForPoints);
 }
 
+size_t
+UsdGeomPoints::ComputePositionsAtTimes(
+    std::vector<VtVec3fArray>& positions,
```
sirpalee: Thanks for the heads up. I'll keep this in mind for the next time.

pxr/usd/lib/usdGeom/points.cpp
```diff
@@ -238,4 +238,82 @@ TF_REGISTRY_FUNCTION(UsdGeomBoundable)
...
+                curr[i] = points[i] + velocities[i] * currentMultiplier;
+            }
+        }
+        return sampleCount;
+    } else {
+        if (!pointsAttr.Get(&positions[0], sampleTimes[0])) { return 0; }
+
+        const auto idsAttr = GetIdsAttr();
+        VtArray<long> ids;
+        if (!idsAttr.Get(&ids, sampleTimes[0])) { return 1; }
```
sirpalee: Sure. I'll remove the ID checks for now, and just use the built-in interpolation for points. I'll check for point lengths though. Also, this means I can move the function back to pointBased. (my original implementation was there, but I promoted it to points because of the extra id checks)

third_party/katana/lib/usdKatana/readPoints.cpp
sirpalee: Thanks. Let me know once it's done, and I'll update the change asap.

pxr/usd/lib/usdGeom/points.cpp
```diff
@@ -238,4 +238,82 @@ TF_REGISTRY_FUNCTION(UsdGeomBoundable)
...
+    if (!pointsAttr.GetBracketingTimeSamples(baseTime.GetValue(), &pointsLowerTimeSample,
+                                             &upperTimeSample, &hasValue) || !hasValue) {
+        return 0;
+    }
+
+    VtVec3fArray points;
+    VtVec3fArray velocities;
+
+    // We need to check if there is a queriable velocity at the given point,
+    // if it has almost the same lower time sample and the array lengths are equal.
```
sirpalee: Right, I forgot about the uses of a constant velocity field, or varying frequency. Updated the code to require both times to match.

third_party/katana/lib/usdKatana/readPoints.cpp
spiffmon: OK, I think it's actually pretty simple, given how you've structured the computation so nicely to take 'baseTime' as a separate input to the sample times.
```
const std::vector<double> motionSampleTimes = data.GetMotionSampleTimes(points.GetPointsAttr());

// Instead of consulting shutterOpen/shutterClose to build up sampleTimes, iterate through
// motionSampleTImes, adding currentTime to each element, since the fn returns frame-relative
// times
```

I plan to lobby to get this moved up to applying to all PointBased prims in Katana, but probably OK to leave it just on Points to start.

jtran56:
Filed as internal issue #147139.

third_party/katana/lib/usdKatana/readPoints.cpp
sirpalee: @spiffmon Done! Both using the points attr in the query and all the motion samples, not just the shutter open and close.

spiffmon:
Looks great, Pal - will get this on the queue for merging!

spiffmon:
Hi Pal,
   As one of our developers was looking to your pull request as inspiration for how to improve the readPointInstancer.cpp logic, we noticed there's an important case that's not being handled properly with respect to authored samples.  This probably happens more often for PointInstancers and with Points, but the case is that data for all attributes is only authored at UsdTimeCode::Default(), and we shouldn't penalize such assets from being able to use ComputePositionsAtTimes().

   Concretely (and this is why I didn't catch it at first), it means changing how you use the results of GetBracketingTimeSamples() - the last out parameter is not telling you there's no value, only that there are no timeSamples. What we're proposing is that in the case where points are authored but have no timeSamples, is to treat it exactly as if there were no velocity, as your "else" clause for "if (velocityExists)" already does exactly what we'd expect for this state of authored data (i.e. it will give you back N identical samples).

   Does that make sense?

Thanks!
--spiff

sirpalee:
Yes, absolutely! I'll get the PR updated soon, and keep this in mind for future PRs.

spiffmon:
OK, I promise, this is the **last** thing!  Can you add an optional, out parameter for returning the velocity vector (as VtArray) that you used, if any?  And then make readPoints use that value to populate the Points location's point.v attr?

Here's why:
   Our lighters often want to tweak the scale of the velocity of points, pointInstancers, etc, which is why I originally asked to add the velocityScale param to your computation.  Our fx guys have written their own velocityApply op for the lighters to use, which consumes the first position sample, and velocity, and replaces the rest of the P samples with ones they compute with user-specified velocityScale applied.  We would have loved to have this built into pxrUsdIn, but there's no good way to feed a user-specified velocityScale in.  But at least we can ensure that our "override" velocityAply op will be using the same, correctly sampled velocity vector that your code is computing.

Thanks for bearing with me, Pal!

sirpalee:
@spiffmon Updated the PR with both of the changes requested!

Let me know if you need anything else, eager to update the PR to make sure everybody is happy. :)

spiffmon:
Thanks, Pal!  I think the rest is on us: we are weighing two possibilities for how to continue to support authored velocityScale in Katana that is accessible to pxrUsdIn; once we have one we'll be able to pull in your readPoints.cpp code.

--spiff

sirpalee:
Sorry, pressed the wrong button.

PR #216, opened 2017-06-02
[katana] Use fileInput widget for PxrUsdIn filename

nrusch:
### Description of Change(s)

Adds a 'widget' hint to the Katana PxrUsdIn node so that a `fileInput` widget is used for the 'fileName' by default.


jtran56:
Filed as internal issue #147168.

sunyab:
Hi @nrusch, thanks for the PR! I'd like to close this out in favor of commit 283f715df9db1d21b61ac0d49231e153230a6433, which updated PxrUsdIn to use an assetIdInput widget. This seems preferable for the reasons mentioned in #218.

Please let us know if there's any reason you think we should reconsider. Thanks again!

nrusch:
@sunyab Nope, that all makes sense to me. Thanks.

PR #217, opened 2017-06-05
[usdKatan] Fixing installation directory for katana libs.

sirpalee:
Fixing a small typo in the Katana plugin installation target. It should be Libs, not Lib.


jtran56:
Filed as internal issue #147228.

sunyab:
This was merged in commit 2f7bfe3, thanks!

PR #218, opened 2017-06-05
Improve widget hints for PxrUsdIn.py.

aferrall:
### Description of Change(s)

Very small tweak to PxrUsdIn.py so that its "fileName" parameter uses an assetIdInput widget, allowing type-filtered file browsing.

### Fixes Issue(s)
-

### DIFF
--- a/third_party/katana/python/nodetypes/PxrUsdIn.py
+++ b/third_party/katana/python/nodetypes/PxrUsdIn.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2016 Pixar
+# Copyright 2016-2017 Pixar
 #
 # Licensed under the Apache License, Version 2.0 (the "Apache License")
 # with the following modification; you may not use this file except in
@@ -39,6 +39,8 @@ gb = FnAttribute.GroupBuilder()
 
 gb.set('fileName', '')
 nb.setHintsForParameter('fileName', {
+    'widget' : 'assetIdInput',
+    'fileTypes' : 'usd|usda|usdc|abc',
     'help' : 'The USD file to read.',
 })
 


aferrall:
I realize this is similar to pull request https://github.com/PixarAnimationStudios/USD/pull/216, but the "assetIdInput" provides greater flexibility compared with "fileInput", since it allows customization of the browser based on local Asset plugin implementations.  This is also consistent with existing file loaders like the shipping "Alembic_In" node.

jtran56:
Filed as internal issue #147290.

sunyab:
Hi @aferrall, thanks for the pull request! We ask all contributors to sign a CLA before contributing code to USD. Could you take a look at https://graphics.pixar.com/usd/docs/Contributing-to-USD.html and follow the instructions at the top of the page? Once we've received your CLA, we can have a look at merging in your changes. Thanks!

aferrall:
Hey @sunyab, I've actually already gone through that process and the ILM legal folks told me that everything is setup for the corporate CLA.  I assume that if you have an ILM CLA on file, you don't have my GitHub username associated with it.

Do I need that to come through the same formal channel which sent along the original CLA?  Thanks!

sunyab:
Ah, let me double check about the corporate CLA and get back to you -- I think if we've got that on file, you should be all set. In the meantime, I'll queue this change up for review.

sunyab:
Hi @aferrall, I'd like to close this out in favor of commit 283f715df9db1d21b61, which was coincidentally submitted shortly after your pull request and uses an assetIdInput widget like your change.

However, your change adds .abc files as a supported file type, while the above commit does not. I don't think we should unconditionally add this file type, since USD only supports .abc files if users have built the Alembic plugin. Ideally, we'd only add the file type in if the plugin were available, but I'm not sure how complicated it would be to do such a check.

Is having .abc files listed here important for your use case? If not, I propose we close this out in favor of the above commit. What do you think?

aferrall:
Hey @sunyab  - I'm sure we'll be able to make do without the .abc filetype option.  Thanks!

PR #223, opened 2017-06-15
[usdMaya] Adding an option to put everything into a parent scope.

elrond79:
### Description of Change(s)
This adds an option to the maya usd export, that will place all the created nodes under a user-named Scope.  It's useful because usd's referencing system works on a single top-level node - so if you're exporting a maya scene that has multiple top-level nodes, and want the entire scene to be easily referenced via USD, you need to place it all under one top node.
We're already using this in several places through our pipeline, and think other facilities would likely find it handy as well.

### Included Commit(s)
- https://github.com/PixarAnimationStudios/USD/commit/446b66a9ac2b90c7a7a44645a97d7f84f78b7405

### Fixes Issue(s)
- None, more of a feature addition


jtran56:
Filed as internal issue #147636.

sirpalee:
Note, the PR was updated to work with the recent exportParam changes for the shader exporter and I added tests.

PR #226, opened 2017-06-21
[usdMaya] Loading the shading mode plugins.

sirpalee:
This PR is a general improvement for the shading mode exporter plugin interface.

- Previously there was no functionality to load the plugins. I added a function similar to the primWriter plugin loading code, that's run before any of the exporters are accessed.
- The default values caused a bit of confusion. The GPrim Colors -> displayColor remapping was missing from the export code (both the translator and the usdExport call).
- A request to an invalid export appended empty exporter to the list with the same name. This problem was fixed by accessing the exporter through an iterator rather than std::map's operator[].
- If the manual export command was used instead of the file translator, the shading mode worked incorrectly. The shading modes were accessible by their internal names, which are not documented anywhere. I added support for the custom names on the file/export interface.

### DIFF
--- a/third_party/maya/lib/usdMaya/registryHelper.cpp
+++ b/third_party/maya/lib/usdMaya/registryHelper.cpp
@@ -40,6 +40,8 @@ PXR_NAMESPACE_OPEN_SCOPE
 TF_DEFINE_PRIVATE_TOKENS(_tokens, 
     (mayaPlugin)
     (providesTranslator)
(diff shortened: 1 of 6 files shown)


jtran56:
Filed as internal issue #147855.

sirpalee:
Note the original commit mixed up GPrim Colors and Material Colors shading mode, and exported them incorrectly. The PR was updated.

PR #227, opened 2017-06-21
[USD] Extending the special character support when parsing usda asset paths.

sirpalee:
The following change extends the list of supported characters in the asset path from usda files with !, <, >, = and +. These characters are common for rez users and interacting with rez installed directories are cumbersome otherwise.

Note: < and > is not supported on Windows. However, since the users can't create a path containing these characters anyway, I don't think it's worth having a separate parser for Windows.

I used Flex-2.5.39 and Bison-2.4.1, as requested in the CMake scripts.


jtran56:
Filed as internal issue #147856.

sunyab:
Commit 62ae544 extends the parser to allow asset paths in .usda files to contain any printable character, so this pull request shouldn't be needed anymore. I'm going to close this out, thanks!

sirpalee:
Thank you! This is even better.

PR #228, opened 2017-06-21
[usdMaya] Passing dagPathToUsdPathMap to the material translator.

sirpalee:
The following change passes the Maya DagPath to Usd Prim Path map to the shader exporters. This is useful when implementing the material export the other way around, looking up the materials of the existing nodes, rather than iterating through all the existing shading engines in the scene.

We are relying on this in our usd-arnold package.

### DIFF
--- a/third_party/maya/lib/usdMaya/shadingModeExporter.h
+++ b/third_party/maya/lib/usdMaya/shadingModeExporter.h
@@ -46,7 +46,8 @@ public:
     virtual void DoExport(const UsdStageRefPtr& stage,
                           const PxrUsdMayaUtil::ShapeSet& bindableRoots,
                           bool mergeTransformAndShape,
-                          const SdfPath& overrideRootPath);
+                          const SdfPath& overrideRootPath,
+                          const PxrUsdMayaUtil::MDagPathMap<SdfPath>::Type& dagPathToUsdMap);
 
     PXRUSDMAYA_API
     virtual void Export(const PxrUsdMayaShadingModeExportContext& context);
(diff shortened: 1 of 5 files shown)


jtran56:
Filed as internal issue #147857.

third_party/maya/lib/usdMaya/shadingModeExporter.h
```diff
@@ -46,7 +46,8 @@ class PxrUsdMayaShadingModeExporter {
     virtual void DoExport(const UsdStageRefPtr& stage,
                           const PxrUsdMayaUtil::ShapeSet& bindableRoots,
                           bool mergeTransformAndShape,
-                          const SdfPath& overrideRootPath);
+                          const SdfPath& overrideRootPath,
+                          PxrUsdMayaUtil::MDagPathMap<SdfPath>::Type& dagPathToUsdMap);
```
sunyab: Do we expect shading exporters to introduce new entries in the dagPathToUsdMap parameter? It didn't look like the usd-arnold code ever added entries to this map, for instance.

If we can make this const, that'd be great. If we need it to be non-const, could we change it to be a non-const pointer instead of a reference? That's our usual coding convention for parameters that may be modified.

sirpalee: Good catch! Yes, it should be const ref, I'm going to update the PR.

sirpalee:
@sunyab this was updated a while ago, but for some reason, it didn't show up as the changes were done. I also found something that could be changed to const&, done that too.

sunyab:
Hi @sirpalee, thanks for the update! I'm not sure why this pull request is listing 293 files changed, perhaps there are commits that need to be merged with your branch? In any case, we're ready to merge this one as soon as that's resolved.

sirpalee:
@sunyab I think I managed to export based on an older dev, accidentally. Looking at the changes now, it shows that we have +11 -6 lines changed. That looks right to me.

PR #229, opened 2017-06-22
[usdMaya] Clearing out the prim writers before saving the stage.

sirpalee:
The change is super simple here; we are clearing the prim writers before saving the stage. This change is useful when somebody wants to run cleanup code in a primWriter's destruction. Otherwise, there is no way to track down the last write to get the same effect.

For example, we have been using feature successfully with some of our writers to optimize animation data before saving the stage.

### DIFF
--- a/third_party/maya/lib/usdMaya/usdWriteJob.cpp
+++ b/third_party/maya/lib/usdMaya/usdWriteJob.cpp
@@ -382,6 +382,10 @@ void usdWriteJob::endJob()
         // prim for the export... usdVariantRootPrimPath
         mJobCtx.mStage->GetRootLayer()->SetDefaultPrim(defaultPrim);
     }
+    // Running post export function on all the prim writers.
+    for (auto& primWriter: mJobCtx.mMayaPrimWriterList) {
+        primWriter->postExport();
+    }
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #147936.

third_party/maya/lib/usdMaya/usdWriteJob.cpp
```diff
@@ -348,11 +348,13 @@ void usdWriteJob::endJob()
         // prim for the export... usdVariantRootPrimPath
         mJobCtx.mStage->GetRootLayer()->SetDefaultPrim(defaultPrim);
     }
+    // clear this so that no stage references are left around
+    // and running code in destructors
+    mJobCtx.mMayaPrimWriterList.clear();
```
sunyab: I think having an explicit hook would be preferred to doing stuff in a destructor here - I'd rather have an explicit call to some sort of "PostExport" callback (or virtual function on the MayaPrimWriter base class)

sirpalee:
@sunyab sorry for the late reply, but I just got the PR updated. I added a separate postExport function that runs before saving the stage.

Issue #211, opened 2017-05-08
install codegenTemplates


elrond79:
### Description of Issue
the codegenTemplates directory is not installed in the bin dir, so any time "usdGenSchema" is run, it needs to be pointed at a source-code checkout (or the directory needs to be manually copied)

This issue is referenced in https://github.com/PixarAnimationStudios/USD/issues/158, but I felt it deserved it's own issue.

### Steps to Reproduce
1. Do a default / fresh usd build + install
2. `cd` to a directory with a valid `schema.usd` file
3. Run `usdGenSchema`
4. It will error if the "-t/--templates" arg is not set, because the default templates are not installed in `${USD_INSTALL_ROOT}/bin/codegenTemplates`

meshula:
+1 This would for sure be a boon to people developing plugins, but relying on a distributed package.

jtran56:
Filed as internal issue #146333.

chadrik:
Are there any complicating factors to this issue or is it as simple as adding some `install` commands to cmake?

PR #231, opened 2017-06-26
[USD] Include codegenTemplates in usdGenSchema.

sirpalee:
### Description of Change(s)
The change includes codegenTemplates in usdGenSchema during the build process. If the codegenTemplates path does not exist, the code falls back to use the dictionary inside usdGenSchema.

This way external plugin devs can work with the installed usdGenSchema without manually copying codegenTemplates, and USD devs can keep working with usdGenSchema inside the source tree.

### Fixes Issue(s)
#211


jtran56:
Filed as internal issue #148278.

sunyab:
Hi @sirpalee, I'd like to close this PR out in favor of commit 043ebbfa7783a14d6575. We were discussing this during the week and we prefer this solution because it allows external devs to run usdGenSchema from a distributed build but doesn't require extra code transformations at build time.

Let us know if you think there's anything here we should reconsider. Thanks!

Issue #158, opened 2017-02-03
Installing Schemas.usda


dbaz:
Hi,
I'm trying to build my schemas library using USDv0.7.2 but it fails to build because the schemas.usda is not installed.

Could we get these base-type schemas.usda installed?

-Baz

jtran56:
Filed as internal issue #142730.

dbaz:
Also, it doesn't seem like codegenTemplates is installed.
I should really be putting PR's in for this kind of thing.
Thanks!

chadrik:
Hi all, has there been any movement on this issue?  If there has not, it would be great to start discussing how this *should* work so that we can possibly make a PR for it.

spiffmon:
Hi Chad,
     We spent a little time discussing it, and got hung up on how to address 
the fact that each schema.usda directly subLayers other schema.usda files.  
Since we don't want to require the src structure to be the same as the inst 
structure, and we still want to allow running usdGenSchema from source, it 
wasn't immediately obvious what the best approach was going to be, so this 
got pushed from "we can just knock it out" to "going to require some thought".

--spiff

chadrik:
Hmm... tricky problem.  What solutions have you come up with? 

Here's a very rough draft that comes to mind:

- create an environment variable for defining a schema search path
- give each schema.usda a unique name, e.g.. usdGeom.schema.usda
- use a custom asset resolver to find schemas on the search path.  e.g.
    ```
    #usda 1.0
    (
        "This file describes the USD Geometric schemata for code generation."
        subLayers = [
            @usd.schema.usda@
        ]
    )
    ```
- set the env var during the build process

spiffmon:
That's along the lines of one of the two approaches we'd discussed, and upon 
further discussion, seems like the more plausible one. Some notes:

  * The ArDefaultResolver should be up to the task - just need to ensure that
    you set its configuring env var PXR_AR_DEFAULT_SEARCH_PATH in
    usdGenSchema *before* causing an instance to get constructed and installed
  * Install the schema.usda files right alongside the generatedSchema.usda. 
    Within schema.usda files, other schema.usda files would be referenced
    like so: @usdGeom/resources/schema.usda@
  * You should be able to populate PXR_AR_DEFAULT_SEARCH_PATH from the
    (possibly slightly massaged) list of paths that PlugRegistry uses to
    search for plugins.  That's not currently exposed, but we approve of
    adding an accessor on PlugRegistry to return the already-held
[...]

PR #232, opened 2017-06-27
[USD] Installing built-in schemas.

sirpalee:
### Description of Change(s)
The following changes are included:
- Installing the schemas next to generatedSchema.usda.
- Massaging the installed schema to work well outside the usd source-tree.
- Adding a new function to PlugRegistry (and its python binding) that allows users to query all the search paths for plugins.
- Modifying usdGenSchema to pass these paths to the default resolver, making the external plugins easier to setup.

### Fixes Issue(s)
#158

### DIFF
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -98,6 +98,23 @@ function(_plugInfo_subst libTarget pluginToLibraryPath plugInfoPath)
     )
 endfunction() # _plugInfo_subst
 
+function(_replaceSublayer_schema NAME schemaSource)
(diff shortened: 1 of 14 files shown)


jtran56:
Filed as internal issue #148279.

sunyab:
Thank you for this PR! It was extremely useful in proving out the search path-based direction for solving this issue. We've just landed several changes to address #158 that we think should take care of this without requiring build-time modifications of the installed schema.usda files. 

I'm going to close this PR out in favor of those changes. Thanks again!

PR #233, opened 2017-06-30
[Katana] Support vec4 primvars.

sirpalee:
Supporting 4 component vector and 4 component vector array primvars in pxrUsdIn. Without this, there are a bunch of errors thrown, and with this, we can support RGBA color sets exported from Maya.


jtran56:
Filed as internal issue #148280.

PR #237, opened 2017-07-12
[usdAbc] Support abc camera clipping range property IO.

fabal:
### Description of Change(s)
Add read/write support for alembic camera clipping range.


jtran56:
Filed as internal issue #148584.

PR #243, opened 2017-07-21
Fix typos in comments.

waywardmonkeys:
This corrects a couple of typos in comments.


jtran56:
Filed as internal issue #148977.

PR #244, opened 2017-07-21
Fix mismatched tags warnings under clang.

waywardmonkeys:
Clang issues a warning when forward declarations don't match the
actual declaration about whether something is a struct or class.


jtran56:
Filed as internal issue #148978.

PR #245, opened 2017-07-21
Fix warnings about missing 'override'.

waywardmonkeys:
When some declarations in a class use 'override' but not all virtual
methods do, clang issues a warning.


jtran56:
Filed as internal issue #148979.

PR #246, opened 2017-07-21
_ExpandSymlinks is only used on Windows, so only build it there.

waywardmonkeys:
This fixes a warning when building on macOS.


jtran56:
Filed as internal issue #148980.

PR #247, opened 2017-07-21
Use enum of correct type for comparing geom style.

waywardmonkeys:
This fixes a warning when building on macOS.


jtran56:
Filed as internal issue #148981.

c64kernal:
Thanks very much, Bruce!  We caught this one a few days ago, the fix should be pushed up soon.

sunyab:
This was fixed as part of commit 92cd4ba4e8d031af786cc5d59a053cb02c126a3c, so I'm going to close this out. Thanks again!

PR #249, opened 2017-07-21
Remove unused private field: _fileBufferSize.

waywardmonkeys:
This fixes a warning about an unused private field on the class.


jtran56:
Filed as internal issue #148983.

PR #250, opened 2017-07-21
Remove unused malloc hook functions on non-Linux platforms.

waywardmonkeys:
This code is only used on Linux, so only build it there.


jtran56:
Filed as internal issue #148984.

PR #251, opened 2017-07-23
HdChangeTracker const correctness improvements.

waywardmonkeys:
This makes a few more methods `const` so that a reference to
the change tracker in the render pass code can also be `const`.


jtran56:
Filed as internal issue #149040.

PR #252, opened 2017-07-26
Export and install pxrTargets.cmake.

fabal:
### Description of Change(s)
Since v0.7.6 usd cmake's targets are not exported anymore. Was this intended? We're relying on this at the moment.
Cheers.
### Fixes Issue(s)
-

### DIFF
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -1299,13 +1299,29 @@ function(_pxr_library NAME)
     #
 
     if(NOT isObject)
-        install(
-            TARGETS ${NAME}
-            LIBRARY DESTINATION ${libInstallPrefix}
-            ARCHIVE DESTINATION ${libInstallPrefix}
-            RUNTIME DESTINATION ${libInstallPrefix}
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #149132.

sunyab:
Hi @fabal, sorry for the trouble and thanks for the pull request! The change to remove the targets was intentional since we couldn't generate the same file for the monolithic build, but we think your change is a better idea.

I noticed that your change doesn't restore the code that installs the pxrTargets.cmake file in pxr/CMakeLists.txt ([diff here](https://github.com/PixarAnimationStudios/USD/commit/3598908bcd6307cb0a6ca0f78657bf4563152b33#diff-ac4b93e42562fe6a5b8f3ca7367fa440)). So with your change, the export file will be generated but not installed. Was that intentional? It seems like we ought to bring that back too, with the same guards as in Private.cmake.

fabal:
Hi @sunyab, you're right I forgot to add it here simply because we're not using pxr/CMakeLists.txt internally to build usd.
Thanks for the heads up.

PR #269, opened 2017-08-23
Allow compilation against PySide or PySide2.

marsupial:
There are actually few changes necessary to get usdview working with Pyside2.

The majority of changes simply have to do with importing the proper [sub]modules for Qt5.
e6b7824 contains the source-level changes necessary, to support HDPI viewport, forwarding to newer method names, and create signal/slots in the sorted order that Qt5 requires.

### DIFF
--- a/cmake/modules/FindPySide.cmake
+++ b/cmake/modules/FindPySide.cmake
@@ -27,25 +27,42 @@ if (NOT PYTHON_EXECUTABLE)
     return()
 endif()
 
-find_program(PYSIDEUICBINARY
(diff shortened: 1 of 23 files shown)


jtran56:
Filed as internal issue #150149.

sunyab:
Hi @marsupial, thanks for this PR! We actually had a similar project in the works internally and this helped drive it forward. We wound up taking parts of your change but tweaked some naming and the logic in the FindPySide cmake module as well -- see commit ebe11b8.

I'm going to close out this PR in favor of that commit. Please let us know if you have any feedback, and thanks again!

PR #272, opened 2017-08-29
usdview: add "Copy Layer identifier" menu item

elrond79:
useful when using Assest Resolvers, where the "real path" may be less useful than the unresolved identifier

### Description of Change(s)
Adds in a new right-click menu item that will copy the layer identifier (ie, unresolved asset path) to the clipboard

### Fixes Issue(s)
- None


jtran56:
Filed as internal issue #150332.

elrond79:
Sorry, updated so the "Layer Stack" tab works now too!

PR #273, opened 2017-08-31
Pr/usd maya translator xformable use tf token

elrond79:
### Description of Change(s)

I noticed that the code for handling the xform operator stack import was doing a lot of string comparison, so I changed it over to use much faster TfToken comparison; almost all the information is already available as TfTokens (or can be made into new static token constants), so it should be a definite improvement.  And since we're now using TfTokens, hashmaps should now be fast, so I threw in an accelerator map lookup.

### Fixes Issue(s)
- Small speed gains - my tests show reproducible improvements of 1-2% when importing a file with ~4000 xforms.

### DIFF
--- a/third_party/maya/lib/usdMaya/testenv/testUsdImportXforms.py
+++ b/third_party/maya/lib/usdMaya/testenv/testUsdImportXforms.py
@@ -31,6 +31,7 @@ from maya import standalone
 
 import os
 import unittest
+import pprint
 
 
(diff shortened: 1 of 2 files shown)


elrond79:
For reference, here are the scripts I used for testing (adapted from the test I added):

https://gist.github.com/elrond79/b7b1a1a36331ce6d6148439a1999939c

jtran56:
Filed as internal issue #150395.

Issue #275, opened 2017-09-05
Maya usd exporter crashes on exporting instance


chloe5608:
### Description of Issue
When exporting a Maya scene with instances to usda, by selecting export all -> export instances, Maya crashes. 
### Steps to Reproduce
1. Load attached carInstances.ma into Maya ([carInstances.ma.zip](https://github.com/PixarAnimationStudios/USD/files/1278684/carInstances.ma.zip))
2. Select File->Export All, check export instances and make references instanceable
3. Crash

### System Information (OS, Hardware)
Mac OS 10.13 (High Sierra), Mac Pro
### Package Versions
Maya 2017
### Build Flags
USD 0.8

chloe5608:
Only check export instances is enough to trigger the crash.

jtran56:
Filed as internal issue #150465.

meshula:
FWIW, the crash log says: "Used null prim".

````
>	libusd.dll!pxrInternal_v0_8__pxrReserved__::Usd_IssueFatalPrimAccessError(const pxrInternal_v0_8__pxrReserved__::Usd_PrimData * p) Line 224	C++
 	libusd.dll!pxrInternal_v0_8__pxrReserved__::UsdAttribute::_CreateSpec(const pxrInternal_v0_8__pxrReserved__::SdfValueTypeName & typeName, bool custom, const pxrInternal_v0_8__pxrReserved__::SdfVariability & variability) Line 274	C++
 	libusd.dll!pxrInternal_v0_8__pxrReserved__::UsdAttribute::_Create(const pxrInternal_v0_8__pxrReserved__::SdfValueTypeName & typeName, bool custom, const pxrInternal_v0_8__pxrReserved__::SdfVariability & variability) Line 314	C++
 	libusd.dll!pxrInternal_v0_8__pxrReserved__::UsdPrim::CreateAttribute(const pxrInternal_v0_8__pxrReserved__::TfToken & name, const pxrInternal_v0_8__pxrReserved__::SdfValueTypeName & typeName, bool custom, pxrInternal_v0_8__pxrReserved__::SdfVariability variability) Line 306	C++
[...]

PR #280, opened 2017-09-08
[usdMaya] Fixing instancing when instances are below other instances.

sirpalee:
### Description of Change(s)
The PR fixes problems with the export instances feature for Maya. The issue happens if a transform is instanced while it has instances below itself, the collapsing code didn't check for the childCount, just for the numberOfInstancesBelow. This also caused transform issues in some other cases.

I also added an extra change to avoid possible crashes, when the xform can't be created.

I'm planning to extend the PR with tests soon, I just wanted to get this change up ASAP, so the reporter of #275 can pull in the change and fix their local issue.

### Fixes Issue(s)
#275


jtran56:
Filed as internal issue #150596.

PR #287, opened 2017-09-19
[maya] factor out matching of xforms into PxrUsdMayaXformStack

elrond79:
### Description of Change(s)
This factors out the code for defining standardized transform stacks into a new class, xformStack.  This class is also responsible for checking compatibility with existing usd xform stacks.  This code is also exposed through the API for use externally.

There's actually nothing in the code which is specific to maya, so in theory it could be promoted to somewhere in the usd core, but it may be premature if it's not used anywhere else yet... and, of course, I'm sure that would warrant a higher level of scrutiny.

Speaking of scrutiny - I'm planning to add some unit tests, but wanted to give a chance for comment now!

### DIFF
--- a/third_party/maya/lib/usdMaya/CMakeLists.txt
+++ b/third_party/maya/lib/usdMaya/CMakeLists.txt
@@ -59,6 +59,7 @@ pxr_shared_library(${PXR_PACKAGE}
         util
         usdPrimProvider
         writeUtil
+        xformStack
 
         MayaPrimWriter
         MayaTransformWriter
@@ -124,6 +125,7 @@ pxr_shared_library(${PXR_PACKAGE}
         wrapRoundTripUtil.cpp
         wrapStageCache.cpp
         wrapUserTaggedAttribute.cpp
+        wrapXformStack.cpp
 
     PYMODULE_FILES
         __init__.py
@@ -167,6 +169,7 @@ pxr_test_scripts(
         testenv/testUsdMayaProxyShape.py
         testenv/testUsdMayaReferenceAssemblyEdits.py
         testenv/testUsdMayaUserExportedAttributes.py
+        testenv/testUsdMayaXformStack.py
         testenv/testUsdMetadataAttributeConverters.py
         testenv/testUsdReferenceAssemblyChangeRepresentations.py
         testenv/testUsdReferenceAssemblySelection.py
@@ -566,3 +569,12 @@ pxr_register_test(testUsdReferenceAssemblySelection
         MAYA_DISABLE_CIP=1
 )
 
+pxr_register_test(testUsdMayaXformStack
+    CUSTOM_PYTHON "${MAYA_BASE_DIR}/bin/mayapy"
+    COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testUsdMayaXformStack"
+    TESTENV testUsdMayaXformStack
+    ENV
+        MAYA_PLUG_IN_PATH=${CMAKE_INSTALL_PREFIX}/third_party/maya/plugin
+        MAYA_SCRIPT_PATH=${CMAKE_INSTALL_PREFIX}/third_party/maya/share/usd/plugins/usdMaya/resources
+        MAYA_DISABLE_CIP=1
+)
--- a/third_party/maya/lib/usdMaya/testenv/testUsdMayaXformStack.py
+++ b/third_party/maya/lib/usdMaya/testenv/testUsdMayaXformStack.py
@@ -1,0 +1,1161 @@
+#!/pxrpythonsubst
+#
+# Copyright 2017 Pixar
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
+# 6. Trademarks. This License does not grant permission to use the trade
+#    names, trademarks, service marks, or product names of the Licensor
+#    and its affiliates, except as required to comply with Section 4(c) of
+#    the License and to reproduce the content of the NOTICE file.
+#
+# You may obtain a copy of the Apache License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
+#
+
+from pxr import Gf, Usd
+
+# XXX: The try/except here is temporary until we change the Pixar-internal
+# package name to match the external package name.
+try:
+    from pxr import UsdMaya
+except ImportError:
+    from pixar import UsdMaya
+
+from maya import cmds
+from maya.api import OpenMaya as OM
+from maya import standalone
+
+import os
+import unittest
+import pprint
+import itertools
+from collections import OrderedDict
+
+class testUsdMayaXformStack(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        standalone.initialize('usd')
+        cmds.loadPlugin('pxrUsd')
+
+    @classmethod
+    def tearDownClass(cls):
+        standalone.uninitialize()
+    
+    # Named so this runs first, so we can be reasonably sure
+    # that UsdGeom hasn't been imported yet - see below
+    # tests:
+    #   PxrUsdMayaXformOpClassification::GetOpType
+    def test00GetOpType(self):
+        translateOp = UsdMaya.XformStack.CommonStack().GetOps()[0]
+        # could import UsdGeom, and compare objects
+        # directly, but had a bug where if you didn't
+        # import UsdGeom, the wrappers for
+        # XformOp::Type weren't created yet... so test
+        # for this
+        self.assertEqual(
+            str(translateOp.GetOpType()),
+            'UsdGeom.XformOp.TypeTranslate')
+        # ok, we've tested the XformOp::Type wrapper,
+        # may as well make sure the actual objects
+        # line up
+        from pxr import UsdGeom
+        self.assertEqual(
+            translateOp.GetOpType(),
+            UsdGeom.XformOp.TypeTranslate)
+
+    # tests:
+    #   PxrUsdMayaXformOpClassification::GetName
+    #   PxrUsdMayaXformOpClassification::GetOpType
+    #   PxrUsdMayaXformOpClassification::IsInvertedTwin
+    #   PxrUsdMayaXformStack::MayaStack
+    #   PxrUsdMayaXformStack::GetOps    
+    def testMayaStack(self):
+        from pxr import UsdGeom
+        mayaStack = UsdMaya.XformStack.MayaStack()
+        self.assertEqual(
+            [(x.GetName(), x.IsInvertedTwin(), x.GetOpType()) for x in mayaStack.GetOps()],
+            [
+                ('translate',             False, UsdGeom.XformOp.TypeTranslate),
+                ('rotatePivotTranslate',  False, UsdGeom.XformOp.TypeTranslate),
+                ('rotatePivot',           False, UsdGeom.XformOp.TypeTranslate),
+                ('rotate',                False, UsdGeom.XformOp.TypeRotateXYZ),
+                ('rotateAxis',            False, UsdGeom.XformOp.TypeRotateXYZ),
+                ('rotatePivot',           True,  UsdGeom.XformOp.TypeTranslate),
+                ('scalePivotTranslate',   False, UsdGeom.XformOp.TypeTranslate),
+                ('scalePivot',            False, UsdGeom.XformOp.TypeTranslate),
+                ('shear',                 False, UsdGeom.XformOp.TypeTransform),
+                ('scale',                 False, UsdGeom.XformOp.TypeScale),
+                ('scalePivot',            True,  UsdGeom.XformOp.TypeTranslate),
+            ]
+        )
+            
+    # tests:
+    #   PxrUsdMayaXformOpClassification::GetName
+    #   PxrUsdMayaXformOpClassification::GetOpType
+    #   PxrUsdMayaXformOpClassification::IsInvertedTwin
+    #   PxrUsdMayaXformStack::CommonStack
+    #   PxrUsdMayaXformStack::GetOps
+    def testCommonStack(self):
+        from pxr import UsdGeom
+        commonStack = UsdMaya.XformStack.CommonStack()
+        self.assertEqual(
+            [(x.GetName(), x.IsInvertedTwin(), x.GetOpType()) for x in commonStack.GetOps()],
+            [
+                ('translate', False, UsdGeom.XformOp.TypeTranslate),
+                ('pivot',     False, UsdGeom.XformOp.TypeTranslate),
+                ('rotate',    False, UsdGeom.XformOp.TypeRotateXYZ),
+                ('scale',     False, UsdGeom.XformOp.TypeScale),
+                ('pivot',     True,  UsdGeom.XformOp.TypeTranslate),
+            ]
+        )
+        
+    # tests:
+    #   PxrUsdMayaXformOpClassification::GetName
+    #   PxrUsdMayaXformOpClassification::GetOpType
+    #   PxrUsdMayaXformOpClassification::IsInvertedTwin
+    #   PxrUsdMayaXformStack::MatrixStack
+    #   PxrUsdMayaXformStack::GetOps
+    def testMatrixStack(self):
+        from pxr import UsdGeom
+        matrixStack = UsdMaya.XformStack.MatrixStack()
+        self.assertEqual(
+            [(x.GetName(), x.IsInvertedTwin(), x.GetOpType()) for x in matrixStack.GetOps()],
+            [('transform', False, UsdGeom.XformOp.TypeTransform)]
+        )
+
+    # tests:
+    #   PxrUsdMayaXformOpClassification::IsCompatibleType
+    #   PxrUsdMayaXformStack::CommonStack        
+    #   PxrUsdMayaXformStack::FindOp
+    def testIsCompatibleType(self):
+        from pxr import UsdGeom
+        commonStack = UsdMaya.XformStack.CommonStack()
+        
+        translateOp = commonStack.FindOp('translate') 
+        for opType in UsdGeom.XformOp.Type.allValues:
+            if opType == UsdGeom.XformOp.TypeTranslate:
+                self.assertTrue(translateOp.IsCompatibleType(opType),
+                                "{} should be compatible with {}".format(
+                                    translateOp.GetName(),
+                                    opType))
+            else:
+                self.assertFalse(translateOp.IsCompatibleType(opType),
+                                "{} should not be compatible with {}".format(
+                                    translateOp.GetName(),
+                                    opType))
+                
+        rotateOp = commonStack.FindOp('rotate')
+        for opType in UsdGeom.XformOp.Type.allValues:
+            if str(opType).startswith('UsdGeom.XformOp.TypeRotate'):
+                self.assertTrue(rotateOp.IsCompatibleType(opType),
+                                "{} should be compatible with {}".format(
+                                    rotateOp.GetName(),
+                                    opType))
+            else:
+                self.assertFalse(rotateOp.IsCompatibleType(opType),
+                                "{} should not be compatible with {}".format(
+                                    rotateOp.GetName(),
+                                    opType))
+    
+    # tests:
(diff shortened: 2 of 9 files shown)


jtran56:
Filed as internal issue #151051.

elrond79:
Just pushed an update for this PR - mostly just adds python wrappers and a bunch of tests (though also renames one of the class aliases).

The commit in this PR is the consolidated version, if you want to see the incremental changes since the last push, you can see them here:

d66310552c88f217988cc67889e2b12a27f41bb9
14028e074da745ae2d34cc413f24c945ca434d0c
00b102e356119cbdf4469e15b359f32adf85a7b7
2148a40c9a5e94aada3e20e0c4e015832853eea5

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,385 @@
...
+    PXRUSDMAYA_API
+    inline static std::vector<OpClassPtr>
+    FirstMatchingSubstack(
+            const std::vector<UsdGeomXformOp>& xformops,
+            MTransformationMatrix::RotationOrder* MrotOrder=nullptr)
+    {
+        return std::vector<OpClassPtr>();
+    }
+
+   /// \brief Runs MatchingSubstack against the given list of stacks
```
sunyab: The variadic template stuff here is neat, but we had some concerns about it because of the duplicated code in the Python wrapping. We'd really like to find a way to avoid that if possible.

Are there performance concerns or other reasons why this method couldn't accept something like a vector<PxrUsdMayaXformStack>? I noticed PxrUsdMayaXformStack objects are non-copyable, but maybe we could relax that?

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,385 @@
...
+        {
+            return stackOps;
+        }
+        return FirstMatchingSubstack(xformops, MrotOrder, otherStacks...);
+    }
+
+private:
+    PxrUsdMayaXformStack(
+            const OpClassList ops,
+            const std::vector<std::pair<size_t, size_t> > inversionTwins,
```
sunyab: Did you mean to pass a const-ref in here instead?

third_party/maya/lib/usdMaya/testenv/testUsdMayaXformStack.py
```diff
@@ -0,0 +1,1123 @@
...
+# language governing permissions and limitations under the Apache License.
+#
+
+from pxr import Gf
+
+from maya import cmds
+from maya.api import OpenMaya as OM
+from maya import standalone
+
+from pxr import Usd, UsdMaya
```
sunyab: Could you change the UsdMaya import here to look like they do in other tests, like:
https://github.com/PixarAnimationStudios/USD/blob/496775539c324af7cb2ea12ddbb80720865b6cb2/third_party/maya/lib/usdMaya/testenv/testUsdMayaReferenceAssemblyEdits.py#L26-L31

This is unfortunately an internal build issue that affects the USD repository; we'd like to fix the issue so this won't be needed in the future, but in the meantime we'd appreciate the change here

third_party/maya/lib/usdMaya/CMakeLists.txt
```diff
@@ -164,6 +166,7 @@ pxr_test_scripts(
         testenv/testUsdMayaProxyShape.py
         testenv/testUsdMayaReferenceAssemblyEdits.py
         testenv/testUsdMayaUserExportedAttributes.py
+        testenv/testUsdMayaXformStack.py
```
sunyab: Thank you for adding a test!

third_party/maya/lib/usdMaya/CMakeLists.txt
```diff
@@ -548,3 +551,13 @@ pxr_register_test(testUsdReferenceAssemblySelection
...
 
+pxr_register_test(testUsdMayaXformStack
+    CUSTOM_PYTHON "${MAYA_BASE_DIR}/bin/mayapy"
+    COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testUsdMayaXformStack"
+    TESTENV testUsdMayaXformStack
+    ENV
+        MAYA_PLUG_IN_PATH=${CMAKE_INSTALL_PREFIX}/third_party/maya/plugin
+        MAYA_SCRIPT_PATH=${CMAKE_INSTALL_PREFIX}/third_party/maya/share/usd/plugins/usdMaya/resources
+        MAYA_DISABLE_CIP=1
+        MAYA_APP_DIR=./maya_profile
```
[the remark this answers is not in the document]
elrond79: Oops, that was accidentally included. I'll get rid of it.

third_party/maya/lib/usdMaya/xformStack.h
elrond79: Well, haven't done any benchmarking, but copying the entire stare object would end up copying multiple vectors, so I'd like to avoid it.  I could make it take a vector of pointers, but then I'd either have to have it take raw pointers, or switch PxrUsdMayaXformStack to use either weak or ref ptrs.

I guess right now I'm leaning toward just making PxrUsdMayaXformStack  derive from TfRefBase, switch everything to use refPtrs instead of refs, and then make FirstMatchingSubstack take a vector of those.

I'm also thinking I may just switch PxrUsdMayaXformOpClassification to use refPtrs, instead of weak ptrs... both for consistency with FirstMatchingSubstack, and because when I first chose weak ptrs, and because I've since realized that refPtrs are slightly lighter-weight (other than the initial copy from static memory to heap-allocated, but that will only be done once per stack).  Again, will have to update everything to use refptrs... but will probably simplify things, and don't have to make any assumptions about how things are allocated.

third_party/maya/lib/usdMaya/testenv/testUsdMayaXformStack.py
elrond79: Sure, easy fix...

third_party/maya/lib/usdMaya/xformStack.h
elrond79: Oops, yes, that should be a const-ref. Good catch.

third_party/maya/lib/usdMaya/xformStack.h
elrond79: So, after having found that you can't (or at least, it's difficult) to python-wrap const-ref-ptrs, I think I'm just going to make FirstMatchingSubstack take a vector of raw ptrs.  It feels slightly yucky to use raw pointers, but it also feels like overkill to introduce weak pointers purely so they can be used with this function.  Let me know if you'd prefer weak pointers, though (or any other approach), and I'll change it.

elrond79:
Addressed the code review notes, as well as converted over both classes to be "value types", that store a refPtr to the actual data. Since both are immutable classes, this seemed a good compromise between safety / performance (ie, refPtr is lighter weight than weakPtr) / and ease-of-python-wrapping.  Thanks to Alex Mohr for the suggestion!

Once again, this is the collapsed PR - the (new) incremental changesets can be seen here:

https://github.com/LumaPictures/USD/commit/728dfed8ff9bc91464194b03c4e63da3f26c4cf6
https://github.com/LumaPictures/USD/commit/711a54732569c5487b79373e5bc929d223ec528b
https://github.com/LumaPictures/USD/commit/40ad81be3cdede9ac346118772f0fd6bcb0edbe7
https://github.com/LumaPictures/USD/commit/c6db468b2243d0f089508ed883b8d4f0f563370d
https://github.com/LumaPictures/USD/commit/5aa4fa8c24c138ab639b80bc44acce46dcdbb998

sunyab:
Thanks! I've pulled these changes over again and am running them through builds now

sunyab:
Unfortunately this change is causing some internal test failures for usdExport. I'm looking at this further now, I will see if I can package up the test and send it your way to get your take on it.

elrond79:
Ok... so the test break is due to specifying rotateAxis using a single-axis (rotateX) xformOp.  I had set rotateAxis to only accept rotateXYZ ops... because, unlike the "rotate" xform, it has no support for changing the rotation order (and, in fact, under the hood it's stored as a quaternion anyway!)... but it should at least be able to accept single-axis rotates, if not things like rotateZXY.

However... on thinking about it a bit more, I think this highlights one of the current shortcomings of my design - the maya-centric way in which it's handling rotation order.  Granted, it's currently only used for Maya... but it was designed in a way that should be usable by other applications.  For rotation order, though, it assumes that is set based off of the "rotate" op... and in fact, it sets a maya-only class (MTransformationMatrix::RotationOrder) to indicate what it what set to.

In a more general case, it's entirely possible for an xform stack to have 2 (or more) rotation ops, all of which may have their own independent rotation orders... so I think I'm going to get rid of handling of rotation-order as part of the API.  Instead, the paradigm will be:

- any 3-axis rotate in a stack is compatible with any-other 3-axis rotate (or 1-axis rotate)... since, at minimum, it is always possible to convert between them
- it will be the responsibility of the user of the API to check the rotate-order of the matched xforms, and either set a rotate order in their app, or convert, as appropriate to their context
- therefore, the optional RotationOrder* pointer will be dropped from the API for MatchingSubstack / FirstMatchingSubstack

I should have these changes (and some more tests) by EOD

spiffmon:
Hi Paul,
   Apologies as I haven't reviewed the code, but based on the last set of comments about rotationOrder, I just wanted to confirm that it will still be possible, using the new API, to encode triple-axis rotations in USD, as they will be three times faster to read than three single-axis, which adds up at scale. 

--Spiff

elrond79:
So, made the changes I mentioned above.  rotateAxis can now handle single-axis rotations, as well as will automatically cross-convert from non-XYZ-ordered triple-axis rotates.

@spiffmon absolutely, triple-axis rotates are still recorded as such.  I just meant that if you have a USD file that records it's rotateAxis as rotateYZX, this new version will automatically convert that to rotateXYZ, which is the only format that transform nodes will accept as input.  It means that the maya stack is now slighly more tolerant (though an xformOp:rotateYZX:rotateAxis will not round-trip - ie, it will import fine, and maintain the same transformation / rotation, but when you re-ouput a new USD, it will be xformOp:rotateXYZ:rotateAxis).

elrond79:
Oh, also, in case you want to see incremental commits:

https://github.com/LumaPictures/USD/commit/9bb528be373d4d044b5a394a90e65b6daa21588a
https://github.com/LumaPictures/USD/commit/3215095c5988979fcdf36955b1f0f85f8ccd73e7

third_party/maya/lib/usdMaya/testenv/testUsdMayaXformStack.py
```diff
@@ -0,0 +1,1161 @@
+#!/pxrpythonsubst
+#
+# Copyright 2016 Pixar
```
sunyab: 2017 :)

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+            break;
+            default:
+                return defaultRotOrder;
+            break;
+        }
+    }
+
+    PxrUsdMayaXformStack(
+            const OpClassList& ops,
+            const std::vector<std::pair<size_t, size_t> >& inversionTwins,
```
sunyab: Should this be `const std::vector<IndexPair>& ...`?

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+    // Don't want to accidentally make a copy, since the only instances are supposed
+    // to be static!
+    explicit PxrUsdMayaXformStack(const PxrUsdMayaXformStack& other) = default;
+    explicit PxrUsdMayaXformStack(PxrUsdMayaXformStack&& other) = default;
+
+    PXRUSDMAYA_API
+    OpClassList const & GetOps() const;
+
+    PXRUSDMAYA_API
+    std::vector<std::pair<size_t, size_t> > const & GetInversionTwins() const;
```
sunyab: Should we use `const std::vector<IndexPair>& ...` here as well?

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+
+    PXRUSDMAYA_API
+    size_t GetSize() const;
+
+
+    /// \brief  Finds the index of the Op Classification with the given name in this stack
+    /// \param  opName the name of the operator classification we  wish to find
+    /// \param  isInvertedTwin the returned op classification object must match
+    ///         this param for it's IsInvertedTwin() - if an op is found that matches
+    ///         the name, but has the wrong invertedTwin status, nullptr is returned
```
sunyab: Looks like this should be "NO_INDEX is returned" instead of "nullptr is returned"

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+    /// return  Undex to the op classification object with the given name (and
```
sunyab: Minor typo: "Undex" instead of "Index"

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+    ///         inverted twin state); will be NO_INDEX if no match could be found.
+    PXRUSDMAYA_API
+    size_t FindOpIndex(const TfToken& opName, bool isInvertedTwin=false) const;
+
+    /// \brief  Finds the Op Classification with the given name in this stack
+    /// \param  opName the name of the operator classification we  wish to find
+    /// \param  isInvertedTwin the returned op classification object must match
+    ///         this param for it's IsInvertedTwin() - if an op is found that matches
+    ///         the name, but has the wrong invertedTwin status, nullptr is returned
+    /// return  Pointer to the op classification object with the given name (and
```
sunyab: Does this return OpClass::NullInstance instead of a pointer?

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+    ///         returned (in the order they appear in this stack). If found, but
+    ///         not as part of an inverted twin pair, the first result will point
+    ///         to the found classification, and the second will be NO_INDEX.  If
+    ///         no matches are found, both will be NO_INDEX.
+    PXRUSDMAYA_API
+    const IndexPair& FindOpIndexPair(const TfToken& opName) const;
+
+    /// \brief  Finds the Op Classification(s) with the given name in this stack
+    /// \param  opName the name of the operator classification we  wish to find
+    /// return  A pair of pointers to op classification objects with the given name;
```
sunyab: Same here -- returns pair of OpClass::NullInstance?

third_party/maya/lib/usdMaya/xformStack.h
```diff
@@ -0,0 +1,315 @@
...
+    ///         if the objects are part of an inverted twin pair, then both are
+    ///         returned (in the order they appear in this stack). If found, but
+    ///         not as part of an inverted twin pair, the first result will point
+    ///         to the found classification, and the second will be null.  If
+    ///         no matches are found, both pointers will be nullptr.
+    PXRUSDMAYA_API
+    const OpClassPair FindOpPair(const TfToken& opName) const;
+
+    /// \brief Returns a list of pointers to matching XformOpDefinitions for this stack
```
sunyab: Doc for returned value needs to be updated, since it's no longer a list of pointers

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+                || opType == UsdGeomXformOp::TypeRotateYZX
+                || opType == UsdGeomXformOp::TypeRotateZXY
+                || opType == UsdGeomXformOp::TypeRotateZYX
+                || opType == UsdGeomXformOp::TypeRotateX
+                || opType == UsdGeomXformOp::TypeRotateY
+                || opType == UsdGeomXformOp::TypeRotateZ);
+    }
+
+    IndexMap _buildInversionMap(
+            const std::vector<std::pair<size_t, size_t> >& inversionTwins)
```
sunyab: `const std::vector<IndexPair>&` here too?

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+// linker symbol...
+constexpr size_t PxrUsdMayaXformStack::NO_INDEX;
+
+class _PxrUsdMayaXformStackData : public TfRefBase
+{
+public:
+    typedef TfRefPtr<_PxrUsdMayaXformStackData> RefPtr;
+
+    RefPtr static Create(const OpClassList& ops,
+            const std::vector<std::pair<size_t, size_t> >& inversionTwins,
```
sunyab: `const std::vector<IndexPair>& ...` here?

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+            bool nameMatters)
+    {
+        return RefPtr(new _PxrUsdMayaXformStackData(
+                ops, inversionTwins, nameMatters));
+    }
+
+    _PxrUsdMayaXformStackData(
+            const OpClassList& ops,
+            const std::vector<std::pair<size_t, size_t> >& inversionTwins,
```
sunyab: `const std::vector<IndexPair>& ...` here?

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+                        ops, inversionTwins, nameMatters))
+{
+}
+
+OpClassList const &
+PxrUsdMayaXformStack::GetOps() const {
+    return _sharedData->m_ops;
+}
+
+std::vector<std::pair<size_t, size_t> > const &
```
sunyab: `const std::vector<IndexPair>& ...` here?

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+            // Inverted twin pairs will always have same name, so can skip one
+            if (op.IsInvertedTwin()) continue;
+
+            for (auto attrName : op.CompatibleAttrNames())
+            {
+                auto indexPair = _makeInversionIndexPair(i, inversionMap);
+                // Insert, and check if it already existed
+                if (!result.insert({attrName, indexPair}).second)
+                {
+                    // The attrName was already in the lookup map...
```
sunyab: Our coding convention is to not use exceptions in our code, so we'd prefer not to throw here. If this is something that should never happen (i.e., it's an invariant in the code and not reliant on user input), we prefer to use TF_VERIFY to indicate that something is wrong, like:

```
TF_VERIFY(result.insert({attrName, indexPair}).second, 
          "AttrName %s already found in attrName lookup map", 
          attrName.GetText());
```
This will cause the error output to be emitted if the condition is false, but allows the code to continue executing so that it doesn't cause the program to crash.

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+            auto indexPair = _makeInversionIndexPair(i, inversionMap);
+            // Insert, and check if it already existed
+            if (!result.insert({op.GetName(), indexPair}).second)
+            {
+                // The op name was already in the lookup map...
+                // ...this shouldn't happen, error
+                std::string msg = TfStringPrintf(
+                        "Op classification name %s already found in op lookup map",
+                        op.GetName().GetText());
+                throw std::invalid_argument(msg);
```
sunyab: See note about exceptions and TF_VERIFY above

third_party/maya/lib/usdMaya/xformStack.cpp
```diff
@@ -0,0 +1,748 @@
...
+                            _buildOpNamesToIdxs(m_ops, m_inversionMap)),
+                    m_nameMatters(nameMatters)
+    {
+        // Verify that all inversion twins are of same type, and exactly one is marked
+        // as the inverted twin
+        for (auto& pair : m_inversionTwins)
+        {
+            const PxrUsdMayaXformOpClassification& first = m_ops[pair.first];
+            const PxrUsdMayaXformOpClassification& second = m_ops[pair.second];
+            if (first.GetName() != second.GetName())
```
sunyab: See note about exceptions and TF_VERIFY above

elrond79:
Ok, addressed the notes. Incremental commit can be found here:

https://github.com/LumaPictures/USD/commit/bf06845daf91b21e4511dceb25a2dd68b07df3b9

Also made one other small tweak - marking the python wrapper for XformStack as noncopyable:

https://github.com/LumaPictures/USD/commit/cdac1b80ed82b9bca1374756028e9c94a4354fa7

sunyab:
Just wanted to note I had added a follow-up commit ddd583f55ad8b467c62e3b3d01ceee3859f70a0f to fix a build issue on Windows. I also had to remove the noncopyable change to the XformStack's Python wrapping, as it caused the associated unit test to fail.

elrond79:
Doh - my mistake.  Completely forgot that I had switched to the shared-data-ref-ptr paradigm, and that copies were now ok / preferred!  In any case, thanks!

PR #297, opened 2017-10-04
[usdKatana] Allowing custom parent scope names for Katana materials.

sirpalee:
### Description of Change(s)
The change allows any parent scope names for materials. Previously it had to be Looks. The behavior can be controlled via an environment variable, and it's turned off by default to maintain current behavior.

### Fixes Issue(s)
None reported.


chadrik:
`USD_KATANA_ALLOW_CUSTOM_SCOPE_NAMES_FOR_MATERIALS` env var is a little wordy. How about  `USD_KATANA_ALLOW_CUSTOM_MATERIAL_SCOPES`?   Does anyone have a better suggestion?

jtran56:
Filed as internal issue #151978.

sirpalee:
@chadrik Good one! I replaced the env name with your suggestion.

PR #302, opened 2017-10-12
Allow compiling the Houdini plugin against either Houdini 16.0 or Houdini 16.5.

marktucker:
Changes to allow compiling the Houdini plugin against either Houdini 16.0 or Houdini 16.5.

Also fixed a couple of other minor issues related to building this plugin under Windows: removing the use of "not" in place of "!", added an include, and used a lambda to call pclose() in shaderWrapper.cpp because MSVC wouldn't accept pclose() directly as the destructor function for the FILE * shared pointer.


marktucker:
Sorry, just realized I did this on master instead of dev branch...

PR #303, opened 2017-10-12
Allow compiling the Houdini plugin against either Houdini 16.0 or Houdini 16.5.

marktucker:
Trying this pull request again, hopefully with the right branches this time...

Changes to allow compiling the Houdini plugin against either Houdini 16.0 or Houdini 16.5.

Also fixed a couple of other minor issues related to building this plugin under Windows: removing the use of "not" in place of "!", added an include, and used a lambda to call pclose() in shaderWrapper.cpp because MSVC wouldn't accept pclose() directly as the destructor function for the FILE * shared pointer.


jtran56:
Filed as issue #152348.

marktucker:
I pulled in the latest changes published to dev, and made a few minor additions to my changes. Primarily changed the use of UT_ScopedPtr (which is deprecated) to UT_UniquePtr. Added a virtual destructor to NameMatcher because not having it there was causing compile errors for me on Linux (even though it seems no NameMatcher object is ever instantiated anywhere).

PR #307, opened 2017-10-17
[usdHoudini] Explicitly linking against Houdini libs

sirpalee:
### Description of Change(s)
The change explicitly links OP_gusd.so against Houdini libraries. Without this, we observed crashes in MPlay and hython. Those two apps are not loading all the required libraries like the main application does.

### Fixes Issue(s)
None reported.


jtran56:
Filed as internal issue #152498.

third_party/houdini/lib/gusd/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ pxr_shared_library(${PXR_PACKAGE}
         usdRi
         usdShade
         usdUtils
+        ${HOUDINI_OPZ_LIBRARY}
+        ${HOUDINI_UI_LIBRARY}
```
sunyab: I wasn't able to find any documentation about what the various Houdini .so's provide, but our FX team pointed out that the Makefile included in the Houdini dev kit under toolkit/makefiles/Makefile.linux links against:

`-lHoudiniUI -lHoudiniOPZ -lHoudiniOP3 -lHoudiniOP2 -lHoudiniOP1 -lHoudiniSIM -lHoudiniGEO -lHoudiniPRM -lHoudiniUT`

Should we do the same, even if OPZ and UI happen to be sufficient for now?

cmake/modules/FindHoudini.cmake
```diff
@@ -78,6 +78,35 @@ find_path(HOUDINI_LIB_DIRS
...
+            Houdini${HOUDINI_LIB}
+        HINTS
+            "${HOUDINI_LIB_DIRS}"
+        DOC
+            "Houdini's ${HOUDINI_LIB} library path"
+        NO_CMAKE_SYSTEM_PATH
+    )
+
+    if (HOUDINI_${HOUDINI_LIB}_LIBRARY)
+        list(APPEND HOUDINI_LIBRARIES ${HOUDINI_${HOUDINI_LIB}_LIBRARY})
```
sunyab: Could we add the new outputs to the "Out:" documentation at line 42?

third_party/houdini/lib/gusd/CMakeLists.txt
sirpalee: Sounds good. That would make it development more future proof, especially if we start relying on more API functions. Loading libraries and such outside houdini is more of a rare thing, so it's unlikely it would be tested by houdini devs.

cmake/modules/FindHoudini.cmake
sirpalee: Sure.

sirpalee:
@sunyab Updated PR with the requested changes. Please let me know if you need anything else!

third_party/houdini/lib/gusd/CMakeLists.txt
```diff
@@ -14,6 +14,15 @@ pxr_shared_library(${PXR_PACKAGE}
...
         usdShade
         usdUtils
+        ${HOUDINI_UI_LIBRARY}
+        ${HOUDINI_OPZ_LIBRARY}
+        ${HOUDINI_OP3_LIBRARY}
+        ${HOUDINI_OP2_LIBRARY}
+        ${HOUDINI_OP1_LIBRARY}
+        ${HOUDINI_SIM_LIBRARY}
+        ${HOUDINI_GEO_LIBRARY}
+        ${HOUDINI_RPM_LIBRARY}
```
sunyab: One typo here -- I think this should be HOUDINI_PRM_LIBRARY?

sirpalee:
@sunyab Ups! Fixed typo.

PR #317, opened 2017-10-27
provide a way to override the usd imaging scene delegate

robthebloke:
### Description of Change(s)
This change provides a way for clients of UsdImaging to use a derived UsdImagingDelegate. We need to make this change to allow the AL_USDMayaPlugin to provide its own scene delegate to allow us to transparently map OpenGL handles from Maya/VP2 _(e.g. file textures, image planes, results from previous render passes, etc, etc)_ through to the GLSL shaders used within UsdHydra. We have a prototype of this up an running, and would like to start rolling this out to the public branch of AL_USDMaya. 

In effect, it allows us to set up some additional data within our USD files, e.g.

```
    def Shader "imagePlaneTest"
    {
        uniform token info:id = "HwUvTexture_1"
        color3f outputs:colour
        custom string mayakind = "imagePlane"
        custom string mayanode = "imagePlaneShape1"
        float2 imgp_offset = (0, 0)
        float2 imgp_scale = (0.5, 0.5)
        float imgp_rotate = 0
    }
   
    def Shader "fileTextureTest"
    {
        rel connectedSourceFor:uv = </material/uvcoords.outputs:result>
        uniform token info:id = "HwUvTexture_1"
        color3f outputs:colour
        float2 uv
        custom string mayakind = "fileTexture"
        custom string mayanode = "file1"
    }
```

Which are intercepted by our custom UsdImagingDelegate, and provides some customised TextureResource classes to map the maya VP2 data into UsdImaging, e.g. 

```c++
GLuint MayaFileTextureResource::GetTexelsTextureId()
{
  MHWRender::MRenderer* const renderer = MHWRender::MRenderer::theRenderer();
  MHWRender::MTextureManager* const textureMgr = renderer ? renderer->getTextureManager() : nullptr;
  const MHWRender::MTexture* texture = 0;

  if (textureMgr)
  {
    const MHWRender::MTexture* const texture = textureMgr->acquireTexture(m_fileTexture, true);
    if(texture)
    {
      const void* const handle = texture->resourceHandle();
      if(handle)
      {
        return *((const GLuint*)handle);
      }
    }
  }
  return 0;
}
```

### Fixes Issue(s)
- Provides a way for users of UsdImaging to override the behaviour of the imaging delegate used internally within UsdImaging.

### DIFF
--- a/pxr/usdImaging/lib/usdImaging/delegate.h
+++ b/pxr/usdImaging/lib/usdImaging/delegate.h
@@ -434,6 +434,11 @@ public:
     USDIMAGING_API
     bool IsInInvisedPaths(const SdfPath &usdPath) const;
 
+protected:
+  
+    USDIMAGING_API
+    UsdStageRefPtr GetStage() { return _stage; }
+    
 private:
     // Internal friend class.
     class _Worker;
--- a/pxr/usdImaging/lib/usdImagingGL/hdEngine.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/hdEngine.cpp
@@ -60,7 +60,8 @@ UsdImagingGLHdEngine::UsdImagingGLHdEngine(
         const SdfPath& rootPath,
         const SdfPathVector& excludedPrimPaths,
         const SdfPathVector& invisedPrimPaths,
-        const SdfPath& delegateID)
+        const SdfPath& delegateID,
+        UsdImagingDelegateCreatorFunc delegateCreatorFunc)
     : UsdImagingGLEngine()
     , _renderIndex(nullptr)
     , _selTracker(new HdxSelectionTracker)
@@ -77,6 +78,7 @@ UsdImagingGLHdEngine::UsdImagingGLHdEngine(
     , _invisedPrimPaths(invisedPrimPaths)
     , _isPopulated(false)
     , _renderTags()
+    , _delegateCreatorFunc(delegateCreatorFunc)
 {
     // _renderIndex, _taskController, and _delegate are initialized
     // by the plugin system.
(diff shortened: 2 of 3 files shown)


jtran56:
Filed as internal issue #153071.

meshula:
This looks very handy :)

davidgyu:
Hi Rob,

Instead of specializing the UsdImagingDelegate, our plan to support data access like this it to expose a way to extend the handling of custom shader types via adapters.

Within UsdShade, the "info:id" attribute designates the behavior of the Shader. So, while "HwUvTexture_1" means to read a texture from the filesystem, additional shader/texture types can be defined to support mapping onto Maya image planes, render target attachment textures, etc.

Unfortunately, this isn't quite ready to go, but this is active work in progress within the Hydra core team.  How urgent is the need for this in AL_UsdMaya?

Thanks!
-David

murphyeoin:
Hi David, many apologies on behalf of the AL_USDMaya team for not getting back to you on this, it slipped through the cracks.
The internal use case for this work was to support camera image plane projection texturing onto geometry. At AL, we can of course keep this  as a patch on top of USD, but we were hoping to push a chunk of work to our opensource repository to support the overall goal of better Hydra/VP2 integration - so I suppose the answer partially depends on who else out there needs this functionality, and what your approximate timeframe is for the adapter solution.
Thanks a lot,
Eoin

robthebloke:
Hi, bringing this PR up again, because we've hit a brick wall. Currently there is still no way to access or override the behaviour of the UsdImagingDelegate that is buried within the UsdImagingGLEngine class (which is what this PR solves). This means we can't add any custom rendering passes to support Shadows/SSAO/Dof for Maya VP2 (well, we could drop UsdImaging in favour of using Hydra directly I suppose, but that would seem to be a little self defeating)

davidgyu:
Hi Rob and Eoin, 

We discussed this PR again today, specifically with respect to the use cases (Shadows, SSAO, DoF, etc) that you mentioned.  In our applications we've been thinking of these as part of the Task system rather than as extensions of our UsdImaging scene delegate operation (see for example the recent improvements to HdxTaskController to support Shadows, OIT, AOVs, color correction, etc).

As you can see, there's not much remaining in UsdImagingGLEngine, since we've now implemented most of the rendering using HdxTaskController. You can do this as well. This would be our recommended way to use your own scene delegate, to implement custom rendering tasks, etc.

Thanks!

PR #318, opened 2017-10-30
Auto deleter for pointers stolen from a UT_String needs to be a callback deleter

marktucker:
This misuse of PRM_AutoDeleter was caught by running through an address sanitizer which pointer out the use of malloc to allocate and delete to free the stolen string.

The GA_Offset casting is from a build of Houdini where GA_Offset is a class, not a simple typedef to int. Often this catches bad code. The code in GU_Usd.cpp actually looks a bit suspicious, but I can't find any evidence it is actually called, so perhaps this function should be deleted?


jtran56:
Filed as internal issue #153072.

PR #319, opened 2017-10-30
Fix a number of minor warnings in the Houdini plugin code.

marktucker:
Some typos in include guards.
Adding override keyword consistently on virtual overrides.
Placing assignments in if() expressions inside parentheses.
Removing unuse m_purposeSet member data.
Pre-declare a class as a class, not as a struct.

These warnings were all reported by clang.


jtran56:
Filed as internal issue #153073.

PR #320, opened 2017-11-02
[maya] fix for exporting orientation attribute for doubleSided meshes

nxkb:
### Description of Change(s)
If a mesh is "doubleSided", do not export an "opposite" attribute. The "opposite" attribute is disabled and disregarded in maya, so it should not be set in usd. (Before "opposite" was getting exported regardless of "doubleSided".)

### Fixes Issue(s)
This is a fix for getting different sidedness across DCCs.

When a mesh is doubleSided, maya ignores the "opposite" sidedness, and treats the mesh the same as if it were unchecked. However, once you've exported the mesh to USD it would pick up on the disabled checkboxes state and will give you a different result. Other DCC's that consume the USD file will then have a different picture of the mesh than existed in maya.


jtran56:
Filed as internal issue #153300.

PR #322, opened 2017-11-08
disable testUsdExportRenderLayerMode on maya-2018

elrond79:
### Description of Change(s)
disable testUsdExportRenderLayerMode on maya-2018

### Fixes Issue(s)
- These tests will always fail on Maya 2018, since old-style render layers are disabled


jtran56:
Filed as internal issue #153508.

sunyab:
Hi @elrond79, we'd like to close out this pull request in favor of commit f7eeac9ed58e0eebb43a49a3c90289de1a7f985c, which should keep this test running for now. Let me know if you feel we should go in a different direction, thanks!

elrond79:
Sure, looks good to me!

Issue #126, opened 2016-12-22
maya plugin: support for stripping namespaces in "usdExport"


nxkb:
Hello! I wanted to poke in on this issue which will impact a lot of people's Maya integration and guage the interest in a pull request to address it. 
### Description of Issue
When exporting usd out of maya, we will need to be able to strip a variable number of namespaces on referenced nodes. Currently they become part of the prim name, which makes it problematic later on to figure out whether they were originally namespaces or actually part of the node names. I believe this feature is pretty critical for studios that are doing animation in maya...

If you start with a referenced rigg (that also contains another reference), this is your hierarchy in maya:
```
wolf1:werewolf_top
 - wolf1:def:werewolf_mesh
```
and you export usd:
```
 wolf1_werewolf_top
 - wolf1_def_werewolf_mesh
```
This makes things tricky... later operations like name based shader assignment will have to work around this extra prefix on _every_ prim. So to match the node, you would need a rule like:
```*/*_werewolf_top/*_def_werewolf_mesh```

However, this would be the hierarchy with support for stripping (1) level of namespaces:
```
 werewolf_top
 - def_werewolf_mesh
```
which means you could have a much simpler expression to match the same node (for all your werewolves!):
```*/werewolf_top/def_werewolf_mesh```

One approach to solve this would be to add a custom export Chaser that adjusts exporting node names to strip a certain number of namespaces from the prim name. However, this behavior may be common enough to warrant something more built-in than just a custom Chaser. I'm sure many users would like to skip the setup of a chaser to get this behavior. 

So I'm wondering:
Are there any efforts already under way to handle this issue? Would you suggest adding support to the maya plugin directly or going through the chaser plugin system? Would we be able to contribute this back to the USD repo? 

Thanks!
-- nick kb

jtran56:
Filed as internal issue #141438.

spiffmon:
Hi Nick,
   We'd like to first investigate whether we can have full roundtripping support for Maya namespaces, i.e. something like, preserving in USD prim metadata, on **export**, the actual Maya nodename and namespaces; then on **import**, for any prim that has that metadata, record on the Maya node the USD prim name, which would in turn be used on re-export rather than re-computing a new USD name from the Maya name (pros and cons of that behavior... not sure... but the initial export/import seems reasonable).

Would that address your needs, or would there still be a need/desire for selective stripping of Maya namespaces?

nxkb:
Hi Spiff,
[...]

PR #327, opened 2017-11-16
[maya] add stripNamespaces flag to usdExport

nxkb:
This PR adds the option to strip maya namespaces from resulting prim names in the usdExport maya command.

### Description of Change(s)
If you have a node in maya named: `|root_node|one:two:nodename`
the default name for the exported prim is: `/root_node/one_two_nodename` (unchanged)
But if this new flag is provided the prim name will be: `/root_node/nodename`

If any dag nodes would result in the same prim path after stripping namespaces, an error will be raised.

### Fixes Issue(s)
issue #126:
- Often in maya, artists use references with different namespaces. Without this flag, there is no way to get exports to have the same hierarchy when exported. This causes problems later on for things like material assignment.
- This also allows exporting an identical hierarchy from a maya scene (rig file) and a referenced version of that file (animation file with referenced rig) so that geometry will be accessible at the same prim path.

### Limitations
This is a scaled back version of some initial designs that nonetheless satisfies our current needs, but here are some of the missing pieces that could be added later. 
1.) ability to strip a specific number of namespaces rather than all or none
2.) storing namespaces as prim metadata to allow reconstituting maya namespace for roundtripping

### Justification
We have found maya namespaces to be redundant when using USD. They are an opposite approach to the same problem: where maya inserts extra pieces to names in every level of a hierarchy to keep node names unique, USD solves it by renaming a single top node of a reference and using full paths. Mayas namespace solution makes it more complicated and less flexible to use matching expressions in modern applications like katana. 

The way that USD uses the name of the referencing prim rather than the referenced prim enables us to keep uniqueness without maya namespaces. If we remove all of them, then we have fewer problems because we have the most similar hierarchies across levels of maya referencing. The main need for a solution without limitation 1 is when identical maya references are parented to the same node. For now we are able to avoid that easily by ensuring that references always have a unique parent node. However, it would be easy to add another argument for the number of namespaces to strip.

Although round tripping is always a good goal for an exchange format, we don't yet have a use case that demands it. It is also not yet supported as is because the names will be different (`:` vs `_`) when reimporting an exported .usd file. If the need arises, roundtripping could be added on top of this current PR by modifying the usdExport and usdImport commands to use metadata for maya namespaces.

### DIFF
--- a/third_party/maya/lib/usdMaya/JobArgs.cpp
+++ b/third_party/maya/lib/usdMaya/JobArgs.cpp
@@ -262,6 +262,9 @@ JobExportArgs::JobExportArgs(
                 PxrUsdExportJobArgsTokens->mergeTransformAndShape)),
         normalizeNurbs(
             _Boolean(userArgs, PxrUsdExportJobArgsTokens->normalizeNurbs)),
+        stripNamespaces(
+            _Boolean(userArgs,
+                PxrUsdExportJobArgsTokens->stripNamespaces)),
         parentScope(
             _AbsolutePath(userArgs, PxrUsdExportJobArgsTokens->parentScope)),
         renderLayerMode(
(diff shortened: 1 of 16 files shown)


jtran56:
Filed as internal issue #153933.

elrond79:
Pushed new version to fix merge conflicts

elrond79:
pushed new version to fix some merge conflicts

PR #328, opened 2017-11-16
[Usd] Change visibility of pcp child nodes iterator.

fabal:
### Description of Change(s)
Change the visibility of pcp child nodes iterator.
We came across this while building [AL_USDMaya](https://github.com/AnimalLogic/AL_USDMaya/blob/develop/lib/AL_USDMaya/AL/usdmaya/cmds/LayerCommands.cpp#L731) on Windows where symbols are hidden by default.
The same symbols from the reverse iterator should also probably be exposed.

### Fixes Issue(s)
- Unable to build and use the iterator on Windows


jtran56:
Filed as internal issue #153935.

PR #336, opened 2017-11-29
Selection in VP2 for usdmaya plugin. Also increasing perforamnce for …

dalgos-adsk:
…VPS (Tumble of 1000 spehere from 13fps to 60+fps)

### Description of Change(s)
Increase of FPS on tumble in viewport 2 (1000 spheres : 13.1fps to 66.8fps)
Initial work for picking on the viewport

### DIFF
--- a/third_party/maya/lib/pxrUsdMayaGL/CMakeLists.txt
+++ b/third_party/maya/lib/pxrUsdMayaGL/CMakeLists.txt
@@ -29,6 +29,9 @@ pxr_shared_library(${PXR_PACKAGE}
     PUBLIC_CLASSES
         batchRenderer
         hdRenderer
+		usdTaskDelegate
+        usdShapeRenderer
+        usdBatchRenderer
         proxyDrawOverride
         proxyShapeUI
         softSelectHelper
--- a/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
+++ b/third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
@@ -21,14 +21,14 @@
 // KIND, either express or implied. See the Apache License for the specific
 // language governing permissions and limitations under the Apache License.
 //
-#include "pxr/pxr.h"
 #include "pxrUsdMayaGL/proxyDrawOverride.h"
+#include "pxrUsdMayaGL/usdBatchRenderer.h"
 
-#include "pxr/base/tf/envSetting.h"
-#include "pxrUsdMayaGL/batchRenderer.h"
-#include "usdMaya/proxyShape.h"
-
+#include <maya/MSelectionContext.h>
+#include <maya/MHWGeometryUtilities.h>
+#include <maya/MViewport2Renderer.h>
 #include <maya/MBoundingBox.h>
+#include <maya/MObjectHandle.h>
 #include <maya/MDagPath.h>
 #include <maya/MDrawContext.h>
 #include <maya/MFnDependencyNode.h>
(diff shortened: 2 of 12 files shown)


jtran56:
Filed as internal issue #154660.

third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
```diff
@@ -41,36 +41,74 @@
...
 MHWRender::MPxDrawOverride* 
 UsdMayaProxyDrawOverride::Creator(const MObject& obj)
 {
-    UsdMayaGLBatchRenderer::Init();
+    UsdBatchRenderer::Init();
     return new UsdMayaProxyDrawOverride(obj);
 }
 
+MHWRender::DrawAPI
+UsdMayaProxyDrawOverride::supportedDrawAPIs() const
```
mattyjams: This was actually added in commit a7e8bfd983ad0783cd970ae7c1e0eb2a81eca4b1

third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
```diff
@@ -194,11 +262,418 @@ UsdMayaProxyDrawOverride::supportedDrawAPIs() const
...
+	drawBox(1.0, GL_LINE_LOOP);
+	glMatrixMode(GL_PROJECTION);
+	glPopMatrix();
+	glMatrixMode(GL_MODELVIEW);
+	glPopMatrix();
+	glPopAttrib(); // GL_ENABLE_BIT | GL_CURRENT_BIT | GL_LIGHTING_BIT
+}
+
+void
+UsdMayaProxyDrawOverride::_SetupLighting(
```
mattyjams: This method should probably also be removed in favor of another utility method in px_vp20:
https://github.com/PixarAnimationStudios/USD/blob/022f7ef54ff8ce15120f34e61e078c1daf471153/third_party/maya/lib/px_vp20/utils.h#L54

third_party/maya/lib/pxrUsdMayaGL/usdTaskDelegate.cpp
```diff
@@ -0,0 +1,456 @@
...
+	(idRenderTask)
+	(renderTask)
+	(shadowTask)
+	(selectionTask)
+	(simpleLightTask)
+	(camera)
+);
+
+namespace {
+	class ShadowMatrix : public HdxShadowMatrixComputation
```
mattyjams: This is awesome that you're tackling this too! I was actually playing with something similar internally, but I hadn't yet finished it. I **think** that the camera needs a windowPolicy set on it though for this to work, and I don't see that happening down below. Was that left out of this change, maybe?

Also, a few of the values used to setup the frustum look like they're hard-coded and should likely instead be derived from properties of the light.

third_party/maya/lib/pxrUsdMayaGL/usdTaskDelegate.cpp
```diff
@@ -0,0 +1,456 @@
...
+    GetRenderIndex().GetChangeTracker().MarkTaskDirty(
+        renderTaskId, HdChangeTracker::DirtyCollection);
+
+    // update render params in the value cache
+    HdxRenderTaskParams taskParams =
+                _GetValue<HdxRenderTaskParams>(renderTaskId, HdTokens->params);
+
+    // update params
+    taskParams.overrideColor         = overrideColor;
+    taskParams.wireframeColor        = GfVec4f(0, 0.001238, 0.11666, 1.0);
```
mattyjams: Where does this color value come from?

third_party/maya/lib/pxrUsdMayaGL/CMakeLists.txt
```diff
@@ -29,6 +29,9 @@ pxr_shared_library(${PXR_PACKAGE}
     PUBLIC_CLASSES
         batchRenderer
         hdRenderer
+		usdTaskDelegate
+        usdShapeRenderer
+        usdBatchRenderer
```
mattyjams: `UsdBatchRenderer` looks like it may have begun life as a copy of `UsdMayaGLBatchRenderer` and then has all of your performance improvements and refactoring added. Does it make sense for both to continue to exist? Should `UsdMayaGLBatchRenderer` be deprecated/removed now in favor of `UsdBatchRenderer`?

third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
```diff
@@ -194,11 +262,418 @@ UsdMayaProxyDrawOverride::supportedDrawAPIs() const
...
+		glVertex3fv(&v[faces[i][0]][0]);
+		glVertex3fv(&v[faces[i][1]][0]);
+		glVertex3fv(&v[faces[i][2]][0]);
+		glVertex3fv(&v[faces[i][3]][0]);
+		glEnd();
+	}
+}
+
+void
+UsdMayaProxyDrawOverride::_RenderBounds(
```
mattyjams: This method and the drawBox() helper function that it uses look like ones we removed a while ago since they use fixed-function OpenGL API.

There's a utility method in px_vp20 that should probably be used instead:
https://github.com/PixarAnimationStudios/USD/blob/022f7ef54ff8ce15120f34e61e078c1daf471153/third_party/maya/lib/px_vp20/utils.h#L59

third_party/maya/lib/pxrUsdMayaGL/proxyDrawOverride.cpp
```diff
@@ -194,11 +262,418 @@ UsdMayaProxyDrawOverride::supportedDrawAPIs() const
...
+		GfMatrix4d(projectionMat.matrix),
+		viewport
+	);
+}
+
 void
 UsdMayaProxyDrawOverride::draw(const MHWRender::MDrawContext& context, const MUserData *data)
 {
-    _GetBatchRenderer().Draw( context, data );
+	// Draw callback
```
mattyjams: Is it necessary that all of this moves out of the batch renderer and into the draw override here? This and a few other things that migrated into the draw override seem like they could be more generally useful if kept closer to the batch renderer. I think our hope is that the draw override delegates as much as possible to the batch renderer so that it's easier for shape and other plugin nodes that want to image using Hydra to leverage the batch renderer in their own draw overrides.

I know that there is currently a lot of stuff jammed into batchRenderer.h/.cpp, so I think migrating out the shape renderer and the task delegate to separate files makes a lot of sense, but I wonder whether it makes sense to keep more of the heavy lifting of the draw inside the batch renderer still.

Also, I know you'll hate to hear me say it, but we do still have a handful of use cases that necessitate using the legacy viewport from time to time, though we're working to get rid of those and get fully migrated to Viewport 2.0. In the meantime though, if this logic is moved out of the batch renderer and into the draw override, it would also need to be duplicated in the proxy shape UI that the legacy viewport uses.

mattyjams:
Hey Sebastien!

Sorry for the delay, but I'm taking a look at your changes now. There's a lot of great stuff in here, so thanks very much!

I'm wondering though whether you think it might be possible to carve up the changes into a few separate, smaller pull requests. I think that might help speed up the review process, and make the motivation for some of the changes more clear.

Correct me if I'm wrong, but it seems like each of these could possibly be a separate pull request:
- Adding the use of `isAlwaysDirty=false` to `UsdMayaProxyDrawOverride`
- Implementing `wantUserSelection()` and `userSelect()` on `UsdMayaProxyDrawOverride` (I **think** this should be possible separate from some of the structural changes in this PR?)
- Replacing the queue empty-ness based checking for whether to render/select or not with your much cleaner frame stamp-based solution
- Implementing shadow support
- Other refactoring and changes necessary to get the performance gains you've achieved

Also, it looks as if your fork of the repo may have fallen a bit out of date, and a few bits of code that I believe we've either removed or replaced seem to have snuck back in.

One other minor gripe, but there seems to be a mix of whitespace usage in some of the changes. Could we have you ensure that the code contains spaces instead of tabs, and clean up some inconsistent indentation in the changes (four spaces per indent level).

I've commented on a few higher-level things, but let me know whether it's possible for you to split this request up a bit, as well as rebase it against the current dev branch. In the meantime, I'll start giving your changes a try internally.

Thanks again for all the great work!

sunyab:
We've integrated a number of the changes from this pull request in some of the recent work to refactor and improve Maya/Hydra batch rendering and believe we're in a much better place now. I'm going to close this PR out, thanks again!

PR #337, opened 2017-12-01
Pr/clang maya2018 fixes

elrond79:
### Description of Change(s)
Just some clang fixes for usdMaya and maya-2018

Note that I don't actually know if the entire project builds with clang + maya2018, these were just some errors I was getting from a clang-based static analysis tool... but I confirmed that they raised clang build errors for the specific compilation units in question (which are now fixed).


elrond79:
Oops, just realized these exact issues were already addressed in https://github.com/PixarAnimationStudios/USD/commit/e5549a498d4a9d259b25892883e5a57cf4048e36

PR #339, opened 2017-12-02
[usdMaya] add PreExport / PostExport to PxrUsdMayaShadingModeExporter

elrond79:
### Description of Change(s)
This adds two new virtual methods, PreExport and PostExport, which provide points at which sub-classes can customize the shader export process.

### Fixes Issue(s)
- A recent commit made DoExport no longer a virtual method; while this ensured a more unified export process, it also meant a loss of flexibility for shader exporters which need to do some sort of setup + tear down. This commit restores the ability for third-party subclasses to do such things.


jtran56:
Filed as internal issue #154662.

PR #347, opened 2017-12-10
Allow user to specify boost version

meshula:
### Description of Change(s) 

I'm using boost 1.65.1, and like the convenience of using the standard build script. This modification doesn't change the default behavior, but if boost version is specified on the command line, USD will build with the specified version.

Also handles toolset version for VS2015 and VS2017, because boost 1.65.1 insists on toolset 141 for VS2017.

OpenImageIO 1.7.17 is the minimum compatible with boost 1.65.1

### Fixes Issue(s) 

build script is hard coded to boost 1.61 and using the 140 toolset with VS2015 and VS2017

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -373,12 +373,10 @@ ZLIB = Dependency("zlib", InstallZlib, "include/zlib.h")
 ############################################################
 # boost
 
-if Linux():
-    BOOST_URL = "http://downloads.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.tar.gz"
-elif Windows() or MacOS():
-    BOOST_URL = "http://downloads.sourceforge.net/project/boost/boost/1.61.0/boost_1_61_0.tar.gz"
-
 def InstallBoost(context, force):
+    boost_version = context.boost.replace(".", "_")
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #154998.

build_scripts/build_usd.py
```diff
@@ -1143,6 +1155,19 @@ def ForceBuildDependency(self, dep):
...
     os.environ['PYTHONPATH'] = os.pathsep.join(paths)
 
+# The default installation of boost on Windows puts headers in a versioned 
+# subdirectory, which we have to account for here. In theory, specifying 
+# "layout=system" would make the Windows install match Linux/MacOS, but that 
+# causes problems for other dependencies that look for boost.
+if Windows():
+    # .0 versions of boost don't end in _0.
+    boost_version = "_".join(context.boost.split("."))
+    if boost_version[-2:] == ".0":
```
meshula: Typo - this should be == "_0":

sunyab:
We'd prefer to keep build_usd.py to a curated set of dependencies for easier testing and validation, especially given the relationships between dependencies like the one between boost and OpenImageIO identified in this PR. This sort of customization (even at an "at-your-own-risk" level) opens us up to more support load; if users want to use different versions they can edit the build script locally, which should make it clear they're in unguarded waters. :)

So, I'm going to close this PR out. Thanks!

meshula:
Makes sense to me, the curated build environment today works for me very well in all the configurations I'm dealing with!

PR #351, opened 2017-12-14
[usdmaya] MashInstancer can only run on maya-2018

elrond79:
### Description of Change(s)

Basically, fixes a test that would fail with maya < 2018.


jtran56:
Filed as internal issue #155170.

PR #353, opened 2017-12-14
mark the GfMatrix*.Get method const

elrond79:
### Description of Change(s)
Pretty self explanatory. I wanted to use GfMatrid4d.Get from a const instance, but couldn't.

### Fixes Issue(s)
- None reported


jtran56:
Filed as internal issue #155174.

PR #354, opened 2017-12-14
[usdGeom] allow UsdGeomXformable::GetLocalTransformation(resetsXformStack=nullptr)

elrond79:
### Description of Change(s)
If there's no xform ops, you can get a crash if you use a nullptr for resetsXformStack.

### Fixes Issue(s)
- Crash if no xform ops, and you pass resetsXformStack=nullptr

### DIFF
--- a/pxr/usd/lib/usdGeom/xformable.cpp
+++ b/pxr/usd/lib/usdGeom/xformable.cpp
@@ -687,7 +687,7 @@ UsdGeomXformable::GetLocalTransformation(
         return false;
 
     if (opOrderVec.size() == 0) {
-        *resetsXformStack = false;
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #155175.

sunyab:
Closing this pull request in favor of commit 417956efd. We think that the resetsXformStack parameter should not be optional because it's a key bit of data that consumers need to consider when using the result of GetLocalTransformation. The referenced commit fixes the null pointer dereference so bad code won't crash, but the function will now emit a coding error in this case (as had been documented in the header previously).

Please let us know if you think we should reconsider. Thanks!

PR #356, opened 2017-12-20
Fix building issue about OpenEXR and runtime issue about Houdini 16.5.

YKZDY:
### Description of Change(s)
- Add INSTALL_RPATH and INSTALL_RPATH_USE_LINK_PATH in InstallOpenEXR.
- Add boost_system linking dependence for gusd in Houdini plugin.
### Fixes Issue(s)
- Fix the runtime dependence issue between IlmBase and OpenEXR.
- Since Houdini16.5, the stock Boost C++ libraries are no longer shipped with the Houdini Development Kit, so here we need to explicitly bind boost_system with gusd in CMake.


YKZDY:
Sorry again... Just recognize I try to merge into the master branch, re-request then.....

PR #357, opened 2017-12-20
Fix building issue about OpenEXR and runtime issue about Houdini 16.5

YKZDY:
### Description of Change(s)
- Add INSTALL_RPATH and INSTALL_RPATH_USE_LINK_PATH in InstallOpenEXR.
- Add boost_system linking dependence for gusd in Houdini plugin.
### Fixes Issue(s)
- Fix the runtime dependence issue between IlmBase and OpenEXR.
- Since Houdini16.5, the stock Boost C++ libraries are no longer shipped with the Houdini Development Kit, so here we need to explicitly bind boost_system with gusd in CMake.


jtran56:
Filed as internal issue #155305.

sunyab:
Hi @YKZDY, I'm testing out your pull request now. Just one initial question:

- In gusd/CMakeLists.txt, do we need to link against boost::filesystem instead of (or in addition to?) boost::system? AFAICT we don't use boost::system directly in gusd, but we do use boost::filesystem, which uses boost::system internally.

Thanks!

YKZDY:
Hi @sunyab, I haven't checked the codes in Houdini plugin. But when I try to build the USD against the H16.5 before, the program told me it cannot find the "boost::system", so I commit that change which will fix this linking error on my machine.
If you think the "boost::filesystem" would be the precise library then it would be great to switch to it. Just make sure if everything could work with H16.5 properly. :)

andrewkaufman:
@sunyab any chance of 5d9e2b6 (or a similar "boost::filesystem" change) making it into the next release? I think linking to either one should solve the H16.5 issue. Though perhaps the conflicting changes on the dev branch (linking to several native Houdini libs) already solves the problem?

FWIW, here is the error I get when building from 0.8.2 and loading in Houdini 16.5.268:

`libgusd.so: error: symbol lookup error: undefined symbol: _ZN5boost6system16generic_categoryEv (fatal)`

sunyab:
@andrewkaufman Yep! We've already made the change to link against boost::filesystem internally. It should be pushed up to the dev branch shortly and (barring any issues) will be in the next release.

sunyab:
Commit 1bed49675 adds an explicit link to boost::filesystem, which should fix the symbol lookup errors. 

@YKZDY, I haven't merged your build script changes for OpenEXR yet, as I'd like to do a little more testing. How did the OpenEXR / IlmBase dependency issue show up for you?

YKZDY:
@sunyab, it interrupted the building process on my machine, since OpenEXR requires IImBase as its linking dependency. If I remember correctly, the exception message was like `cannot find libIex.so.12`.

c64kernal:
I believe we are in a better place now, the Houdini plugins have been separated out of this repo and we can't reproduce the OpenEXR issue.

PR #359, opened 2017-12-21
 [plug] fix for loading of same plugin twice 

elrond79:
### Description of Change(s)

This modifies the way that plugin locking is done, to avoid a race condition where the same plugin is loaded twice, causing errors and/or crashes.

(Note: The first commit consolidates some code related to plugin creation - this was just so that I didn't have to repeat adding the same code three times.)

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/358

### DIFF
--- a/pxr/base/lib/plug/plugin.cpp
+++ b/pxr/base/lib/plug/plugin.cpp
@@ -56,6 +56,15 @@ using std::pair;
 using std::string;
 using std::vector;
 
+// Set this to nonzero to insert a pause for debugging, to make registration
+// collisions more likely
+#define PXRPLUG_TEST_REGISTRATION_SLEEP_MILLISECS 0
+
+#if PXRPLUG_TEST_REGISTRATION_SLEEP_MILLISECS > 0
+#include <chrono>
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #155354.

gitamohr:
Hi elrond79 -- thanks for this PR.  I'm integrating this change internally but I've made some modifications.  I've removed the code related to dropping the lock around constructing the PlugPlugin object, since that object's constructor doesn't really do much; in particular it doesn't invoke any foreign code.  So just holding the lock the entire time gives a nice simplification and avoids any possibility of a race.  I have kept your refactoring of the three NewPlugin functions.  Thanks!

elrond79:
Cool, makes sense... I had left in the dropping of the lock around the constructor since that's what the original code did, and I assumed that it was expensive... but if not, then yeah, just holding the lock will certainly simplify things a lot!

sunyab:
Closing this pull request out in favor of commit 9ad3c35. Thanks!

PR #366, opened 2018-01-17
[usdMaya] fix usdExport to account for camera shake attribute

nxkb:
### Description of Change(s)
Adds maya camera's shake offset to USD camera aperture offset if shake is enabled.

This documentation indicates that simply adding the values should be sufficient: https://help.autodesk.com/view/MAYAUL/2018/ENU/?guid=GUID-C3EBB008-7DBE-4B9D-B9AC-1DA974CEFE15

### Fixes Issue(s)
- Exported Maya cameras that use shakeEnabled, shakeHorizontal and shakeVertical will not line up with USD camera.


nxkb:
reopen now that its properly based on dev :)

jtran56:
Filed as internal issue #155963.

Issue #349, opened 2017-12-12
[Katana] Instancing produces random results in pxrUsdIn when "as sources and instances" is enabled


sirpalee:
### Description of Issue
When using instances that instance instances themselves pxrUsdIn fails to create the sources properly when instanceMode is set to "as sources and instances".

The issue happens because of the change introduced in 523ee0a3 . If you look at pxrUsdIn.cpp:503, you can see why it fails. The loop properly iterates through the masters, but the locationParent is going to be the same for all the masters, so the last master (in the case of the example scene, m19) will always overwrite usdPrimPath and usdPrimName, thus the following loop (line 528) will only iterate through one master.

### Steps to Reproduce
1. Load up the attached file in katana
2. Set instanceMode to "as sources and instances"
3. Expand the scene graph, and inspect /root/world/geo/Masters/master/variants. There is only one instance, m19.

### Package Versions
Latest dev.

sirpalee:
Updated the title of the ticket and attached a simpler file to reproduce the issue. The random behaviour can be seen on only one level of instancing.

[instancing_bug_generated_single.usda.zip](https://github.com/PixarAnimationStudios/USD/files/1550398/instancing_bug_generated_single.usda.zip)

jtran56:
Filed as internal issue #154999.

nrusch:
@sirpalee Was this also fixed by https://github.com/PixarAnimationStudios/USD/commit/7172773bd3651132ce3ae9c93a0336280950fda4?

c64kernal:
Closing this as we are in the process of deprecating the Katana plugins from the USD repo.

PR #367, opened 2018-01-19
[usdKatana] Fixing instances and sources mode for pxrUsdIn.

sirpalee:
### Description of Change(s)
The change aims to fix the issue found with pxrUsdIn's "as sources and instances". I simply extended the usdPrimPath and usdPrimName parameters on the location so they can now be used as arrays, and the intermediate op iterates through all the elements. I'm a bit unsure about collecting the prim paths and prim names for each location parent in the first part of the PR because I don't know if there is a possibility of multiple location parent names.

### Fixes Issue(s)
#349


sirpalee:
@stevelavietes @chadrik

jtran56:
Filed as internal issue #156101.

elrond79:
No longer needed, after: 7172773bd3651132ce3ae9c93a0336280950fda4

PR #369, opened 2018-01-20
[Pxr] Only look for pyside if python is enabled

superfunc:
### Description of Change(s)
Remove pyside lookup when python is off.

### Fixes Issue(s)
- Doesn't look up dependencies that won't be used.


sunyab:
I incorporated a version of this fix into commit 3a387cd4, so I'm going to close this out in favor of that change. Thanks!

Issue #170, opened 2017-03-03
Can we eliminate the use of boost shared pointers? (BOOST_SP_USE_STD_ATOMIC problem)


meshula:
There is a difficult USD configuration issue around the use of boost shared pointers.

This flag 
```
BOOST_SP_USE_STD_ATOMIC
```
exists as a preprocessor directive, and it is impossible to tell if a compiled version of boost was made with it set to 1 or not.

If you are trying to build USD against a supplied version of boost, instead of one built specifically for USD, it's extremely difficult to narrow the issues down to this flag, and then to work out how to set it.

I would request that we don't use boost shared pointers at all, since for the most part the functionality is covered by std or tbb.

jtran56:
Filed as internal issue #143729.

waywardmonkeys:
I have a branch that seems to do this. It is currently, without all of the requisite cleanup, modifying about 170 files.

To avoid repeatedly rebasing this ... is there a good time to submit it as a pull request?

meshula:
I just wanted to check in and say that this is still an issue.

nvmkuruc:
@meshula @jesschimein I think this issue can be closed.

PR #372, opened 2018-01-23
Switch from boost::shared_ptr to std::shared_ptr.

waywardmonkeys:
This converts from using `boost::shared_ptr` and friends
(like `weak_ptr`) over to using the `std` equivalents.

Fixes issue #170.


waywardmonkeys:
Once this lands, I have some follow up patches that fix up other usages of `boost` that can be replaced with `std` equivalents.

jtran56:
Filed as internal issue #156190.

waywardmonkeys:
This had merge conflicts ... so I updated it and force-pushed this branch.

c64kernal:
Thank for this PR. It was well ahead of its time and we've finally done some of this work now, and we'll continue doing it.

PR #373, opened 2018-01-23
[UsdImaging] Remove unnecessary boost::noncopyable

superfunc:
### Description of Change(s)
Remove use of boost noncopyable.

### Fixes Issue(s)
Reduces dependency on boost.


jtran56:
Filed as internal issue #156191.

PR #374, opened 2018-01-23
[Imaging] Remove unnecessary boost::noncopyable

superfunc:
Same rationale as #373


jtran56:
Filed as internal issue #156369.

PR #375, opened 2018-01-24
build_usd: Output what the cmake generator configuration is.

waywardmonkeys:
When using `--generator` with `build_usd.py`, this makes it more clear in the output what was selected.


jtran56:
Filed as internal issue #156370.

PR #376, opened 2018-01-24
cmake: Fix building with ninja.

waywardmonkeys:
Ninja and cmake get unhappy when there is a custom command output
and a custom target with the same name resulting in a cyclic
dependency between them. This isn't an issue with cmake's support
for Makefiles.

### DIFF
--- a/cmake/macros/Public.cmake
+++ b/cmake/macros/Public.cmake
@@ -100,13 +100,13 @@ function(pxr_python_bin BIN_NAME)
     endif()
 
     # Add the target.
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #156371.

waywardmonkeys:
Any update on this landing?

sunyab:
@waywardmonkeys - Yes, I pulled this change down locally for a sanity check earlier this week and all looked good. This should land with the next push of changes early next week. Thanks!

PR #377, opened 2018-01-24
Fix warnings about missing parentheses.

waywardmonkeys:
This happens when the result of a comparison is used as the
condition within an if statement in clang (and gcc?).


jtran56:
Filed as internal issue #156372.

PR #378, opened 2018-01-24
Fix format security warnings under clang.

waywardmonkeys:
This wants to get a literal string for the format argument and
then to do the substitution itself.


jtran56:
Filed as internal issue #156373.

PR #379, opened 2018-01-24
Disable the undefined-var-template warning in clang.

waywardmonkeys:
This emits a warning about something that can potentially be
a link time problem.


jtran56:
Filed as internal issue #156374.

elrond79:
I'm confused... your description says it emits a warning about something that could be a problem at link time.  That sounds like a useful warning, why do we want to disable it?  Is there already a situation in the codebase that's emitting this warning, and there's not a easy fix to make it stop?

waywardmonkeys:
It warns about something that, if it were an issue, would result in a linked error. But in this codebase, there is no actual issue. I don’t recall seeing an easy fix to address the code side.

sunyab:
Closing this pull request in favor of #646. Thanks!

PR #380, opened 2018-01-24
Fix -Wmissing-braces warnings.

waywardmonkeys:
These happen with clang.


jtran56:
Filed as internal issue #156375.

sunyab:
Marking this as addressed in the next release. A bunch of the warnings that were being issued were due to bugs in clang that have been fixed in newer versions (see https://bugs.llvm.org/show_bug.cgi?id=21629). The remaining warnings were fixed up with the other commits against this issue. Thanks!

PR #381, opened 2018-01-24
Fix mismatched tags warnings.

waywardmonkeys:
These happen under clang.


jtran56:
Filed as internal issue #156376.

waywardmonkeys:
I updated this to apply to current `dev` branch. This is a really simple change and it is needed because the original class definitions are as `struct`, not `class`, so forward declarations should match. This fixes warnings when building with clang on macOS.

PR #382, opened 2018-01-24
Use override as other things in this file do.

waywardmonkeys:
This fixes a warning under clang.


jtran56:
Filed as internal issue #156377.

Issue #286, opened 2017-09-16
Katana: PointInstancer Op produces recursive hierarchy in some cases


nrusch:
### Description of Issue

With certain stage configurations, the Katana reader Op for `PointInstancer` prims will generate an infinitely recursive scene graph hierarchy below the `"usd point instancer"` location, and will not generate `"instance source"` locations for the prototype prims.

To reproduce, simply read the USD file in the attached [PointInstancerTest.zip](https://github.com/PixarAnimationStudios/USD/files/1307954/PointInstancerTest.zip) into Katana, and expand the hierarchy under the `USD_TEST_INSTANCER ` location.

If the `material:binding` attribute values are removed from the prototype source prims, the stage seems to load as expected.

### Package Versions

USD 0.8.0

sirpalee:
I was digging around a bit and found why the infinite recursion happens.

The example scene looks like this. The materials placed into their own separate scope by the Arnold shader exporter from usd-arnold.

```
-root
--- xform
----- instancer
--- xform
----- instances
--- materials
```

When the instancer exporter runs, it tries to build both the instance sources and the assigned materials by looking for a common prefix for both. Since the common prefix is above the instancer in our case, the instancer builder triggers its export once again.

A quick patch like this gets rid of the infinite recursion; it's just quickly checking if the common prefix is above the instancer. As expected, this breaks all the material bindings.

```
diff --git a/third_party/katana/lib/usdKatana/readPointInstancer.cpp b/third_party/katana/lib/usdKatana/readPointInstancer.cpp
index 7a9dbfa8..4b93f71e 100644
--- a/third_party/katana/lib/usdKatana/readPointInstancer.cpp
+++ b/third_party/katana/lib/usdKatana/readPointInstancer.cpp
@@ -361,8 +361,8 @@ PxrUsdKatanaReadPointInstancer(
     //
     // Validate instancer data.
     //
-
-    const std::string instancerPath = instancer.GetPath().GetString();
+    const auto instancerSdfPath = instancer.GetPath();
+    const std::string instancerPath = instancerSdfPath.GetString();

     UsdStageWeakPtr stage = instancer.GetPrim().GetStage();

@@ -568,7 +568,7 @@ PxrUsdKatanaReadPointInstancer(
                 {
                     const SdfPath &commonPrefix =
                             protoPath.GetCommonPrefix(materialPath);
-                    if (commonPrefix.GetString() == "/")
[...]

PR #384, opened 2018-01-24
Fixing infinite recursion when instances and materials are outside the instancer's scope

sirpalee:
### Description of Change(s)
The following changes cases when infinite recursion happens in pxrUsdIn when sources are outside the scope of the pointInstancer. The change also deals with corner-case material assignments, when the materials are outside the primitive's scope. For that, we just expect the material to be built by something else (it'll be in all our cases), we don't do any additional checks.

### Fixes Issue(s)
- #286


slooper:
This has bitten us as well. Thanks for the fix!

jtran56:
Filed as internal issue #156381.

PR #387, opened 2018-01-26
[usdHoudini] need to link against boost_system lib

elrond79:
...this only became a problem with houdini-16.5, which renamed it's internal boost
to hboost

as a result, we didn't link against boost "for free" when houdini started up...
...so we need to explicitly tell the houdini library it needs to link!

### Fixes Issue(s)
- plugin would not load on houdini-16.5

### DIFF
--- a/third_party/houdini/lib/gusd/CMakeLists.txt
+++ b/third_party/houdini/lib/gusd/CMakeLists.txt
@@ -14,6 +14,7 @@ pxr_shared_library(${PXR_PACKAGE}
         usdRi
         usdShade
         usdUtils
+        ${Boost_SYSTEM_LIBRARY}
         ${HOUDINI_UI_LIBRARY}
         ${HOUDINI_OPZ_LIBRARY}
         ${HOUDINI_OP3_LIBRARY}


jtran56:
Filed as internal issue #156383.

sunyab:
Closing this out in favor of commit 1bed49675. Please let us know if this doesn't ersolve the issue for you, thanks!

elrond79:
Hi sunya - so, it seems you replaced our adding of Boost_SYSTEM_LIBRARY with Boost_FILESYSTEM_LIBRARY instead. Is this intentional? In our testing, your patch still raises an error when attempting to launch houdini-16.5; conversely, it seems to launch fine with SYSTEM, but no FILESYSTEM linkage.

sunyab:
Yes, I had intentionally linked against boost::filesystem since gusd was using that directly, not boost::system. In my own testing that seemed sufficient to fix the Houdini 16.5 errors, so I'm not sure why you were still seeing them. 

In any case, I've committed another change internally that adds the boost::system link which should be pushed up shortly. Sorry for the trouble!

elrond79:
Hmm, weird that we both seemed to need a different library! Probably some detail of our exact environment (LD_LIBRARY_PATH, or order in which houdini plugins are loaded, or something?)

Anyway, requiring both should fix it. Thanks!

sunyab:
For reference, this was just pushed in commit 92ffe54bb7acbcae99c5ca061234c344e580b46e.

PR #388, opened 2018-01-26
[Hd] Fix for import modules better than PySide and PySide2

sho7noka:
### Description of Change(s)
Houdini include Qt.py in hutil modules.
So i think not good about Houdini version check.

### Fixes Issue(s)
- use Qt.py in hutil modules.


jtran56:
Filed as internal issue #156384.

PR #390, opened 2018-01-27
[Pxr] Add UsdStage::GetObjectAtPath.

superfunc:
### Description of Change(s)

This allows a convenient way to get items in a stage from paths,
whether they are property paths or prim paths.

### Fixes Issue(s)

This was requested on the interest forum a while back, don't have the link handy.

### DIFF
--- a/pxr/usd/lib/usd/stage.cpp
+++ b/pxr/usd/lib/usd/stage.cpp
@@ -1746,6 +1746,32 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
     return UsdPrim(primData, proxyPrimPath);
 }
 
+UsdObject
+UsdStage::GetObjectAtPath(const SdfPath &path) const
+{
+    // Maintain consistent behavior with GetPrimAtPath
+    if (!path.IsAbsolutePath()) {
+        return UsdObject();
+    }
+
+    const bool isPrimPath = path.IsPrimPath();
+    const bool isPropPath = !isPrimPath && path.IsPropertyPath();
+    if (!isPrimPath && !isPropPath) {
+        return UsdObject();
+    }
+
+    // A valid prim must be found to return either a prim or prop
+    if (isPrimPath) {
+        return GetPrimAtPath(path);
+    } else if (isPropPath) {
+        if (auto prim = GetPrimAtPath(path.GetPrimPath())) {
+            return prim.GetProperty(path.GetNameToken());
+        }
+    }
+
+    return UsdObject();
+}
+
 Usd_PrimDataConstPtr
 UsdStage::_GetPrimDataAtPath(const SdfPath &path) const
 {
--- a/pxr/usd/lib/usd/stage.h
+++ b/pxr/usd/lib/usd/stage.h
@@ -728,6 +728,32 @@ public:
     USD_API
     UsdPrim GetPrimAtPath(const SdfPath &path) const;
 
+    /// Return the UsdObject at \p path, or an invalid UsdObject if none exists.
+    ///
+    /// If \p path indicates a prim beneath an instance, returns an instance
+    /// proxy prim if a prim exists at the corresponding path in that instance's 
+    /// master. If \p path indicates a property beneath a child of an instance, 
+    /// returns a property whose parent prim is an instance proxy prim.
+    ///
+    /// Example:
+    ///
+    /// \code
+    ///if (UsdObject obj = stage->GetObjectAtPath(path)) {
+    ///    if (UsdPrim prim = obj.As<UsdPrim>()) {
+    ///        // Do things with prim
+    ///    }
+    ///    else if (UsdProperty prop = obj.As<UsdProperty>()) {
+    ///        // Do things with property. We can also cast to
+    ///        // UsdRelationship or UsdAttribute using this same pattern.
+    ///    }
+    ///}
+    ///else {
+    ///    // No object at specified path
+    ///}
+    /// \endcode
+    USD_API
+    UsdObject GetObjectAtPath(const SdfPath &path) const;
+
 private:
     // Return the primData object at \p path.
     Usd_PrimDataConstPtr _GetPrimDataAtPath(const SdfPath &path) const;
--- a/pxr/usd/lib/usd/testenv/testUsdPrims.py
+++ b/pxr/usd/lib/usd/testenv/testUsdPrims.py
@@ -61,6 +61,33 @@ class TestUsdPrim(unittest.TestCase):
             assert a == b
             assert hash(a) == hash(b)
 
+            # check for prims/props that exist
+            p = s.GetObjectAtPath(u'/foo')
+            assert p
+            assert type(p) is Usd.Prim
+
+            a = s.GetObjectAtPath(u'/foo.attr')
+            assert a
+            assert type(a) is Usd.Attribute
+
+            r = s.GetObjectAtPath(u'/foo.relationship')
+            assert r 
+            assert type(r) is Usd.Relationship
+
+            # check for prims/props that dont exist
+            p = s.GetObjectAtPath(u'/nonexistent')
+            assert not p
+            assert type(p) is Usd.Prim
+
+            a = s.GetObjectAtPath(u'/foo.nonexistentattr')
+            assert not a
+            assert type(a) is Usd.Property
+
+            r = s.GetObjectAtPath(u'/foo.nonexistentrelationship')
+            assert not r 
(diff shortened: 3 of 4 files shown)


jtran56:
Filed as internal issue #156422.

pxr/usd/lib/usd/stage.h
```diff
@@ -700,6 +700,33 @@ class UsdStage : public TfRefBase, public TfWeakBase {
...
+    /// Return the UsdObject at \p path, or an invalid UsdObject if none exists.
+    ///
+    /// If \p path indicates a prim beneath an instance, returns an instance
+    /// proxy prim if a prim exists at the corresponding path in that instance's 
+    /// master.
+    ///
+    /// Suggested use:
+    ///
+    /// \code
+    /// if (UsdObject myObj = stage.GetObjectAtPath(path)) {
```
sunyab: We can just use UsdObject::As in this example, since As is documented to return an invalid object if it's called on an invalid object, like:

```
if (UsdObject obj = stage->GetObjectAtPath(path)) {
    if (UsdPrim prim = obj.As<UsdPrim>()) {
        // Do things with prim
    }
    else if (UsdProperty prop = obj.As<UsdProperty>()) {
        // Do things with property. We can also cast to
        // UsdRelationship or UsdAttribute using this same pattern.
    }
}
else {
    // No object at specified path
}
```

sunyab: We can just use UsdObject::As in this example, since As is documented to return an invalid object if it's called on an invalid object, like:
```
if (UsdObject obj = stage->GetObjectAtPath(path)) {
    if (UsdPrim prim = obj.As<UsdPrim>()) {
        // Do things with prim
    }
    else if (UsdProperty prop = obj.As<UsdProperty>()) {
        // Do things with property. We can also cast to
        // UsdRelationship or UsdAttribute using this same pattern.
    }
}
else {
    // No object at specified path
}
```

pxr/usd/lib/usd/stage.h
```diff
@@ -700,6 +700,33 @@ class UsdStage : public TfRefBase, public TfWeakBase {
     USD_API
     UsdPrim GetPrimAtPath(const SdfPath &path) const;
 
+    /// Return the UsdObject at \p path, or an invalid UsdObject if none exists.
+    ///
+    /// If \p path indicates a prim beneath an instance, returns an instance
+    /// proxy prim if a prim exists at the corresponding path in that instance's 
+    /// master.
+    ///
+    /// Suggested use:
```
sunyab: Instead of 'Suggested use:', can we say 'Example:' instead?

pxr/usd/lib/usd/stage.cpp
```diff
@@ -1732,6 +1732,36 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
...
+UsdObject
+UsdStage::GetObjectAtPath(const SdfPath &path) const
+{
+    // Maintain consistent behavior with GetPrimAtPath
+    if (!path.IsAbsolutePath()) {
+        return UsdObject();
+    }
+
+    bool isPrimPath = path.IsPrimPath();
+    bool isPropPath = path.IsPropertyPath();
```
sunyab: Can we make these bools const? Also, can we avoid calling IsPropertyPath() if isPrimPath is already true:
```
const bool isPrimPath = path.IsPrimPath();
const bool isPropPath = !isPrimPath && path.IsPropertyPath();
```

pxr/usd/lib/usd/stage.cpp
```diff
@@ -1732,6 +1732,36 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
...
+    if (!isPrimPath && !isPropPath) {
+        return UsdObject();
+    }
+
+    // A valid prim must be found to return either a prim or prop
+    auto prim = GetPrimAtPath(path.GetPrimPath());
```
sunyab: Can we avoid calling GetPrimPath() on paths we already know are prim paths?

pxr/usd/lib/usd/stage.cpp
```diff
@@ -1732,6 +1732,36 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
...
+    if (prim) {
+        if (isPrimPath) {
+            return prim.As<UsdObject>();
```
sunyab: Shouldn't need to explicitly cast to UsdObject using As() here and below, the implicit derived-to-base conversion should be sufficient.

pxr/usd/lib/usd/stage.h
```diff
@@ -700,6 +700,33 @@ class UsdStage : public TfRefBase, public TfWeakBase {
     USD_API
     UsdPrim GetPrimAtPath(const SdfPath &path) const;
 
+    /// Return the UsdObject at \p path, or an invalid UsdObject if none exists.
+    ///
+    /// If \p path indicates a prim beneath an instance, returns an instance
```
sunyab: Can we add to this doc to mention properties and instance proxies as well? Something like,

```
If \p path indicates a property beneath a child of an instance, returns a property
whose parent prim is an instance proxy prim.
```

superfunc: Thanks for the notes, I'll get an updated branch pushed up this weekend.

superfunc: Sorry for the delay @sunyab, updated

pxr/usd/lib/usd/stage.cpp
```diff
@@ -1722,6 +1722,36 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
...
+
+    const bool isPrimPath = path.IsPrimPath();
+    const bool isPropPath = !isPrimPath && path.IsPropertyPath();
+    if (!isPrimPath && !isPropPath) {
+        return UsdObject();
+    }
+
+    // A valid prim must be found to return either a prim or prop
+    if (isPrimPath) {
+        if (auto prim = GetPrimAtPath(path)) {
```
sunyab: I think we can just return GetPrimAtPath(path) here and avoid unnecessarily checking the validity of the prim here.

pxr/usd/lib/usd/stage.cpp
```diff
@@ -1722,6 +1722,36 @@ UsdStage::GetPrimAtPath(const SdfPath &path) const
...
+            return prim;
+        }
+    } else if (isPropPath) {
+        if (auto prim = GetPrimAtPath(path.GetPrimPath())) {
+            if (auto prop = prim.GetProperty(path.GetNameToken())) {
```
sunyab: Similar here; we need to check the validity of the prim, but we should be able to just return prim.GetProperty(...) and not need to check the validity of the property object before returning it.

pxr/usd/lib/usd/testenv/testUsdPrims.py
```diff
@@ -61,6 +61,33 @@ def test_Basic(self):
...
+            assert type(a) is Usd.Attribute
+
+            r = s.GetObjectAtPath(u'/foo.relationship')
+            assert r 
+            assert type(r) is Usd.Relationship
+
+            # check for prims/props that dont exist
+            p = s.GetObjectAtPath(u'/nonexistent')
+            assert not p
+            assert type(p) is Usd.Object
```
sunyab: These asserts are failing now. I think this assert needs to be:
```
assert type(p) is Usd.Prim
```

pxr/usd/lib/usd/testenv/testUsdPrims.py
```diff
@@ -61,6 +61,33 @@ def test_Basic(self):
...
+
+            a = s.GetObjectAtPath(u'/foo.nonexistentattr')
+            assert not a
+            assert type(a) is Usd.Object
```
sunyab: assert type(p) is Usd.Property

pxr/usd/lib/usd/testenv/testUsdPrims.py
```diff
@@ -61,6 +61,33 @@ def test_Basic(self):
...
+
+            r = s.GetObjectAtPath(u'/foo.nonexistentrelationship')
+            assert not r 
+            assert type(r) is Usd.Object
```
sunyab: assert type(r) is Usd.Property

superfunc:
Sorry about that! I forgot CI isn't running the tests, I'll get this pushed in a few

superfunc:
Ok @sunyab should be good to go. I also smashed it down into one commit.

Issue #80, opened 2016-10-14
Deprecate TF_FOR_ALL


superfunc:
The ```TF_FOR_ALL``` macro is no longer necessary in most cases, and can be replaced with C++11's range-for. 

cc @PeterShinners

Cewein:
I guess this is still relevant, with a quick search I see there is a lot of file still using `TF_FOR_ALL`. I will jump on the bandwagon and help you @superfunc

there is also the case of the `TF_REVERSE_FOR_ALL` that can be discused I guess.

PR #398, opened 2018-01-28
[Gf] Remove TF_FOR_ALL usage.

superfunc:
### Description of Change(s)
Replaces uses of TF_FOR_ALL macro with standard for each construct introduced in C++11.

### Fixes Issue(s)
Towards #80 , many uses left.


jtran56:
Filed as internal issue #156433.

PR #399, opened 2018-01-28
[Usd] Make jinja message of type status.

superfunc:
### Description of Change(s)

This makes it match the warning in Gf.
This also changes the capitalization to match Gf.


jtran56:
Filed as internal issue #156434.

superfunc:
Given the load of PRs, and the relative unimportance of this one, closing so the PRs can get down to a reasonable count.

PR #403, opened 2018-02-04
[Usd] Add more informative error message in sg.

superfunc:
This is part of a continuing effort to make schemaGen friendlier to use.

### Description of Change(s)

UsdGenSchema disallows schemas from inheriting from multiple
parents. Currently this error manifests as:

```
ERROR: 'NoneType' object has no attribute 'typeName'
```

Needless to say, this isn't particularly helpful. This commit
changes the error to manifest as follows(given an example schema
defined at /MyC):

```
ERROR: Invalid schema definition at </MyC>
Schemas can only inherit from one other schema at most. This schema inherits from 2 (/MyBase, /Xformable).
See https://graphics.pixar.com/usd/docs/api/_usd__page__generating_schemas.html for more information.
```


jtran56:
Filed as internal issue #156818.

Issue #362, opened 2018-01-10
Houdini: Add st primvar <-> uv attribute conversion


nrusch:
The Maya USD integration plugin includes special-case handling of the `st` primvar for geometry: it reads `st` into Maya's default `map1` UV set on import, and likewise exports `map1` to the `st` primvar.

The `pxrUsdIn` Katana op also reads the `st` primvar in a way that it is generally interpreted as a UV set, though this could almost be considered a happy accident (though I'm reasonably sure it wasn't) based on the fact that A) the op reads primvars into the `geometry.arbitrary` group, and B) many renderer plugins interpret `geometry.arbitrary.st` as a UV set.

Currently, the USD Houdini plugin does not treat the `st` primvar in any special way, so artists wishing to utilize UVs in "standard" ways must manually copy the primvar data to one of Houdini's `uv` attributes on import (as well as converting it from a `vector2` to a `vector`), and subsequently reverse that process on export if they wish to modify UV data in Houdini.

There was a [previous discussion](https://groups.google.com/d/msg/usd-interest/Paz_H4MjI0o/qXysKd-OBQAJ) that touched on this topic, where there was mention of some uncertainty surrounding the current status quo with regard to USD's handling (or not) of UV data. However, with no alternative core changes on the horizon, I think it would be worthwhile to add support for automatic conversion between the `st` primvar and Houdini's `uv` attribute to the `Gusd` libraries to provide a more intuitive "out-of-the-box" interchange workflow. Making the conversion optional would help avoid interference with more customized site-specific workflows.

jtran56:
Filed as internal issue #155727.

spiffmon:
Hey Nathan,
    Sorry for the delay here!  Our Fx devs have given their OK for adding
this conversion.  We already do something similar for displayColor <-> Cd ,
and they ask you follow the same pattern (i.e. if both st and uv appear on
a prim, prefer uv over st when importing).

--spiff

nrusch:
Will do. I should hopefully have a PR ready in the next few days.

nrusch:
One question: Since the imported/exported primvars are governed by a pattern string (which defaults to `"Cd"`), I'm wondering if the following logic sounds sensible with regard to respecting the pattern:

- If both `uv` and `st` match the pattern, and `primvars:uv` exists on the prim, read it into `uv`.
[...]

PR #405, opened 2018-02-06
[houdini] Add basic support for writing GT_TYPE_TEXTURE attrs in 16.5+

nrusch:
### Description of Change(s)

Houdini 16.5 introduced a new `GT_TYPE_TEXTURE` storage type, which it uses for newly-created `uv` attributes. This change adds basic support for writing float attributes tagged with this type as `Float3/Float3Array` primvars.

### DIFF
--- a/third_party/houdini/lib/gusd/GT_Utils.cpp
+++ b/third_party/houdini/lib/gusd/GT_Utils.cpp
@@ -134,6 +134,12 @@ struct GtDataToUsdTypename
             = SdfValueTypeNames->Float3;
         m_typeLookup[KeyType(GT_STORE_REAL32, GT_TYPE_NONE, 3, true)]
             = SdfValueTypeNames->Float3Array;
+#if (GUSD_VER_CMP_2(>=,16,5))
+        m_typeLookup[KeyType(GT_STORE_REAL32, GT_TYPE_TEXTURE, 3, false)]
+            = SdfValueTypeNames->Float3;
+        m_typeLookup[KeyType(GT_STORE_REAL32, GT_TYPE_TEXTURE, 3, true)]
+            = SdfValueTypeNames->Float3Array;
+#endif
         // VectorFloat
         m_typeLookup[KeyType(GT_STORE_REAL32, GT_TYPE_VECTOR, 3, false)]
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internali ssue #156819.

spiffmon:
Hi Nathan,
   Assuming GT_TYPE_TEXTURE is actually indicative of texture coordinates
(online documentation seems lacking), could you provide some more context
of where further commits might take this feature? Some concerns I have are:

* texture coordinates should all be primvars in USD (we hope to provide a
TextureCoordinate "Role" soon.

* Exporting 2D uvs as float3's is going to be pretty tough on interchange;
I don't think we want to have to teach all importers and renderers  how to
do the conversion.

* If this new type is what comes out of a (nicely documented) UV Texture
sop, do any of the parameters (e.g. "Texture Type" need to be recorded onto
the primvar, or is all the input info baked into the TextureCoordinate
result?


--spiff

nrusch:
Hey @spiffmon,

This change only meant to preserve the status quo by allowing Gusd to export Houdini UV sets in 16.5+. This new type tag is now applied to the `uv` attributes that are created by Houdini's built-in UVW Unwrap tool, etc, so without this change, the export process will throw a `Can't find type name for primvar XX` warning and skip the attribute.

Prior to 16.5, these `uv` attributes were tagged as `GT_TYPE_VECTOR` instead; hence the use of `Float3`/`Float3Array` as the companion USD type.

I will reply to #362 with more thoughts on the attribute type conversions, name remapping, etc. as I am currently wading through all of that.

-Nathan

nrusch:
Between artists using the Houdini plugins more in production and my work with the Gusd code, I've found a number of holes in the current datatype support on both the import and export sides (e.g. RGBA attributes tagged as `GT_STORE_REAL32`/`GT_TYPE_COLOR` not exporting at all).

Thus, I'm closing this in anticipation of shortly opening a new PR with a broader set of changes to support a more complete set of `GT_Storage/GT_Type` export combinations and `SdfValueTypeName` import mappings, as well as adding awareness of `GT_TYPE_TEXTURE`.

PR #406, opened 2018-02-07
[usdHoudini] Creating a shading mode registry for houdini

sirpalee:
### Description of Change(s)
The PR is our initial take on a usdMaya-like shading mode registry for Houdini. To have the most flexibility for implementing all kinds of shader exporters, we decided to inject custom code into GusdROP_usdoutput::bindAndWriteShaders , via a simple function pointer. For simplicity, I avoided the class-based approach of usdMaya, because usdoutput's shader binding already happens in bulk.

We haven't tested this change with USD-Arnold Houdini related plugins, so currently, this PR is only opened for feedback, until we can confirm the current approach covers our needs.

### Fixes Issue(s)
No reported issues.

### DIFF
--- a/third_party/houdini/lib/gusd/CMakeLists.txt
+++ b/third_party/houdini/lib/gusd/CMakeLists.txt
@@ -41,6 +41,7 @@ pxr_shared_library(${PXR_PACKAGE}
         purpose
         refiner
         scopeWrapper
+        shadingModeRegistry
         shaderWrapper
         stageCache
         stageEdit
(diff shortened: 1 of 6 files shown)


jtran56:
Filed as internal issue #156961.

sirpalee:
Pushed back a working version (comments for the new classes are still missing), I think the change is ready for review. It does work with our plugin in usd-arnold. (https://github.com/LumaPictures/usd-arnold/blob/master/houdini/arnoldShadingExporter.cpp)

elrond79:
pushed a new version which adds some more documenation

elrond79:
pushed new version to fix some merge conflicts

c64kernal:
Houdini plugins have been removed from this repo. Closing this one out.

PR #408, opened 2018-02-08
add support for Boost_USE_STATIC_LIBS on Windows. Fixes #407

### LINKED ISSUE #407: Can't link against custom static boost build
### Description of Issue

When building USD against pre-built libraries using CMake, `Boost_USE_STATIC_LIBS` option is ignored, resulting in linking errors if the prebuilt boost libs were built as static libs.

The fix is fairly simple, and I have a fix (and pull request) ready to allow users to use the `Boost_USE_STATIC_LIBS`. I'm posting this issue as per the "Contributing to USD" guide :)

### Steps to Reproduce

1. Build Boost as static libraries only.
2. Build USD using the following options during configuration: `-DBOOST_ROOT=<path_to_boost> -DBoost_USE_STATIC_LIBS=On`
3. Results in `LINK : fatal error LNK1104: cannot open file 'boost_regex-vc140-mt-1_61.lib'`

Windows 10 (but probably on any platform)

### Package Versions

0.8.x and dev branch

ix-dcourtois:
### Description of Change(s)

`msvcdefaults.cmake` is always forcing linking the dynamic boost libs (dlls) This commit allows the user to specify the `Boost_USE_STATIC_LIBS` option to disable this behavior.

If the option is not set in CMake, the behavior does not change.

### Fixes Issue(s)

- #407

### DIFF
--- a/cmake/defaults/msvcdefaults.cmake
+++ b/cmake/defaults/msvcdefaults.cmake
@@ -82,8 +82,11 @@ _add_define("NOMINMAX")
 _add_define("YY_NO_UNISTD_H")
 
 # Forces all libraries that have separate source to be linked as
-# DLL's rather than static libraries on Microsoft Windows.
-_add_define("BOOST_ALL_DYN_LINK")
+# DLL's rather than static libraries on Microsoft Windows, unless
+# explicitely told otherwise.
+if (NOT ${Boost_USE_STATIC_LIBS})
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #156963.

ix-dcourtois:
Closing, I just realized that I submitted this PR against master instead of dev branch. I'll re-submit it against the correct branch.

PR #412, opened 2018-02-19
Feature: CMake option to override PXR_PLUGINPATH_NAME

ix-dcourtois:
### Description of Change(s)

This pull request exposes a CMake option to allow the user to modify the default `PXR_PLUGINPATH_NAME` env var to a custom one. It also documents this environment variable, since it's used by USD, but wasn't mentioned anywhere in the documentation.

I wasn't sure where to put the option, so if this PR needs modifications, I'm open to any suggestions.

Note: this pull request follows this thread of the `usd-interest` Google group: https://groups.google.com/forum/#!topic/usd-interest/K-Ore2N1xg0

### Fixes Issue(s)
- N/A


jtran56:
Filed as internal issue #157346.

cmake/defaults/CXXDefaults.cmake
```diff
@@ -82,6 +82,11 @@ else()
     set(PXR_PYTHON_SUPPORT_ENABLED "0")
 endif()
 
+# Set plugin path environment variable name
+if (PXR_PLUGINPATH_NAME)
```
sunyab: - Did you mean for this (and below) to be PXR_OVERRIDE_PLUGINPATH_NAME?

- I think we should use the _add_define helper here if we can, something like:

if (PXR_OVERRIDE_PLUGINPATH_NAME)
    _add_define(PXR_PLUGINPATH_NAME=${PXR_OVERRIDE_PLUGINPATH_NAME})
endif()

- Does this work properly if PXR_OVERRIDE_PLUGINPATH_NAME is set to a path that has whitespace in it, like "/tmp/USD Plugins" or something like that? I'm wondering if we might have to put quotes around the ${PXR_OVERRIDE_PLUGINPATH_NAME} above.

- For organization, could you move this block of code up to where other defines are added, around line 50?

ix-dcourtois: Hi, yes, you're totally right about that (at first I was planning to use the `PXR_PLUGINPATH_NAME` variable as the option name but I thought that would be a bit missleading, thus the `OVERRIDE` name for the option)
And the weird thing is: it seems to work as it is (that's why I didn't notice that I forgot to rename those) Not sure if my setup is broken or if it's a mystery of CMake.

Anyway to answer your points:

* totally right, as I said, a mistake, I'll change that
* noted, I'll use `_add_define`
* don't know. I'll have to check (I don't know how CMake behaves in this case)
* sure

ix-dcourtois:
I just noticed the changes in BUILDING.md ! I'm sorry, I didn't notice those changes ! I have no idea where they came from, the only chunk I commited was the paragraph about override plugin plath... I don't even have them in my current branch :/

I'm going to check what happened (and how I can fix this)

cmake/defaults/CXXDefaults.cmake
sunyab: Re: the spaces in the env setting name -- of course, I totally misread that. Thanks!

PR #413, opened 2018-02-22
[Usd] Add better static assert.

superfunc:
### Description of Change(s)

This static assert makes it clearer that get cannot be called
with a const T*. Before this, the static assert below would fail, but
the output error is much less clear.


waywardmonkeys:
Would this work better with a usage of `std::enable_if` or `std::conditional` ? Both are in C++11, so hopefully they have enough compiler support!

superfunc:
I don't think so, for a couple of reasons. First, I think `static_assert` produces nicer error messages. Second, the pattern in usd/attribute is to use `static_assert` for these kind of clauses so this matches that.

jtran56:
Filed as internal issue #157639.

PR #414, opened 2018-02-22
[usdHoudini]: Fix usage of nonexistent VtArray::GetSize() method

nrusch:
### Description of Change(s)

Fixes references to a nonexistent `VtArray::GetSize()` method by `GusdGT_VtArray`, `GusdUT_VtArrayRO`, and `GusdUT_VtArrayRW`. I'm guessing this was an old method, and that these instances may have been missed during a refactor.

PR #415, opened 2018-02-23
[usdHoudini] Translate st primvar to/from Houdini uv attribute

nrusch:
### Description of Change(s)

Adds basic translation between USD's `st` primvar and Houdini's `uv` attribute during import (unpacking) and export. This allows USD UV data to be preserved through (or manipulated by) Houdini, and for UVs generated in Houdini to be exported to USD's canonical primvar, without manual attribute wrangling.

#### Import

The unpacking logic will currently prefer a `uv` primvar if both `st` and `uv` are found and are selected for import. Otherwise, if `st` is found and is a `Float2Array` type, it will be "promoted" to a `Float3Array` in transit and applied to Houdini's `uv` attribute.

If `st` is *not* a `Float2Array`, it will still be remapped to Houdini's `uv` attribute, but no type conversion will be done and a warning will be issued. This is considered an edge case, and the current behavior may be considered undesirable, so it can be augmented or removed if necessary.

#### Export

The export logic is similar to import, although a bit more selective. If a `uv` attribute is being written from Houdini, and it is a float attribute scoped to either points or vertices with a tuple size >= 2, it will be translated to a `Float2Array` and written to the `st` primvar. It may be preferable for this check to be further limited to only consider an attribute with a tuple size of exactly 3.

#### Future

Eventually, we should be able to replace the uses of the `Float2Array` type name with a forthcoming `textureCoordinate2f ` (or similar) type, in order to make the behaviors even more explicit. Also, when reading `st` data to `uv` in Houdini 16.5+ (and possibly when reading `uv` as-is), I think we will want to mark it as `GT_TYPE_TEXTURE` (new in 16.5), in order to align with Houdini's own behavior for application-generated UV data.

### Fixes Issue(s)

Addresses #362, as well as fixing some references to a nonexistent `VtArray::GetSize()` method by other `Gusd` classes, which was temporarily open as #414.

### DIFF
--- a/third_party/houdini/lib/gusd/GT_VtArray.h
+++ b/third_party/houdini/lib/gusd/GT_VtArray.h
@@ -233,7 +233,7 @@ void
 GusdGT_VtArray<T>::swap(ArrayType& o)
 {
     _array.swap(o);
-    _size = _array.GetSize();
+    _size = _array.size();
     _UpdateDataPointer(false);
 }
 
--- a/third_party/houdini/lib/gusd/UT_VtArray.h
+++ b/third_party/houdini/lib/gusd/UT_VtArray.h
@@ -54,7 +54,7 @@ public:
     typedef typename ArrayType::const_reverse_iterator  const_reverse_iterator;
 
     GusdUT_VtArrayRO(const ArrayType& array)
-        : _array(array), _size(array.GetSize()), _data(array.cdata())
+        : _array(array), _size(array.size()), _data(array.cdata())
         { UT_ASSERT(_size == 0 || _data != NULL); }
 
     const T&                operator()(exint i) const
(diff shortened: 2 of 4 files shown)


jtran56:
Filed as internal issue #157640.

spiffmon:
I believe we ran into non-trivial test failures when incorporating this change, but our fx devs have been too overwhelmed with production work to dig in too deeply.

chrisrydalch:
@spiffmon what were some of the test failures encountered?

c64kernal:
We think this was addressed by commit 92996e58f926abafbf1084ba2b10371e7c6604b9.  I'll close out this PR, but if we missed something, please let me know!

nrusch:
Hey @c64kernal,

I haven't been staying on top of the Gusd changes, so I may have missed something, but I don't believe this one was addressed by the linked commit. This PR was dealing specifically with mapping USD's canonical UV primvar to Houdini's standard `uv` attribute to get (relatively) seamless UV interchange between USD and Houdini, in much the same way that `usdMaya` maintains a special awareness of the `"map1"` UV set.

That said, I realize these changes may not be worth keeping around, since the Gusd code has evolved quite a bit since it was opened. However, I do think #362 remains relevant.

c64kernal:
@nrusch  -- my sincere apologies, I got two issues mixed up.  You're totally right.  Re-opening this.

nrusch:
No longer relevant due to the Gusd code being taken over by SideFX.

c64kernal:
Thank you @nrusch

PR #422, opened 2018-03-09
Make the matching of alembic properties to the types expected by the USD

marktucker:
importer less strict. In particular, ignore the "interpretation" set on
the Alembic property. This doesn't affect the underlying data types, so
the data will still be readable by USD without issue. So if an Alembic file
has marked its point positions with a "vector" interpretation instead of
a "point" interpretation, the importer will still load them.


jtran56:
Filed as internal issue #158217.

PR #424, opened 2018-03-11
Fix simple lighting shader to compile under OSX

marktucker:
The OSX GLSL compiler doesn't like preprocessor directives like:
#if foo
when the macro "foo" has not been defined. This was showing up in the simple lighting shader when checking a couple of "HD_HAS_xxx" conditions, causing the shader to fail to compile. These check should be of the form:
#if defined(foo)
as all other checks for "HD_HAS_xxx" conditions are in other shaders.


jtran56:
Filed as internal issue #158260.

PR #425, opened 2018-03-12
Guard against empty tokenization in ApplyCollection

donboie:
### Description of Change(s)

Ensure ApplyCollection doesn't access an empty vector when called with an invalid name.


spiffmon:
Nice catch, Don - thank for finding and fixing!

--spiff

jtran56:
Filed as internal issue #158295.

sunyab:
We coincidentally fixed this up internally in commit 7319e3b00, so I'm going to close this out in favor of that change. Thanks again for the PR @donboie!

PR #426, opened 2018-03-13
Fixing IsRelativePath call in AnchorRelativePath

sirpalee:
### Description of Change(s)
The change updates ArDefaultResolver::AnchorRelativePath so it'll call the virtual IsRelativePath function, instead of explicitly calling ArDefaultResolver::IsRelativePath. This helps developers who want to replace ArDefaultResolver.


jtran56:
Filed as internal issue #158296.

PR #428, opened 2018-03-15
Use tbb::enumerable_thread_specific instead of C++ thread_local keyword

marktucker:
to make thread local data. OSX compilers don't all support thread_local, and this bit of code is
the only thing preventing USD from building on all OSX platforms we wish to support. I believe there
should be no impact on behavior from this change, and tbb::enumerable_thread_specific is already
used elsewhere in the USD code.

### DIFF
--- a/pxr/base/lib/tf/scopeDescription.cpp
+++ b/pxr/base/lib/tf/scopeDescription.cpp
@@ -32,6 +32,7 @@
 #include "pxr/base/arch/threads.h"
 
 #include <tbb/spin_mutex.h>
+#include <tbb/enumerable_thread_specific.h>
 
 #include <algorithm>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #158391.

meshula:
I imagine you need to build with Xcode 7, but I thought I'd leave this note here for others following the thread: Xcode 8 and greater have supported TLS since 2016 (https://developer.apple.com/videos/play/wwdc2016-405/?time=354)

marktucker:
Yes. It's really more of a legacy thing for us than a critical requirement (some devs are reluctant to upgrade older hardware), but given how simple the change is, I figured it wouldn't hurt to remain compatible with the older Xcode?

gitamohr:
Using enumerable_thread_specific here is going to be quite a bit slower than using thread_local, at least on linux.  It might not matter for most practical cases but making the scope description stuff as fast as possible was a goal so we could use them in more performance-sensitive code.  For example, we use them in the very low level data access routines in the usd crate file format implementation.

marktucker:
Ah, okay. I didn't realize there was a significant performance difference. Given the relative rate of occurrence of thread_local vs. tbb::enumerable_thread_specific in the USD code (3 of the former and about 15 of the latter) I assumed they were more or less interchangeable, and this instance was an unintended deviation from the majority approach. I'm sure I can work around this if you want to just reject this pull request. Thanks for the explanation.

sunyab:
Closing this PR out given the above discussion on thread_local vs. enumerable_thread_specific performance. Thanks all!

PR #429, opened 2018-03-15
Fix a bug where calling Reload on an Alembic file layer would not cause

marktucker:
hydra to update the display of the geometry until there was some other
edit to the stage.


jtran56:
Filed as internal issue #158393.

Issue #409, opened 2018-02-08
ALEMBIC_DIR not used when an Alembic lib is found in default paths


ix-dcourtois:
### Description of Issue

If you have a version of Alembic library installed in the default paths (`/usr/lib/` for instance on Linux, or in a folder accessible through the `%PATH%` env var on Windows) then the build CMake script will always use this one, not matter what you specify in `-DALEMBIC_DIR=...`

There are a few problems with the `FindAlembic.cmake` module:
- `find_library` is using `PATHS` to handle the user provided `ALEMBIC_DIR`, and those are used only if CMake cannot find the lib in the default paths. So if you have an Alembic lib installed in one of those default, the script will always link this one instead of the user provided one. This can be fixed by using `HINTS` instead of `PATHS`: Hints take precedence on default paths.
- `LIBRARY_PATHS` starts with the standard `/usr/lib`, etc. paths. So same as before, even if we use hints and specify a custom `ALEMBIC_DIR`, the one installed in `/usr/lib` will be used.
- The include directories are located using `find_path` at the end of the file, but with totally different parameters than the `find_library` call. And neither `PATHS` nor `HINTS` are used, thus any Alembic headers installed in default paths are found before the ones in `ALEMBIC_DIR\include`

I can work on a pull request, but I'd like to know if I should change the order of the paths where Alembic is looked for: I think it's best to first check in the user provided paths, and only when not found check the standard/default ones.

### Steps to Reproduce

1. On Windows, update `PATH` to point to a folder where an Alembic lib can be found (which happens if you have a previous install of USD, see the requirements here: http://graphics.pixar.com/usd/docs/Maya-USD-Plugins.html)
2. Build USD using `-DALEMBIC_DIR=<another_alemic_folder>` with another Alembic lib, and check that the Alembic library used is the one in your `PATH`

### System Information (OS, Hardware)

Windows 10 (but probably happens on all platforms)

### Package Versions

0.8.3

jtran56:
Filed as internal issue #156964.

sunyab:
This fix was released with v0.8.5, closing this out. Thanks!

PR #436, opened 2018-03-22
fix for issue 409: FindAlembic now uses the user provided one.

ix-dcourtois:
### Description of Change(s)

The FindAlembic.cmake module now looks for user provided Alembic dir before trying the standard paths. This allows to have a version of Alembic installed, while still being able to build USD against another locally built version.

### Fixes Issue(s)

- #409


jtran56:
Filed as internal issue #158734.

PR #439, opened 2018-03-23
[Plug] Update fwd decl.

superfunc:
### Description of Change(s)

Commit 9ad3c35335a38db94736ab3ee13b6fafa50956c0 changed this from
being a struct to a class. On certain compilers(clang in this case)
it warns on forward declarations in which the declared type doesn't
match the one in the definition.

### Fixes Issue(s)
- None


jtran56:
Filed as internal issue #158737.

PR #440, opened 2018-03-24
Fix GetDisplayName() for Anonymous Layers.

superfunc:
### Description of Change(s)

It was broken when the provided identifier had a colon in it,
which is legal in a posix file path.

e.g.

```python
from pxr import Usd
s = Usd.Stage.CreateInMemory('foo:foo.usda')
s.GetRootLayer().GetDisplayName()
```

Should return `foo:foo.usda`, but currently returns an empty string

### Fixes Issue(s)
- None

### DIFF
--- a/pxr/usd/lib/sdf/assetPathResolver.cpp
+++ b/pxr/usd/lib/sdf/assetPathResolver.cpp
@@ -212,6 +212,10 @@ Sdf_GetAnonLayerDisplayName(
     // We want to find the second occurence of ':', traversing from the left,
     // in our identifier which is of the form anon:0x4rfs23:displayName
     auto fst = std::find(identifier.begin(), identifier.end(), ':');
+    if (fst == identifier.end()) {
+        return std::string();
+    }
+
     auto snd = std::find(fst + 1, identifier.end(), ':');
     if (snd == identifier.end()) {
         return std::string();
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #158778.

superfunc:
Yeah I'll get that up tonight :v:

pxr/usd/lib/sdf/assetPathResolver.cpp
```diff
@@ -209,9 +209,15 @@ string
 Sdf_GetAnonLayerDisplayName(
     const string& identifier)
 {
-    if (std::count(identifier.begin(), identifier.end(), ':') == 2)
-        return identifier.substr(identifier.rfind(':') + 1);
-    return std::string();
+    // We want to find the second occurence of ':', traversing from the left,
+    // in our identifier which is of the form anon:0x4rfs23:displayName
+    auto fst = std::find(identifier.begin(), identifier.end(), ':');
```
sunyab: Can we bail here if fst == identifier.end() for safety? The old code handled the case where there were no ":" characters in the identifier (even if this should never happen), so I think it'd be good to follow suit.

superfunc: Done, updated. Thanks :)

PR #443, opened 2018-03-27
[maya] remove duplicate usdTranslatorExport_AppendFromTextField mel func

elrond79:
### Description of Change(s)

So, in our pr/export_parent_scope pull request (https://github.com/PixarAnimationStudios/USD/pull/223), we accidentally introduced a duplicate MEL function. Looks like we both implemented it, and then when we did a rebase, we didn't pick up the duplicated code. Apologies!


jtran56:
Filed as internal issue #158861.

Issue #410, opened 2018-02-09
Euler filter for XformCommonAPI


AlexSchwank:
### Description of Issue
When using more advanced xform ops, getting rotation values via the XformCommonAPI can lead to Euler flips. 

**eulerTest.usda [zip](https://github.com/PixarAnimationStudios/USD/files/1710506/eulerTest.zip)**

```
//...
float3 xformOp:rotateXYZ.timeSamples = {
    0: (0, 80, 0),
    1: (0, 90, 0),
    2: (0, 100, 0),
}
uniform token[] xformOpOrder = ["xformOp:translate", "xformOp:translate:rotatePivotOffset", "xformOp:translate:rotatePivot", "xformOp:rotateXYZ:rotateOffset", "xformOp:rotateXYZ", "!invert!xformOp:translate:rotatePivot", "xformOp:translate:scalePivotOffset", "xformOp:translate:scalePivot", "xformOp:transform:shear", "xformOp:scale:scaleOffset", "xformOp:scale", "!invert!xformOp:translate:scalePivot"]
//...
```

**XformCommonAPI::GetXformVectors() result**

```
0: (0, 80, 0)
1: (0, 90, -0)
2: (180, 80, 180)
```

For now, we can fix the rotation for example with Mayas MEulerRotation::setToClosestSolution() but Usd itself should include a 

feature for Euler filters.

Thanks,
Alex

### Steps to Reproduce

```
>>> from pxr import Usd, UsdGeom
>>> stage = Usd.Stage.Open("/path/to/eulerTest.usda")
>>> prim = stage.GetPrimAtPath("/myXform")
>>> xformAPI = UsdGeom.XformCommonAPI(prim)
>>> xformAPI.GetXformVectors(0)
(Gf.Vec3d(0.0, 0.0, 0.0), Gf.Vec3f(0.0, 80.0, 0.0), Gf.Vec3f(1.0, 1.0, 1.0), Gf.Vec3f(0.0, 0.0, 0.0), UsdGeom.XformCommonAPI.RotationOrderXYZ)
>>> xformAPI.GetXformVectors(1)
(Gf.Vec3d(0.0, 1.0, 0.0), Gf.Vec3f(0.0, 90.0, -0.0), Gf.Vec3f(1.0, 1.100000023841858, 1.0), Gf.Vec3f(0.0, 0.0, 0.0), UsdGeom.XformCommonAPI.RotationOrderXYZ)
[...]

PR #450, opened 2018-04-09
[maya] added euler filter for imports

AlexSchwank:
### Description of Change(s)
Added Euler filter to Maya import to fix potential Euler flips for animated rotations. 
The filter is using Maya's `MEulerRotation::setToClosestSolution()`.
Added _useEulerFilter_ option to toggle the filter.  

This helps with #410 but obviously only on the Maya side.

### DIFF
--- a/third_party/maya/lib/usdMaya/translatorXformable.cpp
+++ b/third_party/maya/lib/usdMaya/translatorXformable.cpp
@@ -288,7 +288,19 @@ static bool _pushUSDXformOpToMayaXform(
                                 PxrUsdMayaXformStack::RotateOrderFromOpType<MEulerRotation::RotationOrder>(
                                         xformop.GetOpType());
                         MEulerRotation eulerRot(xValue[i], yValue[i], zValue[i], MrotOrder);
+
+                        // For all rotations except the first one, we can check for Euler flips.
+                        // The first rotation can't be validated against a previous value so we always take it as it is.
+                        if (args.GetUseEulerFilter() && i > 0)
+                        {
+                            size_t previousRotationIndex = i - 1;
+                            MEulerRotation previousRotation(xValue[previousRotationIndex],
+                                                            yValue[previousRotationIndex],
+                                                            zValue[previousRotationIndex], MrotOrder);
+                            eulerRot.setToClosestSolution(previousRotation);
+                        }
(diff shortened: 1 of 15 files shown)


jtran56:
Filed as internal issue #159381.

third_party/maya/lib/usdMaya/translatorXformable.cpp
```diff
@@ -288,7 +288,19 @@ static bool _pushUSDXformOpToMayaXform(
...
+                        // For all rotations except the first one, we can check for Euler flips.
+                        // The first rotation can't be validated against a previous value so we always take it as it is.
+                        if (args.GetUseEulerFilter() && i > 0)
+                        {
+                            size_t previousRotationIndex = i - 1;
+                            MEulerRotation previousRotation(xValue[previousRotationIndex],
+                                                            yValue[previousRotationIndex],
+                                                            zValue[previousRotationIndex], MrotOrder);
+                            eulerRot.setToClosestSolution(previousRotation);
+                        }
```
elrond79: These changes won't actually do nothing, because rotateAxis is internally stored as a quaternion, so 360 flips change nothing.  ie, try this in a python prompt:

```
cmds.setAttr('persp.rotateAxis', 720, 0, 0)
print cmds.getAttr('persp.rotateAxis')
```

elrond79:
A note: given that the changes in _pushUSDXformOpToMayaXform actually do nothing (see my note), the only changes are to _pushUSDXformToMayaXform... which should only get invoked for xform stacks that aren't compatible with the maya xform stack.

So the behavior currently is that if eulerFilter is on, then non-compatible xforms will be filtered, and compatible ones won't be.  This is actually the behavior I would want by default, but might be somewhat confusing... there might be situations where someone passes in a compatible xform stack (ie, nothing but a rotate X of [350, 360, 10], and might actually want euler filtering on.  Currently, they won't get it.  I think the ideal behavior would be to make the option a tri-mode:

on: always filter, for both compatible and incompatible xform stacks
auto: filter incompatible stacks, don't filter compatible stacks (default)
off: never filter either

AlexSchwank:
Thanks @elrond79 for your notes and for posting the explanation why we need the Euler filter.
I'm not sure how often one would want to filter compatible xform stacks but I can certainly add it. While doing that, I'll take another look at `_pushUSDXformOpToMayaXform`.

PR #451, opened 2018-04-09
[python] replaced single quoted doc strings

AlexSchwank:
### Description of Change(s)
Replaced single quoted docstrings with double qoutes.


jtran56:
Filed as internal issue #159382.

PR #455, opened 2018-04-09
[python] mutable args, duplicate dict keys. list/dict literals

AlexSchwank:
### Description of Change(s)
 - fixed mutable function args
 - fixed duplicate dict keys
 - used list/dict literals


jtran56:
Filed as internal issue #159386.

PR #459, opened 2018-04-10
[python] removed unnecessary backslashes

AlexSchwank:
### Description of Change(s)
Removed unnecessary backslashes.


jtran56:
Filed as internal issue #159452.

PR #463, opened 2018-04-17
[usdAlembic] Preferring st over uv attribute if present for UV writes.

sirpalee:
### Description of Change(s)
The following change modifies the behaviour of the alembicWriter plugin so that it prefers the "st" attribute over the "uv" attribute. When exporting USD from any 3rd party application, the plugins write uvs to the "st" attribute not to "uv".

### Fixes Issue(s)
No issues reported.

### DIFF
--- a/pxr/usd/plugin/usdAbc/CMakeLists.txt
+++ b/pxr/usd/plugin/usdAbc/CMakeLists.txt
@@ -62,11 +62,12 @@ pxr_plugin(usdAbc
 )
 
 pxr_test_scripts(
-    testenv/testUsdAbcAlembicData.py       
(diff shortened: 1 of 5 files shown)


jtran56:
Filed as internal issue #159711.

spiffmon:
Hey Pal,
   This change looks good, though it would be great if there were a unit test for it?

sirpalee:
Thanks! I didn't notice the tests when doing the original bugfix but pushed back a fixed version now. Also, while implementing the test suite I saw some incorrect behaviour, so I ended up redesigning how the remapping of st -> uv works. No more leftover primvars, and the case when the uv and st attribute are both on the mesh the export behaves better.

PR #465, opened 2018-04-17
[usdHoudini] Fixing treeview for USD Import and not installing the gusd python files twice

sirpalee:
### Description of Change(s)
The PR improves on a recent change that introduced a gusd py module to the build and started installing the gusd python files in the standard location. There are two main differences, changing how UsdImport.pypanel imports the treeview module, so there won't be an error message using the new module, secondly stop installing the same python files into third_party/Houdini/scripts/python/gusd. Installing the same files twice could confuse users, and error messages, as these scripts try to load the _gusd module, even if it's not present.

### Fixes Issue(s)
No reported issues.

### DIFF
--- a/third_party/houdini/lib/gusd/CMakeLists.txt
+++ b/third_party/houdini/lib/gusd/CMakeLists.txt
@@ -115,12 +115,3 @@ install(
     DESTINATION
         ${PXR_INSTALL_SUBDIR}/config/Icons
 )
-
-install( 
-    FILES
-        __init__.py
-        treeview.py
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #159713.

sunyab:
Hi @sirpalee, I'm going to close this out in favor of commit 81d8955, which fixes the issues you described but changes the 'gusd' package to 'Gusd'. Thanks again!

sirpalee:
Thanks for looking into this!

Issue #97, opened 2016-11-02
Hydra can't open tx textures


sirpalee:
Hydra can't open tx textures. It's commonly used, so it might be a good idea just to enable it. It's just adding the tx extension to the supported formats by Glf_OIIOImage.

```

---
 pxr/imaging/lib/glf/plugInfo.json | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/pxr/imaging/lib/glf/plugInfo.json b/pxr/imaging/lib/glf/plugInfo.json
index ce86c17..bb5a20f 100644
--- a/pxr/imaging/lib/glf/plugInfo.json
+++ b/pxr/imaging/lib/glf/plugInfo.json
@@ -6,7 +6,7 @@
                 "Types": {                                                                  
                     "Glf_OIIOImage" : {
                         "bases": ["GlfImage"],
-                        "imageTypes": ["bmp", "exr", "jpg", "png", "tif", "zfile"],
+                        "imageTypes": ["bmp", "exr", "jpg", "png", "tif", "zfile", "tx"],
                         "precedence": 0
                     },
                     "GlfPtexTexture" : {
-- 
2.7.4
```

jtran56:
Filed as internal issue #139514.

meshula:
This one is interesting. Wouldn't it make sense for glf to ask the linked OIIO for a list of readable imageTypes, instead of hard coding it in the plugInfo.json file?

sirpalee:
@meshula Right now hydra doesn't support all the data kinds that oiio does (like field3d), and also ptex is implemented in a separate class. So you'll end up manually specifying the formats you are interested in (or excluding some of them), like now.

meshula:
Makes sense. Thanks

c64kernal:
Thanks yeah -- and there's a bit of a chicken and egg problem where OIIO is implemented as a "plugin" in our system that we would (in theory) avoid loading altogether until we knew we needed it (based on matching manually listed extensions in the plugInfo file).  However, for now that isn't being taken advantage of, of course, hence the word plugin in quotes :)

davidgyu:
One bit of confusion is that both ".tx" and ".tex" filename extensions have been used to designate files created by prman's 'txmake' using prman's proprietary texture image format. OIIO has support for ".tx" files compatible with OIIO's 'maketx'.

[...]

PR #467, opened 2018-04-17
[usdImaging] Enabling tx support in imaging.

sirpalee:
### Description of Change(s)
The PR add tx to the list of supported textures to fix issue #97 . The conclusion there was to manually maintain the supported texture format lists because listing all the supported formats from OIIO would lead loading the incorrect texture formats using Glf_OIIOImage. (like ptex)

### Fixes Issue(s)
- #97


jtran56:
Filed as internal issue #159716.

PR #470, opened 2018-04-21
[usd] Add dumpcrate utility to build scripts.

superfunc:
### Description of Change(s)

usddumpcrate wasn't being built by cmake, this change fixes that.

### Fixes Issue(s)
- No filed issues I know of.


jtran56:
Filed as internal issue #159945.

PR #472, opened 2018-04-23
Fix typos.

waywardmonkeys:
Fix more typos.


jtran56:
Filed as internal issue #159947.

PR #473, opened 2018-04-23
Change "OS X" text to "macOS"

waywardmonkeys:
Apple calls their OS "macOS" now, so follow suit.


jtran56:
Filed as internal issue #159948.

waywardmonkeys:
Updated to apply to `HEAD`.

PR #475, opened 2018-04-25
[usdMaya] add filterTypes option to avoid exporting certain node types

elrond79:
### Description of Change(s)
Adds a new flag, "filterTypes", which allows you to avoid exporting nodes of certain maya node types.
This is handy for, ie, constraints, which are exported as transforms, but generally have no xform data of their own.  Particularly nice if you use mergeTransformAndShape, and want to avoid having your camera renamed just because it has a constraint on it.  (I would like to set filterTypes="constraint" by default, but this would technically break backwards compatibility.  It seems unlikely anyone is relying on exporting the constraint nodes themselves, since all their useful data is NOT exported, but I suppose anything is possible...)
Likely also useful for other scenarios, as well (ie, maybe you don't want to export cameras, or place3dTexture nodes, etc)

### Fixes Issue(s)
- If you have mergeTransformAndShape on, adding a constraint to a camera will change the USD camera node's name - ie, from "/persp" to "/persp/perspShape"; setting filterTypes="constraint" will avoid this


jtran56:
Filed as internal issue #160429.

third_party/maya/lib/usdMaya/JobArgs.cpp
```diff
@@ -417,6 +423,44 @@ const VtDictionary& JobExportArgs::GetDefaultDictionary()
...
+    }
+
+    for (unsigned int i=0; i < inheritedTypes.length(); ++i) {
+        if (inheritedTypes[i].length() == 0) continue;
+        id = MNodeClass(inheritedTypes[i]).typeId().id();
+        if (id == 0) {
+            // Unfortunately, the returned list will often include weird garbage, like
+            // "THconstraint" for "constraint", which cannot be converted to a MNodeClass,
+            // so just ignore these...
+            // MGlobal::displayError(MString("Given inherited excluded node type '") + inheritedTypes[i]
```
sunyab: Is keeping this commented-out displayError call intentional?

third_party/maya/lib/usdMaya/JobArgs.h
```diff
@@ -168,12 +171,30 @@ struct JobExportArgs
     PXRUSDMAYA_API
     static const VtDictionary& GetDefaultDictionary();
 
+    PXRUSDMAYA_API
```
sunyab: For style consistency with the rest of the API, could these be "AddFilteredTypeName", "GetFilteredTypeIds", and "ClearFilteredTypeIds"?

third_party/maya/lib/usdMaya/JobArgs.h
```diff
@@ -168,12 +171,30 @@ struct JobExportArgs
...
 private:
     PXRUSDMAYA_API
     JobExportArgs(
         const VtDictionary& userArgs,
         const PxrUsdMayaUtil::ShapeSet& dagPaths,
         const GfInterval& timeInterval);
+
+    // Maya type ids to avoid exporting; these are
+    // EXACT types, though the only exposed way to modify this,
+    // addFilteredTypeName, will also add all inherited types
```
sunyab: Need to fix up this reference to addFilteredTypeName to AddFilteredTypeName per above

third_party/maya/lib/usdMaya/JobArgs.h
sunyab: Also, could we add docs to AddFilteredTypeName about the inherited type behavior, e.g:
```
/// Adds type name to filter out during export. This will also add all 
/// inherited types (so if you exclude "constraint", it will also exclude 
/// "parentConstraint")
```

third_party/maya/lib/usdMaya/JobArgs.h
```diff
@@ -168,12 +171,30 @@ struct JobExportArgs
...
+    // (so if you exclude "constraint", it will also exclude
+    // "parentConstraint")
+    std::set<unsigned int> filteredTypeIds;
```
sunyab: Our coding style uses "_" as a prefix for private members, so could we rename this to _filteredTypeIds?

third_party/maya/lib/usdMaya/usdTranslatorExport.mel
```diff
@@ -142,6 +142,11 @@ global proc int usdTranslatorExport (string $parent,
         checkBox -l "Export Visibility" exportVisibilityCheckBox;
 
         checkBox -l "Renderable Only" renderableOnlyCheckBox;
+
+        textFieldGrp -l "Filtered Node Types"
+            -annotation ("Do not filter nodes of this type, or their children;\n" +
```
sunyab: This annotation seems incorrect, should the description be something like "Filter out nodes of this type and their children" or "Do not import nodes of this type or their children"?

third_party/maya/lib/usdMaya/JobArgs.cpp
elrond79: It was - I liked that it was sort of descriptive of what was going on, but it's not really necessary. I'll get rid of it.

sirpalee:
Updated the PR to make it work with the latest usdMaya changes.

sunyab:
Merged this PR, closing!

PR #478, opened 2018-05-02
[Usd] Add debug information for schema reg.

superfunc:
### Description of Change(s)
This adds debug information that I've come to find helpful when developing new, out of branch schemas.

### Fixes Issue(s)
- None filed.


jtran56:
Filed as internal issue #160325.

PR #479, opened 2018-05-03
Export more methods from the light adapter class so it can be subclassed

marktucker:
Export more methods from the light adapter class so it can be subclassed in other libraries.


jtran56:
Filed as internal issue #160402.

PR #480, opened 2018-05-05
[Usd] Provide a quiet mode for usdGenSchema.

superfunc:
### Description of Change(s)
Pretty non-invasive change to provide a quiet mode to usdGenSchema. This would be really useful to have. While this could be done with whatever script is calling it, it gives users a much simpler interface for doing so at a low cost to us(in terms of code complexity etc).

### Fixes Issue(s)
- None filed.


jtran56:
Filed as internal issue #160428.

PR #486, opened 2018-05-09
[HD] Setting up a default fallbackValue for shader parameters when connected.

sirpalee:
### Description of Change(s)
In cases when shader parameters are connected to textures, but no fallbackValue is set, CodeGen fails to generate the right function signature (missing the return type) causing a compilation error. The change uses the type of the input value and sets the fallbackValue to that type's default Value when a fallback value is not authored on the shader.

### Fixes Issue(s)
Nothing reported.


jtran56:
Filed as internal issue #160609.

PR #487, opened 2018-05-11
[Pcp] Remove use of std::auto_ptr.

superfunc:
### Description of Change(s)
This is deprecated in C++11, and removed in C++17, so we shouldn't rely
on it (source: https://en.cppreference.com/w/cpp/memory/auto_ptr).

Note that this uses shared_ptr, we could use unique_ptr if we relied on
a newer version of boost, as far as I know.

### Fixes Issue(s)
- Future proofing.

### DIFF
--- a/pxr/usd/lib/pcp/wrapCache.cpp
+++ b/pxr/usd/lib/pcp/wrapCache.cpp
@@ -45,14 +45,14 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
 
-static std::auto_ptr<PcpCache>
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #160644.

sunyab:
We decided not to take this pull request, as boost::python does not officially support std::shared_ptr until 1.63 -- see boostorg/python#29. On closer inspection though, it turns out we can simplify this code and remove std::auto_ptr anyway. See commit d1d115c7. Closing, thanks!

superfunc:
As long as `auto_ptr` goes away :)

PR #488, opened 2018-05-11
[USD] Supporting animated textures in hydra.

sirpalee:
### Description of Change(s)
The PR adds support for animated textures. There are three changes I had to do to make this happen.

- The texture id hashing function only takes the full path to the texture property into account (+ some extra salt). I modified the function to also use the texture path for the hash.
- The hydraMaterialAdapter was only checking for animated parameters on itself, but it wasn't traversing connection. Note, I did not add fully recursive traversal here, with the current limitation of shader networks this is not required.
- When marking the material dirty, I also mark any texture shader with animated file path dirty. Currently, I'm only doing that in one of the branches in UsdImagingGLHydraMaterialAdapter::MarkDirty. Should I do it for both cases?

At this moment there is no test for this new feature. I'll update the PR soon, but before that, I would like to get some feedback about the implementation itself and the last point in the meantime.

### Fixes Issue(s)
no issue reported

### DIFF
--- a/pxr/usdImaging/lib/usdImagingGL/hydraMaterialAdapter.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/hydraMaterialAdapter.cpp
@@ -39,6 +39,8 @@
 #include "pxr/usd/usdHydra/tokens.h"
 #include "pxr/usd/usdShade/connectableAPI.h"
 
+#include <functional>
+
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #160645.

sirpalee:
Attached a very basic example.

[animated_textures.tar.gz](https://github.com/PixarAnimationStudios/USD/files/2061477/animated_textures.tar.gz)

PR #502, opened 2018-05-22
 Add --build-debug option for building with debugging information

BrianHarris:
Add --build-debug option for building with debugging information
Fix iterator access after container modification (fixes debug assert)


jtran56:
Filed as internal issue #161094.

PR #503, opened 2018-05-22
Automatically set PYTHON_EXECUTABLE

BrianHarris:
Automatically set PYTHON_EXECUTABLE based on the python used to launch the build script


jtran56:
Filed as internal issue #161095.

BrianHarris:
Ah, you're right. -DUSE_PYTHON=OFF was added in 0.8.4 (Internal change: 1831920) and our fix was originally added in 0.8.3. We never removed our change after updating to latest.

sunyab:
Thanks for the update!

PR #504, opened 2018-05-22
[usdMaya] fix for VP2 selection when MAYA_VP2_USE_VP1_SELECTION is off

elrond79:
### Description of Change(s)

Currently, if using "pure" VP2 selection (ie, MAYA_VP2_USE_VP1_SELECTION is not
set), then if the "softSelectable" attribute is off, the object cannot be
selected by clicking on it in the viewport.

This is because proxyShape attempts to control soft-selection via
MPxSurfaceShape::getShapeSelectionMask().  Unfortunately, the intent of this
function is to control which selection filters the shape responds to, for any
selection (ie, normal + soft selection), and it is only a quirk / bug of
MAYA_VP2_USE_VP1_SELECTION mode that this setting is queried for soft selection,
but not normal selection.

To make the softSelectable attr work in both pure VP2 and
MAYA_VP2_USE_VP1_SELECTION modes, we now check if object-mode soft-selection is
currently on, and if the object is set to be not selectable; if so, we return
an empty mask, disabling selection.

### Fixes Issue(s)
- Selection when using "pure" VP2 mode (no MAYA_VP2_USE_VP1_SELECTION), and softSelectable attribute is off

### DIFF
--- a/third_party/maya/lib/pxrUsdMayaGL/batchRenderer.h
+++ b/third_party/maya/lib/pxrUsdMayaGL/batchRenderer.h
@@ -49,6 +49,7 @@
 #include <maya/MDrawContext.h>
 #include <maya/MDrawRequest.h>
 #include <maya/MObjectHandle.h>
+#include <maya/MMessage.h>
 #include <maya/MSelectionContext.h>
 #include <maya/MTypes.h>
 #include <maya/MUserData.h>
@@ -217,6 +218,11 @@ public:
             const GfMatrix4d& projectionMatrix,
             GfVec3d* hitPoint);
 
+    /// Returns whether soft selection for proxy shapes is currently enabled.
+    PXRUSDMAYAGL_API
+    inline bool GetObjectSoftSelectEnabled()
+    { return _objectSoftSelectEnabled; }
+
 private:
 
     friend class TfSingleton<UsdMayaGLBatchRenderer>;
@@ -279,6 +285,12 @@ private:
             MHWRender::MDrawContext& context,
             void* clientData);
 
+    /// Record changes to soft select options
+    ///
+    /// This callback is just so we don't have to query the soft selection
+    /// options through mel every time we have a selection event.
+    static void _OnSoftSelectOptionsChangedCallback(void* clientData);
+
     /// Perform post-render state cleanup.
     ///
     /// For Viewport 2.0, this method gets invoked by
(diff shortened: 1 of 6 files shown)


jtran56:
Filed as internal issue #161149.

third_party/maya/lib/pxrUsdMayaGL/batchRenderer.h
```diff
@@ -47,8 +47,10 @@
 
 #include <maya/M3dView.h>
 #include <maya/MDrawContext.h>
+#include <maya/MDrawContext.h>
```
sunyab: Duplicate include from above

third_party/maya/lib/pxrUsdMayaGL/batchRenderer.h
```diff
@@ -217,6 +219,13 @@ class UsdMayaGLBatchRenderer : public TfSingleton<UsdMayaGLBatchRenderer>
             const GfMatrix4d& projectionMatrix,
             GfVec3d* hitPoint);
 
+    /// Gets the UsdMayaGLSoftSelectHelper that this batchRenderer maintains.
```
sunyab: Looks like this docstring needs to be updated, since this returns a bool and is also used outside of PxrMayaHdShapeAdapter (in proxyShapeDelegate.cpp)

third_party/maya/lib/usdMaya/proxyShape.cpp
```diff
@@ -1009,13 +1032,29 @@ UsdMayaProxyShape::~UsdMayaProxyShape()
...
+
+    // However, due to a bug / quirk, it could be used to specifically control
+    // whether the object was SOFT-selectable if you were using
+    // MAYA_VP2_USE_VP1_SELECTON; in this mode, this setting is NOT querierd
+    // when doing "normal" selection, but IS queried when doing soft
+    // selection.
+
+    // Unfortunately, it is queried for both "normal" selection AND soft
+    // selection if you are using "true" VP2 selection.  So in order to
+    // control soft seleciton, in both modes, we keep track of whether
```
sunyab: Small typo: "seleciton" instead of "selection"

third_party/maya/lib/usdMaya/proxyShape.cpp
```diff
@@ -1009,13 +1032,29 @@ UsdMayaProxyShape::~UsdMayaProxyShape()
...
+    // we currently have object soft-select enabled, and then return an empty
+    // selection mask if it is, but this object is set to be non-soft-selectable
+
+    static MSelectionMask emptyMask;
```
sunyab: Could we make these two statics be static const?

elrond79:
Done! Thanks for catching those.  If you want to see the incremental changes, look here:

https://github.com/LumaPictures/USD/commit/9e8f1ea8f4e0b47198b157bb69fa5937b490baa0

PR #514, opened 2018-05-29
First crack at a standard USD volume schema

marktucker:
Adding schemas for Fields, which represent a single component of a Volume (density, color, etc). FieldAsset subclass specifies an asset path for the field data. OpenVDBAsset and Field3DAsset subclasses of FieldAsset add additional attributes for specifying a single field within an OpenVDB or Field3D file.

The Volume schema itself is an arbitrary set of relationships (field:fieldname) that allow individual field primitives to be tied to the volume with names that can be used to tie the fields to shader parameters.


jtran56:
Filed as internal issue #161285.

sunyab:
Closing this pull request out in favor of combined pull request #567.

PR #519, opened 2018-06-01
Removing the whitespace on the standards compliant flags - MSVC

boberfly:
Description of Change(s)
Removing the whitespace on the standards compliant flags to fix build runs which use NMake Makefiles or NMake Makefiles JOM on Windows.

This is against the dev branch.

Fixes Issue(s)
pxr\base\lib\arch\CMakeFiles\arch.dir\flags.make(5) : fatal error U1033: syntax error : '/Zc' unexpected

Cheers


jtran56:
Filed as internal issue #161426.

PR #520, opened 2018-06-01
[usdAbc] texture coordinates' topology (e.g. indices) is now preserved

ix-dcourtois:
### Description of Change(s)

This pull request fixes texture coordinates loaded through the Alembic plugin. Previously the UVs were expanded, meaning that the topology information (basically the indices) were lost, resulting in bad UVs when using subdivisions.

The pull request introduces a way to read indices from indexed Alembic properties, and use it on texture coordinates.

### Fixes Issue(s)

- No issue on GitHub, but the subject was raised in the Google Groups forum: https://groups.google.com/forum/#!topic/usd-interest/ZdWkMKTHZu4


pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -3086,11 +3134,25 @@ _ReadPolyMesh(_PrimReaderContext* context)
...
-    context->AddProperty(
-        _GetUVPropertyName(),
-        _GetUVTypeName(),
-        _CopyGeneric<IV2fGeomParam, GfVec2f>(
-            context->ExtractSchema("uv")));
+
+    // special case for texture coordinates: we want to preserve the topology,
+    // so we need to extract the indices properties if they exist.
+    auto uvProperty = context->ExtractSchema("uv");
+    if (uvProperty.Cast<IV2fGeomParam>().isIndexed()) {
```
spiffmon: I'm not super familiar with the Alembic API, but if there's any chance that Cast() can fail, maybe test the result before dereferncing?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -3086,11 +3134,25 @@ _ReadPolyMesh(_PrimReaderContext* context)
...
+        context->AddProperty(
+            TfToken(_GetUVPropertyName().GetString() + ":indices"),
+            SdfValueTypeNames->IntArray,
+            _CopyIndices<IV2fGeomParam>(uvProperty));
+    } else {
+        context->AddProperty(
+            _GetUVPropertyName(),
+            _GetUVTypeName(),
+            _CopyGeneric<IV2fGeomParam, GfVec2f>(uvProperty));
+    }
```
spiffmon: This is a nice chunk of code that:
a) is duplicated below
b) since UsdGeom and AbcGeom both support indexing for all primvars, we might want to apply to other primvars in the future...

Can you factor it out into a templated helper function, please?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -3086,11 +3134,25 @@ _ReadPolyMesh(_PrimReaderContext* context)
...
+    // special case for texture coordinates: we want to preserve the topology,
+    // so we need to extract the indices properties if they exist.
+    auto uvProperty = context->ExtractSchema("uv");
+    if (uvProperty.Cast<IV2fGeomParam>().isIndexed()) {
+        context->AddProperty(
+            _GetUVPropertyName(),
+            _GetUVTypeName(),
+            _CopyGeneric<IV2fGeomParam, GfVec2f, false>(uvProperty));
+        context->AddProperty(
+            TfToken(_GetUVPropertyName().GetString() + ":indices"),
```
spiffmon: There's a UsdGeomTokens->indices, which you can use in combination with [SdfPath::JoinIdentifier)[https://github.com/PixarAnimationStudios/USD/blob/722956654a7090c23ac33d6e52ae93e05e38f94b/pxr/usd/lib/sdf/path.h#L597]

jtran56:
Filed as internal issue #161427.

ix-dcourtois:
Hi @spiffmon 

Thanks for taking the time to review the code. I knew it was going to need a lot of modifications, but I didn't know where to start :)

- about the duplication of code and generic primvars: are you suggesting a `_ReadGenricPrimvars` generic method that I could just invoke in both `_ReadPolyMesh` and `_ReadSubd` and would take care of indices ? 

- about the dereferencing, it should already be taken care of, see : https://github.com/PixarAnimationStudios/USD/blob/722956654a7090c23ac33d6e52ae93e05e38f94b/pxr/usd/plugin/usdAbc/alembicReader.cpp#L472

- Yes, I did notice the arrival of the texture coordinate role, greatly appreciated :) But I didn't notice a role API in the Alembic plugin. If it's not too complicated, I could add an optional role to the `_ReadGenricPrimvars` helper ?

- And for the tests, I'll add one. It's just that I'm building USD against our libs, and I do not have Python support enabled (we're using a static boost build, which causes a lot of problems on Windows with Python)

Thanks again for the feedback (not sure how long it will take to fix the aforementioned problems, I do not have much time allocated for this issue currently, but I'll do my best !)

spiffmon:
>
>    -
>
>    about the duplication of code and generic primvars: are you suggesting
>    a _ReadGenricPrimvars generic method that I could just invoke in both
>    _ReadPolyMesh and _ReadSubd and would take care of indices ?
>
> ​Yes, or even a method on the context, like context->AddPrimvar(), which
would always check for indexing?​


>
>    -
>    -
>
>    about the dereferencing, it should already be taken care of, see :
>    https://github.com/PixarAnimationStudios/USD/blob/
>    722956654a7090c23ac33d6e52ae93e05e38f94b/pxr/usd/plugin/
>    usdAbc/alembicReader.cpp#L472
>    <https://github.com/PixarAnimationStudios/USD/blob/722956654a7090c23ac33d6e52ae93e05e38f94b/pxr/usd/plugin/usdAbc/alembicReader.cpp#L472>
>
> ​Ah, yes - thanks for pointing that out!​


>
>    -
>    -
>
>    Yes, I did notice the arrival of the texture coordinate role, greatly
>    appreciated :) But I didn't notice a role API in the Alembic plugin. If
>    it's not too complicated, I could add an optional role to the
>    _ReadGenricPrimvars helper ?
>
> ​It's taken care of by _GetUVTypeName()​


>
>    -
>    -
>
>    And for the tests, I'll add one. It's just that I'm building USD
>    against our libs, and I do not have Python support enabled (we're using a
>    static boost build, which causes a lot of problems on Windows with Python)
>
> ​Got it.  As long as you can generate the USD file corresponding to your
abc test asset, you can just follow the pattern of testUsdAbcInstancing,
which does a baseline comparison, and we can make any corrections needed to
ensure the test runs properly.
​

​--spiff​

spiffmon:
Looks awesome, @ix-dcourtois - thanks much!  We'll get this in the PR queue!

PR #524, opened 2018-06-08
[usd] Block Authoring Prims on Inactive Ancestor.

superfunc:
### Description of Change(s)

Currently, authoring a prim via DefinePrim with a pre-existing non-active ancestor would throw a TfRuntimeError, _but_ would still complete the authoring. To reproduce run:

```
from pxr import Tf, Usd
p = s.DefinePrim('/p')
p.SetActive(False)
try:
    g = s.DefinePrim('/p/g')
except Tf.ErrorException as e:
    print 'oops: ', str(e)
print '-------'
print s.GetRootLayer().ExportToString()
```

and get

```
oops:
        Error in 'pxrInternal_v0_8__pxrReserved__::UsdStage::_DefinePrim' at line 3076 in file pxr/usd/lib/usd/stage.cpp : 'Failed to define UsdPrim </p/g>'
-------

def "p" (
   active = false
)
{
   def "g"
   {
   }
}
```

With this change the behavior now results in a more reasonable error message, and the authoring does not take place.

```
oops:
	Error in 'pxrInternal_v0_8__pxrReserved__::UsdStage::_DefinePrim' at line 3172 in file pxr/usd/lib/usd/stage.cpp : 'Failed to create UsdPrim </p/g>. Cannot author a UsdPrim under an inactive ancestor </p>.'
-------

def "p" (
    active = false
)
{
}
```

### Fixes Issue(s)
- None filed, ran into in the wild

### DIFF
--- a/pxr/usd/lib/usd/stage.cpp
+++ b/pxr/usd/lib/usd/stage.cpp
@@ -1239,6 +1239,18 @@ UsdStage::_ValidateEditPrimAtPath(const SdfPath &primPath,
         return false;
     }
 
+    if (primPath.IsRootPrimPath()) {
+        return true;
+    } else {
+        SdfPath parentPath = primPath.GetParentPath();
+        if (!GetPrimAtPath(parentPath).IsActive()) {
+            TF_CODING_ERROR("Failed to create UsdPrim <%s>. "
+                            "Cannot author a UsdPrim under an inactive ancestor <%s>.", 
+                            primPath.GetText(), parentPath.GetText());
+            return false;
+        }
+    }
+
     return true;
 }
 
@@ -3181,20 +3193,10 @@ UsdStage::_DefinePrim(const SdfPath &path, const TfToken &typeName)
     }
 
     // Define all ancestors.
-    auto parentPath = path.GetParentPath();
-    if (!_DefinePrim(parentPath, TfToken())) {
+    if (!_DefinePrim(path.GetParentPath(), TfToken())) {
         return UsdPrim();
     }
 
-    // Disallow authoring a prim under an inactive ancestor
-    // Note that GetPrimAtPath() is safe here due to the way _DefinePrim
-    // recurses up to the toplevel prim at 3167.
-    if (!GetPrimAtPath(parentPath).IsActive()) {
-        TF_WARN("Failed to create UsdPrim <%s>. "
-                "Cannot author a UsdPrim under an inactive ancestor <%s>.", 
-                path.GetText(), parentPath.GetText());
-        return UsdPrim();
-    }
     
     // Now author scene description for this prim.
     TfErrorMark m;
--- a/pxr/usd/lib/usd/stage.cpp
+++ b/pxr/usd/lib/usd/stage.cpp
@@ -3187,11 +3187,12 @@ UsdStage::_DefinePrim(const SdfPath &path, const TfToken &typeName)
     }
 
     // Disallow authoring a prim under an inactive ancestor
+    // Note that GetPrimAtPath() is safe here due to the way _DefinePrim
+    // recurses up to the toplevel prim at 3167.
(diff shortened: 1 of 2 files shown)


pxr/usd/lib/usd/stage.cpp
```diff
@@ -3154,12 +3154,24 @@ UsdStage::_DefinePrim(const SdfPath &path, const TfToken &typeName)
...
 
     // Define all ancestors.
-    if (!_DefinePrim(path.GetParentPath(), TfToken()))
+    auto parentPath = path.GetParentPath();
+    if (!_DefinePrim(parentPath, TfToken())) {
         return UsdPrim();
+    }
+
+    // Disallow authoring a prim under an inactive ancestor
+    if (!GetPrimAtPath(parentPath).IsActive()) {
```
spiffmon: @superfunc - If path is more than one level deep beneath an inactive prim, its *parent* prim won't even exist.  So at a minimum, you should also be checking the result of GetPrimAtPath(), first.  You could go further, and, if that fails, walk up the ancestor paths to find the actually inactive prim to report in your error, which I *think* should be a TF_WARN rather than a RUNTIME error.  The error below is RUNTIME because we have no clue why our action failed, so it's a "system failure".  In this case, though, we know what the problem was: you tried to do something you shouldn't have.

Thanks - this is a nice addition!
--spiff

spiffmon: Ah, you're totally right, @superfunc - I didn't expand out the code in the review and so didn't notice it was *self* recursion!  It would be good to clarify in the new comment at line 3167, though, that due to the recursion we are guaranteed that there will be a prim at parentPath (though the paranoid in me still wants you to check the prim first before calling IsActive() :-)

jtran56:
Filed as internal issue #161683.

superfunc:
Ok I've updated with a comment explaining the recursion safety, and added some test cases. Lmk if theres anything else 👍

pxr/usd/lib/usd/stage.cpp
```diff
@@ -3154,12 +3154,25 @@ UsdStage::_DefinePrim(const SdfPath &path, const TfToken &typeName)
...
+    }
 
     // Define all ancestors.
-    if (!_DefinePrim(path.GetParentPath(), TfToken()))
+    auto parentPath = path.GetParentPath();
+    if (!_DefinePrim(parentPath, TfToken())) {
         return UsdPrim();
+    }
+
+    // Disallow authoring a prim under an inactive ancestor
```
sunyab: So this fixes UsdStage::DefinePrim, but UsdStage::OverridePrim has the same issue (even though it doesn't issue the same runtime error that DefinePrim does).

I think this logic may instead belong in UsdStage::_ValidateEditPrimAtPath, which gets called by UsdStage::_IsValidPathForCreatingPrim, which gets called by both DefinePrim and OverridePrim. There's a separate UsdStage::_ValidateEditPrim function, but I don't believe we need this logic there -- that function gets a valid UsdPrim, which means that prim must already be active. 

If we move this logic into _ValidateEditPrimAtPath, we should check that it's not called anywhere that might be a tight loop -- a cursory look suggest we'd be OK. Also, a coding error would be appropriate instead of a warning in that function.

pxr/usd/lib/usd/testenv/testUsdPrims.py
```diff
@@ -408,6 +408,36 @@ def test_GetCachedPrimBits(self):
         assert not stage.GetPrimAtPath(
             deactivatedScope.GetPath().AppendChild('child'))
 
+    def test_DefineOverInactive(self):
```
sunyab: Per above, it would be great to verify that OverridePrim behaves the same way.

superfunc:
This looks to be fixed now, closing!

PR #526, opened 2018-06-11
Katana plugins on Windows

JamieKenyonFoundry:
### Description of Change(s)

Fixes up the building of the Katana plugins on Windows.

One thing to shout out here is the deliberate change in /cmake/macros/Private.cmake:

```
     set(apiPublic "")
     set(apiPrivate ${uppercaseName}_EXPORTS=1)
     if(NOT _building_monolithic AND args_TYPE STREQUAL "STATIC")
-        set(apiPublic PXR_STATIC=1)
+#        set(apiPublic PXR_STATIC=1)
     endif()
```

I've had to do this because there are 3 static libraries that the usdKatana target links against. The above PXR_STATIC compile definition is a public property on each of the 3 static libraries and is therefore propagated to usdKatana, which is a dynamic library.

This causes usdKatana to attempt to statically link all of its link dependencies rather than import.

Whilst my above "fix" works for us, it shouldn't make its way into the public repo.

Lastly, I've written a bootstrap function in third_party/katana/lib which bootstraps various parts of Geolib, such as FnAttribute.

### DIFF
--- a/cmake/modules/FindKatanaAPI.cmake
+++ b/cmake/modules/FindKatanaAPI.cmake
@@ -31,9 +31,15 @@
 # KATANA_API_SOURCE_DIR    Path to the Katana API source directories
 # KATANA_API_VERSION       Katana API version
 
+if(WIN32)
+    set(KATANA_EXECUTABLE bin/katanaBin.exe)
+else()
+    set(KATANA_EXECUTABLE katana)
+endif()
+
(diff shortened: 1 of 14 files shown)


jtran56:
Filed as internal issue #161742.

sunyab:
Closing this pull request in favor of #560.

PR #528, opened 2018-06-13
[HdSt] Enhance the current implementation of batching in HdSt

yjang:
With the current aggregation strategy, there is a chance to end up with one batch per draw item, which could affect rendering performance.
For example, if you have a scene with many draw items, the draw items could map onto the same batchMap key which alternate into multiple different aggregated buffers.
This PR provides managing multiple aggregated buffers for each batchMap key to reduce the number of draw batches.

---------------------------
The test scene has
 - rprims: 8029
 - instancer: 7429 (1 prototype for each instancer)
 - instances: 28132
 - textures: 231

and everything is static with an animated camera.

The current implementation generated
 - 156 draw batches,
 - 8029 draw items

and it attempted to append a batch 3 times.

This PR generated
 - 7 draw batches,
 - still 8029 draw items

and it attempted to append a batch 11,319 times.
The batchMap had 3 keys and the depth of each key's vector was 1, 5, and 1.
The test scene was able to be rendered upto 10fps performance improvement.

### DIFF
--- a/pxr/imaging/lib/hdSt/commandBuffer.cpp
+++ b/pxr/imaging/lib/hdSt/commandBuffer.cpp
@@ -44,6 +44,7 @@
 #include <boost/functional/hash.hpp>
 
 #include <functional>
+#include <unordered_map>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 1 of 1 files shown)


jtran56:
GitHub #161889.

c64kernal:
Thanks so much @yjang for the awesome PR!  This is now available in dev, we couldn't do a direct merge so we adapted it a bit.  Please let us know if you run into any issues!
Thanks again!

PR #529, opened 2018-06-13
[usd] Fix metadata authoring bug.

superfunc:
### Description of Change(s)

Previously, authoring metadata from python, such as in CustomData, USD would allow python lists, but would not convert them to the proper VtArray type. This would cause layers to be unreadable, unserializable etc.

For example, 

```
from pxr import Usd

p = s.DefinePrim('/p')

p.SetCustomDataByKey('int_array', [1,2,3,4])
print s.ExportToString()
```

Produces the following broken file (notice the missing typename):
```
#usda 1.0
(
    doc = """Generated from Composed Stage of root layer 
"""
)

def "p" (
    customData = {
         int_array = [1, 2, 3, 4]
    }
)
{
}
```

Following my change, we produce a working file:

```
#usda 1.0
(
    doc = """Generated from Composed Stage of root layer 
"""
)

def "p" (
    customData = {
        int[] int_array = [1, 2, 3, 4]
    }
)
{
}
```

My change also disallows unknown list types, so bogus authoring can't get in anymore(e.g. if someone tried to author a list of Foo()s).

### Fixes Issue(s)
- None filed.

### DIFF
--- a/pxr/usd/lib/usd/pyConversions.cpp
+++ b/pxr/usd/lib/usd/pyConversions.cpp
@@ -34,6 +34,20 @@
 #include "pxr/usd/sdf/payload.h"
 #include "pxr/usd/sdf/schema.h"
 
+#include "pxr/base/gf/range1d.h"
+#include "pxr/base/gf/range1f.h"
+#include "pxr/base/gf/range2d.h"
+#include "pxr/base/gf/range2f.h"
+#include "pxr/base/gf/range3d.h"
+#include "pxr/base/gf/range3f.h"
+#include "pxr/base/gf/interval.h"
+#include "pxr/base/gf/matrix2d.h"
+#include "pxr/base/gf/matrix2f.h"
+#include "pxr/base/gf/matrix3d.h"
+#include "pxr/base/gf/matrix3f.h"
+#include "pxr/base/gf/matrix4d.h"
+#include "pxr/base/gf/matrix4f.h"
+
 #include "pxr/base/tf/pyLock.h"
 #include "pxr/base/tf/pyUtils.h"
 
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #161890.

superfunc:
Rebased and clean with dev.

superfunc:
@spiffmon Sunya and I discussed this one offline; do you see anything wrong with this approach? I'm not as familiar with stuff in Vt so I wasn't sure.

spiffmon:
Hi @superfunc,
   Nothing wrong with the approach, however:
1. Nothing prevents heterogeneous python lists; you are only checking the
type of the first VtValue element, and then assuming all the rest are the
same, and doing UncheckedGet's, which will error (once upon a time
terminate the program!).  So you should ensure you have a homogenous vector
prior to calling _ToVtArray
2. Can you add support for array of SdfAssetPath's?
3. Does SdfSpec::SetInfo have the same problem ?

--spiff

superfunc:
1,2: Very good points, missed that. Checking on each element seems expensive, I'll see if there is a boost python way to check if a list is homogeneous.
3: Not sure, I'll check.

superfunc:
@spiffmon Yeah, looks like the problem lives in Sdf, as well:

```python
>>> p.SetInfoDictionaryValue('customData', 'int_array', ["foo",2,3,4])
>>> print l.ExportToString()
#sdf 1.4.32

def "p" (
    customData = {
         int_array = [foo, 2, 3, 4]
    }
)
{
}
```

I'll get a separate PR together that fixes that one as well (will clean this one up later this week).

superfunc:
Actually, I wonder if the problem lives lower, in VtValues casting functions that this relies on. I'll need to dig a bit more to be sure.

superfunc:
Looks like @gitamohr fixed this in his referenced commit, closing out.

PR #532, opened 2018-06-14
[sdf] Add functions for finding registered formats.

superfunc:
### Description of Change(s)

This provides FindAllFileFormats(), returning a set of TfTypes corresponding to the registered formats, and FindAllFileFormatExtensions(), returning the set of all extensions for the registered types.

This has been useful in populating GUIs and will probably have other use cases I haven't imagined yet.

### Fixes Issue(s)
- None filed, just fulfilling my dreams

### DIFF
--- a/pxr/usd/lib/sdf/fileFormat.cpp
+++ b/pxr/usd/lib/sdf/fileFormat.cpp
@@ -251,6 +251,7 @@ SdfFileFormat::WriteToString(
     return false;
 }
 
+/* static */
 std::string
 SdfFileFormat::GetFileExtension(
     const std::string& s)
@@ -270,6 +271,14 @@ SdfFileFormat::GetFileExtension(
     return extension.empty() ? s : extension;
 }
 
+/* static */
+std::set<std::string>
+SdfFileFormat::FindAllFileFormatExtensions()
+{
+    return _FileFormatRegistry->FindAllFileFormatExtensions();
+}
+
+/* static */
 SdfFileFormatConstPtr
 SdfFileFormat::FindById(
     const TfToken& formatId)
@@ -277,6 +286,7 @@ SdfFileFormat::FindById(
     return _FileFormatRegistry->FindById(formatId);
 }
 
+/* static */
 SdfFileFormatConstPtr
 SdfFileFormat::FindByExtension(
     const std::string& extension,
--- a/pxr/usd/lib/sdf/testenv/testSdfFileFormat.py
+++ b/pxr/usd/lib/sdf/testenv/testSdfFileFormat.py
@@ -1,0 +1,53 @@
+#!/pxrpythonsubst
+#
+# Copyright 2018 Pixar
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
+# 6. Trademarks. This License does not grant permission to use the trade
+#    names, trademarks, service marks, or product names of the Licensor
+#    and its affiliates, except as required to comply with Section 4(c) of
+#    the License and to reproduce the content of the NOTICE file.
+#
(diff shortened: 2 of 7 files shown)


jtran56:
Filed as internal issue #161892.

pxr/usd/lib/sdf/testenv/testSdfFileFormat.py
```diff
@@ -0,0 +1,91 @@
...
+    def test_StaticMethods(self):
+        print 'Testing static methods on SdfFileFormat...'
+
+        # FindById
+        # Note that the id and extension are the same in our case
+        sdfFileFormat = Sdf.FileFormat.FindById('sdf')
+        self.assertTrue(sdfFileFormat)
+        self.assertEqual(sdfFileFormat.GetFileExtensions(), ['sdf'])
+
+        usdFileFormat = Sdf.FileFormat.FindById('usd')
```
sunyab: Since this test lives in sdf, it shouldn't make any assertions involving usd* file formats since those are all downstream of this library.

pxr/usd/lib/sdf/wrapFileFormat.cpp
```diff
@@ -129,6 +130,14 @@ void wrapFileFormat()
         .def("GetFileExtension", &This::GetFileExtension)
         .staticmethod("GetFileExtension")
 
+        .def("FindAllFileFormatExtensions", &This::FindAllFileFormatExtensions,
+             return_value_policy<TfPySequenceToTuple>())
```
sunyab: I think using TfPySequenceToList to return the results as a list instead of a tuple would be more conventional.

pxr/usd/lib/sdf/fileFormat.cpp
```diff
@@ -257,13 +258,39 @@ SdfFileFormat::GetFileExtension(
     return extension.empty() ? s : extension;
 }
 
+/* static */
+std::set<std::string>
+SdfFileFormat::FindAllFileFormatExtensions()
+{
```
sunyab: I think this code will crash if the plugins for the file formats returned by FindAllFileFormats are not loaded first. In that case, FindAllFileFormats will return TfTypes that have been declared in plugInfo.json files but not yet defined by a call to TfType::Define, which means they won't have a factory set, which means factory->New() will die.

pxr/usd/lib/sdf/fileFormat.cpp
```diff
@@ -257,13 +258,39 @@ SdfFileFormat::GetFileExtension(
...
+            result.insert(e);
+        }
+    }
+    return result;
+}
+
+/* static */
+std::set<TfType>
+SdfFileFormat::FindAllFileFormats()
+{
```
sunyab: Was this API just added as a helper for the function above? If so, I'd rather remove it since it's a pretty thin cover around pre-existing API.

pxr/usd/lib/sdf/fileFormat.cpp
sunyab: Also, ideally we would not need to actually load the file format plugins to determine the available extensions, since that's all declared in plugInfo.json files. In other words, if I were to import Sdf in Python and call this function, it shouldn't cause Usd to load.

pxr/usd/lib/sdf/fileFormat.cpp
```diff
@@ -257,13 +258,39 @@ SdfFileFormat::GetFileExtension(
...
+    std::set<TfType> result;
+    TfType::Find<SdfFileFormat>().GetAllDerivedTypes(&result);
```
sunyab: This should be using PlugRegistry::GetAllDerivedTypes instead to ensure that all plugins are discovered first. See Sdf_FileFormatRegistry::_RegisterFormatPlugins.

superfunc:
@sunyab Smushed these into one commit, lmk if you need anything else to get this one merged.

superfunc:
Rebased this one with current dev; lmk if there were follow up changes you wanted.

PR #535, opened 2018-06-18
Katana plugin asset resolver fixes

ablev-dwa:
### Description of Change(s)

In the Katana USD plugin this replaces direct calls to the Resolver with use of the pre-resolved path in an SdfAssetPath. This primarily fixes filenames of relative assets from usd files that are not in the current directory.

First patch does the minimal change to fix the bug. Subsequent patches remove any attempt to use the resolver, and then remove redundant or unnecessary flags from the attribute converter function, these are not needed now that the resolver has already been run correctly by Sdf Layer, and correct the behavior in some (obscure?) cases such as arrays of assets.


jtran56:
Filed as internal issue #162204.

PR #536, opened 2018-06-19
WIP: Python3 support

dgovil:
**NOTE**: This is a non functional pull request. I made it to try and get community help with the porting process.

### Description of Change(s)

Hi,
I'm trying to update the build system to work agnostically with Python3. I realize that it's not a priority at the moment as all DCC apps are on 2.7 for a few more years. However at a cursory glance I didn't see anything that would be Python2 specific in USD itself outside of the build script, and I figured it was a decent project to undertake.

This is just a WIP pull request so I can posit some questions (below). I'll clean up before making a true PR.

### Fixes Issue(s)
- Build script runs under both Python 2 and 3
- Update oiio from 1.7.14 to 1.7.19 to support ffmpeg 4.x
- Automatically include Python include paths so that boost can build against it.

### Issues

This is where I was hoping someone may be able to help me (when you get a moment of course). All the dependencies, and USD itself seems to build.
However when linking, it seems to fail. Which is odd because I see the requisite lyb<name>.dylib in the lib directory. Specifically I see `install/build/USD/pxr/base/lib/tf/libtf.dylib`

```
[ 47%] Linking CXX shared library _tf.so
Undefined symbols for architecture x86_64:
  "boost::python::detail::init_module(char const*, void (*)())", referenced from:
      _init_tf in module.cpp.o
      _initlibtf in module.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Full error log: https://www.dropbox.com/s/ao8yp8wmh6c5a78/log.txt?dl=0

Cheers!


jcowles:
Couple thoughts:

1) I would squash all the crufty merge changes
2) Does "Update oiio from 1.7.14 to 1.7.19 to support ffmpeg 4.x" have anything to do with Python3?
3) It's problematic to merge this without actually confirming that USD/Python3 runs correctly -- people will assume it does. It would be awesome to confirm it does actually work (unit tests pass, etc).

This is awesome! I would love to see support for Python3 -- I think the need for it will begin to increase over time as modern apps/libs begin to drop support for Python2 (e.g. PyTorch, which I am using currently).

jcowles:
Makes sense -- I still think it would be good to break out the OIIO change into a separate PR.

The link error is due to the fact that Boost changes the signature based on the Python version:
https://www.boost.org/doc/libs/1_46_1/boost/python/module_init.hpp

dgovil:
Ah that's both good to know and unfortunate. 

I'd love to dig in deeper on this, but I just started a new job which unfortunately precludes me from working on this any more.

I'll close the PR for now, but maybe if someone has time, it didn't look too terrible to port based on that file you linked to.

PR #541, opened 2018-06-25
[Imaging] Improving shadow support in HdxTaskController

sirpalee:
### Description of Change(s)
The change adds shadow task generation to HdxTaskController and sets up the camera Window Policy so shadows can be rendered via the HdStream backend.

This is useful for people, who want to build their own renderer using the base classes and not rely on feeding the render index from an USD scene.

### DIFF
--- a/pxr/imaging/lib/hdx/shadowTask.cpp
+++ b/pxr/imaging/lib/hdx/shadowTask.cpp
@@ -153,44 +153,19 @@ HdxShadowTask::_Sync(HdTaskContext* ctx)
     if (!renderIndex.IsSprimTypeSupported(HdPrimTypeTokens->simpleLight)) {
         return;
     }
-    
-    // Extract the HD lights used to render the scene from the 
-    // task context, we will use them to find out what
-    // lights are dirty and if we need to update the
-    // collection for shadows mapping
 
-    // XXX: This is inefficient, need to be optimized
-    SdfPathVector lightPaths;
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #162211.

sirpalee:
Fixed 80 column violations and ordered the shadow task after the simple lighting task.

sirpalee:
I updated the pull request with a simple change. When mixing simple lights with usd lux lights we found that shadow task generation does not work when using the HdSt render delegate. This happens, because HdSt internally translates usd lux lights to simple lights, so the total number of simple lights won't match the simple lights in the render index.

I had to change how accessing certain data (shadow rprim collections) happens and effectively skip lights with no simple light sprim in the render index. See the changes to shadowTask.cpp.

sirpalee:
Improved the changes to shadowTask.cpp, including removal of dead code and a simplified approach to checking against lights.

sirpalee:
I added a new function to the task controller that lets the user change the shadow task params.

sirpalee:
@sunyab @spiffmon What's the reason for leaving out the SetShadowParams function from the PR?

poljere:
I think this got lost during merge process as we didn't see any code using it, sorry about that. We can bring it back, and just out of curiosity, which parameters are you actually using in that struct?

sirpalee:
@poljere for now we are controlling cull settings.

PR #549, opened 2018-07-07
Allow this library to build against Houdini 17

marktucker:
Trying to bring this library in line with the needs of Houdini 17:
* Several changes to packed primitive function signatures
* Support float16 UT_Quaternions and vectors
* Exporting a few symbols and classes needed by H17 plugins. Removing
  export claim from GusdDefaultArray which is a header-only template class
  and so shouldn't claim to export any symbols
* GEO_IOTranslator accepts usd, usda, or usdc extensions
* GEO_IOTranslator loads proxy and default prims, rather than just proxy
* USD_CustomTraverse intialization invoked by an explicit function call
  (which is called from gustInitialize)
* USD_XformCache env var initialization fix, most likely peculiar to the
  way that H17 causes this library to be loaded
* Convert uses of HDK_API_VERSION to SYS_VERSION_FULL_INT. This latter
  variable is a more reliable reflection of the Houdini build version,
  and so should be used in preference to HDK_API_VERSION
* Use GA and GT constants for Houdini geometry attribute names where
  applicable. Use GT_Offset constants instead of assuming GT_Offset==int
* In printf functions, use SYS_PRI constants for reliable cross-platform
  printing of 64 bit values
* More error checking when converting Houdini attributes to/from USD
  attributes, and when calculating prim xforms

### DIFF
--- a/third_party/houdini/lib/gusd/GEO_IOTranslator.cpp
+++ b/third_party/houdini/lib/gusd/GEO_IOTranslator.cpp
@@ -71,7 +71,8 @@ checkExtension(const char* name)
     UT_String nameStr(name);
     if(nameStr.fileExtension()
      &&  (!strcmp(nameStr.fileExtension(), ".usd")
-       || !strcmp(nameStr.fileExtension(), ".usda"))) {
+       || !strcmp(nameStr.fileExtension(), ".usda")
+       || !strcmp(nameStr.fileExtension(), ".usdc"))) {
         return true;
     }
     return false;
@@ -110,11 +111,13 @@ fileLoad(GEO_Detail* gdp, UT_IStream& is, bool ate_magic)
     // all the top level prims.
     auto defPrim = stage->GetDefaultPrim();
     if(  defPrim ) {
-        GusdGU_PackedUSD::Build( *detail, fileName, defPrim.GetPath(), f );
+        GusdGU_PackedUSD::Build(*detail, fileName, defPrim.GetPath(), f, NULL,
+	    GusdPurposeSet(GUSD_PURPOSE_DEFAULT | GUSD_PURPOSE_PROXY));
     }
     else {
         for( const auto &child : stage->GetPseudoRoot().GetChildren() ) {
-            GusdGU_PackedUSD::Build( *detail, fileName, child.GetPath(), f );
+            GusdGU_PackedUSD::Build(*detail, fileName, child.GetPath(), f, NULL,
+		GusdPurposeSet(GUSD_PURPOSE_DEFAULT | GUSD_PURPOSE_PROXY));
         }
     }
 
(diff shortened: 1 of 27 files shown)


spiffmon:
Sweet - thank you, Mark!

--spiff

jtran56:
Filed as internal issue #162656.

third_party/houdini/lib/gusd/GEO_IOTranslator.cpp
```diff
@@ -71,7 +71,8 @@ checkExtension(const char* name)
     UT_String nameStr(name);
     if(nameStr.fileExtension()
      &&  (!strcmp(nameStr.fileExtension(), ".usd")
-       || !strcmp(nameStr.fileExtension(), ".usda"))) {
```
superfunc: Note we should check for something more robust once #532 is merged.

marktucker: Sorry, this was just a stop gap to cover a particular common case. Anything more general would definitely be better, thanks.

third_party/houdini/lib/gusd/stageCache.cpp
```diff
@@ -56,21 +56,39 @@
...
-
-
-TF_DEFINE_ENV_SETTING(GUSD_STAGEMASK_ENABLE, true,
-                      "Enable use of stage masks when accessing prims from "
-                      "the cache. Note that disabling this feature may "
-                      "be very detrimental to performance when separately "
-                      "querying many prims with variant selections "
-                      "(or other types of stage edits).");
-
+// The follofing declarations are equivalent to using the TF_DEFINE_ENV_SETTING
```
sunyab: Hey @marktucker, I'm looking at merging this pull request in now and was looking at these changes. Could you give any more details about how you ran into this issue -- what platform, etc? I think we'd like to pull this fix into tf/envSetting.h itself rather than have it here, but I was hoping to get a little more context. Thanks!

sunyab: Thanks Mark, that did help quite a bit. I'm testing a change that folds your fix into tf now. One other question: what compiler were you using when you ran into this? Were you using Visual Studio 2017?

marktucker: Visual Studio 2017. Version 15.7.4 to be precise.

sunyab: OK, that seems to line up with what I've been seeing in godbolt. From the following thread, it sounds like this could be a bug in MSVC 2017:

https://developercommunity.visualstudio.com/content/problem/76198/vs-2017-compiler-creates-broken-debug-build-using.html

In the meantime, I've adopted your fix into tf/envSetting.h internally and will push that bit out in a separate change. Thanks again!

PR #550, opened 2018-07-09
[UsdView] Adding resource path containing shader resources to the default search path.

sirpalee:
### Description of Change(s)
The PR loads all the resource paths containing shaders in usdview. This way users are able to attach default shaders (or any kind of shader resources from a plugin) without specifying the full path.

### Fixes Issue(s)
No issues reported.

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appController.py
+++ b/pxr/usdImaging/lib/usdviewq/appController.py
@@ -1064,6 +1064,16 @@ class AppController(QtCore.QObject):
 
     def _openStage(self, usdFilePath, populationMaskPaths):
 
+        # We are iterating through the plugin registry to add anything
+        # containing shaders to the default search path.
+        resourcePaths = set()
+        from pxr import Plug
+        pr = Plug.Registry()
+        for t in pr.GetAllPlugins():
+            if t.metadata.get('ShaderResources') is not None:
+                resourcePaths.add(t.resourcePath)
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #162657.

spiffmon:
Hi @sirpalee ,
    You have the right goal here, but the approach needs to be a little different, and unfortunately, HdStream needs some love before it can make use of "the right" solution.

    That solution is that renderers should find shaders through the SdrRegistry, which the new HdMaterial networks support, but which HdStream does not yet use - hdPrman is the only render delegate to make use of them so far, and it is still private, at this point.

   But there *is* a missing piece that is very similar to what you are proposing, that you could tackle if you wanted to!  SdrRegistry relies on the foundation module "Ndr" (generic Node Definition Registry) for functionality like discovering where shaders reside.  What we should be doing in https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/lib/ndr/filesystemDiscovery.cpp#L35 is adding exactly the paths you are finding in your change onto the _searchPaths.

    That will allow the SdrRegistry to find all shaders whose extensions are listed in the env var PXR_NDR_FS_PLUGIN_ALLOWED_EXTS . Note that we also don't yet have a definition *parser* for glslfx, so there will likely be errors when you fetch a node definition for a glslfx shader, as it will attempt to parse it, but won't find a parser.

PR #554, opened 2018-07-10
Bug ssbo glsl

BenjaminCoquelle:
### Description of Change(s)
updating glsl compute shader to follow specification. SSBO needs to layout (std430) buffer semantics in shaders

### Fixes Issue(s)
Without this change it doesn't render on AMD gpu as compiler strictly follows the specs.


jtran56:
Filed as internal issue #162819.

sunyab:
This looks like #555, but targeting the master branch instead of dev. Changes aren't merged directly into master, and #555 is slated for merging into dev, so I'm going to close this out. Please let me know if this is mistaken!

PR #555, opened 2018-07-10
fix layout for SSBO, according to glsl specs it is mandatory to use s…

BenjaminCoquelle:
### Description of Change(s)
updating glsl compute shader to follow specification. SSBO needs to layout (std430) buffer semantics in shaders

### Fixes Issue(s)
Without this change it doesn't render on AMD gpu as compiler strictly follows the specs.


poljere:
Thanks @BenjaminCoquelle! We will take a look.

PR #557, opened 2018-07-12
[usdMaya] Add new test confirming exports can be used as Value Clips

nxkb:
### Description of Change(s)
Adds an end to end test that makes sure UsdMaya will export files that can be used as value clips.

This test has been useful in the past to expose a few issues with value resolution in value clips and making sure maya writes all the correct samples. Both discovered issues are resolved now so this test passes with no problems. It simply checks that the samples of certain attributes are identical whether the files are exported as value clips and stitched or as a single usd file.

### Fixes Issue(s)
- no tests for using maya files as value clips

### DIFF
--- a/third_party/maya/lib/usdMaya/CMakeLists.txt
+++ b/third_party/maya/lib/usdMaya/CMakeLists.txt
@@ -166,6 +166,7 @@ pxr_shared_library(${PXR_PACKAGE}
 )
 
 pxr_test_scripts(
+        testenv/testUsdExportAsClip.py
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #162821.

c64kernal:
Hi @nxkb -- thanks so much for this PR, it's great.  It'll get merged soon, but I wanted to let you know that we modified it a bit to remove luma specific things that found their way into the .ma file, and we replaced a use of `basestring` with `str`.  Thanks again!

PR #565, opened 2018-07-18
[usdImaging] valueCache - avoid storing iterators to prevent crash

elrond79:
### Description of Change(s)
We were having a crash with our custom sceneDelegate, and tracked the problem down to UsdImagingValueCache using iterators that had since become invalid.

My change here simply alters things so that we store the map key, instead of an iterator, which should always be safe; however, it will be slower when iterating over the items to cleanup in _GarbageCollect, as we need to re-do the lookup for each item.  My hope is that speed is less important during this phase (thus the reason why deletion is delayed?), and this won't be a problem.

If you believe this will create performance issues, we can discuss alternative solutions, but I wanted to at least get this discussion started.

### Fixes Issue(s)
- Crash when doing things that cause items to be added to the render index

### DIFF
--- a/pxr/usdImaging/lib/usdImaging/valueCache.h
+++ b/pxr/usdImaging/lib/usdImaging/valueCache.h
@@ -65,6 +65,9 @@ public:
         SdfPath _path;
         TfToken _attribute;
     public:
+        Key()
+        {}
+
         Key(SdfPath const& path, TfToken const& attr)
             : _path(path)
             , _attribute(attr)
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #163025.

poljere:
Matthew Williams from our team took a look at this, here are some thoughts: 

There is a performance concern here, as it changes the algorithm from O to O(n log n). The reason for the deferred deletion isn't performance as the comment above infers, but thread safety (tbb's container delete's need to be single threaded).

The fact you are hitting an invalid iterator here, indicates you have a bug probably elsewhere in the scene delegate (probably a double-extract?), which this change would not remove.

While we can acknowledge the double-extract is a difficult problem to identify and it does leave us open to these sort of crashes (maybe we can add some duplication detection to _GarbageCollect - possibly behind an envvar?), I'm concerned about the performance regression a change like this would make to work around a coding error.

elrond79:
Oh, damn... I had missed the fact that we were using a tbb::concurrent_unordered_map here, not a normal map, and that inserts on that don't invalidate existing iterators.  My apologies!  Our crash may have been caused by a double extract, I'll close this PR and look into that.
...though I may be making a new PR to check for double-extracts as you suggest!  Either guarded by an env var, or just by being in debug mode, perhaps...

PR #517, opened 2018-05-31
Adding hydra support for USD volume schema

marktucker:
Note that this pull request won't build without PR #514, but I was asked to submit it in advance of that one being accepted.

This PR provides usdImaging adapters for volume and field prims, Hd base classes for field bprims and volume rprims, and new scene delegate and prim adapter virtuals for fetching field information from a volume primitive.

PR #567, opened 2018-07-20
Combined UsdVol schema and Hydra changes

marktucker:
This pull request supercedes #514 and #517. It includes all revisions from those pull requests, plus a couple of conflict resolutions, plus a large number of changes based on feedback from the USD team. This PR is not yet complete, with more changes on the Hydra side likely still to come.


jtran56:
Filed as internal issue #163093.

PR #568, opened 2018-07-23
Expose the normals attribute through hydra on meshes

marktucker:
Expose the normals attribute through hydra on meshes using almost the same code used for exposing this attribute on basis curves. The only difference is retrieving the interpolation metadata from the normals attribute for meshes, which isn't necessary on basis curves.

If there is a normals primvar on the mesh, I have been told that this primvar should take precedence over the normals attribute from the UsdGeomPointBased schema. This commit (like the basis curves adapter) does not do this, but it is necessary code in any case, and does handle the simple situation where a mesh uses the schema attribute, and the render delegate expects to have access to it.

### DIFF
--- a/pxr/usdImaging/lib/usdImaging/gprimAdapter.cpp
+++ b/pxr/usdImaging/lib/usdImaging/gprimAdapter.cpp
@@ -52,25 +52,6 @@ TF_REGISTRY_FUNCTION(TfType)
     // No factory here, GprimAdapter is abstract.
 }
 
-static HdInterpolation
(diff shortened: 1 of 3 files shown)


jtran56:
Filed as internal issue #163098.

c64kernal:
Thanks @marktucker!  Hang tight, and some changes should be landing soon from @tcauchois that may clean things up for you here.

sunyab:
This should be addressed by commit b779a46dc93c46fba5ad252f5db017eca5017e4e. Closing this out, thanks!

PR #580, opened 2018-08-05
[Gf] Add deprecation attribute to deprecated code.

superfunc:
### Description of Change(s)

This will ensure people get the warning when compiling against USD if
they are using this code. I have left the doxygen comments in, as
doxygen will not currently pick up on these attributes to annotate
the generated docs.


jtran56:
Filed as internal issue #164137.

PR #582, opened 2018-08-06
[build_usd] print tracebacks on error if verbosity >= 3

elrond79:
### Description of Change(s)
Mostly handy if debugging build_usd.py itself - but could be handy in other circumstances. And it makes sense not to swallow the traceback if verbosity is this high.

### Fixes Issue(s)
- Tracebacks swallowed even on high verbosity

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -65,6 +65,9 @@ def PrintCommandOutput(output):
         sys.stdout.write(output)
 
 def PrintError(error):
+    if verbosity >= 3 and sys.exc_info()[1] is not None:
+        import traceback
+        traceback.print_exc()
     print "ERROR:", error
 
 # Helpers for determining platform


jtran56:
Filed as internal issue #164139.

build_scripts/build_usd.py
```diff
@@ -64,6 +64,9 @@ def PrintCommandOutput(output):
         sys.stdout.write(output)
 
 def PrintError(error):
+    if verbosity >= 3 and getattr(sys, "exc_value", None) is not None:
```
c64kernal: Thanks for this PR, @elrond79!  exc_value has been deprecated for quite a while (https://docs.python.org/2/library/sys.html#sys.exc_value), any chance there's a formulation here with exc_info() that you'd be happy with?

elrond79: Oh, wow... how did I not know that exc_value/etc had been deprecated?? It's not like I've ever used python-1.5!

Anyway.. yeah, updated the PR to use exc_info. Thanks!

c64kernal: @sunyab never misses a beat :)  Thanks so much @elrond79!

PR #590, opened 2018-08-15
Tweaks to make build_usd.py work with ninja

elrond79:
### Description of Change(s)

Initially aimed to make it work on windows, but subsequently tested / got it working on all 3 platforms: windows, linux (ubuntu), and MacOS 

### Fixes Issue(s)
- build_usd.py fails with --generator=Ninja

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -203,6 +203,7 @@ def RunCMake(context, force, extraArgs = None):
 
     # On Windows, we need to explicitly specify the generator to ensure we're
     # building a 64-bit project. (Surely there is a better way to do this?)
+    # TODO: figure out exactly what "vcvarsall.bat x64" sets to force x64
     if generator is None and Windows():
         if IsVisualStudio2017OrGreater():
             generator = "Visual Studio 15 2017 Win64"
@@ -839,7 +840,17 @@ def InstallOpenSubdiv(context, force, buildArgs):
         # Add on any user-specified extra arguments.
         extraArgs += buildArgs
 
-        RunCMake(context, force, extraArgs)
+        oldGenerator = context.cmakeGenerator
+    
+        # OpenSubdiv seems to error when building on windows w/ Ninja...
+        # ...so just use the default generator (ie, VisualStudio on windows)
+        # until someone can sort it out
+        if oldGenerator == "Ninja" and Windows():
+            context.cmakeGenerator = None
+        try:
+            RunCMake(context, force, extraArgs)
+        finally:
+            context.cmakeGenerator = oldGenerator
 
 OPENSUBDIV = Dependency("OpenSubdiv", InstallOpenSubdiv, 
                         "include/opensubdiv/version.h")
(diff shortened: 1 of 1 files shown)


chadrik:
Paul is making pull requests while at Siggraph.  Respect :)

meshula:
Should be fixed in OpenEXR 2.3. Maybe the 2.2 patches should be conditional on version.

elrond79:
True - but build_usd is currently hardcoded to 2.2... and even if it were switched to 2.3, and we forget to remove this patch, it will just end up as a do-nothing patch in that case.

superfunc:
+1 on the need for this one. Ninja builds are significantly faster on my local machine, and would really help dev process for me.

build_scripts/build_usd.py
```diff
@@ -201,14 +201,29 @@ def RunCMake(context, force, extraArgs = None):
 
     generator = context.cmakeGenerator
 
+    # OpenSubdiv seems to error when building on windows w/ Ninja...
```
sunyab: I'd prefer to have library-specific behavior set up in that library's Install... function, rather than putting it in RunCMake. That way, all of the quirks for building a specific library are contained in a single place for ease of inspection.

So, would it be possible to move this logic into InstallOpenSubdiv instead? I was thinking you could do this check in that function and if it's required, then temporarily set context.cmakeGenerator to None, call RunCMake, then reset it back to its original value (probably in a try-finally block).

sunyab: If we did this, you could revert the changes that add currentDependency to the context object as well as the above change.

build_scripts/build_usd.py
```diff
@@ -201,14 +201,29 @@ def RunCMake(context, force, extraArgs = None):
...
     # On Windows, we need to explicitly specify the generator to ensure we're
     # building a 64-bit project. (Surely there is a better way to do this?)
+    # TODO: figure out exactly what "vcvarsall.bat x64" sets to force x64
     if generator is None and Windows():
         if IsVisualStudio2017OrGreater():
             generator = "Visual Studio 15 2017 Win64"
         else:
             generator = "Visual Studio 14 2015 Win64"
 
+    if generator == 'Ninja':
```
sunyab: For simplicity, instead of branching can we just always do:
```
cmakeBuildType = '-DCMAKE_BUILD_TYPE=Release '
buildConfig = '--config Release '
```
or does the Ninja generator not like the '--config Release' bit?

build_scripts/build_usd.py
```diff
@@ -646,6 +677,34 @@ def InstallOpenEXR(context, force, buildArgs):
 
     ilmbaseSrcDir = os.path.join(srcDir, "IlmBase")
     with CurrentWorkingDirectory(ilmbaseSrcDir):
+        # openexr 2.2 has a bug with Ninja:
```
sunyab: Thanks for documenting this!

build_scripts/build_usd.py
elrond79: Done and done in latest changes!

build_scripts/build_usd.py
elrond79: I think Ninja will just safely ignore it, so made it non-branching as requested...

PR #593, opened 2018-08-18
[Build] Only build boost bits required for OIIO when building OIIO.

superfunc:
### Description of Change(s)
This dependency was removed, so we don't need to build a shared
library or install its headers in the build script anymore.

### Fixes Issue(s)
- None filed.

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -494,7 +494,6 @@ def InstallBoost(context, force, buildArgs):
             'threading=multi', 
             'variant=release',
             '--with-atomic',
-            '--with-date_time',
             '--with-filesystem',
             '--with-program_options',
             '--with-system',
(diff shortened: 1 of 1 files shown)


superfunc:
Hey G :) 

I'm not sure, thanks for bringing it up, I'll run it through an oiio build.

superfunc:
@c64kernal You were right, amended the PR to do it conditionally, thanks for spotting it.

c64kernal:
Thanks Josh!  We might want to do the same with date_time while we're at it...

superfunc:
Yeah that seems reasonable, added as a separate commit.

c64kernal:
Thanks @superfunc!

c64kernal:
@superfunc -- the latest dev branch should have all this functionality merged in (in some form or another) I'll close this out, but if there's more you're missing, please let me know!

PR #594, opened 2018-08-19
[Build] Make examples+tutorials optional.

superfunc:
### Description of Change(s)

This makes the examples set (usdObj, etc) optional to build. This is desirable for applications embedding a USD build which want a minimal set of components.

### Fixes Issue(s)
- None filed.


jtran56:
Filed as internal issue #164199.

PR #597, opened 2018-08-21
[USD] Adding UDIM support to HdSt

sirpalee:
### Description of Change(s)
The following PR adds UDIM support to HdSt. There are a few outstanding issues yet to be solved, but since this is a big change I think it's work starting to review early.

Issues.
- I don't really like the isUdim (and isPtex) flags all over the place. While it would be good to get rid of them, I think it should be done after this PR.
- Same goes for the dynamic casts to get the texture and layout ids. I think these should be moved to the base glf texture class, as it's OpenGL specific already. For now having a buffer/texture to store texels and one for layouts is a good overall approximation for multiple types of textures. Alternatively, we could request ids based on tokens and have a function to iterate through textures in the base glf texture class.
- I have to do some performance tests wrt loading large amounts of udim textures and make sure we only load the data when required.
- Texture origins are not yet handled properly.
- Probably I need to do more checks in codeGen regarding the OpenGL version / caps.

### Fixes Issue(s)
No reported issues.

### DIFF
--- a/pxr/imaging/lib/glf/CMakeLists.txt
+++ b/pxr/imaging/lib/glf/CMakeLists.txt
@@ -81,6 +81,7 @@ pxr_library(glf
         texture
         textureHandle
         textureRegistry
+        udimTexture
         uniformBlock
         utils
         uvTexture
--- a/pxr/imaging/lib/glf/contextCaps.cpp
+++ b/pxr/imaging/lib/glf/contextCaps.cpp
@@ -62,11 +62,13 @@ GlfContextCaps::GlfContextCaps()
     : glVersion(0)
     , coreProfile(false)
 
+    , maxArrayTextureLayers(0)
     , maxUniformBlockSize(0)
     , maxShaderStorageBlockSize(0)
     , maxTextureBufferSize(0)
     , uniformBufferOffsetAlignment(0)
 
+    , arrayTexturesEnabled(false)
     , shaderStorageBufferEnabled(false)
     , bufferStorageEnabled(false)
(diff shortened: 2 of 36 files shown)


sirpalee:
Significantly updated the PR:

- Minimized the use of auto to be in line with the repo's coding style.
- Removed the HwUdimTexture_1 token and the old ways of setting up the texture.
- Changed the default wrapping of udim textures to black.
- Only supporting the uppercase udim token as specified in the Preview Surface's specs.
- _GetMipLevels was moved to an anonymous function rather than lambda as it's not capturing anything.
- Removed IsUdim and IsPtex functions all over the codebase and created a new enum, HdTextureType to replace these functions and flags.

jtran56:
Filed as internal issue #USD-4706.

elrond79:
Resolved merge conflicts and rebased.  Also fixed an issue with udims + usdz, and updated to use ArResolverScopedCache

For the incremental changes, see:
https://github.com/LumaPictures/USD/commit/926c9c98466f50cc814d736e1750969e3735fe40
https://github.com/LumaPictures/USD/commit/dc1ca2a01cc9bb782ba783df0dfa564a95540c3f

sirpalee:
Updated the PR to work with the latest dev branch. There are a few changes as well.
- Changed types handling pixel counts and such to unsigned int in the UdimTexture class.
- Fixed a bug regarding loading one more mip level than required.
- Missing udim tiles now display a black color instead of the first tile.
- Handling cases with 0 memory request by generating a 1x1 image for each tile.
- Fixed some formatting issues.

PR #599, opened 2018-08-22
Adding python access to the IsSupported method on renderer plugins.

marktucker:
Use this method to decide which renderers to show in the usdview menu.

Unrelated, but I couldn't bring myself to make a separate pull request for it: also including <functional> in renderThread.h for definition of std::function.

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appController.py
+++ b/pxr/usdImaging/lib/usdviewq/appController.py
@@ -1222,6 +1222,15 @@ class AppController(QtCore.QObject):
                     if action.text() == self._stageView.rendererPluginName:
                         action.setChecked(True)
                         break
+                # Then display an error message to let the user know something
+                # went wrong, and disable the menu item so it can't be selected
+                # again.
+                for action in self._ui.rendererPluginActionGroup.actions():
+                    if action.pluginType == plugin:
+                        self.statusMessage(
+                            'Renderer not supported: %s' % action.text())
+                        action.setText(action.text() + " (unsupported)")
+                        action.setDisabled(True)
 
     def _configureRendererPlugins(self):
         if self._stageView:
--- a/pxr/imaging/lib/hd/renderThread.h
+++ b/pxr/imaging/lib/hd/renderThread.h
@@ -29,6 +29,7 @@
 
 #include <atomic>
 #include <condition_variable>
+#include <functional>
 #include <mutex>
 #include <thread>
 
--- a/pxr/usdImaging/lib/usdImagingGL/engine.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/engine.cpp
@@ -700,6 +700,13 @@ UsdImagingGLEngine::GetRendererPluginDesc(TfToken const &id) const
 
 /* virtual */
 bool
+UsdImagingGLEngine::GetRendererPluginIsSupported(TfToken const &id) const
+{
+    return false;
+}
+
+/* virtual */
+bool
 UsdImagingGLEngine::SetRendererPlugin(TfToken const &id)
 {
     return false;
--- a/pxr/usdImaging/lib/usdImagingGL/engine.h
+++ b/pxr/usdImaging/lib/usdImagingGL/engine.h
@@ -371,6 +371,10 @@ public:
     USDIMAGINGGL_API
     virtual std::string GetRendererPluginDesc(TfToken const &id) const;
 
+    /// Checks if a renderer plugin is supported in this runtime environment.
+    USDIMAGINGGL_API
+    virtual bool GetRendererPluginIsSupported(TfToken const &id) const;
+
     /// Set the current render-graph delegate to \p id.
     /// the plugin will be loaded if it's not yet.
     USDIMAGINGGL_API
--- a/pxr/usdImaging/lib/usdImagingGL/gl.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/gl.cpp
@@ -298,6 +298,13 @@ UsdImagingGL::GetRendererPluginDesc(TfToken const &id) const
 
 /* virtual */
 bool
+UsdImagingGL::GetRendererPluginIsSupported(TfToken const &id) const
+{
+    return _engine->GetRendererPluginIsSupported(id);
+}
+
+/* virtual */
+bool
 UsdImagingGL::SetRendererPlugin(TfToken const &id)
 {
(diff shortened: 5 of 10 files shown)


jtran56:
Filed as internal issue #USD-4705.

c64kernal:
Hi @marktucker, thanks for the PR!  One of the reasons we haven't done this already is that we desperately try not to load plugins in usdview, when not needed, because of load time to dlopen some plugins (and also the arbitrary cost of running unknown code).  This PR forces the system to load every hydra backend plugin regardless of whether or not the user was ever going to switch to it.

The other, much more distant but still valid concern is in order to avoid folks scratching their heads as to why their plugin didn't get discovered, we would probably want to gray out the not supported plugins rather than omit them entirely.  However, this still has the performance problem of course, so we didn't really think about it too much.

From a "user exploring the system for the first time" perspective, I totally see how a grayed out menu option for the unsupported plugins is more friendly, but for someone launching usdview 100's of times a day, our general philosophy is to try to avoid any extra overhead that's not absolutely necessary.

Hope that makes sense and please let us know your thoughts.

marktucker:
I hadn't thought about the increasing load times. I would have assumed the plugin libraries were already being loaded even before the plugin is instantiated the first time. I can't speak to the question of how much this is going to increase load times, and I imagine it's going to vary wildly from studio to studio. The case of Pixar is likely to be the worst case, but that's just a guess.

Anyway, my counter-argument would be that each studio controls which render plugins are available by deciding which ones to build, and which ones to put in their plugin path. And yes, I realize that argument cuts both ways, as a plugin that will be unsupported in a particular environment could be excluded by adjusting the plugin path rather than relying on IsSupported to dynamically check if the plugin is supported. Though I would have thought this was the purpose of IsSupported... Anyway, not a very strong argument.

My other argument is that selecting a plugin that looks for all the world like it is an acceptable choice from that menu, only to have absolutely nothing happen isn't great behavior... Maybe a dialog or some other notification at that point that the chosen renderer isn't supported?

And certainly greying out the menu items rather than hiding them makes a lot of sense (though that begs for an error message being returned by IsSupported so the user can be told why the plugin isn't available).

I still think that exposing the IsSupported function is useful (in case someone wants to modify their own usdview's behavior). And I am willing to adjust the UI portion of this pull request to grey out the unsupported render menu items, if you are swayed by my weak argument about it being okay to create all the plugins. I'm also willing to use the IsSupported call to test if the renderer is supported and bring up an error dialog if it isn't (or if the switch to that renderer fails for some other unknown reason). Just let me know if you'd like me to do either of these things.

spitzak:
Could IsSupported be based on the names of the available plugins? Some
python api that reads all the plugin directories and returns the filenames
would work for this.

marktucker:
I went with a status message instead of a popup, since that seems to be the normal way of reporting errors in usdview. I also disable the menu item and add an "unsupported" suffix in the menu in case the user missed the status message. I removed the IsSupported calls when building the menu.

I am using IsSupported elsewhere, so would appreciate it being left in the API, FWIW.

pxr/usdImaging/lib/usdviewq/appController.py
```diff
@@ -1222,6 +1222,15 @@ def _rendererPluginChanged(self, plugin):
...
                         break
+                # Then display an error message to let the user know something
+                # went wrong, and disable the menu item so it can't be selected
+                # again.
+                for action in self._ui.rendererPluginActionGroup.actions():
+                    if action.pluginType == plugin:
+                        self.statusMessage(
+                            'Renderer not supported: %s' % action.text())
+                        action.setText(action.text() + " (unsupported)")
+                        action.setDisabled(True)
```
c64kernal: Not a huge deal, but I think we can 'break' here, like above?

c64kernal:
Thanks so much, @marktucker  that all sound good.  And to be clear, you'd like to leave the whole chain of IsSupported added in this change (including the python wrappings)?

c64kernal:
(If you only literally need IsSupported that we already have, then no problem of course) My only concern with keeping the python wrapping is somehow someone coming in later not knowing that it's being depended on and deleting it as dead code... perhaps we can work it in a test and leave a comment or something like that?

pxr/usdImaging/lib/usdviewq/appController.py
marktucker: Yes, certainly.

marktucker:
I'm really only interested in the C++ call being left in, and I'm happy with it as-is.

I thought the Python wrapper might be useful, but maybe it's a bad idea to leave it in since a) it is unused, and b) it does require loading the plugin, which isn't really clear from the function. Perhaps I should just close this pull request and submit one that just does the error message and menu editing if an attempt to set the renderer fails?

Or if you do want to keep the Python wrapper that fact that it loads the plugin should probably be documented, though I'm not sure where to add such documentation.

c64kernal:
Thanks Mark -- yeah, in general, if it's code that we're not using, it'd be better to not have it in there at all, so I'd be in favor of getting rid of the python wrapper.  As for the C++ call, I wonder if adding a call to it in testUsdImagingGLBasicDrawing.cpp for coverage would make sense?  That way it'd be covered by something and would be less likely to be mistaken by someone as dead code... so maybe two pull requests if that makes sense?

marktucker:
IsSupported is actually called in hdEngine::SetRendererPlugin, so it is actually called indirectly by testUsdImagingGLBasicDrawing already, and the purpose of the function is pretty clear from the comment in SetRendererPlugin... I don't really see how an explicit call in the test would do anything to further clarify that this method shouldn't be deleted...

c64kernal:
Sorry Mark, I'm a bit confused, I thought you wanted to keep the new GetRendererPluginIsSupported that you added.  Those series of new functions were the ones I thought we were talking about...

marktucker:
The confusion is my fault I'm sure... I only added UsdImagingGLHdEngine::GetRendererPluginIsSupported to put a Python wrapper around it. I have no interest in having that method in C++.

The HdxRendererPlugin::IsSupported virtual method is the only one I care about and don't want removed. It sounded like you were talking about maybe removing it someday in this comment: https://github.com/PixarAnimationStudios/USD/pull/599#issuecomment-422498210

c64kernal:
Ah!  Got it!  Thank you!  We're all on the same page.  (I think as long as we think HdxRendererPlugin::IsSupported is useful and backends make good use of it, I don't think we have any intention of removing it).

PR #609, opened 2018-08-31
New usdresolve helper script

aloysbaillet:
### Description of Change(s)
- Added python wrapper for `ArResolver::IsRelativePath`
- New `usdresolve` command line script that resolves an asset path using the standard ArResolver methods.


jtran56:
Filed as internal issue #USD-4720.

pxr/usd/bin/usdresolve/usdresolve.py
```diff
@@ -0,0 +1,97 @@
...
+def _ConfigureAssetResolver(args, resolver):
+    if args.configureAssetPath:
+        resolver.ConfigureResolverForAsset(args.configureAssetPath)
+        resolverContext = resolver.CreateDefaultContextForAsset(args.configureAssetPath)
+    else:
+        resolver.ConfigureResolverForAsset(args.inputPath)
+        resolverContext = resolver.CreateDefaultContextForAsset(args.inputPath)
+    return resolverContext
+
+def _AnchorRelativePath(args, resolver):
```
sunyab: - ArResolver::Resolve is expected to only deal with files and not directories(*), since not all asset systems may necessarily have a notion of a directory. So I think we shouldn't call Resolve on the anchor path or os.getcwd() here. Instead, I think we should just call AnchorRelativePath on the paths as given by the user, and let the call to Resolve later on deal with it.

- I wonder if it would be better to only anchor the path if the user explicitly specifies args.anchorPath and not force the use of os.getcwd(). That way, users can see what Resolve returns for a given inputPath without the script adding on extra behavior. If they want to anchor against cwd, they could do something like ```usdresolve --anchorPath `pwd` ```.

So I think we'd be left with:
```
def _AnchorRelativePath(args, resolver):
    if args.anchorPath and resolver.IsRelativePath(args.inputPath):
       return resolver.AnchorRelativePath(args.anchorPath, args.inputPath)
    return args.inputPath
```

pxr/usd/bin/usdresolve/usdresolve.py
```diff
@@ -0,0 +1,97 @@
...
+import argparse, sys, os
+
+def _Msg(msg):
+    sys.stdout.write(msg + '\n')
+
+def _Err(msg):
+    sys.stderr.write(msg + '\n')
+
+def _ConfigureAssetResolver(args, resolver):
+    if args.configureAssetPath:
```
sunyab: Maybe simplify to:
```
    configurePath = args.configureAssetPath or args.inputPath
    resolver.ConfigureResolverForAsset(configurePath)
    return resolver.CreateDefaultContextForAsset(configurePath)
```

PR #614, opened 2018-09-12
[Cmake] Don't blindly overwrite CMAKE_MODULE_PATH.

marsupial:
### Description of Change(s)
Make sure CMAKE_MODULE_PATH passed via command line has precedence over USD.

### Fixes Issue(s)
If there are non standard paths/ installs CMAKE_MODULE_PATH is a much easier route than specifying multiple locations, but USD is currently overwriting any CMAKE_MODULE_PATH passed to it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,8 @@ if (NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
     endif()
 endif()
 
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/defaults
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
+                      ${CMAKE_SOURCE_DIR}/cmake/defaults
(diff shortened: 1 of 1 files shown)


meshula:
I think, but am not positive, that CMake searches first to last? If that is true, wouldn't you want the existing CMAKE_MODULE_PATH to have the lowest precedence? 

The documentation is ... lacking. https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html

If this change were to be made, it would probably also be preferable that the cmake scripts in USD be prefixed with USD, e.g. USD_Public.cmake, USD_Packages.cmake, etc., as there is a greater than zero probability that if you allow longer CMAKE_MODULE_PATHS, conflicting files may appear in the total search path.

marsupial:
If you don't use `-DCMAKE_MODULE_PATH=`, then `CMAKE_MODULE_PATH` is empty.
So like $PATH, you're trying to override behavior in favor of something custom/better.

jtran56:
Filed as internal issue #USD-4737.

PR #618, opened 2018-09-13
[sdf] Remove boost/noncopyable

superfunc:
### Description of Change(s)

Chipping away at the mountain of boost.


jtran56:
Filed as internal issue #USD-4738.

Issue #610, opened 2018-09-03
Embedded Third Party Libs as System Libs?


superfunc:
### Description of Issue
There are a few libs embedded in USD, e.g. rapidjson. From time to time, these libraries throw warnings, such as:

```bash
usd/pxr/base/lib/js/rapidjson/internal/stack.h:43:5: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
#if RAPIDJSON_HAS_CXX11_RVALUE_REFS
```

Given that these warnings are essentially out of USD's control, it seems like we may want to mark these libraries as system libraries in CMake, which would muffle these warnings.

jtran56:
Filed as internal issue #USD-4721.

PR #620, opened 2018-09-15
Add define for rapidjson

superfunc:
See discussion in #610, suggested by the wise @meshula


jtran56:
Filed as internal issue #USD-4741.

pxr/base/lib/js/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ pxr_library(js
         types.h
 )
 
+# This removes warnings generated by rapidjson
+target_compile_definitions(js PRIVATE RAPIDJSON_HAS_CXX11_RVALUE_REFS=1)
```
sunyab: Instead of adding this define here, can we add a #define in js/json.cpp, near where the other RAPIDJSON defines are located? rapidjson includes are contained to this one file, and it'd be good to centralize all of the necessary configuration there.

superfunc: Yeah that sounds fine by me 👍

superfunc:
Added the defines, this should work fine. My one concern would be MSVC compliance in certain areas; it'd be great if we could turn on an appveyor build for windows so this could be automatically verified here.

superfunc:
@c64kernal Note this was updated, the [requested changes] tag just didn't get removed.

c64kernal:
Hey @superfunc and @meshula -- we're wondering what compilers give you this error and why the rapidjson detection mechanism already in place isn't picking these up automatically?

meshula:
It's been a while since I looked at it, but I think the problem is that this chunk in rapidjson.h does not detect clang properly? And similarly for all the other values that @superfunc set by hand.

````
#ifndef RAPIDJSON_HAS_CXX11_RVALUE_REFS
#if defined(__clang__)
#define RAPIDJSON_HAS_CXX11_RVALUE_REFS __has_feature(cxx_rvalue_references) && \
    (defined(_LIBCPP_VERSION) || defined(__GLIBCXX__) && __GLIBCXX__ >= 20080306)
#elif (defined(RAPIDJSON_GNUC) && (RAPIDJSON_GNUC >= RAPIDJSON_VERSION_CODE(4,3,0)) && defined(__GXX_EXPERIMENTAL_CXX0X__)) || \
      (defined(_MSC_VER) && _MSC_VER >= 1600)

#define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
#else
#define RAPIDJSON_HAS_CXX11_RVALUE_REFS 0
#endif
#endif // RAPIDJSON_HAS_CXX11_RVALUE_REFS

````

superfunc:
> @superfunc set by hand.

Hand crafted, artisanal changes, these `#define`s are my art.

c64kernal:
Artisanal as they are, and they are beautiful, I'm still wondering why that chunk of code doesn't detect clang properly.  I'm running on AppleClang 10.0.x (the default with Xcode 10.1 (10B61)) and I don't see this error.

superfunc:
Let me run a clean build and see what it does.

superfunc:
Confirmed on a clean build this morning, still see these warnings, just going through a clean build via build_usd.py on mac.

superfunc:
Version info

```Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin```

c64kernal:
Thanks for the confirmation @superfunc -- it'd be great to dig in to why the code in rapidjson that is meant to detect the supported features automatically aren't working for you.

superfunc:
Yeah, that's not a great use of time in my opinion. If someone else wants to dig, feel free, but I don't wanna spend any more time on something this trivial.

c64kernal:
Okay, thanks @superfunc, that's fair.  On our side, it's hard for us to take a change we don't reasonably understand, so we'll close out this PR, but if you have time to get to the bottom of it in the future, please feel free to reopen.

PR #622, opened 2018-09-18
[usdview] Enable OpenGL 3 or 4 on OS X when possible.

marsupial:
### Description of Change(s)
Use Qt's ability to create an OpenGL 3/4 context when possible.
Disable features OS X cannot handle (hiliting and line-stiplling, for separate reasons)
More robust error checking and exception handling to release resources on failure.

### Fixes Issue(s)
Drawing via OpenGL with usdview on OS X was rather limited (atrocious probably being a better word).

### DIFF
--- a/pxr/imaging/lib/glf/contextCaps.cpp
+++ b/pxr/imaging/lib/glf/contextCaps.cpp
@@ -346,6 +346,25 @@ GlfContextCaps::_LoadCaps()
     }
 }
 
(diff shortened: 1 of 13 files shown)


jtran56:
Filed as internal issue #USD-4743.

sunyab:
Closing out old PR, the fixes here should no longer be needed with the support for Metal on MacOS as well as the removal of the legacy engine in fc24f3ff4e.

PR #623, opened 2018-09-19
Add missing export attributes to SdfLayerDelegate and SdfData

sopvop:
### Description of Change(s)

Adds export attributes missing from SdlLayerStateDelegate and SdfData. These are necessary to make https://github.com/LumaPictures/usd-qt compile on windows with MSVC

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/lib/sdf/layerStateDelegate.h
+++ b/pxr/usd/lib/sdf/layerStateDelegate.h
@@ -60,26 +60,35 @@ class SdfLayerStateDelegateBase
     , public TfWeakBase
 {
 public:
+    SDF_API
     virtual ~SdfLayerStateDelegateBase();
 
+    SDF_API
     bool IsDirty();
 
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #USD-4768.

pxr/usd/lib/sdf/layerStateDelegate.h
```diff
@@ -60,81 +60,106 @@ class SdfLayerStateDelegateBase
     , public TfWeakBase
 {
 public:
+    SDF_API
```
sunyab: Could you add:
```
#include "pxr/usd/sdf/api.h"
```
to the includes right below "pxr/pxr.h", at line 28?

sopvop:
BTW, recreating pull request was not necessary. You can change target branch if you click edit button to the right of request title.

PR #624, opened 2018-09-19
Use GUSD_API on these whole classes rather than individual functions.…

marktucker:
This is required to build the Houdini plugins using gcc5.4 which only exports
the vtable for a class if the whole class is exported (and so the whole
class must be exported to allow the use of dynamic_cast to these classes
in the Houdini plugins).

I'm hoping these additional exports will be non-controversial because they live
off in the gusd library rather than part of the USD core...


jtran56:
Filed as internal issue #USD-4769.

PR #625, opened 2018-09-19
Change the define used to identify the Houdini version

marktucker:
Use SYS_VERSION_FULL_INT define to figure out what version of Houdini is
being built against rather than HDK_API_VERSION. This is a more predictable
value that is always tied exactly to the Houdini version number.


jtran56:
Filed as internal issue #USD-4770.

PR #626, opened 2018-09-19
Fix the conversion of Quaternion USD attributes to GT attributes

marktucker:
Fix the conversion of Quaternion and Quaternion array attributes in USD
attributes to equivalent GT attribute data. Previously the USD data was
extracted as GfVec4 data, which USD would refuse to do.


jtran56:
Filed as internal issue #USD-4771.

c64kernal:
Apologies @marktucker -- do you have a version of this with the conflicts resolved?  If so, I'll try to get it merged to avoid future conflicts.

c64kernal:
Thanks @marktucker -- when I was trying to merge it, it was the quaternion reversal that I was wondering about.  I'll make a note internally too.  Thanks again!

PR #628, opened 2018-09-20
Add include for std::function in eventTree.h.

marktucker:
Include <functional> for definition of std::function used in this header.

The absence of this include caused a build failure using gcc7.3.


jtran56:
Filed as internal issue #USD-4772.

PR #621, opened 2018-09-16
Remove boost/noncopyable

superfunc:
See related PRs for rationale (e.g. #481 #484 #618 #482 )

PR #484, opened 2018-05-08
[Js] Remove boost noncopyable.

superfunc:
### Description of Change(s)
See message in previous commits of the same type.

### Fixes Issue(s)
- None filed, just cleanup

PR #630, opened 2018-09-20
[Kind] Remove unnecessary boost

superfunc:
### Description of Change(s)
See my other art at #621 #484 #481 #618

This goes about removing one of the easier pieces of boost in the USD repo (those that
can be replaced with standard C++11 constructs).


jtran56:
Filed as internal issue #USD-4774.

Issue #631, opened 2018-09-20
[Houdini] rightHanded mesh packed prim with vertex attributes render incorrectly in viewport


marktucker:
### Description of Issue
When a USD mesh with rightHanded orientation and a vertex primvar for color, opacity, or normals is rendered as a packed primitive in the Houdini viewport, the vertex attributes are set on the wrong vertices. When the geometry is unpacked to polygons, the attributes are assigned to the correct attributes.

### Steps to Reproduce
1. Open the attached vertex_colored_grid.hip file into Houdini with the Houdini plugins enabled.
2. Notice that as you move the display flag from file1 to unpack1 and back, the geometry changes. File1 is showing the USD packed primitives. Unpack1 is showing the unpacked polygons.

[vertex_colored_grid.zip](https://github.com/PixarAnimationStudios/USD/files/2403076/vertex_colored_grid.zip)

jtran56:
Filed as internal issue #USD-4775.

sunyab:
Closing old issue -- USD no longer ships with a Houdini plugin. (Plus it looks like #632 fixed this anyway but GitHub failed to close the issue automatically).

PR #632, opened 2018-09-20
Issue #631: fixing vertex attributes on rightHanded meshes when refining for the viewport

marktucker:
When refining a USD mesh packed primitive, if we have vertex attributes
and we had to reveress the winding order (i.e. the USD primitive's
orientation attribute is "rightHanded"), we need to reorganize the
vertex attribute data whether we are refining for the viewport or not.


spitzak:
What if there already is an indirect index array?

marktucker:
I haven't actually tested that case, so I don't know if it works or not. This issue is just about the difference that viewport refinement doesn't work one case where refinement to polygons works. It would be good to know if refinement (to viewport and polygons) works with indirect primvars, but if not I think that warrants a separate issue.

jtran56:
Filed as internal issue #USD-4777.

PR #635, opened 2018-09-21
Better error reporting if the user picks an unsupported renderer in usdview

marktucker:
If the user picks a Renderer in usdview, but USD fails to set this renderer
as current (because the runtime environment doesn't support it), display an
error status message, disable the Renderer in the menu, and put the suffix
"(unsupported)" on the menu item (in case the user missed the status
message).

This is a partial replacement for PR #599.


c64kernal:
This is really nice, thanks Mark!

sunyab:
Filed as internal issue #USD-4778.

PR #641, opened 2018-10-01
Fix python wrappers to be explicit for RTLD_GLOBAL environments.

abucior:
I recently ran into some problems with the Python bindings to USD. Our Python configuration uses RTLD_GLOBAL for legacy reasons.

The problem with this is that multiple USD projects contain functions with the same name. For example, there are multiple "wrapRegistry()" and "wrapNode()" functions defined in various places. We were seeing problems where, for example, ndr's module wrap function ended up calling sdr's wrapRegistry() instead of ndr's, and ndr's wrapRegistry() was never called. This led to various runtime errors.  Our workaround was to rename wrapRegistry() in sdr as "wrapSdrRegistry()", for example, and use TF_WRAP(SdrRegistry) in the module wrapper. We made a handful of changes like this to make all the wrap calls explicit.

### Description of Change(s)

This change makes all python wrappers uniquely named.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/lib/plug/module.cpp
+++ b/pxr/base/lib/plug/module.cpp
@@ -29,9 +29,9 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 TF_WRAP_MODULE
 {
-    TF_WRAP( Notice );
-    TF_WRAP( Plugin );    
-    TF_WRAP( Registry );    
+    TF_WRAP( PlugNotice );
+    TF_WRAP( PlugPlugin );    
+    TF_WRAP( PlugRegistry );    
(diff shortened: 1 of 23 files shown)


jtran56:
Filed as internal issue #USD-4800.

c64kernal:
Hi @abucior -- we think with commit https://github.com/PixarAnimationStudios/USD/commit/f61f1035c67cbb174789c9222e6dbd0a6a4d7a6e this would no longer be needed.  Please reopen if for some reason you're still running into issues that require this PR.  Thanks!

Issue #649, opened 2018-10-10
[Houdini] USD Import with Custom Traversal on Visibility seems wrong


marktucker:
When using the USD Import SOP in Houdini, with the Traversal set to Custom, and using the "Visibility" based traversal, the results seem wrong. If the hierarchy is like this:
```
/foo
  /bar            invisible
    /spangle      inherited
```
and the traversal by prim type or kind causes /foo/bar/spangle to be imported, it gets imported whether the "Visibility" test is "True" or "False". I would expect it to be treated as if it were invisible.

jtran56:
Filed as internal issue #USD-4820.

PR #650, opened 2018-10-10
[usdHoudini] Changing GusdUSD_Utils::ImageablePrimIsVisible to use ComputeVisibility

marktucker:
Changing GusdUSD_Utils::ImageablePrimIsVisible to use ComputeVisibility
to test if the prim is visible by looking at ancestor primitives, rather
than only looking at the visibility attribute of the prim itself.
Fixes Issue #649.


jtran56:
Filed as internal issue #USD-4821.

sunyab:
This pull request was superceded by the fix in commit 148dbd1, so I'm going to close this out. Thanks!

PR #652, opened 2018-10-11
Adds a fix for the stripNamespaces feature

elrond79:
### Description of Change(s)

The latest changes to how mergeTransform is handled internally broke our
stripNamespaces feature.  This adds a fix for that.

### Fixes Issue(s)
- Error when exporting usd with stripNamespace and mergeTransform options both on

### DIFF
--- a/third_party/maya/lib/usdMaya/writeJob.cpp
+++ b/third_party/maya/lib/usdMaya/writeJob.cpp
@@ -350,6 +350,65 @@ UsdMaya_WriteJob::_BeginWriting(const std::string& fileName, bool append)
         }
     }
 
+    // If stripNamespaces is on, this validates that we don't have a
+    // name collision as a result (and maintains mUsdPathToDagPathMap,
+    // which is only used to do this validity / name collision check,
+    // and so is only populated if stripNamespaces is on)
+    auto checkStrippedNamespaceIsUnique = [this] (
+            const UsdMayaPrimWriter& primWriter,
+            const UsdPrim& usdPrim)
+            -> bool {
+        auto foundPair = mUsdPathToDagPathMap.find(usdPrim.GetPath());
+        if (foundPair != mUsdPathToDagPathMap.end()){
+            if (mJobCtx.mArgs.mergeTransformAndShape) {
+                // If we're merging, we may have two primwriters with different
+                // dag paths mapping to the same prim - the transform and the
+                // shape. Check for this...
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-4823.

third_party/maya/lib/usdMaya/util.cpp
```diff
@@ -1106,7 +1106,10 @@ UsdMayaUtil::MDagPathToUsdPath(
     std::replace(usdPathStr.begin(), usdPathStr.end(), '|', '/');
     std::replace(usdPathStr.begin(), usdPathStr.end(), ':', '_'); // replace namespace ":" with "_"
 
-    SdfPath usdPath(usdPathStr);
+SdfPath
```
sunyab: Hey @elrond79, it looks like there might have been some kind of weird merge error here? It looks like the MDagPathToUsdPath function stomped over some part of the above function.

sunyab: And also, is this new function needed? I didn't see it called in this change anywhere..

elrond79:
Got rid of the uneeded function as requested...

third_party/maya/lib/usdMaya/writeJob.cpp
```diff
@@ -350,6 +350,65 @@ UsdMaya_WriteJob::_BeginWriting(const std::string& fileName, bool append)
         }
     }
 
+    // If stripNamespaces is on, this validates that we don't have a
```
sunyab: For readability we'd prefer this check to be in a separate static function above instead of in a large lambda here. Perhaps something like:
```
namespace {

// If stripNamespaces is on, this validates that we don't have a
// name collision as a result (and maintains mUsdPathToDagPathMap,
// which is only used to do this validity / name collision check,
// and so is only populated if stripNamespaces is on)
static
bool
_CheckStrippedNamespaceIsUnique(
    const UsdMayaPrimWriter& primWriter,
    const UsdPrim& usdPrim,
    const UsdMayaJobExportArgs& jobArgs,
    const UsdMayaWriteJobContext& jobContext,
    TfHashMap<SdfPath, MDagPath, SdfPath::Hash>* usdPathToDagPathMap)
{
    // ...
}

} // end anonymous namespace
```

c64kernal:
Closing this one in favor of #683 -- which we'll merge soon.  Let us know if you run into any trouble!  Thanks!

PR #663, opened 2018-10-16
Add ability to register selection callbacks in usdview

aloysbaillet:
### Description of Change(s)
The UsdviewApi class now includes two new methods: `AddPrimSelectionChangedCallback()` and `RemovePrimSelectionChangedCallback()`. These methods allow third-party plugins to connect Qt slots with the `signalPrimSelectionChanged` Qt signal that is emitted by the internal `SelectionDataModel` object.
This allows plugin views to synchronise themselves as the application selection changes.

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/usdviewApi.py
+++ b/pxr/usdImaging/lib/usdviewq/usdviewApi.py
@@ -191,6 +191,12 @@ class UsdviewApi(object):
         """DEPRECATED Returns the old settings object."""
 
         return self.__appController._settings
+    
+    def AddPrimSelectionChangedCallback(self, callback):
+        self.__appController._dataModel.selection.signalPrimSelectionChanged.connect(callback)
+
+    def RemovePrimSelectionChangedCallback(self, callback):
+        self.__appController._dataModel.selection.signalPrimSelectionChanged.disconnect(callback)
 
     def ClearPrimSelection(self):
         self.__appController._dataModel.selection.clearPrims()


jtran56:
Filed as internal issue #USD-4840.

spiffmon:
Thanks, Aloys - this is clearly needed!  Can we plus the low-level Qt pattern with RAII to ensure the signal gets properly disconnected when the plugin widget dies?  I.e. something like CreateSelectionSignalConnector(callback) that returns an object whose __del__ will disconnect?

aloysbaillet:
Hi Spiff, thanks this makes sense indeed!
I'm not actually the author of the PR but I discussed with JP (who is) and we'll try to have some updates soon!

aloysbaillet:
From JP:
Connect to the File > Open event
Expose the signalStageReplaced Qt signal, so that clients can respond to the File > Open operation and update to reflect thew newly-opened USD stage.
In a similar manner to the signalPrimSelectionChanged signal, we introduce two new methods on UsdviewApi:
AddStageReplacedCallback(): Connect a slot to this signal.
RemoveStageReplacedCallback(): Disconnect a slot from this signal.
Introduce RAII pattern for Qt signal-slot connection
There was also a request to encapsulate these in an RAII-style resource idiom. It is unclear if this adds any value, as the Qt documentation suggests signals and slots are automatically disconnected when QObjects are deleted: http://doc.qt.io/archives/qt-4.8/qobject.html#dtor.QObject
In any case, we have attempted to add this pattern via CreateStageReplacedConnection() and CreateSelectionChangedConnection(), each of which returns a Qt signal/slot pair encapsulated by a UsdviewSignalSlotConnection object.
However, in our plugin testing, we found that the del() method was never being called. According to Python documentation, we cannot guarantee that Python objects will ever be garbage collected, so Python doesn't inherently support the RAII idiom.

spiffmon:
Thank you, JP, and Aloys!

Given the information you reported, it seems like the RAII mechanism I asked for is not useful indeed, and sets up conflicting terminology in the API.  Could I ask you to remove it, and also provide a brief doc string for the remaining four new methods?

Thanks, looking forward to accepting this!
--spiff

aloysb:
I don't think you wanted to tag me here @aloysbaillet 

Nice to meet another Aloys though :)

PR #664, opened 2018-10-16
Update JealousFocus() method to handle window modality flag

aloysbaillet:
### Description of Change(s)
The **AppEventFilter.JealousFocus()** focus method has been updated to also consider the **windowModality()** setting on the widget. The role of this method is to return a flag that indicates whether the widget of interest jealously maintains its input focus. If this method returns True then the auto-focus system (which usually auto-focuses the widget under the cursor as the user moves the mouse around the screen) is _not_ applied. The previous implementation only considered the widget type. However, when plugin views are implemented via PyQt (rather than PySide), all widgets in the plugin view are wrapped in a manner that they are only considered to be instances of QWidget, and deeper instance type information is lost. 
Testing the intended modality of a given widget via the **windowModality** flag is a useful way to communicate the intended focus behaviour of third-party plugin views.

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appEventFilter.py
+++ b/pxr/usdImaging/lib/usdviewq/appEventFilter.py
@@ -93,7 +93,9 @@ class AppEventFilter(QtCore.QObject):
                 isinstance(w, QtWidgets.QComboBox) or
                 isinstance(w, QtWidgets.QTextEdit) or
                 isinstance(w, QtWidgets.QAbstractSlider) or
-                isinstance(w, QtWidgets.QAbstractSpinBox))
+                isinstance(w, QtWidgets.QAbstractSpinBox) or
+                isinstance(w, QtWidgets.QWidget) and w.windowModality() in [QtCore.Qt.WindowModal,
+                                                                            QtCore.Qt.ApplicationModal])
             
     def SetFocusFromMousePos(self, backupWidget):
         # It's possible the mouse isn't over any of our windows at the time,


jtran56:
Filed as internal issue #USD-4841.

spiffmon:
This change sounds like it makes sense, Aloys!  My only question is whether it can _replace_ the existing type-checks altogether?  That would certainly be more robust and slightly more performant.

aloysbaillet:
Hi Spiff,

I'm replying by proxying JP who wrote the code... his answer to your question is:
"
We that we cannot reliably remove the existing type checks, as there is no guarantee that the client code will have set the modality flag in the same way we have done. To ensure backwards compatibility, we should keep the old tests as well as introducing a new test.

More generally, we should probably introduce a more formal mechanism in which plugin clients can opt in or out of jealous focus. For me, jealous focus feels  quite aggressive and somewhat unexpected (or at least not strictly standard) UI behaviour, so plugin views should be able to completely opt-out of it if it doesn’t work for them.
"

spiffmon:
Thanks, Aloys and JP - that reasoning makes sense.  With regards to jealous
focus... coming from X-windows origins, the majority of folks here find
pointer-focus to be more concise and RSI-friendly than click-focus.  But
pointer focus is immensely annoying for text/numerical inputs, so ever
since Marionette (menv) days, we've done this jealous focus thing for
inputs, and finally brought it to usdview.  We've tried to find a better
way to enact it in Qt, but no luck so far.

--spiff

spitzak:
At R&H I worked some on this and came up with a scheme that uses Qt's
global shortcuts.

The result is that a keystroke is first sent to the focus widget (such as a
text field), and if it is "uninterested" in it, then it sends it to any
widget that registered that shortcut, preferring the smallest widget
surrounding the mouse, then around the focus.

I believe the implementation required the global focus handler to use rtti
on the focus widget and do different things becuase Qt lacked any working
"I used this keystroke" api. Also the shortcuts had to be registered with
an all-new api because the global shortcut api did not provide a table of
assigned shortcuts. One nice thing about this new api was it was also
designed so you had to make up a menu entry for every shortcut in your
widget, so all the possible shortcuts were always documented in the popup
or pulldown menus. A problem with it is that it won't work with generic Qt
widgets that have shortcuts.

I think it moved the focus if the shortcut was handled, and some widgets
also grabbed the focus on mouse wheel movement. However otherwise we did
not mess with Qt's focus management.

PR #666, opened 2018-10-17
[USD] Missing code / features from the merged UDIM PR.

sirpalee:
### Description of Change(s)
Some changes were not properly merged in from the UDIM pr, like:
- Using unsigned ints for handling pixel counts/dimensions instead of ints.
- Layout data is not getting offset tile ids so that udim tiles will be offset by one.

I also fixed a misleading comment in Codegen.cpp and handled udim images with a null memory request.


jtran56:
Filed as internal issue #USD-4843.

PR #672, opened 2018-10-23
[Usd] Cleanup root dir templates

superfunc:
### Description of Change(s)

This uses githubs `.github` dir to store templates, and splits up bugs from features. This cleans up the repository a little, as these are things new users shouldn't have to concern themselves with.


jtran56:
Filed as internal issue #USD-4859.

c64kernal:
Thanks very much for the awesome suggestion @superfunc ... we've landed a change to clean up the root dir that doesn't include the separation of bug vs. feature because we're not sure what that buys practically right now.  We'll close this once that change makes it to dev, hopefully very soon.  Thanks again and great idea!

Issue #673, opened 2018-10-23
[Houdini] Running ginfo on a USD file crashes


marktucker:
### Description of Issue

Running Houdini's "ginfo" utility on a USD file uses the GEO_IOTranslator from the gusd library. This calls
some global Houdini "CH" library function which rely on the existence of a "CH_Manager" object. But
ginfo (unlike houdini or hython) does not create a CH_Manager, so calling these functions leads to a
crash.

### Steps to Reproduce

1. Make sure USD is built and installed with Houdini support.
2. Run "ginfo foo.usd" on any USD file from a shell that has access to Houdini. This will crash.

jtran56:
Filed as internal issue #USD-4860.

sunyab:
This fix was released in version 19.01. Closing!

PR #674, opened 2018-10-23
[Houdini] Test for a CH_Manager before calling CH_Manager functions.

marktucker:
This fixes a crash when running ginfo on a USD file. ginfo does not create
any node/channel infrastructure, so trying to call global CH_Manager methods
leads to a crash.

If no CH_Manager exists, use a default of "1" for the sample time.

This fixes Issue #673


jtran56:
Filed as internal issue #USD-4861.

PR #683, opened 2018-11-01
[usdMaya] fix bogus name clash errors when stripping namespaces

sopvop:
When exporter strips namespaces is checks if path was already visited and if so raises an error.

However when we also merge transforms and shapes, then it's not an error to map both of them to same primitive.

### Description of Change(s)

In addition to checking if path was already visited we check if path is parent or child of visited path.

### Fixes Issue(s)
Exporting to usd with namespace stripping and merging transforms and shapes results in error and does not work.
```
// Error: Multiple dag nodes map to the same prim path after stripping namespaces: |foo|ns:bar - |foo|ns:barShape  -- Runtime Error in _BeginWriting at line 396 of /tmp/source/third_party/maya/lib/usdMaya/writeJob.cpp //
```

### DIFF
--- a/third_party/maya/lib/usdMaya/CMakeLists.txt
+++ b/third_party/maya/lib/usdMaya/CMakeLists.txt
@@ -177,6 +177,7 @@ pxr_test_scripts(
         testenv/testUsdExportShadingModeDisplayColor.py
         testenv/testUsdExportShadingModePxrRis.py
         testenv/testUsdExportSkeleton.py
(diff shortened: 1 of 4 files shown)


sopvop:
The way it is implemented will miss error message. Will push update tomorrow.

sopvop:
Now it's ready for review.

jtran56:
Filed as internal issue #USD-4891.

PR #685, opened 2018-11-02
Add a check for a valid UsdPrim when looping over affectedPrims in _RefreshObject

marktucker:
Prevent a crash when processing a stage update in hydra in which a prim is
deleted from the stage, and a collection which points directly to that prim
is updated. This would crash while processing the change to the collection,
as it would access the underlying UsdPrim from the index's PrimInfo map
without checking that it is still valid. To get around this, test the
validity of the UsdPrims that show up in the affectedPrims list.

### Fixes Issue(s)
684


jtran56:
Filed as internal issue #USD-4893.

PR #692, opened 2018-11-12
[USD] Invert the value of Simple Light's position when used as a directional light.

sirpalee:
### Description of Change(s)
When the .w coordinate of the simple light's position is set to zero, the simple light is supposed to behave as directional light. However, the direction needs to be inverted, so the rest of the math in `simpleLighting.glslfx` and `previewSurface.glslfx` behaves as it should. (the logic expects the calculated light direction to point "towards" the light)

### Fixes Issue(s)
No reported issues.

### DIFF
--- a/pxr/imaging/lib/glf/shaders/simpleLighting.glslfx
+++ b/pxr/imaging/lib/glf/shaders/simpleLighting.glslfx
@@ -266,7 +266,7 @@ integrateLightsDefault(vec4 Peye, vec3 Neye, LightingInterfaceProperties props)
         vec4 Plight = lightSource[i].position;
 
         vec3 l = (Plight.w == 0.0)
-                    ? normalize(Plight.xyz)
+                    ? normalize(-Plight.xyz)
                     : normalize(Plight - Peye).xyz;
 
         vec3 h = normalize(l + vec3(0,0,1));    // directional viewer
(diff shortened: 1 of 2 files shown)


spitzak:
This seems strange since the other alternative is Plight-Peye. If the light is infinitely far away then Peye can assume to be at the origin and the result is normalize(Plight), not normalize(-Plight).
I think possibly the code calculating the light from this vector for directional lights needs to be fixed instead?

sirpalee:
This is not documented in the SimpleLight class, or anywhere, so it's non-trivial to decide. If the intended behaviour is to make the light infinitely far away when w is zero, then yes, the current code is right. If Plight is supposed to act as a "direction" in case of w being zero, then the proposed change is correct.

HdSt does not support distant lights (it maps other lights in HdStLight::_ApproximateAreaLight to a simple light), so I had no idea what's the preferred behaviour.

sirpalee:
From OpenGL's fixed function pipeline: 

> The position is transformed by the modelview matrix when glLight is called (just as if it were a point), and it is stored in eye coordinates. If the w component of the position is 0, the light is treated as a directional source. Diffuse and specular lighting calculations take the light's direction, but not its actual position, into account, and attenuation is disabled. Otherwise, diffuse and specular lighting calculations are based on the actual location of the light in eye coordinates, and attenuation is enabled. The initial position is (0, 0, 1, 0); thus, the initial light source is directional, parallel to, and in the direction of the −z axis.

Which matches the current behaviour in the source. So yeah, the current behaviour makes sense if we follow the old fixed function pipeline.

I'm going to close this PR and invert the behaviour in our tool because I assume simple light was modelled after the fixed function pipeline.

PR #698, opened 2018-11-18
[usdMaya] Importing instances.

### LINKED ISSUE #142: maya plugin: support export and import of instances
@spiffmon just called to my attention that the maya plugin does not currently support export/import of instances (in issue #141).  I'm making this ticket to track progress.  

do you know of any incompatibilities between Maya's and USD's implementations of instancing which would make this task more difficult?

sirpalee:
### Description of Change(s)
The PR fixes #142, by adding new instance modes to usdImport. Since recreating the exported Maya hierarchy during import is non-trivial, I'm providing three modes when importing instances. "As Transform" ignores the instances and creates a transform only (this is the old behaviour), "Flatten" where the meshes are de-instanced, but the resulting hierarchy is close to Maya's and "Build Sources" where we create the non-deterministic Masters generated by USD.

The PR is not finished yet, it's missing tests, and the code handling the creation of an instance could be moved to a utility module. Before fixing that, it would be good to figure out if we are happy with the current modes or want anything more. An interesting challenge is to recreate the hierarchy from the USD file, which is mostly possible, but finding the "real" master of the scene is non-deterministic.

### Fixes Issue(s)
#142

### DIFF
--- a/third_party/maya/lib/usdMaya/importCommand.cpp
+++ b/third_party/maya/lib/usdMaya/importCommand.cpp
@@ -70,6 +70,9 @@ UsdMayaImportCommand::createSyntax()
     syntax.addFlag("-ar",
                    UsdMayaJobImportArgsTokens->assemblyRep.GetText(),
                    MSyntax::kString);
+    syntax.addFlag("-im",
+                   UsdMayaJobImportArgsTokens->instanceMode.GetText(),
+                   MSyntax::kString);
     syntax.addFlag("-md",
                    UsdMayaJobImportArgsTokens->metadata.GetText(),
                    MSyntax::kString);
--- a/third_party/maya/lib/usdMaya/jobArgs.cpp
+++ b/third_party/maya/lib/usdMaya/jobArgs.cpp
@@ -559,6 +559,14 @@ UsdMayaJobImportArgs::UsdMayaJobImportArgs(
                 UsdMayaJobImportArgsTokens->shadingMode,
                 UsdMayaShadingModeTokens->none,
                 UsdMayaShadingModeRegistry::ListImporters())),
+        instanceMode(
+            _Token(userArgs,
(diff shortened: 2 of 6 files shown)


jtran56:
Filed as internal issue #USD-4922.

sirpalee:
@spiffmon @sunyab Any feedback regarding the currently implemented modes before I move on adding tests?

c64kernal:
Hey @sirpalee  -- here's the feedback from our sets team:

Some notes on functionality:

- Should the masters be organized under a group in Maya so that they don't clutter the root (since they will show up in the Outliner)?

-  The code invis's the masters when it creates the instanced duplicates; why not just invis the masters to begin with, when they're created?

- Perhaps the nicest behavior would be to just delete the masters (instead of just invis'ing them) in Maya once we have made all the instanced duplicates (since Maya does implicit instancing with no need for explicit masters anyways) -- this has the effect of making it so that a Maya scene exported with instances should get imported back the same (modulo some shape names). If we deleted the masters at the end of import, then the other two bullet points above are moot.

Some nits on style:

- Suggestion that the type _PrimReaders to be something more descriptive, e.g., _PrimReaderMap.

- Do we want the tokens AsTransform, BuildSources, etc. to be lowerCased or UpperCased? UpperCased matches the capitalization for the assemblyRep parameter, but unfortunately assemblyRep is a bit of an outliner if you look at all the export options, which use lowerCased tokens.

Thanks again!

sirpalee:
Hey @c64kernal thanks for the feedback! Deleting the masters after importing sounds like the right thing to do, I'll get on that right now, and tackle the two remaining points as well.

c64kernal:
Closing this out as development for the Maya USD plugins moved to https://github.com/Autodesk/maya-usd. Please resubmit in that repo for future development.

PR #702, opened 2018-11-22
[Houdini] Fix for compilation errors caused by missing iostream includes.

marktucker:
Include iostream where std::cout or std:cerr are used, and remove using
declarations for these classes in files where they are not used.

These errors showed up trying to build this library against Houdini 17.4.84.
Presumably caused by some combination of include cleanup in the Houdini
and USD baselines.


jtran56:
Filed as internal issue #USD-4939.

PR #710, opened 2018-12-01
[Usd] Send stage notification change on mute

superfunc:
### Description of Change(s)
It seems like this notification should also get sent during mute/unmute operations.
I ran into this while writing some widgets that need to respond to layer muting, and was surprised it wasn't firing.


jtran56:
Filed as internal issue #USD-4950.

PR #717, opened 2018-12-07
[Usdview] Query available formats for Open

superfunc:
### Description of Change(s)
This uses #532 to query the available formats for the open dialog in usdview. This is useful for external consumers of USD who have file format plugins for legacy or otherwise external formats.

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appController.py
+++ b/pxr/usdImaging/lib/usdviewq/appController.py
@@ -2267,13 +2267,19 @@ class AppController(QtCore.QObject):
             t.PrintTime('tear down the UI')
 
     def _openFile(self):
-        (filename, _) = QtWidgets.QFileDialog.getOpenFileName(self._mainWindow, "Select file",".")
+        extensions = Sdf.FileFormat.FindAllFileFormatExtensions()
+        fileFilter = "USD Compatible Files (" + " ".join("*." + e for e in extensions) + ")" 
+        (filename, _) = QtWidgets.QFileDialog.getOpenFileName(
+            self._mainWindow,
+            caption="Select file",
+            dir=".",
+            filter=fileFilter,
+            selectedFilter=fileFilter)
+
         if len(filename) > 0:
-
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-4973.

pxr/usdImaging/lib/usdviewq/appController.py
```diff
@@ -2267,13 +2267,19 @@ def _cleanAndClose(self):
             t.PrintTime('tear down the UI')
 
     def _openFile(self):
-        (filename, _) = QtWidgets.QFileDialog.getOpenFileName(self._mainWindow, "Select file",".")
-        if len(filename) > 0:
+        extensions = Sdf.FileFormat.FindAllFileFormatExtensions()
+        fileFilter = "USD Compatible Files (" + " ".join("*." + e for e in extensions) + ")"
```
spiffmon: @superfunc - my cincern here is that if the list of supported types is long, the .usd* native types might get lost at the end.  Can you flatten the set to a list and pull the usd types out to the front?

pxr/usdImaging/lib/usdviewq/appController.py
```diff
@@ -2355,6 +2361,9 @@ def _reopenStage(self):
         QtWidgets.QApplication.setOverrideCursor(QtCore.Qt.BusyCursor)
 
         try:
+            # Pause the stage view while we update
+            self._stageView.setUpdatesEnabled(False)
+
```
spiffmon: Unrelated change, but I understand why you bundled it 8-)

pxr/usdImaging/lib/usdviewq/appController.py
superfunc: Yeah good note.

pxr/usdImaging/lib/usdviewq/appController.py
superfunc: :)

PR #723, opened 2018-12-13
Hydra houdini plugin

spitzak:
### Description of Change(s)
For experimental use, uses the Hydra renderer in the Houdini viewport for gUSD Packed Prims (ie the output of the USD Import SOP). See previous pull request.

### Fixes Issue(s)
This has been updated to fix ROP usdoutput when the input contains gUSD Packed Prims, by duplicating a work-around in the collector code from gUSD.


jtran56:
Filed as internal issue #USD-4976.

sunyab:
Closing out this PR since USD no longer distributes a Houdini plugin in favor of native support in Houdini Solaris. Thanks!

PR #727, opened 2018-12-20
Remove references to Python from usdImaging

sunyab:
### Description of Change(s)
Removes includes and linkage to Python from usdImaging.
The usdImaging library no longer has any Python-related
code, so these are no longer needed.


jtran56:
Filed as internal issue #USD-4982.

PR #728, opened 2018-12-20
Cleanup usdImagingGL CMakeLists.txt

sunyab:
### Description of Change(s)
 Cleanup for CMakeLists.txt

- Remove PYTHON_INCLUDE_PATH from CMakeLists
PYTHON_INCLUDE_PATH was deprecated in favor of
PYTHON_INCLUDE_DIRS in CMake v2.8.0, and the
minimum required version for USD is v2.8.8.

- Use PYTHON_LIBRARIES in CMakeLists.txt
PYTHON_LIBRARY is a user-specified input for the
FindPython module and should not be used directly.
Use the PYTHON_LIBRARIES variable instead.

### DIFF
--- a/pxr/usdImaging/lib/usdImagingGL/CMakeLists.txt
+++ b/pxr/usdImaging/lib/usdImagingGL/CMakeLists.txt
@@ -7,7 +7,6 @@ if (NOT ${PXR_ENABLE_GL_SUPPORT})
     return()
 endif()
 
-# Note: both python include directives are included for compatibility.
 pxr_library(usdImagingGL
     LIBRARIES
         gf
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-4983.

sunyab:
I'm actually no longer sure about the PYTHON_LIBRARY vs PYTHON_LIBRARIES distinction, so I'm going to close this out for now.

meshula:
FWIW, isn't the convention in Cmake that FOO_LIBRARY means "the library that is FOO" and FOO_LIBRARIES means "FOO_LIBRARY plus things that must travel with it for a build to succeed, like tbb or whatever"

PR #735, opened 2019-01-07
[USD] Allowing animated textures for drawModeAdapter

sirpalee:
### Description of Change(s)
The PR adds support for animated textures when using the draw mode to override drawing.

### Fixes Issue(s)
None reported.


jtran56:
Filed as internal issue #USD-4995.

PR #737, opened 2019-01-08
[UsdVolImaging] Fixing UsdVolImaging API defines and exposing private tokens

sirpalee:
### Description of Change(s)
The PR fixes two problems. The first one is to define a separate API macro for the UsdImagingVol, which was using UsdImaging's API macro before. While this won't lead to issues, it goes against the practice seen in other modules. The other improvement is to expose the openvdbAsset and field3dAsset tokens, which are used as bprim types for Hydra Render delegates. Without exposing these tokens, the render delegates have to redefine them.

Refer to the changes in our Arnold Render Delegate to see how exposing the tokens simplifies the code: https://github.com/LumaPictures/usd-arnold/compare/usd_vol_tokens .

### Fixes Issue(s)
None reported.

### DIFF
--- a/pxr/usdImaging/lib/usdVolImaging/CMakeLists.txt
+++ b/pxr/usdImaging/lib/usdVolImaging/CMakeLists.txt
@@ -8,6 +8,10 @@ pxr_library(usdVolImaging
     PUBLIC_CLASSES
         openvdbAssetAdapter
         field3dAssetAdapter
+        tokens
+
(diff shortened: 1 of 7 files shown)


jtran56:
Filed as internal issue #USD-4997.

c64kernal:
Thanks very much @sirpalee -- this is great, we'll get this merged in soon!

PR #742, opened 2019-01-11
Publish the normals attribute or a normals primvar from points primitives through Hydra

marktucker:
Publish the normals attribute or a normals primvar from points primitives through
Hydra using the same code used for curve primitives. Generally the normals of
a points primitive will not be used by a renderer, but it is rather
unexpected that this information isn't even accessible for a render that
does want to use this information (such as to render the points as disks).


marktucker:
I can't figure out if something actually went wrong with my change here, or if there was some mysterious build system failure?

jtran56:
Filed as internal issue #USD-5015.

c64kernal:
@marktucker thanks for this PR!  When we merge it in, we'll also make the change to _IsBuiltinPrimvar to return true for primvarsNormals as well.

PR #743, opened 2019-01-11
Expose the velocities attribute through Hydra to render delegates

marktucker:
Expose the velocities attribute through Hydra to render delegates for
PointBased prims (Mesh, BasisCurves, and Points). This will allow renderers
to generate deformation motion blur even in the face of changing topology.

### DIFF
--- a/pxr/imaging/lib/hd/changeTracker.cpp
+++ b/pxr/imaging/lib/hd/changeTracker.cpp
@@ -625,6 +625,7 @@ HdChangeTracker::IsAnyPrimvarDirty(HdDirtyBits dirtyBits, SdfPath const &id)
     bool isDirty = (dirtyBits & (DirtyPoints|
                                  DirtyNormals|
                                  DirtyWidths|
+                                 DirtyVelocities|
                                  DirtyPrimvar)) != 0;
     _LogCacheAccess(HdTokens->primvar, id, !isDirty);
     return isDirty;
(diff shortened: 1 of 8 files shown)


marktucker:
I can't figure out if something actually went wrong with my change here, or if there was some mysterious build system failure?

c64kernal:
It's not your fault!  We're working on getting Travis back up and running.

jtran56:
Filed as internal issue #USD-5016.

c64kernal:
Hey @marktucker -- we're hoping to start reviewing this after the 19.03 release (in a couple of weeks).  One thing I wanted to mention is that we've simplified the API in 19.03 so that we can avoid some duplicated code.  We'll take care of cleaning that up in this change.  I'll let you know if the team has any more feedback once we start reviewing.  Thanks Mark!

marktucker:
Thank you. I felt like there was a lot of copy-pasting of code, but I'm a big proponent of:
a) not rocking the boat, and
b) making my code consistent with the  code around it
So glad to hear this is going to be simplified, and I appreciate your offer to clean up this change along with the rest. Thanks!

c64kernal:
Thanks @marktucker -- totally agreed.  One thing we're a bit concerned about is that we don't use this and have no test for it, so if we botch this with the refactoring, please forgive us :)

c64kernal:
Closing this one, it's been supplanted by #824

PR #745, opened 2019-01-16
[USD] Updating material x test to match the latest spec

sirpalee:
### Description of Change(s)
Updating the material x test to accommodate for a change in the MaterialX spec: changing the name of the `mix` node's parameter from `mask` to `mix`.


jtran56:
Filed as internal issue #USD-5017.

PR #747, opened 2019-01-17
Allow round tripping of namespaced primvars from USD -> Houdini geometry -> USD

marktucker:
Allow round tripping of namespaced primvars from USD -> Houdini
geometry -> USD. To do this, we convert the USD attribute name with the
namespace separator into a form that follows the rules for Houdini
geometry attributes using the UT_VarEncode class. This class is only
available in Houdini 17.5 and above, so this attribute name encoding is
only enabled when building against Houdini 17.5 or greater.

When SOP geometry is being converted back to USD attributes, the decode
method of UT_VarEncode is used to turn any encoded attribute names back
into their original form.

### DIFF
--- a/third_party/houdini/lib/gusd/primWrapper.cpp
+++ b/third_party/houdini/lib/gusd/primWrapper.cpp
@@ -1085,7 +1085,7 @@ GusdPrimWrapper::loadPrimvars(
                     gtData = new GT_DAIndirect( remapIndicies, gtData );
                 }
                 if( point ) {
-                    *point = (*point)->addAttribute( attr_name.c_str(), gtData, true );
+                    *point = (*point)->addAttribute( attrname.c_str(), gtData, true );
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


spitzak:
What does VarEncode actually do? Is the name completely mangled or does it just change punctuation?

marktucker:
It uses an encoding scheme based on "punycode" to take an arbitrary string and turn it into something that is just alphanumeric characters and underscores, safe for a SOP attribute name, but in a way that is reversible. So for example "foo:bar" becomes "xn__foobar_rla".

c64kernal:
Hey @marktucker  -- does this PR compile for you?  I'm wondering if there's a typo in attr_name?  Thanks!

marktucker:
It most certainly does not compile... Sorry about that. The process for me to make changes to the gusd library and push them up to this repository is convoluted to say the least. I obviously messed up something along the way. New commit submitted to fix the problem.

c64kernal:
No worries at all, thanks for the fix Mark!

PR #754, opened 2019-01-25
Add support for Alembic Facesets as UsdGeomSubsets in usdAbc

abucior:
### Description of Change(s)

This work is a continuation of Adam Ferrall-Nunge's work in https://github.com/PixarAnimationStudios/USD/pull/219, which added support for Alembic facesets to USD. At that time they were represented as Faceset data on the parent prim, which was subsequently read by pxrUsdIn and expanded back to faceset nodes. That PR was never accepted because UsdGeomSubset was introduced which became the more appropriate choice for representation of Alembic Facesets. 
To a large extent this simplified the problem: The exact same hierarchy is maintained between Alembic and USD with only the type changing (Alembic "Faceset" to USD "GeomSubset"). pxrUsdIn already has support for GeomSubset, so there was no need to make any pxrUsdIn changes for Katana.

### Testing

I've added a new unit test that performs round-tripping of faceset data from USD to Alembic and back again, verifying that the data is preserved as best as possible. It's not lossless, due to schema differences between Alembic and USD, but it's as good as can be represented, in my opinion.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/plugin/usdAbc/CMakeLists.txt
+++ b/pxr/usd/plugin/usdAbc/CMakeLists.txt
@@ -66,6 +66,7 @@ pxr_test_scripts(
     testenv/testUsdAbcBugs.py
     testenv/testUsdAbcCamera.py
     testenv/testUsdAbcConversionSubdiv.py
+    testenv/testUsdAbcFaceset.py
     testenv/testUsdAbcIndexedProperties.py
     testenv/testUsdAbcInstancing.py
     testenv/testUsdAbcUvReadWrite.py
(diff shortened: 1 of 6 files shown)


c64kernal:
Thanks for the PR @abucior!  Do you mind changing the base branch to the dev branch?

abucior:
> Thanks for the PR @abucior! Do you mind changing the base branch to the dev branch?

Hi George. Sorry, I had a bit of trouble changing the base branch. Instead I've created a new pull request based off of the dev branch, and I'm closing this pull request. 
https://github.com/PixarAnimationStudios/USD/pull/758

c64kernal:
No worries, thanks so much @abucior!

PR #759, opened 2019-02-01
[usdKatana] Changing the inputType for Texcoord primvars to vector.

sirpalee:
### Description of Change(s)
The PR changes how the conversion of texCoord2f/texCoord3f/texCoord2f[]/texCoord3f[] works in pxrUsdIn. Previously, we created arbitrary attributes with the inputType of float / double and elementSize respective of the tuple size of the primvar. This was changed to inputType of vector2/3.

The reason for this change is: there is no other way to create vector2 typed arbitrary values in Katana, as texcoord2f/h/d are the only attribute specs with a non-empty role name. texCoord3f was included in the change for consistency.

### DIFF
--- a/third_party/katana/lib/usdKatana/utils.cpp
+++ b/third_party/katana/lib/usdKatana/utils.cpp
@@ -607,12 +607,12 @@ _KTypeAndSizeFromUsdVec2(TfToken const &roleName,
 {
     if (roleName == SdfValueRoleNames->Point) {
         *inputTypeAttr = FnKat::StringAttribute("point2");
-    } else if (roleName == SdfValueRoleNames->Vector) {
+    } else if (roleName == SdfValueRoleNames->Vector ||
+               roleName == SdfValueRoleNames->TextureCoordinate) {
         *inputTypeAttr = FnKat::StringAttribute("vector2");
     } else if (roleName == SdfValueRoleNames->Normal) {
         *inputTypeAttr = FnKat::StringAttribute("normal2");
-    } else if (roleName == SdfValueRoleNames->TextureCoordinate ||
-               roleName.IsEmpty()) {
+    } else if (roleName.IsEmpty()) {
         *inputTypeAttr = FnKat::StringAttribute(typeStr);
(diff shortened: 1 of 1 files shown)


chadrik:
The background on this is that we ran into problems rendering secondary UV sets from maya in KtoA.  

There are a number of ways to solve this:

1. modify KtoA to treat inputType of `float` and elementSize of 2 as `vector2`.  Mike Farnsworth at Solid Angle reports that KtoA used to do this, but the feature was removed because it caused more problems.  From their release notes:  "the hack attribute geometry.arbitrary.<name>.strictType was added for force KtoA to skip this automatic type inference. Given that the arbitrary data should be formed correctly in the first place, and then have either geometry.arbitrary.<name>.inputType or geometry.arbitrary.<name>.outputType set to indicate a particular interpretation, the automatic type deduction is removed."
2. use a location decorator to convert the data to `vector2`.  doing this avoids the need for this PR, but it doesn't solve the problem for other renderers.
3. check for `usd.role` in KtoA. KtoA already supports prman specific attributes because of usdKatana (like the one specifying polygon winding order), so there is already a precedent.  I like this solution, however, usdMaya and usdHoudini are not setup to export `texCoord2f`.  

The approach taken in this PR felt the "most correct".

jtran56:
Filed as internal issue #USD-5057.

stevelavietes:
While I agree with this change, I want to hold off on merging it until the shipping version of RfK recognizes the "vector2" inputType. It's a small change there but the release cycles of the two projects are not synced.

(I was also concerned that katana's "UV Viewer" might be inputType-sensitive -- but it doesn't appear to be at the moment.)

chadrik:
Thanks, Steve. Do you have an approximate release date so that we can check
back around then?

stevelavietes:
It looks like the next RfK release is scheduled for early May. I will make sure the gating change to RfK is present in that release.

c64kernal:
Closing as the Katana plugin has been moved out of this repo to https://github.com/TheFoundryVisionmongers/KatanaUsdPlugins.

PR #760, opened 2019-02-07
[hdx] TaskController::TestIntersections allows selection by nearestToCamera

elrond79:
### Description of Change(s)

This allows HdxTaskController::TestIntersection to return results that are closest (by depth) to the camera, instead of just closest to the camera center.

Exposing this would be useful for AL_USDMaya.


jtran56:
Filed as internal issue #USD-5076.

c64kernal:
Thanks so much @elrond79 -- this is great, we'll try to get this in soon!

PR #761, opened 2019-02-07
Squashed commits of Draco plugin for USD.

igorvytyaz:
### Description of Change(s)
* Initial version of Draco plugin for USD.
* Fixing copyright date.
* Fixing more copyright dates.
* Renaming Draco CMake option in USD build script.
* Fixing attribute indices with polygon recovery and refactoring.
* Checking quantization bits option range, zero indicates no quantization.
* Adding compression level option.
* Changing hole_faces and added_edges attribute types to uint8_t.
* Deleting Draco decoder before mesh translation starts.
* Finding Draco library and making it available as a build variable.
* Updating copyright headers.
* Updating Draco plugin version to match Draco release version.


jtran56:
Filed as internal issue #USD-5077.

elrond79:
Was curious about how this plugin works. Does it compress data in "standard" usdc files, or does it allow creating / reading of another file format?  If the former, is there a standard way to identify draco-compressed files? What sort of compression does it give over a normal usdc - say, for some of the standard models here:

https://casual-effects.com/data/

spiffmon:
Hi Paul,
   I'll leave it to Igor to discuss compression results, but the high-level
picture is that for any given USD file, the compressor application produces
one .drc file for each well-defined Mesh present in the original, plus a
new USD file with all the Draco-handled data removed, replaced by a
reference to the appropriate .drc file, with a FileFormat plugin
translating the drc files on the fly.  The assembly of resulting files
could be put into a single usdz file, but that convenience hasn't been done
yet, iirc.

--spiff

igorvytyaz:
This PR is obsolete.

New PR is here: https://github.com/PixarAnimationStudios/USD/pull/912

c64kernal:
Thanks @igorvytyaz closing this one.

PR #768, opened 2019-02-21
[usdview] Added a check to _rangeBegin(End)Changed to only update if the value changes

ianlawson:
### Description of Change(s)

Fixes a bug that would reset the current frame when the start/end frame widgets loses focus and no changes were made.


jilliene:
Filed as internal issue #USD-5094

PR #774, opened 2019-02-25
Add DSO export visibility to UsdImagingGLEngine._UpdateHydraCollection 

murphyeoin:
### Description of Change(s)
When upgrading AL_USDMaya to USD 19.01, see PR [here](https://github.com/AnimalLogic/AL_USDMaya/pull/133), we (actually, our friends at Luma who did the work) ended up calling _UpdateHydraCollection (see [here](https://github.com/AnimalLogic/AL_USDMaya/blob/0a119b251ea11c3b4d91c5aa6551c6c300312009/lib/AL_USDMaya/AL/usdmaya/nodes/Engine.cpp#L63))
This worked without problems in Linux, but failed on Windows due to the method's symbol not being exported

### Fixes Issue(s)
- Add DSO export visibility to UsdImagingGLEngine._UpdateHydraCollection in usdImagingGL/engine.h

### DIFF
--- a/pxr/usdImaging/lib/usdImagingGL/engine.h
+++ b/pxr/usdImaging/lib/usdImagingGL/engine.h
@@ -411,6 +411,7 @@ protected:
 
     // Create a hydra collection given root paths and render params.
     // Returns true if the collection was updated.
+    USDIMAGINGGL_API
     static bool _UpdateHydraCollection(HdRprimCollection *collection,
                           SdfPathVector const& roots,
                           UsdImagingGLRenderParams const& params,


c64kernal:
Thanks @murphyeoin -- we'll get this merged in the next push to dev.

One thing I'd love to talk about at some point is what it would take to get UsdImagingGLEngine to be a final class.  Internally we don't derive from it, but we've resisted turning anything private or final because we know there are several cases out there that make use of this pattern.  A pattern I'd love to promote instead is to use a UsdImagingGLEngine as a helper in your own engine-like classes, rather than a base class of them. I think we'd need to better understand the use cases for the derivation and probably expose more public functionality on UsdImagingGLEngine.  Anyway, idle thoughts for now :)

Filed as internal ticket USD-5089

jilliene:
Filed as internal issue #USD-5098

murphyeoin:
Thanks @c64kernal, your idea sounds good! And thanks for helping us out with such a quick response
Eoin

PR #775, opened 2019-02-25
[Houdini] Provide support for adding in-memory stages to the GusdStageCache

marktucker:
Adding support for the explicit addition of stages to the GusdStageCache. The use case here is that we have in-memory stages (with anonymous root layers) from which we would like to construct USD packed primitives. Rather than having the GusdStageCache construct new (masked) stages from these in-memory stages, it makes more sense to just tell the cache that these full stages are already in memory.

Also had to fix the cache invalidation code to properly deal with anonymous layer identifiers so that these in-memory stages can be modified, and the cache properly invalidated in response.


jilliene:
Filed as internal issue #USD-5099

PR #777, opened 2019-02-28
[Houdini] Changes for future Houdini release related to caching packed primitive bounding boxes

marktucker:
As of the next major Houdini release, GU_PackedImpl contains a bounding box object for caching a 
calculated bounding box. Therefore, if building against this future release of Houdini:
* Removed redundant bounding box cache from GusdGU_PackedUSD, since it
  can use the one in GU_PackedImpl, and getBounds shouldn't use the cache.
* Made sure that GusdGU_PackedUSD::getBounds always initializes the box,
  even if returning false

For all currently shipping versions of Houdini, the behavior of this code is unchanged.

### DIFF
--- a/third_party/houdini/lib/gusd/GU_PackedUSD.cpp
+++ b/third_party/houdini/lib/gusd/GU_PackedUSD.cpp
@@ -499,7 +499,7 @@ GusdGU_PackedUSD::intrinsicType() const
 {
     // Return the USD prim type so it can be displayed in the spreadsheet.
     UsdPrim prim = getUsdPrim();
-    return UT_StringHolder( prim.GetTypeName().GetText() );
+    return GusdUSD_Utils::TokenToStringHolder( prim.GetTypeName() );
 }
 
 const UT_Matrix4D &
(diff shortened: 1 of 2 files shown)


third_party/houdini/lib/gusd/GU_PackedUSD.cpp
```diff
@@ -638,38 +642,56 @@ GusdGU_PackedUSD::save(UT_Options &options, const GA_SaveMap &map) const
...
+                box )) {
+            return true;
+        }
+#else
+        if( GusdBoundsCache::GetInstance().ComputeUntransformedBound(
+                        prim,
+                        UsdTimeCode( m_frame ),
                     purposes,
                         m_boundsCache )) {
                 box = m_boundsCache;
```
spitzak: This patch could be a lot shorter if in both cases you put the result into box, then for the older version you copied box to m_boundscache. Ie:
```
    if (GuseBoundsCache::Get...(..., box)) {
    #if SYS_VERSION_FULL_INT < 0x12000000
      m_boundsCache = box;
    # endif
      return true;
   }

```

marktucker: Agreed, thank you for the suggestion!

jilliene:
Filed as internal issue #USD-5126

PR #783, opened 2019-03-05
[usdview] fix for Reset View

elrond79:
### Description of Change(s)
Selecting the "View > Reset View" menu item in usdview was giving errors. This fixes that.

### Fixes Issue(s)
- Error doing "View > Reset View" menu item in usdview

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appController.py
+++ b/pxr/usdImaging/lib/usdviewq/appController.py
@@ -746,7 +746,7 @@ class AppController(QtCore.QObject):
 
             self._ui.actionFrame_Backwards.triggered.connect(self._retreatFrame)
 
-            self._ui.actionReset_View.triggered.connect(self._resetView)
+            self._ui.actionReset_View.triggered.connect(lambda: self._resetView())
 
             self._ui.topBottomSplitter.splitterMoved.connect(self._cacheViewerModeEscapeSizes)
             self._ui.primStageSplitter.splitterMoved.connect(self._cacheViewerModeEscapeSizes)


spiffmon:
Hi @elrond79 - under what circumstances does this fail currently?  "Reset View" works fine for me locally with pyside2 usdview on linux.

elrond79:
Odd... just tested with usd-0.19.01 (latest "stock" build we have) and our build of the latest from dev, and the both show the same thing - when you click "Reset View", nothing happens, and this pops up in the terminal:

```
  File "/luma/dev/paulm/rez-release/usd/0.19.5.luma01/platform-linux/gcc-6.3/lib/python/pxr/Usdviewq/appController.py", line 2101, in _resetView
    item = self._getItemAtPath(selectPrim.GetPath())
AttributeError: 'bool' object has no attribute 'GetPath'
```

These were both running with pyside-2.5.11, on linux (CentOS 7).

The error makes sense - if you look at the docs for QAction, the triggered signal has a bool parameter:

https://doc.qt.io/qt-5/qaction.html#triggered

...so presumably this is getting passed into the callback we connect (self._resetView).  When _resetView have any arguments, I'm guessing Qt just automatically called it without any arguments, but since it was changed to accept an (optional) arg, Qt presumably now passes along it's "checked" param.  Thus it complaining about selectPrim (a bool) not having GetPath().

Of course, that doesn't explain why you're not seeing the same behavior...

jilliene:
Filed as internal issue #USD-5127

sunyab:
@elrond79, I was also unable to reproduce this. When I tried your patch in a local build, I got the following error when I hit "Reset View": 

`TypeError: <lambda>() takes exactly 1 argument (0 given)`

Which implies that our callback isn't getting any arguments passed to it.

I'm wondering if the PySide version we're testing on is older than what you're using. In the usdview interpreter, could you try:
```
import PySide2
print PySide2.__version__
```
and see what that gives you? I've tried it on two builds, one of which was `2.0.0~alpha0` and the other was `5.6.0a1.dev1524150472`.

elrond79:
So, our PySide2 version is 5.11.0.

I updated the commit so that it just now wraps _resetView in a lambda with no args, which should work for both cases...

elrond79:
closed in favor of:
https://github.com/PixarAnimationStudios/USD/pull/868

PR #786, opened 2019-03-07
[Glf] Fixed an empty GlfSimpleShadowArray instance being leaked.

ppenenko:
### Description of Change(s)

An instance of GlfSimpleShadowArray created by GlfSimpleLightingContext was missing a TfCreateRefPtr wrapper which caused it to be leaked. In the pxrUsd Maya plugin, this happened on every frame because of a temporary lighting context used to retrieve lighting settings from Maya.

By the way, is TfRefPtr here to stay, or can it be replaced with boost or standard pointers in the future? TfCreateRefPtr is unfortunately very easy to miss.

### DIFF
--- a/pxr/imaging/lib/glf/simpleLightingContext.cpp
+++ b/pxr/imaging/lib/glf/simpleLightingContext.cpp
@@ -69,7 +69,7 @@ GlfSimpleLightingContext::New()
 }
 
 GlfSimpleLightingContext::GlfSimpleLightingContext() :
-    _shadows(new GlfSimpleShadowArray(GfVec2i(1024, 1024), 0)),
+    _shadows(TfCreateRefPtr(new GlfSimpleShadowArray(GfVec2i(1024, 1024), 0))),
     _worldToViewMatrix(1.0),
     _projectionMatrix(1.0),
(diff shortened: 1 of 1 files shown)


c64kernal:
Great catch, @ppenenko  -- thank you.  We don't appear to have a record of a CLA from you, would you mind sending one in (or asking your company to amend a previous CLA to include you).

ppenenko:
Oh sorry, I thought our corporate CLA covered me already, I'll look into it.

c64kernal:
Thanks @ppenenko!

jilliene:
Filed as internal issue #USD-5135

c64kernal:
Just an update on this, @ppenenko  -- we think this change is great, but after integrating it locally we found some crashes in our local testing.  We think this is due to some other part of our code base that isn't doing some right, but we'll have to track that down before we can merge this back in.

c64kernal:
Okay we resolved it, thanks for your patience.  We'll get this merged in again soon.  (The problem ended up being in our internal test code itself).

ppenenko:
Cool, thanks for the update. On my end I've requested to amend the CLA, hope it went through by now.

c64kernal:
Yes, it all went through, we're all set on the CLA, thanks again!

PR #787, opened 2019-03-08
Fix pxrConfig.cmake

meshula:
### Description of Change(s)

Fixes the installation of the pxrConfig.cmake file for both static and shared USD build variants

- Adds a mechanism to export PXR_STATIC when required
- Adds DbgHelp and shlwapi to the Windows static build
- pxrConfig.cmake unconditionally includes cmake scripts that describe the libraries, so it is necessary to ensure associated cmake files are also exported properly as part of the installation process
- fix selection logic around monolithic and non-monolithic build when generating config files

### Fixes Issue(s)

- pxrConfig.cmake should enable find_package(pxr) and remove the need to write ad hoc FindUSD.cmake scripts. pxrConfig.cmake references files that were not actually exported, making the config file non-functional

- pxrConfig.cmake targets were not set up for a static build at all

- Does not address creating a valid pxrConfig.cmake for monolithic builds. usd_m is created by concatenating object files which shouldn't actually be individually exported. A different mechanism, using tricks, is likely necessary to create a pxrConfig.cmake for monolithic builds. According to this issue:  https://gitlab.kitware.com/cmake/cmake/issues/17357 - the exact situation of usd_m is not currently supported.

### DIFF
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -1278,8 +1278,8 @@ function(_pxr_library NAME)
                 DESTINATION ${headerInstallPrefix}
             )
         endif()
-    else()
-        if(BUILD_SHARED_LIBS AND NOT PXR_BUILD_MONOLITHIC)
+    elseif(NOT PXR_BUILD_MONOLITHIC)
+        if(BUILD_SHARED_LIBS)
             install(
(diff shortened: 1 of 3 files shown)


jilliene:
Filed as internal issue #USD-5136

PR #788, opened 2019-03-09
Display unloaded prims

jjeziorski:
### Description of Change(s)
This change makes Hydra delegates consider unloaded prims, which when used with authored `extentHint` data provides solution for visualization of unloaded assets with bounding boxes (unless other explicit opinion was expressed).

This approach introduces visually appealing workflow for lazy scene loading requested by artists.

### Fixes Issue(s)
- Allows to see unloaded prims in Hydra

### DIFF
--- a/pxr/usdImaging/lib/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/lib/usdImaging/delegate.cpp
@@ -135,6 +135,7 @@ UsdImagingDelegate::UsdImagingDelegate(
     , _appWindowPolicy(CameraUtilMatchVertically)
     , _coordSysEnabled(parentIndex
                        ->IsSprimTypeSupported(HdPrimTypeTokens->coordSys))
+    , _displayUnloadedPrimsWithBounds(false)
 {
     // Provide a callback to the _coordSysBindingCache so it can
     // convert USD paths to Hydra ID's.
@@ -188,7 +189,7 @@ UsdImagingDelegate::_IsDrawModeApplied(UsdPrim const& prim)
     // Draw mode is only applied on models that are components, or which have
     // applyDrawMode = true.
     UsdModelAPI model(prim);
-    bool applyDrawMode = false;
+    bool applyDrawMode = !prim.IsLoaded();
     TfToken kind;
     UsdAttribute attr;
     if (model.GetKind(&kind) && KindRegistry::IsA(kind, KindTokens->component))
@@ -207,6 +208,9 @@ TfToken
 UsdImagingDelegate::_GetModelDrawMode(UsdPrim const& prim)
 {
     HD_TRACE_FUNCTION();
+
+    // Draw unloaded prims as bounds.
+    if (!prim.IsLoaded()) { return UsdGeomTokens->bounds; }
 
     // Draw modes can only be applied to models.
     if (!prim.IsModel()) { return UsdGeomTokens->default_; }
@@ -234,7 +238,9 @@ UsdImagingDelegate::_AdapterLookup(UsdPrim const& prim, bool ignoreInstancing)
     //    threads.
 
     TfToken adapterKey;
-    if (!ignoreInstancing && prim.IsInstance()) {
+    if (_displayUnloadedPrimsWithBounds && !prim.IsLoaded()) {
+        adapterKey = UsdImagingAdapterKeyTokens->drawModeAdapterKey;
+    } else if (!ignoreInstancing && prim.IsInstance()) {
         adapterKey = UsdImagingAdapterKeyTokens->instanceAdapterKey;
     } else if (_hasDrawModeAdapter && _enableUsdDrawModes &&
                _IsDrawModeApplied(prim)) {
@@ -648,7 +654,7 @@ UsdImagingDelegate::_Populate(UsdImagingIndexProxy* proxy)
         TF_DEBUG(USDIMAGING_CHANGES).Msg("[Repopulate] Root path: <%s>\n",
                             usdPath.GetText());
 
-        UsdPrimRange range(_GetUsdPrim(usdPath));
+        UsdPrimRange range(_GetUsdPrim(usdPath), _getDisplayPredicate());
         for (auto iter = range.begin(); iter != range.end(); ++iter) {
             if (!iter->GetPath().HasPrefix(_rootPrimPath)) {
                 iter.PruneChildren();
@@ -1116,7 +1122,7 @@ UsdImagingDelegate::_ResyncUsdPrim(SdfPath const& usdPath,
         // If this path was not pruned by a parent, discover all prims that were
         // newly added with this change.
         if (!prunedByParent) {
-            UsdPrimRange range(prim);
+            UsdPrimRange range(prim, _getDisplayPredicate());
 
             for (auto iter = range.begin(); iter != range.end(); ++iter) {
                 if (prunedByParent) {
@@ -1579,6 +1585,22 @@ UsdImagingDelegate::SetWindowPolicy(CameraUtilConformWindowPolicy policy)
                                                     cachePath, &indexProxy);
             }
         }
+    }
+}
+
+void
+UsdImagingDelegate::SetDisplayUnloadedPrimsWithBounds(bool displayUnloaded)
+{
+    if (_hdPrimInfoMap.size() > 0) {
+        TF_CODING_ERROR("SetDisplayUnloadedPrimsWithBounds() was "
+                        "called after population; this is currently "
+                        "unsupported...");
+    } else if (!_hasDrawModeAdapter) {
+        TF_CODING_ERROR("This delegate does not have draw mode "
+                        "adapter; unloaded prims cannot be displayed "
+                        "with bounds...");
+    } else {
+        _displayUnloadedPrimsWithBounds = displayUnloaded;
     }
 }
 
--- a/pxr/usdImaging/lib/usdImaging/delegate.h
+++ b/pxr/usdImaging/lib/usdImaging/delegate.h
@@ -267,6 +267,15 @@ public:
     USDIMAGING_API
     void SetWindowPolicy(CameraUtilConformWindowPolicy policy);
 
+    /// Sets display of unloaded prims as bounding boxes.
+    /// Unloaded prims will need bounds definition in order to see them.
+    /// Effective only for delegates that support draw modes.
+    USDIMAGING_API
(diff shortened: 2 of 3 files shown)


jilliene:
Filed as internal issue #USD-5137

jdiversen:
Hi, if I could vote for this feature, this sounds very helpful.
Regards, Jason

marktucker:
FWIW, I'd also really like to see this PR adopted.

spiffmon:
Hi all,
   Apologies for not responding to this earlier.  The goals here are good ones, but I don't think we can accept this change because it opens the door to serious violations of user-expectations (e.g. posing diffs).  Here's an elaborate example, though there are many simpler ones, also.
   Let's say we have a simple LuxoLamp asset that uses a transform hierarchy to pose the head of the lamp, and it has no light bulb - the asset comes with a built-in payload, like the assets in our Kitchen_set scene.  We reference that model into a set, and now in the set-context, with the lamp loaded, we add a Mesh for the bulb, as a child of the Head prim, and the tool we use to do it adds typeless defs between the model root prim and the new bulb prim.  Everything looks and behaves great!
  But now, when you unload the LuxoLamp, the bulb will now remain and be imaged, **but it will pop to the model's origin**.  We want to preserve the property that if you see something in the viewport, you're seeing it portrayed accurately, to within the limitations of the particular renderer's capabilities and the selected LOD for the object.

   The proposed changes to _GetModelDrawMode() don't seem problematic to me (though I'm not the domain expert), but I think a more extensive reconsideration of the population traversal would be needed to pull this off, and I don't think we have that work scheduled anytime soon.

-- @spiffmon

jjeziorski:
Hi @spiffmon,

Thank you for detailed answer. I see your point.

I've rebased this PR against latest version and modified it so that behavior is (hopefully) more predictable i.e. does not rely on model requirements.

I've also changed it so that it's an opt-in behavior change.

Admittedly, it still suffers from point 3 you've mentioned, but I'm not sure how we could address it. Anyway, it's not a major issue for us.

Please let us know if you find this approach acceptable.

pxr/usdImaging/lib/usdImaging/delegate.cpp
```diff
@@ -1582,6 +1588,22 @@ UsdImagingDelegate::SetWindowPolicy(CameraUtilConformWindowPolicy policy)
     }
 }
 
+void
+UsdImagingDelegate::SetDisplayUnloadedPrimsWithBounds(bool displayUnloaded)
+{
+    if (_hdPrimInfoMap.size() > 0) {
+        TF_CODING_ERROR("SetDisplayUnloadedPrimsWithBounds() was "
+                        "called after population; this is currently "
+                        "unsupported...");
```
spiffmon: Remove ellipses from the end of this error and the next?

pxr/usdImaging/lib/usdImaging/delegate.h
```diff
@@ -267,6 +267,15 @@ class UsdImagingDelegate : public HdSceneDelegate, public TfWeakBase {
     USDIMAGING_API
     void SetWindowPolicy(CameraUtilConformWindowPolicy policy);
 
+    /// Sets display of unloaded prims as bounding boxes.
+    /// Unloaded prims will need bounds definition in order to see them.
```
spiffmon: We should be more explicit here: Bounds for an unloaded prim will only be drawn if either the prim is a UsdGeomBoundable that has an authored _extent_ attribute, or if the prim is a \ref UsdPrim::IsModel() "model" and has an authored _extentsHint_ attribute (see UsdGeomModelAPI::GetExtentsHint()).

pxr/usdImaging/lib/usdImagingGL/drawModeAdapter.cpp
```diff
@@ -1092,11 +1092,23 @@ UsdImagingGLDrawModeAdapter::_ComputeExtent(UsdPrim const& prim) const
...
+        UsdGeomBBoxCache bboxCache(
+            UsdTimeCode::EarliestTime(), purposes, true);
+        return bboxCache.ComputeUntransformedBound(prim).ComputeAlignedBox();
+    } else {
+        GfRange3d extent;
+        if (auto attr = UsdGeomModelAPI(prim).GetExtentsHintAttr()) {
+            VtVec3fArray extentHints;
+            if (attr.Get(&extentHints, UsdTimeCode::EarliestTime()) &&
+                extentHints.size() == 2) {
+                extent = GfRange3d(extentHints[0], extentHints[1]);
```
spiffmon: This is incorrect - or rather, it works only for very simple models.  Please reread the documentation for GetExtentsHint(), as it can be an array of up to 8 elements.  In the clause above you are looking for proxy and render purposes in addition to default, so I would recommend that you merge those three extents together here, if present.  If you wanted to add a helper UsdGeomModelAPI::ComputeExtentsHintForPurposes() that _would_ return a simple extent, I wouldn't object, though I also wouldn't require it.

pxr/usdImaging/lib/usdImagingGL/drawModeAdapter.cpp
```diff
@@ -1092,11 +1092,23 @@ UsdImagingGLDrawModeAdapter::_ComputeExtent(UsdPrim const& prim) const
...
-    UsdGeomBBoxCache bboxCache(UsdTimeCode::EarliestTime(), purposes, true);
-    return bboxCache.ComputeUntransformedBound(prim).ComputeAlignedBox();
+    if (prim.IsLoaded()) {
+        TfTokenVector purposes = { UsdGeomTokens->default_,
+            UsdGeomTokens->proxy, UsdGeomTokens->render };
+        UsdGeomBBoxCache bboxCache(
+            UsdTimeCode::EarliestTime(), purposes, true);
+        return bboxCache.ComputeUntransformedBound(prim).ComputeAlignedBox();
+    } else {
+        GfRange3d extent;
```
spiffmon: As I alluded to in the comment for the header file, you can check here to see if prim.IsA<UsdGeomBoundable>(), and if so, see if it has an authored extent you can use (this is probably an uncommon case - at least it would be for us, but maybe it's worth it?)

pxr/usdImaging/lib/usdImagingGL/drawModeAdapter.cpp
```diff
@@ -1092,11 +1092,23 @@ UsdImagingGLDrawModeAdapter::_ComputeExtent(UsdPrim const& prim) const
...
+        if (auto attr = UsdGeomModelAPI(prim).GetExtentsHintAttr()) {
+            VtVec3fArray extentHints;
+            if (attr.Get(&extentHints, UsdTimeCode::EarliestTime()) &&
```
tcauchois: Do we need a null check for attr here?

pxr/usdImaging/lib/usdImaging/delegate.cpp
```diff
@@ -1582,6 +1588,22 @@ UsdImagingDelegate::SetWindowPolicy(CameraUtilConformWindowPolicy policy)
     }
 }
 
+void
+UsdImagingDelegate::SetDisplayUnloadedPrimsWithBounds(bool displayUnloaded)
```
tcauchois: A question for spiff, do we need to check _enableUsdDrawModes here?

pxr/usdImaging/lib/usdImaging/delegate.cpp
```diff
@@ -188,7 +189,7 @@ UsdImagingDelegate::_IsDrawModeApplied(UsdPrim const& prim)
     // Draw mode is only applied on models that are components, or which have
     // applyDrawMode = true.
     UsdModelAPI model(prim);
-    bool applyDrawMode = false;
+    bool applyDrawMode = !prim.IsLoaded();
```
tcauchois: We only ever call _IsDrawModeApplied from _AdapterLookup, and in the case of this feature, we early-out before we get to that call.  I'd revert this line; it shouldn't affect your feature.

tcauchois:
I left a few specific comments in the code, but it looks pretty good.

We have a few hazard zones in usdImaging, and it would be good to test your change against those:
1.) When you load/unload a prim, does the "Resync" function correctly switch between the bbox and the loaded prim?  Both for load/unload of an ancestor pivot and a gprim.
2.) If you unload instance prototypes, do those show a bbox?  instanceAdapter.cpp and pointInstancerAdapter.cpp both have their own duplicate population traversals for populating prototypes.

pxr/usdImaging/lib/usdImaging/delegate.cpp
```diff
@@ -234,7 +238,9 @@ UsdImagingDelegate::_AdapterLookup(UsdPrim const& prim, bool ignoreInstancing)
     //    threads.
 
     TfToken adapterKey;
-    if (!ignoreInstancing && prim.IsInstance()) {
+    if (_displayUnloadedPrimsWithBounds && !prim.IsLoaded()) {
```
tcauchois: By putting this as the first "if" case, if your unloaded prim is a native instance, you won't get the benefit of instancing.  I think that's much simpler, and may be fine.

I'm curious what would happen if you swap those two cases, though.  By my read, the unloaded prim goes into the instance adapter, gets bound to a hydra instancer, and then the prototype population loop would end up right back in _AdapterLookup, but this time with "ignoreInstancing" set, and so it would populate the unloaded prim as a draw mode object.  If you fix the UsdPrimRange in instanceAdapter.cpp to use the new predicate, it might just work automagically...

pxr/usdImaging/lib/usdImagingGL/drawModeAdapter.cpp
marktucker: I think the`if (auto attr = ...)` code does a null pointer check already?

spiffmon: Yeah, it returns an object not a pointer, but the object's validity _is_ guarded by the check of the auto local.

pxr/usdImaging/lib/usdImaging/delegate.cpp
spiffmon: I _think_ it's fine to leave them decoupled... the doc for this method states it won't be effective if draw modes are not disabled, and by leaving them decoupled, the client can configure them in either order without frustration.

marktucker:
I used the AL changes as a base (adjusted for the new directory structure), then tried to address the comments raised here. Submitted as a separate pull request: https://github.com/PixarAnimationStudios/USD/pull/1145

c64kernal:
Closing this one in favor of PR #1145 where this work is progressing.

PR #791, opened 2019-03-12
[usdSkel] [usdMaya] fix skin import with explicit joint orders

elrond79:
### Description of Change(s
This fixes the maya import of skin bindings that set an explicit joint order

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/789 (partially - HdxComputation is not fixed)


jilliene:
Filed as internal issue #USD-5137

c64kernal:
Thanks for the PR, @elrond79 -- we'll merge it in soon.  We changed a couple of very minor style things, and one call to Remap() that should have been RemapTransforms().

Thanks again!

PR #797, opened 2019-03-14
[usdview] Added a check to _stepSizeChanged to only update if the val…

ianlawson:
…ue changes

### Description of Change(s)

Small fix similar to the changes made in:

https://github.com/PixarAnimationStudios/USD/pull/768


jilliene:
Filed as internal issue #USD-5166

PR #804, opened 2019-03-20
[imaging] make image extension case-insensitive

elrond79:
### Description of Change(s)
so we can use, ie, .PNG, .TIF, etc
oiio already can handle this, and Glf_StbImage::_GetFilenameExtension() already converts to lowercase as well

### Fixes Issue(s)
- Loading of .PNG, .TIF, etc (ie, non-lower-case extensions)

### DIFF
--- a/pxr/imaging/lib/glf/imageRegistry.cpp
+++ b/pxr/imaging/lib/glf/imageRegistry.cpp
@@ -68,7 +68,8 @@ GlfImageRegistry::_ConstructImage(std::string const & filename)
     static GlfImageSharedPtr NULL_IMAGE;
 
     // Lookup the plug-in type name based on the filename.
-    TfToken fileExtension(ArGetResolver().GetExtension(filename));
+    TfToken fileExtension(
+            TfStringToLower(ArGetResolver().GetExtension(filename)));
 
     TfType const & pluginType = _typeMap->Find(fileExtension);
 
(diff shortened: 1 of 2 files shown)


pxr/imaging/lib/glf/imageRegistry.cpp
```diff
@@ -68,7 +68,8 @@ GlfImageRegistry::_ConstructImage(std::string const & filename)
     static GlfImageSharedPtr NULL_IMAGE;
 
     // Lookup the plug-in type name based on the filename.
-    TfToken fileExtension(ArGetResolver().GetExtension(filename));
+    TfToken fileExtension(
+            TfStringToLower(ArGetResolver().GetExtension(filename)));
```
spitzak: It would be better to not change the case when printing the unknown image type error message. You could put to tolower in the call to Find(), or you could change the error message to print the entire filename (which is what the Python version is doing).

jilliene:
Filed as internal issue #USD-5169

c64kernal:
Thanks for the fix @elrond79  -- we merged this and it should land in dev soon after 20.02 is released. In order to address @spitzak's excellent point, I added some extra debug output that gives the exact name of the file before any filtering, so hopefully that should address the review feedback.  Thank you both.

elrond79:
Great, thanks both!

PR #807, opened 2019-03-26
Fix compiler warnings generated on Windows when using the VtValue::Get method.

marktucker:
Fix compiler warnings generated on Windows when using the VtValue::Get
method. Mark the whole Vt_DefaultValueFactory class specializations for
matrix and vector types as "VT_API". Then in the .cpp file, specify VT_API
on the Invoke function definition. This ensures that the specializations
are exported, and external libraries calling these specializations know
that they are being imported.

It was actually generating the correct end result already, but it
complained because it thought that the Invoke method was supposed to be
inlined (but then I guess the linker would find the imported function).

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/USD/issues/806

### DIFF
--- a/pxr/base/lib/vt/value.cpp
+++ b/pxr/base/lib/vt/value.cpp
@@ -568,7 +568,7 @@ VtStreamOut(vector<VtValue> const &val, std::ostream &stream) {
 }    
 
 #define _VT_IMPLEMENT_ZERO_VALUE_FACTORY(r, unused, elem)                \
-template <>                                                              \
+template <> VT_API                                                       \
 Vt_DefaultValueHolder Vt_DefaultValueFactory<VT_TYPE(elem)>::Invoke()    \
 {                                                                        \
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-5177

sunyab:
Just wanted to mention that with this change I got build errors on gcc6 like:

> error: specialization of 'static Vt_DefaultValueHolder Vt_DefaultValueFactory<T>::Invoke() [with T = GfVec4i]' after instantiation

I was able to fix that by just removing the declaration of the invoke specialization in value.h altogether and leaving the change in value.cpp as you have it. That does still fix the warning. However, I also found:

https://developercommunity.visualstudio.com/content/problem/359415/c4506-no-definition-for-inline-function-incorrectl.html

which implies this may be an issue in Visual Studio.

sunyab:
Revisiting some old PRs, I have an alternate fix for this that fixes the compiler warnings on Windows without the build errors on Linux and Mac. I'm going to close this out in favor of that change, which should be landing in the dev branch soon. Thanks!

PR #808, opened 2019-03-27
Better ptex glew find

charlesfleche:
### Description of Change(s)

This PR improves the `FindGLEW` and `FindPTex` cmake modules by allowing searching in custom `lib64` directories for GLEW and PTex include subfolder.

### Fixes Issue(s)
-

### DIFF
--- a/cmake/modules/FindPTex.cmake
+++ b/cmake/modules/FindPTex.cmake
@@ -67,6 +67,10 @@ else ()
             "${PTEX_LOCATION}/include/wdas"
             "$ENV{PTEX_LOCATION}/include"
             "$ENV{PTEX_LOCATION}/include/wdas"
+            "${PTEX_LOCATION}/include/Ptex"
+            "${PTEX_LOCATION}/include/wdas"
+            "$ENV{PTEX_LOCATION}/include/Ptex"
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-5178

cmake/modules/FindPTex.cmake
```diff
@@ -67,6 +67,10 @@ else ()
             "${PTEX_LOCATION}/include/wdas"
             "$ENV{PTEX_LOCATION}/include"
             "$ENV{PTEX_LOCATION}/include/wdas"
+            "${PTEX_LOCATION}/include/Ptex"
+            "${PTEX_LOCATION}/include/wdas"
+            "$ENV{PTEX_LOCATION}/include/Ptex"
+            "$ENV{PTEX_LOCATION}/include/wdas"
```
c64kernal: It seems like  "${PTEX_LOCATION}/include/wdas" and the ENV version are already included above on lines 67 and 69.

charlesfleche: Well spotted, thanks. Fixed in another commit.

c64kernal: That's great, thanks so much, @charlesfleche

PR #824, opened 2019-04-16
Expose velocities attribute to render delegates

marktucker:
This is a greatly simplified replacement for PR #743.

After discussions with the USD team, it was decided that velocities does not need its own dirty bit or dedicated change tracker or value cache functions. Instead, use the pointsDirty bit, and publish it as a regular primvar. Using the pointsDirty bit is reasonable because the velocities attribute, according to USD standards, should be published at the same time samples where the points attribute is published.


jilliene:
Filed as internal issue #USD-5210

Issue #667, opened 2018-10-19
Alembic 1.7.9 support


AlexSchwank:
### Description of Issue
The Alembic API changed a bit from 1.7.7 to 1.7.9:

https://github.com/alembic/alembic/blob/1.7.7/lib/Alembic/AbcCoreOgawa/ReadWrite.h#L72
`ReadArchive( size_t iNumStreams );`

https://github.com/alembic/alembic/blob/1.7.9/lib/Alembic/AbcCoreOgawa/ReadWrite.h#L73
`ReadArchive( size_t iNumStreams, bool iUseMMap );`

So the AlembicReader would need to be updated to support both versions: https://github.com/PixarAnimationStudios/USD/blob/master/pxr/usd/plugin/usdAbc/alembicReader.cpp#L1377

### Steps to Reproduce
1. Try to build USD 18.11 with Alembic 1.7.9

### Package Versions
Tested in 18.09 and 18.11 but the same for older versions too.

### Build Flags
`-DPXR_BUILD_ALEMBIC_PLUGIN=ON`

jtran56:
Filed as internal issue #USD-4850.

elrond79:
I assume now that #825 is merged, this should be closed?

(One caveat - we might want to update the BUILDING.md to note that Alembic-1.7.9 is supported. Or at least make it say 1.5.2+?)

c64kernal:
Thanks @elrond79  -- yeah, good call on closing this one out, and good call on updating the BUILDING.md docs as well.  We were taking a pass through recently and of course a lot of stuff is out of date, so we'll have to take a more concerted look at how we represent these version numbers.  One caveat also is that while we're supporting 1.7.9 after this merge, we only test against 1.5.2 so far.  This is the same problem with the gcc 6+ support that we have too.  We'll need to find the right way to word that properly in the docs.

Thanks again!

PR #825, opened 2019-04-17
Fix compatibility for AbcCoreOgawa::ReadArchive with Alembic>=1.7.9

charlesfleche:
### Description of Change(s)

Alembic 1.7.9 introduced an API change in `Alembic::AbcCoreOgawa::ReadArchive`: a new boolean argument is required. This PR conforms to the new function signature for Alembic >= 1.7.9

The behavior of `Alembic::AbcCoreOgawa::ReadArchive` can be configured by a new environment variable: `USD_ABC_READ_ARCHIVE_USE_MMAP`
By default, `Alembic::AbcCoreOgawa::ReadArchive` keep on using the pre 1.7.9 behavior

### Fixes Issue(s)
- #667

### DIFF
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -98,7 +98,7 @@ TF_DEFINE_ENV_SETTING(
 #if ALEMBIC_LIBRARY_VERSION >= 10709
 TF_DEFINE_ENV_SETTING(
     USD_ABC_READ_ARCHIVE_USE_MMAP, false,
-    "Collapse Xforms containing a single geometry into a single geom Prim in USD");
+    "Use mmap when reading from an Ogawa archive.");
 #endif
 
 namespace {
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -95,6 +95,12 @@ TF_DEFINE_ENV_SETTING(
     USD_ABC_XFORM_PRIM_COLLAPSE, true,
     "Collapse Xforms containing a single geometry into a single geom Prim in USD");
 
+#if ALEMBIC_LIBRARY_VERSION >= 10709
+TF_DEFINE_ENV_SETTING(
+    USD_ABC_READ_ARCHIVE_USE_MMAP, false,
+    "Collapse Xforms containing a single geometry into a single geom Prim in USD");
+#endif
+
 namespace {
 
 using namespace ::Alembic::AbcGeom;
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-5231

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -95,6 +95,12 @@ TF_DEFINE_ENV_SETTING(
     USD_ABC_XFORM_PRIM_COLLAPSE, true,
     "Collapse Xforms containing a single geometry into a single geom Prim in USD");
 
+#if ALEMBIC_LIBRARY_VERSION >= 10709
+TF_DEFINE_ENV_SETTING(
+    USD_ABC_READ_ARCHIVE_USE_MMAP, false,
+    "Collapse Xforms containing a single geometry into a single geom Prim in USD");
```
AlexSchwank: Can you please update the help string?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -95,6 +95,12 @@ TF_DEFINE_ENV_SETTING(
     USD_ABC_XFORM_PRIM_COLLAPSE, true,
     "Collapse Xforms containing a single geometry into a single geom Prim in USD");
 
+#if ALEMBIC_LIBRARY_VERSION >= 10709
```
AlexSchwank: how about updating FindAlembic to generate this automatically?

pxr/usd/plugin/usdAbc/alembicReader.cpp
charlesfleche: Sorry to have let this one slip. Fix is pushed.

pxr/usd/plugin/usdAbc/alembicReader.cpp
charlesfleche: What do you mean ?

pxr/usd/plugin/usdAbc/alembicReader.cpp
AlexSchwank: Thanks :)

pxr/usd/plugin/usdAbc/alembicReader.cpp
AlexSchwank: Right now we would need to define `ALEMBIC_LIBRARY_VERSION` manually but the FindAlembic cmake module could have some logic included to set it automatically when it finds the libraries.  Just an idea though.

charlesfleche: Unless I'm mistaken ALEMBIC_LIBRARY_VERSION is defined in the Alembic headers.

AlexSchwank: Oh, you're right. https://github.com/alembic/alembic/blob/master/lib/Alembic/Util/Config.h.in#L42 perfect :)

Issue #834, opened 2019-04-26
[usdKatana] Broken reading of PointInstancers with prototypes in a separate hierarchy


nrusch:
### Description of Issue

The current `PointInstancer` reading code does not support prototypes that do not live under the instancer prim itself. In particular, one problematic hierarchy pattern we encountered looks like this:

```
/geo
    /point_instancer
    /prototypes
        /Looks
            /shared_proto_material
        /proto_meshes                   # `shared_proto_material` is bound here
            /proto_A                    # prototype target
            /proto_B                    # prototype target
            /proto_C                    # prototype target
``` 

Because the current instancer reading code searches for the common ancestor of the prototype prim and a (possibly inherited) material binding in order to decide where to start building the `instance source` location, the target paths in the resulting instance array are all incorrect.

We had to release a hotfix for this internally for production needs, but I am working on finalizing a pull request.

### System Information (OS, Hardware)

CentOS 7.3.1611

### Package Versions

USD 0.19.5+ (dev @ d63796726481e9dbe3549f54c6865dc0db92b99e)
Katana 3.0v7

jilliene:
Filed as internal issue #USD-5247

nrusch:
For reference/completeness, it seems there may be another tangentially-related issue with the PointInstancer reading code, which I've outlined here: https://github.com/PixarAnimationStudios/USD/pull/835#issuecomment-499274033

nrusch:
I'm going to close this, since I believe the specific issue with the broken `geometry.instanceSource` paths is technically resolved.

There is still an outstanding bug with the prim hierarchy outlined in this issue, but I'm going to open a new issue for that, since the cause is somewhat more general than the prototypes being children of the `PointInstancer`, and I don't want to cause too much confusion by editing this one in place.

PR #835, opened 2019-04-27
[katana] Fix reading of point instancers with external prototypes

nrusch:
### Description of Change(s)

The current logic for converting point instancer prototypes to Katana `instance source` locations was to start from the prototype prim and then walk upward until either a material binding or a (sub)component was found, at which point that location would then become the root of the generated `instance source` hierarchy. This behavior is (subjectively) somewhat unintuitive, but more importantly, it breaks in situations where the prototypes live outside of the point instancer's child hierarchy (as in #834 ).

These changes simplify the logic to the following:

- The prototype prim is always used as the root of the `instance source` location.
- If the prototype prim is not a (sub)component, we walk up the hierarchy looking for a material binding, and if one is found, that material is assigned directly to the `instance source` location.

### Fixes Issue(s)
- #834

### DIFF
--- a/third_party/katana/lib/usdKatana/readPointInstancer.cpp
+++ b/third_party/katana/lib/usdKatana/readPointInstancer.cpp
@@ -373,91 +373,41 @@ PxrUsdKatanaReadPointInstancer(
                 continue;
             }
 
-            // Determine where (what path) to start building the prototype prim
-            // such that its material bindings will be preserved. This could be
-            // the prototype path itself or an ancestor path.
-            //
-            SdfPathVector commonPrefixes;
+            const std::string buildPath = protoPath.GetString();
 
-            // If the proto prim itself doesn't have any bindings or isn't a
-            // (sub)component, we'll walk upwards until we find a prim that
-            // does/is. Stop walking if we reach the instancer or the usdInArgs
-            // root.
+            FnKat::Attribute materialAssignAttr;
+
+            // If the prototype prim is a (sub)component, it should have
+            // materials defined below it.
             //
-            UsdPrim prim = protoPrim;
-            while (prim and prim != instancer.GetPrim() and
-                   prim != data.GetUsdInArgs()->GetRootPrim())
+            TfToken kind;
+            std::string assetName;
+            auto assetAPI = UsdModelAPI(protoPrim);
+
+            const bool hasMaterialChildren = (
+                    assetAPI.GetAssetName(&assetName) and
+                    assetAPI.GetKind(&kind) and (
+                        KindRegistry::IsA(kind, KindTokens->component) or
+                        KindRegistry::IsA(kind, KindTokens->subcomponent)));
+
+            // Otherwise, walk up from the prototype prim looking for a material
+            // binding, and if one is found, assign it directly to the generated
+            // instance source location.
+            //
+            if (!hasMaterialChildren)
             {
-                UsdRelationship materialBindingsRel =
-                        UsdShadeMaterial::GetBindingRel(prim);
-                SdfPathVector materialPaths;
-                bool hasMaterialBindings = (materialBindingsRel and
-                        materialBindingsRel.GetForwardedTargets(
-                            &materialPaths) and !materialPaths.empty());
-
-                TfToken kind;
-                std::string assetName;
-                auto assetAPI = UsdModelAPI(prim);
-                // If the prim is a (sub)component, it should have materials
-                // defined below it.
-                bool hasMaterialChildren = (
-                        assetAPI.GetAssetName(&assetName) and
-                        assetAPI.GetKind(&kind) and (
-                            KindRegistry::IsA(kind, KindTokens->component) or
-                            KindRegistry::IsA(kind, KindTokens->subcomponent)));
-
-                if (hasMaterialChildren)
+                UsdPrim prim = protoPrim;
+                while (prim and prim != instancer.GetPrim() and
+                       prim != data.GetUsdInArgs()->GetRootPrim())
                 {
-                    // The prim has material children, so start building at the
-                    // prim's path.
-                    //
-                    commonPrefixes.push_back(prim.GetPath());
-                    break;
-                }
-
-                if (hasMaterialBindings)
-                {
-                    for (auto materialPath : materialPaths)
-                    {
-                        const SdfPath &commonPrefix =
-                                protoPath.GetCommonPrefix(materialPath);
-                        if (commonPrefix.GetString() == "/" || instancerSdfPath.HasPrefix(commonPrefix))
-                        {
-                            // XXX Unhandled case.
-                            // The prim and its material are not under the same
-                            // parent; start building at the prim's path
-                            // (although it is likely that bindings will be
-                            // broken).
-                            //
-                            commonPrefixes.push_back(prim.GetPath());
-                        }
-                        else
-                        {
-                            // Start building at the common ancestor between the
-                            // prim and its material.
-                            //
-                            commonPrefixes.push_back(commonPrefix);
-                        }
+                    materialAssignAttr = 
+                        PxrUsdKatanaUtils::GetMaterialAssignAttr(prim, data);
+                    if (materialAssignAttr.isValid()) {
+                        break;
                     }
-                    break;
+                    prim = prim.GetParent();
                 }
-
-                prim = prim.GetParent();
             }
-
-            // Fail-safe in case no common prefixes were found.
-            if (commonPrefixes.empty())
-            {
-                commonPrefixes.push_back(protoPath);
-            }
-
-            // XXX Unhandled case.
-            // We'll use the first common ancestor even if there is more than
-            // one (which shouldn't appen if the prototype prim and its bindings
-            // are under the same parent).
-            //
-            SdfPath::RemoveDescendentPaths(&commonPrefixes);
-            const std::string buildPath = commonPrefixes[0].GetString();
 
             // See if the path is a child of the point instancer. If so, we'll
             // match its hierarchy. If not, we'll put it under a 'prototypes'
@@ -483,6 +433,8 @@ PxrUsdKatanaReadPointInstancer(
             //
             sourcesBldr.setAttrAtLocation(relBuildPath,
                     "type", FnKat::StringAttribute("instance source"));
+            sourcesBldr.setAttrAtLocation(relBuildPath,
+                    "materialAssign", materialAssignAttr);
 
             // Author a tracking attr.
             //
--- a/third_party/katana/lib/usdKatana/readPrim.cpp
+++ b/third_party/katana/lib/usdKatana/readPrim.cpp
@@ -62,10 +62,6 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-TF_DEFINE_ENV_SETTING(USD_KATANA_ALLOW_CUSTOM_MATERIAL_SCOPES, false,
-        "Set to true to enable custom names for the parent scope "
-        "of materials. Otherwise only scopes named Looks are allowed.");
-
 TF_DEFINE_ENV_SETTING(USD_KATANA_API_SCHEMAS_AS_GROUP_ATTR, false,
         "If true, API schemas will be imported as group attributes instead "
         "of an array of strings. This provides easier support for CEL "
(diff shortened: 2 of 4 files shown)


jilliene:
Filed as internal issue #USD-5253

nrusch:
@stevelavietes I think this PR is out of date right now due to some of the more recent usdKatana changes (notably 9c21fb7b1e7d940e2cf90f157515a45757f94391), but I wanted to ping this in case it hasn't made it on to your radar yet, as I'm curious what your feelings are about this change conceptually.

spiffmon:
@nrusch - sorry for the delay getting you feedback.  Here's what our PI in Katana expert had to say:

I'm concerned that this change may interfere with prototypes that are defined under the point instancer. Consider the (FX) case where gprims under the same model may be used as prototypes, for example:

prototypes = [
instancer/prototypes/ModelA/Geom/Foo/subdv1,
instancer/prototypes/ModelA/Geom/Foo/subdv2
]

Let's say there's a material binding on Foo that points to instancer/prototypes/ModelA/Looks/FooMaterial. Their change would successfully localize the material binding down to subdv1 and subdv2, but because they've told the reader to start building out its prototypes at the prototypePath, it will start building at ModelA/Geom/Foo/subdv1 and ModelA/Geom/Foo/subdv2, thereby bypassing creation of ModelA/Looks.

The problem that the "start building at the common ancestor" feature tried to solve was not missing material binding attrs, but rather missing Look scopes.

Presumably, in their prototypes-in-a-separate-hierarchy setup, PxrUsdIn builds out the prototypes as it would regular prims, Look scopes and all. This is not how children of point instancers are handled, as PxrUsdIn traversal stops at the UsdGeomPointInstancer level and readPointInstancer is responsible for continuing build-out of any relevant geometry below.

Their setup also seems to imply that both readPrim and readPointInstancer are generating prototype geometry......? ReadPrim would be generating the separate-hierarchy-prototypes and then readPointInstancer would be generating them again, as children of the point instancer. I wonder if their separate-hierarchy-prototypes are tagged with something to prevent this double-build?

Anyway, for our studio setups, we need to ensure that the Look scopes that the prototypes are bound to get built out as material locations in Katana and I think this change breaks that.

nrusch:
Thanks @spiffmon .

I'll update my changes to ensure that the outlined workflow doesn't break. The comments in the original code didn't mention that the branched `Looks` group above a prototype path was a consideration in the current behavior, so I (mistakenly) assumed that it was only really concerned with using (sub)component root locations as prototype paths, rather than arbitrary children of them.

I suspect the end result may require some conditional behavior based on whether each prototype is a child of the instancer, but we'll see how it shakes out.

For completeness:

> Their setup also seems to imply that both readPrim and readPointInstancer are generating prototype geometry......? ReadPrim would be generating the separate-hierarchy-prototypes and then readPointInstancer would be generating them again, as children of the point instancer.

Yes, that's correct. We don't find this to be particularly problematic, since the prototypes are generally set to be invisible regardless.

nrusch:
So in building some test scenes for this, I've found another issue with the current behavior that affects prototypes nested under the instancer as well.

Consider the given example case of two gprims under a nested model being used as prototypes:

```
prototypes = [
    instancer/prototypes/ModelA/Geom/Foo/subdv1,
    instancer/prototypes/ModelA/Geom/Foo/subdv2
]
```

and a material binding on the model pointing to `instancer/prototypes/ModelA/Looks/FooMaterial`.

The prototype building code builds the `instance source` location starting from `instancer/prototypes/ModelA`, but it doesn't modify the paths in the `instance array`'s `geometry.instanceSource` attribute, so they will still be pointing to paths *below* the root of the `instance source` location, which is out of step with Katana's `instance array` attribute conventions.

Now, obviously updating the paths to point to the `instance source` location isn't the solution in this case, since that would make every instance clone the full model hierarchy instead of just a specific target child location. However, it does seem there are wider issues with the PointInstancer reading code than just the lack of support for external prototypes.

I'm not sure how Pixar doesn't run into this problem, unless RfK is more permissive in allowing instance arrays to use non`-instance source` locations as prototypes. @spiffmon, are you able to add any insight here?

emeroo:
Hey there, sorry for the hot potato.

I'll need to refresh my memory on all of our crazy prototype hierarchy situations. I vaguely recall that Katana *does* permit you to specify a geometry.instanceSource entry that is not necessarily of type "instance source". We have an internal test that exercises this case, so I'll take a look at it to see how it's working.

A lot of this trickery is, as mentioned, to support FX debris cases where FX will build out a single model whose children (leaf gprim or interim group) are used as prototypes. So, we had to support any arbitrary child of a model as a potential prototype, while also making sure the point instancer code built out "enough" of the prototype's ancestry for it to be functional. Because we only build out prototypes in one spot (ReadPointInstancer, as opposed to both ReadPointInstancer and ReadPrim), we have to go to extra lengths to determine what is "enough". Whereas, I see in the prototypes-not-as-children case, you're guaranteed to get "enough" via ReadPrim. :-)

I attempted to describe the importance of preserving the Look scope with this comment:
// Determine where (what path) to start building the prototype prim
// such that its material bindings will be preserved.

But I can see how it's unclear which part of the material binding would be lost, i.e. the Look scope itself as opposed to the material assignment path.

nrusch:
Hi @emeroo ,

Thanks for the follow-up.

> I vaguely recall that Katana does permit you to specify a geometry.instanceSource entry that is not necessarily of type "instance source".

I think my interpretation of the [Attribute Conventions documentation](https://learn.foundry.com/katana/3.1/dev-guide/AttributeConventions/Instancing.html#using-instance-arrays) would be that the prototype paths should adhere to the same conventions as regular `instance` locations in terms of sourcing, based on this phrasing:

> Instance sources are defined as above.

> An instance array is a scene graph location declaring multiple instances of one or more instance sources.

So while it doesn't explicitly refer to `instance source`-type locations in the "instance array" section, it does reference the attribute conventions for them.

On our side, KtoA currently follows this pattern, in that it expects all instance array prototypes to be `instance source` locations. While we could modify KtoA to be less stringent in enforcing prototype location types, I'm inclined to try and take the stricter/more formal route, at least as a starting point.

So, between the prototype "type" isssue and our desire to support "out-of-tree" instance prototypes, I think I'm going to try and keep the primary change from this PR (where the prototype path is used directly as the root of each `instance source` location), but add some additional logic to ensure that paths to required materials which are children of the point instancer but not of any prototype path are explicitly built as well.

I'm going to start poking around with this today, but I would love to get any additional thoughts or feedback you might have.

Thanks

nrusch:
So after talking this over some more internally, we're pretty inclined to follow your lead on this in order to try and minimize complications, including pushing changes over the to KtoA side instead if necessary.

With that in mind, I've got a couple of follow-up questions:

- Do you know if there's an authority (either at Pixar or Foundry) on whether using arbitrary location types as instance array prototypes is within the spec? I think I recall hearing from Steve that the instance array attribute conventions are based on Pixar's previously-existing internal representation, but I don't know if this particular aspect was ever codified anywhere...
- Do you know if RfK validates that the prototype locations are children of `instance source` locations? Or does it not care about the type hierarchy at all?

Thanks again

emeroo:
I have indeed confirmed that the canonical debris setup that we have internally is building out its prototypes at the model level and tagging the model as the "instance source", but referring to gprim children of the model as entries in the geometry.instanceSource list. For example:

location with type "instance source"
/root/world/geo/World/fx/Breach/INST_debris/obj_0/Prototypes/prototype_2/AsphaltDebriSmallC

one of the entries in the geometry.instanceSource list:
/root/world/geo/World/fx/Breach/INST_debris/obj_0/Prototypes/prototype_2/AsphaltDebriSmallC/geo/Geom/AsphaltDebriSmallC

This may not be what Katana expects, but works anyway (at least, is not prevented from being handed off to Rfk).

I'm also trying to recall why the "instance source" type wasn't lofted down onto the gprims in this case, perhaps because changing their type from "subdmesh" to "instance source" caused them to no longer render. Or perhaps this was a limitation in how much information we could throw over the fence to prims created by the BuildIntermediate op (the thing that takes over building out prototypes that are children of point instancers).

On first glance, there doesn't seem to be any validation of type when processing an instance array's geometry.instanceSource list. Rfk simply attempts to create an instance link between 2 RixSG nodes (the placeholder node for instance i and the node that corresponds to the scenegraph location that instance i is using). The "instance source" type string *does* affect what Rfk (for RMan22) considers to be an instance source location, however, so there may be unintended consequences for RMan22 that we haven't hit yet... it could be that setups like this have not yet been rendered in 22.

Still digging, thanks for your patience.

emeroo:
Well, it looks like this crazy setup breaks in RMan22 and won't render (whereas in RMan21 it rendered as expected). Looks like you guys unearthed quite the discovery! :sob:

emeroo:
However, pushing the "instance source" type down to the gprim level (/root/world/geo/World/fx/Breach/INST_debris/obj_0/Prototypes/prototype_2/AsphaltDebriSmallC/geo/Geom/AsphaltDebriSmallC) also breaks the render, as the "subdmesh" location type is blasted.

We essentially need a way to specify "this is an instance source" for leaf gprims. It looks like Rfk22 might have a way to do this ("geometry.isInstanceSource" is used to set non-group locations as instance sources). Does KtoA have a similar notion?

nrusch:
> Looks like you guys unearthed quite the discovery!

Whoops! Sorry about that... 🙈 

> We essentially need a way to specify "this is an instance source" for leaf gprims. It looks like Rfk22 might have a way to do this ("geometry.isInstanceSource" is used to set non-group locations as instance sources). Does KtoA have a similar notion?

There isn't currently a way to tag an arbitrary shape location as an instance source, although that sounds like it could be a useful addition. I don't see a ticket for it in the issue tracker, but I can add one.

The closest thing it currently supports is an `instance.ID` attribute, where any locations with the same value will become instances of each other, with the first encountered location being used as the source.

Perhaps an intermediate solution would be to have usdKatana just always generate an `instance source` location above the prototype target path (similar to the approach for prims tagged "instanceable" when generating "master" locations), and modify the paths set in `geometry.instanceSource` accordingly?

emeroo:
Your last solution would work if not for the fact that the geometry.instanceSource list needs to distinguish between sibling subdvs that may both be in use as prototypes, a la instancer/prototypes/ModelA/Geom/Foo/subdv1 and instancer/prototypes/ModelA/Geom/Foo/subdv2. Meaning, if we tagged Foo as the "instance source" and modified the instanceSource list accordingly, we'd get ["instancer/prototypes/ModelA/Geom/Foo", "instancer/prototypes/ModelA/Geom/Foo"] which wouldn't give us the result we wanted. Unless I am misunderstanding?

My gut feeling is not to use instance.ID, as that one is used with implicit instancing, whereas Katana instance arrays and instance sources are examples of explicit instancing. We probably wouldn't want to cross those streams. Also... you can totally have prototypes in a point instancer whose guts are implicitly instanced via instance.ID already, so maybe adding to that notion is asking for trouble.

I actually got our weirdo FX setup (and other non-weirdo ones) to work by setting the prototype's type to "instance source" in the non-gprim case and setting its "geometry.isInstanceSource" attr to 1 in the gprim case. Since it doesn't sound like you guys have gprim-level prototypes, then this would effectively be like setting type="instance source" on your prototypes, meaning that a new KtoA feature probably wouldn't be necessary until you do have gprim-level prototypes...

nrusch:
Sorry, I should have clarified further... I was thinking that when reading into Katana, an additional hierarchy level could be introduced immediately above the prim referenced as a prototype, and that new location would be the `instance source` type. Just a thought based on the way non-array USD instances are handled.

I agree that `instance.ID` is not the solution... I was just mentioning that for completeness (I actually didn't even realize it was part of the core attribute conventions until I went back to the documentation page).

> I actually got our weirdo FX setup (and other non-weirdo ones) to work by setting the prototype's type to "instance source" in the non-gprim case and setting its "geometry.isInstanceSource" attr to 1 in the gprim case.

I'm curious if you have any thoughts on what an ideal long-term solution might look like. In general, I feel like it would be preferable if the behavior used to generate the prototype locations was uniform, rather than varying based on the source prim type.

emeroo:
Ah, I understand now. Introducing an additional hierarchy level is an interesting idea. I like that it introduces consistency, i.e. the same behavior regardless of gprim or group.

For me, because we already introduced the notion of isInstanceSource to RfK to support other cases of gprim-level instance sources (arrays of mesh lights), the pull to use an existing feature is stronger than the pull to maintain consistency with implicit (instance.ID) style instancing. But, I think that's largely in part because I foresee more workflow breakage with the additional scope approach.

For example, adding an additional scope makes the USD path differ from the Katana path, which has historically caused departmental woes due to authoring values (say, material bindings) in terms of the USD hierarchy and having to fix those up as part of import. For example, if your prototype is instancer/prototypes/ModelA and there is a Looks scope below it (instancer/prototypes/ModelA/Looks), then doing the hierarchy adjustment (instancer/prototypes/stub/ModelA) would require fixing up all material bindings using the Looks scope (instancer/prototypes/stub/ModelA/Looks). Granted, there is precedence for doing this sort of fixup, since we insert a bridge location (the "geo" scope) for switching to the BuildIntermediate op (which finishes building out the prototype). This is actually why we authored the "info.usd.sourceUsdPath" (to help assist with reverse engineering a Katana prototype path to its original path in USD). Besides material bindings, there are other workflows in which a non-seamless translation between USD<->Katana has caused disruptions, for example users will often select paths in usdview in an attempt to find corresponding geometry in Katana and vice versa.

Anyway, this is not to shoot down the idea, just to bring up a potential concern. I'll see what the USD folks think.

emeroo:
Hello again!

It sounds like inserting scopes into the original USD hierarchy (thereby creating a Katana path that doesn't match the USD one) messes with production workflows more than I thought. In fact, in some cases, a workaround (reverse engineering the USD path given the Katana path) was never implemented and some Katana-side tools just don't work on locations under point instancer prototypes.

This led me to investigate whether or not we still need the "geo" bridge location that I described above, which (other than PxrUsdIn's starting location) is the reason why prototype paths in Katana don't match up with the USD paths. It used to be the case that the "geo" location didn't have access to the UsdIn private data, so it couldn't represent a scope in the original USD hierarchy. But thanks to a change last year, now it can. I think a lot of users would love to see this extra scope go away.

So, sadly, there is precedence for introducing an extra scope in the original USD hierarchy, but...
a) The existing "geo bridge location for instancer prototypes" example interferes with production workflows, doesn't have workarounds in some cases, can be removed, and internal users would love to see it go away.
b) The existing "build an extra hierarchy scope to support non-array USD instances" example isn't something we encounter because we use instanceMode=expanded (apparently our USD masters aren't stable enough for the "as sources and instances" mode to work). I suspect that if we did use "as source and instances", the same sort of workflows that break down for (a) would become problematic.

So, the good news (hopefully for both of us) is that the point instancer code can be simplified due to no longer needing that "geo" go-between scope (which will likely make your special case material logic, if you're still doing it, easier to inject).

The other good news is that we'll now be properly adhering to the convention that "entries in the geometry.instanceSource list must be instance sources".

The bad news is that injecting an additional hierarchy scope to be able to support gprim prototypes is historically disruptive and we'd prefer to go with a method that permits the hierarchies to match as closely as possible (OR, if not that, then a way to opt in/out). That probably means falling prey to inconsistency by going with the "geometry.isInstanceSource" tag attribute for grpims and "instance source" type for non-gprims.

Were did you guys land in terms of getting support for "out-of-tree" prototypes to work? It sounded like at first you were going to modify the original PR to ensure that any missing Look scopes get created, but then later you said you wanted to "minimize complications"?

nrusch:
Sorry, I haven't had time to fully process your most recent comments... things have been a little frenetic here, but I will try to get back to this tomorrow.

In the meantime...

> Were did you guys land in terms of getting support for "out-of-tree" prototypes to work? It sounded like at first you were going to modify the original PR to ensure that any missing Look scopes get created, but then later you said you wanted to "minimize complications"?

Getting out-of-tree prototypes working was relatively trivial: any bound material locations get built by pxrUsdIn, so we can just find and assign them directly to the `instance source` locations that are generated in the `prototypes` group.

emeroo:
No rush. Thanks for the update!

It's probably better not to have read anything yet, as I've oscillated back and forth on which approach to take. It occurs to me that, while matching USD and Katana hierarchy is ideal and provides many workflow benefits, deviating from the established type="instance source" approach for tagging instance sources may cause unintended consequences for Katana ops that are looking for "instance source" and/or the type the location would have if it wasn't an instance source. So, still waffling over here. Thank you for your patience.

nrusch:
> It occurs to me that, while matching USD and Katana hierarchy is ideal and provides many workflow benefits, deviating from the established type="instance source" approach for tagging instance sources may cause unintended consequences for Katana ops that are looking for "instance source" and/or the type the location would have if it wasn't an instance source.

Yeah, it seems like it would be an awkward bit of inconsistency for anything that's trying to locate the instance sources.

Also, I think it's worth mentioning that the current reading code already diverges from the stage hierarchy in other ways:

- It creates both `usd point instancer` and `instance array` locations for each instancer prim, so the instancer's prim path no longer maps to its Katana location.
- It copies external prototype hierarchy into a `prototypes` group.

In other words, the hierarchy currently only matches the USD stage for the prototypes, and *only* if you nest them under the instancer.

From my perspective (which is admittedly less entangled in the status quo), sacrificing a bit of convenience (by needing to update existing scenes/tools) to get behavior that's both internally consistent and follows Katana's existing conventions seems preferable to introducing an ad hoc convention, and then requiring consumers (i.e. renderer plugin vendors) to support it if they want to be able to use pxrUsdIn out of the box.

emeroo:
I think your perspective is on point. It's good to be less entangled!

We've decided to go with a safer approach and insert a scope into the hierarchy to hold the "instance source" type, thereby preserving the prototype's original type.

We did think about a sneaky side door, though. If the prototype's original type is deemed to be "unimportant" (for example, a USD type that gets turned into "group"), an additional scope won't be inserted, as the original type can merely be set to "instance source". In this way, if point instancer authors choose "unimportant" scopes (say, groups) as their prototypes and put their prototypes under the instancer, they can achieve the golden apple of hierarchy parity. If they don't, they'll still get a result that is both stable and adheres to existing conventions.

nrusch:
That sounds great, and I think your "sneaky side door" is a reasonable one.

Are you or someone at Pixar actively working on these (or planning to)?

emeroo:
Yup, I'm packaging up a change to get what I just described to work.

I haven't tried integrating the original PR into this refactor. Would you like me to give that a try? Or would it be easier for you guys to iterate on this new stuff once you get it?

nrusch:
Nah, don't worry about it. I'll deal with rebasing and testing our changes once your stuff lands.

Thanks!

emeroo:
Hey Nathan! Should I close out this PR given that there will be a different one that builds on top of the recent changes?

nrusch:
I was thinking I would leave it open and then just push a new set of commits once your changes land and I have a chance to build on top of them. That way there's a continuous record of the previous discussion and whatnot.

However, if you'd rather close this one out and have me open a new one down the line for the sake of cleanliness, I'm fine with that.

emeroo:
Let's leave it open, no worries. I like having the discussion history too.

nrusch:
I'm finally getting back to this one, partially because it has reared its head again after the most recent PointInstancer reader updates in 19.11.

The prim hierarchy we're using is essentially the same as the one outlined in https://github.com/PixarAnimationStudios/USD/issues/834, except now the `geometry.instanceSource` paths are correct, but the prototypes don't get built correctly (there are no `instance source` or geometry locations, just empty groups for everything).

I'm in the process of tracking down the underlying cause, and then I may propose a couple of different options for fixes to see what everyone things.

nrusch:
OK, after further investigation, I think I've got a decent handle on the new behavior and what causes it to break. I'm going to open a new issue to discuss some possible solutions, which may in turn lead to some new changes.

Since the conversation on this MR has meandered through a couple of different issues (as well as being somewhat tricky to follow at this point), and since the code changes here aren't particularly relevant anymore, I'm going to close this one out.

Issue #530, opened 2018-06-14
pxrTargets.cmake wanting import libraries when there are none


foundry-markf:
### Description of Issue
I have built USD 0.8.5a with VisualStudio 2015 using the provided build_usd.py script.
I tried to use the cmake config files in the install folder by including them into my CMakeLists.txt file, and resulted in errors looking for DLL import libraries for two DLLs that have no symbol exports (hence the import libs don't exist). The two DLLs are hdStream and usdObj.

### Steps to Reproduce
1. Build and install USD for VisualStudio
2. Create a noddy CMakeLists.txt file containing
```
cmake_minimum_required(VERSION 3.11)
include(pxrConfig)
```
3. Invoke cmake with `cmake -DCMAKE_MODULE_PATH=<USD install dir> .`
4. Observe the errors:
```
CMake Error at D:/thirdparty/usdbuild/cmake/pxrTargets.cmake:381 (message):
  The imported target "hdStream" references the file

     "D:/thirdparty/usdbuild/lib/hdStream.lib"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "D:/thirdparty/usdbuild/cmake/pxrTargets.cmake"

  but not all the files it references.
```
5. You can work around this by removing the reference to hdstream.lib in
`list(APPEND _IMPORT_CHECK_FILES_FOR_hdStream "${_IMPORT_PREFIX}/lib/hdStream.lib" "${_IMPORT_PREFIX}/lib/libhdStream.dll" )`
in `<USD install dir>/cmake/pxrTargets-release.cmake`.
6. Run CMake again, observe these errors
```
CMake Error at D:/thirdparty/usdbuild/cmake/pxrTargets.cmake:381 (message):
  The imported target "usdObj" references the file

     "D:/thirdparty/usdbuild/share/usd/examples/plugin/usdObj.lib"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "D:/thirdparty/usdbuild/cmake/pxrTargets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  D:/thirdparty/usdbuild/pxrConfig.cmake:8 (include)
  CMakeLists.txt:2 (include)
```
[...]

PR #840, opened 2019-05-02
Windows compatibility changes for compiling AL_USD etc.

99alfie:
adding boost chrono build script (required by AL_USD, just so it does not have to be compiled separately and we can use the same boost build as USD)
adding line that forces .lib file generation along with the .dll on Windows for hdStream, usdShaders and usdObj

### Description of Change(s)
force .lib generation on windows for hdStream, usdShaders and usdObj
adding chrono to boost build for convenience

### Fixes Issue(s)
#530

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -575,6 +575,7 @@ def InstallBoost(context, force, buildArgs):
             b2_settings.append("--with-date_time")
             b2_settings.append("--with-system")
             b2_settings.append("--with-thread")
+            b2_settings.append("--with-chrono")
 
         if context.buildOIIO:
             b2_settings.append("--with-filesystem")
--- a/extras/usd/examples/usdObj/CMakeLists.txt
+++ b/extras/usd/examples/usdObj/CMakeLists.txt
@@ -1,6 +1,10 @@
 set(PXR_PREFIX pxr/usd)
 set(PXR_PACKAGE usdObj)
 
+IF(WIN32)
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+ENDIF()
+
 pxr_plugin(${PXR_PACKAGE}
     LIBRARIES
         tf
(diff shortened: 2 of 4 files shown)


meshula:
I think the underlying issues discussed by Sunya here ( https://github.com/PixarAnimationStudios/USD/issues/530#issuecomment-432466952 ) should be addressed directly. If AL_USD needs to actually link those libraries, then wouldn't it make sense to explicitly export the needed symbols similarly to the rest of the USD library?

The reason I'm asking is that introducing DEF files to the system becomes another complicating consideration for those of us building on Windows via automation or other build contraptions - it's another artifact that needs to be accounted for and copied around and possibly noted as input to various build processes.

99alfie:
the problem is that pxrConfig.cmake lists those three libs as part of PXR_LIBRARIES, and by default install, expects those to be present. On Windows, all except those (hdStream etc.) get compiled as .dll with their corresponding .lib files (they only come as .dll) so linking against PXR_LIBRARIES will error. True if they are not supposed to be linked against, they should not appear on that list in the first place. However since that (pxrConfig.cmake) is generated on install, it makes little sense to edit it.

c64kernal:
Thanks @99alfie -- I think what @meshula and @sunyab are saying is that rather than for libs to be created, we should instead fix the build system so that it does not issue the plugins in the pxrConfig.cmake.

Similarly for the chrono addition, my instinct is that we shouldn't hardcode it to the build_usd.py (who knows what other boost dependencies might be needed by other systems).  I'm not sure what the right answer is here, but some possibilities might be to parameterize build_usd.py so that you can pass in "additional boost dependencies" on the command line?  (I'm not suggesting you do that at all, because I'm not sure I even like it).  Part of the issue is the intent for build_usd.py is to quickly bootstrap someone into getting a USD build, it's not really meant to be a build system for an ecosystem depending on USD.  If you have a pipeline that manages lots of dependencies, chances are you want to include USD into the larger build process and customize the build yourself running cmake directly and not using build_usd.py.

99alfie:
ok, boost I agree. In any dev environment that is a bit larger, there would be boost already, so that is not a big deal and not the responsibility of USD.
As with the build script, maybe someone can do this (simple) fix (not issuing the plugins) you suggest. It is simple once you know it but everyone building on Windows runs into this and it takes (at least it took me) quite a while working around it. Closing the PR.

PR #844, opened 2019-05-08
[usdMaya] make UsdMayaTranslatorUtil::CreateShaderNode use shadingNode

elrond79:
### Description of Change(s)
Turns UsdMayaTranslatorUtil::CreateShaderNode into a wrapper over the shadingNode command

The docs said it was intended to mimic this command, but it was lacking some
functionality; in theory, we might have been able to add the (known) missing
functionality, but past experience has shown attempting to replicate this
command not worth the small efficiency gain by avoiding the string conversions
needed by mel commands.

Also eliminated ConnectDefaultLightNode, since it was also mimicking this command

### Fixes Issue(s)
- shadingNodes not showing up in the hypershade
- texture nodes not being hooked into the color management system

### DIFF
--- a/third_party/maya/lib/usdMaya/shadingModePxrRis.cpp
+++ b/third_party/maya/lib/usdMaya/shadingModePxrRis.cpp
@@ -398,14 +398,14 @@ static
 MObject
 _CreateAndPopulateShaderObject(
         const UsdShadeShader& shaderSchema,
-        const bool asShader,
+        const UsdMayaShadingNodeType shadingNodeType,
         UsdMayaShadingModeImportContext* context);
 
 static
 MObject
 _GetOrCreateShaderObject(
         const UsdShadeShader& shaderSchema,
-        const bool asShader,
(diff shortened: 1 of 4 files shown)


elrond79:
Note - this affects some of the Rfm code, which I haven't been able to test...

jtran56:
Filed as internal issue #USD-5282.

c64kernal:
Thanks so much for the PR @elrond79  -- it looks great, we're about ready to merge it in.  One problem we ran into is the token "None" in UsdMayaShadingNodeType -- because of some shenanigans with X11/X.h we have to rename it.  Do you have a preference for an alternative token?

elrond79:
No real preference.  NonShadingNode? GenericNode?

c64kernal:
Thanks @elrond79 -- let's go with NonShadingNode, we'll make the update.  Thanks again!

c64kernal:
Actually since the rest of them don't have a "Node" suffix, how about just NonShading?

elrond79:
Even better!

PR #846, opened 2019-05-10
Allow specifying the material tag for translucent preview surface via an environment variable

sirpalee:
### Description of Change(s)
The PR allows configuring the materialTag for translucent preview surfaces (opacity less than 1 or mapped opacity), which is currently hardcoded to additive. This change lets users to set the material tag to translucent and enable a more expensive, but more accurate way of blending via the new OIT task.

### Fixes Issue(s)
None reported.

### DIFF
--- a/pxr/usdImaging/lib/usdImagingGL/hydraMaterialAdapter.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/hydraMaterialAdapter.cpp
@@ -25,6 +25,8 @@
 #include "pxr/usdImaging/usdImagingGL/package.h"
 #include "pxr/usdImaging/usdImagingGL/textureUtils.h"
 
+#include "pxr/base/tf/envSetting.h"
+
 #include "pxr/usdImaging/usdImaging/debugCodes.h"
 #include "pxr/usdImaging/usdImaging/delegate.h"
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-5283.

elrond79:
After discussion, we decided that it makes more sense to have the default for UsdPreviewSurface be translucent, if opacity < 1.  This makes sense because OIT is very new, so maintaining old behavior is less important, and it's the correct behavior that most users will want / expect.

sirpalee:
@poljere what's your opinion?

poljere:
Hi @sirpalee! @lumonix is refactoring our hydraMaterialAdapter (and material tag) to become a Storm internal behavior instead of a UsdImagingGL, you might have already seen some of his changes. 

In this new world, your idea would fit maybe better as a Storm render setting, of course, there are some "interesting" questions about invalidation to be answered. I think for now, it would be best to wait until the material cleanup is more advanced and then investigate if our idea would work to solve your problem.

c64kernal:
Hey @sirpalee, thanks to the nice work @clach did, we should no longer need this (i.e., we're not longer hard coded to additive). Closing this out, but if I missed something, please reopen! Thanks!

PR #848, opened 2019-05-10
[tests] Checkout sdf and usda files with LF for cross-platform unit testing.

asluk:
### Description of Change(s)

Ensures that sdf and usda test inputs and baselines use LF on all platforms; otherwise testSdfParsing* and testUsdFlatten* fail on Windows.

### Fixes Issue(s)
-

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+*.sdf		text eol=lf
+*.usda		text eol=lf
+


spiffmon:
Hey @asluk  - from the changelog, it looks like you may need to rebase your change?

spiffmon:
Looks like you already did it - thanks!

meshula:
Is the LF (no CRLF) requirement documented (I can't find it)? If not, it would be nice to make a note in the FAQ, in addition to this fix. (https://graphics.pixar.com/usd/docs/USD-Frequently-Asked-Questions.html)

jtran56:
Filed as internal issue #USD-5285.

asluk:
I don't think the LF requirement is documented anywhere, but the source of it is in the explicit "\n" writes here https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/lib/sdf/textFileFormat.cpp#L192

spitzak:
Definitely in favor of text files on all platforms being written with
single ^J characters as the newlines. Readers should ignore ^M characters
in the input so that text editors that don't obey this can still be used.

Writing different files on different platforms is causing unnecessary
differences and makes cross-platform operation more difficult.

meshula:
+1 on modifying .gitattributes in this way to ensure LF when git cloning.

asluk:
Hi all, this would be a nice one to get into 20.05, as it gets all tests passing on Windows. Thanks!

spitzak:
^M should be considered a whitespace character in usda.

Code on one platform should not add whitespace to the end of every line, as it makes unnecessary differences between files written on different platforms. There is very little software left on Windows that cannot deal with just ^J as a newline.

PR #853, opened 2019-05-28
[Hd, UsdImaging, UsdImagingGL] Render delegates explicitly specify material adapter prim type tokens

ppenenko:
We need UsdImagingDelegate to be able to create a custom material
adapter that would create shader resources for our new Maya plug-in
instead of HdSt ones. This was previously impossible because
UsdImagingDelegate selected one of the two hardcoded material adapter
prim type tokens based on the material binding purpose (full or
preview) returned by the render delegate.

We're instead suggesting a new virtual method that returns the prim
type token explicitly, thus allowing the render delegate to specify a
custom material adapter, transparently to UsdImagingDelegate.

### DIFF
--- a/pxr/imaging/lib/hd/renderDelegate.cpp
+++ b/pxr/imaging/lib/hd/renderDelegate.cpp
@@ -83,6 +83,13 @@ HdRenderDelegate::GetMaterialBindingPurpose() const
     return HdTokens->preview;
 }
 
+TfToken
+HdRenderDelegate::GetMaterialAdapterKey() const
+{
+    return GetMaterialBindingPurpose() == HdTokens->preview
+        ? HdTokens->HydraPbsSurface : HdTokens->Material;
+}
+
 TfTokenVector 
 HdRenderDelegate::GetShaderSourceTypes() const
 {
--- a/pxr/imaging/lib/hd/renderDelegate.h
+++ b/pxr/imaging/lib/hd/renderDelegate.h
@@ -316,6 +316,12 @@ public:
     virtual TfToken GetMaterialBindingPurpose() const;
 
     ///
+    /// Returns a primTypeName token used to find a matching material adapter.
+    ///
+    HD_API
+    virtual TfToken GetMaterialAdapterKey() const;
+
+    ///
     /// Returns a token that can be used to select among multiple
     /// material network implementations.  The default is empty.
     ///
--- a/pxr/imaging/lib/hd/tokens.h
+++ b/pxr/imaging/lib/hd/tokens.h
@@ -68,6 +68,7 @@ PXR_NAMESPACE_OPEN_SCOPE
     (hermite)                                   \
     (hidden)                                    \
     (hullIndices)                               \
+    (HydraPbsSurface)                           \
     (indices)                                   \
     (instancer)                                 \
     (instancerTransform)                        \
@@ -81,6 +82,7 @@ PXR_NAMESPACE_OPEN_SCOPE
     (leftHanded)                                \
     (linear)                                    \
     (lightLink)                                 \
+    (Material)                                  \
     (materialParams)                            \
     (nonperiodic)                               \
     (normals)                                   \
--- a/pxr/usdImaging/lib/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/lib/usdImaging/delegate.cpp
@@ -82,8 +82,6 @@ TF_DEFINE_PRIVATE_TOKENS(
     _tokens,
     (instance)
     (texturePath)
-    (Material)
-    (HydraPbsSurface)
 );
 
 // This environment variable matches a set of similar ones in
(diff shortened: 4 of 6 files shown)


spiffmon:
Hi @ppenenko,
    We do not intend for Material Purpose to be extensible in this way, as we think it would overly complicate interchange.  Instead (and apologies because the documentation could really be improved, here), the supported way to specialize Materials for specific renderers is not through material binding, but rather "network selection".  You simply need to implement the existing virtual HdRenderDelegate::GetMaterialNetworkSelector(), which will cause usdImaging to query each materials for specifically namespaced output terminals *before* looking for the generic ones.  

   For example, if your GetMaterialNetworkSelector() returns "myMaya", then for each Material that usdImaging finds, it will first look for a "outputs:myMaya:surface" before it looks for "outputs:surface" to find the proper surface network to use.

   Hopefully this gives you what you need - please let me know if not!  In the upcoming 19.07 release, we will be including our hdPrman render delegate, which makes use of this mechanism (its selector is "ri").

--spiff

ppenenko:
Hi @spiffmon ,

I'm afraid I haven't made it clear what we're trying to achieve, so let me provide some context. Here at Autodesk, we're working on a new proxy shape implementation for Maya based on MPxSubSceneOverride, which means rendering the real-time preview with Maya's native VP2 instead of the existing MPxDrawOverride proxy shape implementation which renders via HdSt and composites with VP2 on the OpenGL level.

In the end, our goal is to match HdSt visually as closely as possible and that includes materials and textures. We're not looking to support a new VP2-specific surface type which GetMaterialNetworkSelector() seems to be designed for, but rather to support the same surface type that HdSt supports, i.e. UsdPreviewSurface. To get the same representation of UsdPreviewSurface with HdMaterialParam's as HdSt does without unnecessary code duplication, we're relying on UsdImagingGL.

Currently, UsdImaging creates a material adapter based on the material binding purpose. There are effectively two hardcoded options: UsdImagingMaterialAdapter if the purpose is "Material" and UsdImagingGLHydraMaterialAdapter if the purpose is "preview". We're not suggesting to extend the set of possible material purposes (our VP2 renderer is "preview" for all intents and purposes) but unfortunately we're stuck with UsdImagingGLHydraMaterialAdapter. It works well in many situations, e.g. we're able to create instances of our implementation of HdMaterial because its creation is delegated to our render delegate. But there doesn't seem to be a way to create instances of our implementation of HdTextureResource - we need a custom material adapter for that.

So, we're suggesting this new mechanism where the render delegate is able to specify explicitly which material adapter it needs but which defaults to the old behavior where that's determined by the material binding purpose reported by the render delegate, for backward compatibility. If there's a better way to instantiate a custom material adapter, we're open to suggestions.

Hope this makes sense.

jilliene:
Filed as internal issue #USD-5311

spiffmon:
Thanks for explaining, @ppenenko  - it's clear now what you're trying to do.  The mechanisms you're trying to modify, though, are on the chopping block for (we expect) the very near future.  That is, we plan to eliminate the whole notion of "switchable material adapters" in the Usd scene delegate - there should be only one, which largely models the existing full/network material adapter, though with fully implemented support for material-based primvar filtering.

So we'd like to propose a different approach, which, though it will have some shortcomings in the near-term, will put your renderer in a perfect position for the upcoming changes.  Essentially, you should configure your render delegate so that it (for now) asks for "full" material purpose, rather than "preview".  As you've noted, that will cause the other material adapter to be selected, which, like the future single adapter, does minimal processing of the material network and resources... basically just translating the UsdShade networks into HdMaterial networks, which means your render delegate will then be in charge of all texture resources and processing.  
There's an example of how this works in Luma's  [usd-arnold repo](https://github.com/LumaPictures/usd-arnold/tree/master/plugin/pxr/imaging/hdAi), and we will soon be releasing hdPrman, which will have a more involved example of HdMaterial processing.

The two current limitations of this approach that we can think of, which are both temporary, until we can do the previously mentioned work, are:
1. If someone has actually authored material:binding:preview bindings, you'll miss them, because you're not looking for them.  AFAWK, there aren't yet any authoring tools for authoring preview bindings, so hopefully we can skate by on this until our reworking happens and you can then switch back to querying "preview" bindings.
2. The UsdImagingGLHydraMaterialAdapter actually does a really aggressive job of filtering primvars based on the Material's usage.  The UsdImagingMaterialAdapter does not yet do any filtering, so your renderer will potentially see lots more primvars, but it can do filtering itself, if it needs to for performance.

You'll likely be writing more code going this route, but we think it's code you'd need to eventually need to write anyway.  Does this make sense?
 
--spiff

ppenenko:
Thanks for the detailed explanation @spiffmon. The two concerns you've mentioned pretty much sum up why we have gone the "preview" route. We did assume that our materials had to have the "preview" purpose for consistency with HdSt, so it's great to know that the consequences of choosing the "full" purpose would currently be limited. We have also tried to share as much code as possible with HdSt, namely for walking UsdShade networks and filtering primvars, so UsdImagingGLHydraMaterialAdapter came in handy. It should be possible to implement this functionality separately, on the Maya plug-in side but that would require some code duplication, which could become a maintenance burden in our quest to match HdSt. On the other hand, this should be a minor concern if this situation is temporary until Hydra switches to the unified material adapter.

So it looks like we we have to go back to the drawing board.

Danod:
Hi @spiffmon ,

Thanks for the detailed explanation above. I am looking into the direction you suggested and have a follow-up question about primvars. It is not clear to me which side should be in charge of the primvar extraction. From the conversation here it looks to me that material adapter will be responsible for it, but in the code I don't see it happen. The check below will fail for UsdPrimvarReader. Could you please elaborate?

https://git.autodesk.com/autodesk-forks/USD/blob/master/pxr/usdImaging/lib/usdImaging/materialAdapter.cpp#L197

Thanks in advance.

spiffmon:
Hi @Danod , 
   Woof - that code was thrown together pretty quickly for an internal milestone, and it absolutely needs to be revisited.  _ExtractPrimvarsFromNode() should be consulting the node's SdrShaderNode to find the primvars it needs.

PR #855, opened 2019-05-29
[usdImagingGL] fix UsdImagingGLEngine constructor w/ args

elrond79:
### Description of Change(s)
Fixes the overload of the UsdImagingGLEngine constructor that took arguments

It did not call _InitGL, like the no-arg constructor. This resulted in an
error: `GlfContextCaps has not been initialized`

### Fixes Issue(s)
- error when calling UsdImagingGLEngine constructor w/ args


jilliene:
Filed as internal issue #USD-5312

PR #856, opened 2019-05-29
[usdImagingGL] allow usdview to use a delegateID for it's UsdImagingDelegate

pmolodo:
### Description of Change(s)

This option has the effect of making the index rprim paths clearly different
from the usd paths, ie:

```
index path: /hydra/cube1
usd path:   /cube1
```

This has two benefits:

One, it makes it easier when debugging code, and using usdview as a tool
for exploring / learning hydra, as it is clearer what type of path you are
looking at.

Two, it enforces best practices on the usdview codebase.  This codebase
will be used as a reference point by many, and by making the two types of
paths different, it forces usdview to handle path conversions "properly",
instead of relying on them being the same.  This change has already
exposed one case where an indexPath was being returned without conversion
as a usdPath

### Note on the PR

This PR depends on / includes this other PR:

https://github.com/PixarAnimationStudios/USD/pull/855

### DIFF
--- a/pxr/usdImaging/lib/usdImagingGL/engine.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/engine.cpp
@@ -170,6 +170,10 @@ UsdImagingGLEngine::UsdImagingGLEngine(
     , _restoreViewport(0)
     , _useFloatPointDrawTarget(false)
 {
+    static std::once_flag initFlag;
+
+    std::call_once(initFlag, _InitGL);
+
     if (IsHydraEnabled()) {
 
         // _renderIndex, _taskController, and _delegate are initialized
--- a/pxr/usdImaging/lib/usdImagingGL/engine.cpp
+++ b/pxr/usdImaging/lib/usdImagingGL/engine.cpp
@@ -674,6 +674,24 @@ UsdImagingGLEngine::GetPrimPathFromInstanceIndex(
                                              instanceContext);
 }
 
+SdfPath
+UsdImagingGLEngine::ConvertCachePathToIndexPath(SdfPath const& cachePath)
+{
+    if (ARCH_LIKELY(_delegate)) {
+        return _delegate->ConvertCachePathToIndexPath(cachePath);
+    }
+    return cachePath;
+}
+
+SdfPath
+UsdImagingGLEngine::ConvertIndexPathToCachePath(SdfPath const& indexPath)
+{
+    if (ARCH_LIKELY(_delegate)) {
+        return _delegate->ConvertIndexPathToCachePath(indexPath);
+    }
+    return indexPath;
+}
+
 //----------------------------------------------------------------------------
 // Renderer Plugin Management
 //----------------------------------------------------------------------------
--- a/pxr/usdImaging/lib/usdviewq/stageView.py
+++ b/pxr/usdImaging/lib/usdviewq/stageView.py
@@ -804,7 +804,8 @@ class StageView(QtOpenGL.QGLWidget):
     def rendererAovName(self):
         return self._rendererAovName
 
-    def __init__(self, parent=None, dataModel=None, printTiming=False):
+    def __init__(self, parent=None, dataModel=None, printTiming=False,
+                 delegateID=None):
 
         glFormat = QtOpenGL.QGLFormat()
         msaa = os.getenv("USDVIEW_ENABLE_MSAA", "1")
@@ -817,6 +818,7 @@ class StageView(QtOpenGL.QGLWidget):
 
         self._dataModel = dataModel or StageView.DefaultDataModel()
         self._printTiming = printTiming
+        self._delegateID = delegateID
 
         self._isFirstImage = True
 
@@ -908,7 +910,11 @@ class StageView(QtOpenGL.QGLWidget):
         # create the renderer lazily, when we try to do real work with it.
         if not self._renderer:
             if self.isValid():
-                self._renderer = UsdImagingGL.Engine()
+                if self._delegateID:
+                    self._renderer = \
+                        UsdImagingGL.Engine('/', delegateID=self._delegateID)
+                else:
+                    self._renderer = UsdImagingGL.Engine()
                 self._handleRendererChanged(self.GetCurrentRendererId())
             elif not self._reportedContextError:
                 self._reportedContextError = True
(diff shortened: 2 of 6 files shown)


jilliene:
Filed as internal issue #USD-5313

pxr/usdImaging/lib/usdImagingGL/engine.cpp
```diff
@@ -637,6 +641,30 @@ UsdImagingGLEngine::GetPrimPathFromInstanceIndex(
...
+UsdImagingGLEngine::ConvertCachePathToIndexPath(SdfPath const& cachePath)
+{
+    if (ARCH_UNLIKELY(_legacyImpl)) {
+        return SdfPath();
+    }
+
+    TF_VERIFY(_delegate);
+
+    return _delegate->ConvertCachePathToIndexPath(cachePath);
+}
```
spiffmon: Instead of two different checks, can we just do a 
```
if (ARCH_LIKELY(_delegate)){
   return _delegate->ConvertCachePathToIndexPath(cachePath);
}

return SdfPath();
```

and similarly below?

pxr/usdImaging/lib/usdviewq/stageView.py
```diff
@@ -908,7 +908,7 @@ def _getRenderer(self):
         # create the renderer lazily, when we try to do real work with it.
         if not self._renderer:
             if self.isValid():
-                self._renderer = UsdImagingGL.Engine()
+                self._renderer = UsdImagingGL.Engine('/', delegateID='/hydra')
                 self._handleRendererChanged(self.GetCurrentRendererId())
```
spiffmon: I love the idea of making this feature available for debugging... but please understand there's a non-trivial performance and memory cost to it, because you are forcing the creation of an entire new hierarchy of SdfPaths for the scene, which is a significant and locky process.  So:
1. I wouldn't want people copying this pattern for every rendering app they write
2. Can we put this behind a command-line debug option?

pxr/usdImaging/lib/usdviewq/stageView.py
```diff
@@ -2096,38 +2096,40 @@ def pickObject(self, x, y, button, modifiers):
...
-                    selectedPrimPath = instancePrimPath
-                    selectedInstanceIndex = absInstanceIndex
+                        selectedIndexPath, selectedInstanceIndex)
+                if instanceIndexPath:
+                    selectedIndexPath = instanceIndexPath
+                    selectedInstanceIndex = instancerIndex
             else:
                 selectedInstanceIndex = ALL_INSTANCES
+                
+            selectedUsdPath = renderer.ConvertIndexPathToCachePath(selectedIndexPath)
```
spiffmon: Thanks much for catching this!

pxr/usdImaging/lib/usdviewq/stageView.py
elrond79: Ouch - I didn't realize there would be a significant performance hit.  I guess I had been hoping that since hydra mostly works on a flattened scene graph, introducing an extra top-level item wouldn't have much of an impact.  I didn't realize that SdfPaths are kept in a global table, like TfTokens... well, good to know, anyway!

Anyway - yes, making it a command line option makes sense.

pxr/usdImaging/lib/usdImagingGL/engine.cpp
elrond79: Sure - if it's kosher to assume that having _delegate be non-zero implies _legacyImpl is zero, then combining into one check makes sense.  We do lose the "warning" / feedback we get from TF_VERIFY when both are zero, however.

It's worth noting, though, that I was following the template of GetPrimPathFromInstanceIndex:

https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/lib/usdImagingGL/engine.cpp#L629

...and this pattern is used by many other methods, as well  - ie, SetRootTransform, SetRootVisibility, etc.

Do you want me to go through and clean up all these as well? Perhaps in a different PR? If so, I assume we can make similar assumptions about the relationship between _taskController and _legacyImpl?

elrond79:
Made this into a flag, and made the conversion methods only check on _delegate.

Made one other change - in the case where _delegate is not defined, they now just return the original path passed in.  This is more likely to be the desired behavior for _legacyImpl.

pxr/usdImaging/lib/usdImagingGL/engine.cpp
```diff
@@ -171,6 +171,10 @@ UsdImagingGLEngine::UsdImagingGLEngine(
     , _restoreViewport(0)
     , _useFloatPointDrawTarget(false)
 {
+    static std::once_flag initFlag;
+
+    std::call_once(initFlag, _InitGL);
+
```
poljere: Good catch! Thanks.

pxr/usdImaging/lib/usdviewq/stageView.py
poljere: Indeed, good catch.

elrond79:
Rebased to fix a merge conflict...

c64kernal:
Hey @elrond79  -- apologies for the back and forth on this.  We've been debating it a bit internally. We think this is a very cool feature and it will actually help keep us honest in tests.  We are struggling with the level of exposure it would be getting if it were presented as a command line argument to usdview.  We're leaning towards the idea of this is more appropriately expressed as a TfEnvSetting -- this would be an env var that you would set for debugging purposes before you started usdview.  What do you think of that?

elrond79:
First, no need to apologize for having a discussion!

As for making it an env var: the disadvantage of hiding it is... less
people know about it. =P

I'm putting myself in the shoes of another 3rd party USD developer who's
trying to write something related to hydra.  (Or maybe just trying to learn
hydra.)  Personally, I found it very helpful to know, simply by looking at
a path, whether it was a "USD" path or an index path.  I think other hydra
developers would as well, so I'd like them to KNOW it's there, so they can
make use of it.

We've hidden it behind a "--debug" flag, so it won't show up to confuse the
"average" user.  And it's in the help string that it "may have a
performance impact" (though I could word that more strongly, if desired).
Is there some other concern / danger related to its use that I'm missing?

- Paul

elrond79:
...or, if the idea is more that having an env var is simply more convenient
(I can see that - ie, you could set it in your .bashrc when you're doing a
lot of hydra dev, or set it before running tests), there's no reason we
can't have both (ie, have the default controlled by an env var).

c64kernal:
Yeah, that's exactly right. I may also be underestimating the number of folks developing Hydra code who would actually benefit from having this being easily discoverable, compared to the number of folks who use usdview.  Granted hiding things behind --debug does help.  We were thinking that this would have nothing to do with usdview, and we could use this env var to help in cases outside of usdview as well.  In our internal test infrastructure, we would set this env var potentially high up our test hierarchy so that all of our tests would get some indirection to weed out path-translation bugs.

elrond79:
> I may also be underestimating the number of folks developing Hydra code who would actually benefit from having this being easily discoverable

Heh, and I'm probably overestimating the number of people with the same needs as me. ("I am legion!").  Whichever direction you decide to go with this is fine with me...

pmolodo:
So noticed that this feature went in via this commit:
e3ec6a7ef8e3bed5db083db2e9a28a843c4f47e2
...some time ago, so closing this PR.  Thanks!

PR #857, opened 2019-05-29
[imaging] reword docstring + param names for GetPathForInstanceIndex

elrond79:
### Description of Change(s)

This was done in the attempt to clarify the purpose / inputs / outputs of
a fairly complex function. The changes introduced in
2d846f6578ec130d87e941ddd04242a594af5d97 helped clarify things, but did
not address this family of functions. Some notes about the changes:

- Index in this function could have 2 different meanings: it could
  refer both to a type of path (ie, id / render-index sdf paths, as
  opposed to cache paths or usd paths), and to instance-indices
  (integer array indices).  Because of this, I'm using "rprimId" instead
  of "index" to denote the type of paths - ie, "protoRprimId"
  instead of "protoIndexPath" (originally: "protoPrimPath").

- I'm replacing "instanceIndex" and "absoluteInstanceIndex" with
  "prototypeIndex" and "instancerIndex", since I feel these names are
  are more descriptive.  (Ie, "instancerIndex", holds the index used in
  the top-level instancer... it does NOT indicate a position within an array
  of all possible leaf-instances created by a given instancer.  In fact, if
  there is a nested hierarchy of instancers, the top-level instancer may have
  fewer possible values than the indices for the prototype.)

- I changed "rprimPath" to "masterCachePath", since rprimPath is fairly
  generic.

- In the docstring, I clarified behavior when there are multiple levels of
  instancing, since this is something I found confusing / unclear.


elrond79:
So, I think I now have a much better understanding of the relationship between usd, cache, and index paths.  However, there's still a number of items that I'm unclear on:

- To start with, some confirmation would be appreciated concerning the translation between cachePath + usdPath.  The docs say that the "cache path is the same as a USD prim path, except for instanced prims, which get a name-mangled encoding." First, does that mean that if we know a path is not involved in instancing in any way, we can always assume usdPaths and cachePaths are identical?  Secondly - if it is instanced, is UsdImagingPrimAdapter::GetPathForInstanceIndex essentially the function that does this conversion?

  If both of these are "yes", then the missing "ConvertCachePathToUsdPath" would look something like:

  ```
  SdfPath ConvertCachePathToUsdPath(cachePath, instanceIndex)
  {
      instancerPath = GetPathForInstanceIndex(
          ConvertCachePathToIndexPath(cachePath), instanceIndex);
      if (!instancerPath) {
          return cachePath;
      }
      return ConvertIndexPathToCachePath(instancerPath);
  }
  ```
  
  If so, then we might want to create such a method... and possibly modify GetPathForInstanceIndex to just take cache paths as input and output, instead of index paths.  (Internally, it already works on cache paths.)

- I chose to use the term "instancers" to refer generically to "prims which generate instances" - ie, either "instanceable-references" or PointInstancers. It's admittedly not the best name, since it's too closely related to PointInstancers... but I couldn't think of a better succinct term to use. Open to suggestions!

- I wasn't entirely sure what to do with the "rprimPath" arg. I renamed it to "masterCachePath" since that's what it seems to be used for - ie, with "instanceable-reference" instancers, it seems to be the cache (usd?) path for the master prim. It is always empty for PointInstancer style instances, however... should it be set to the usd path for the top-level prototype (and we rename the arg to something like "sourceUsdPath")? Or should we leave functionality as is, and perhaps remove the arg from the second overload of UsdImagingPrimAdapter::GetPathForInstanceIndex (since it is never used / filled)?

- The instanceContext only seems to be filled in for instances created via "instanceable references", and not for instances created from instancers. Is this intentional?  Should this be noted in the docs?

- Finally - I said in the doc that "instancerIndex" would have the instance index of the returned top-level instancer, because it's the observed behavior for PointInstancer style instances.  It's not strictly correct, though - instanceable prims seems to always return -1.  There's a note in instanceAdapter.h that they "intentionally leave instancerIndex as it is, so that partial selection of point instancer can be passed through." However, I'm guessing this is a bug - if the user wants to do "exact" selection, they already have everything they need (the original proto path and proto index) - and if they need indices for any other intermediate instancers, this should probably be handled via a list of indices, similar to instanceContext.

tcauchois:
Thanks a bunch for posting this!  The instancing code in usdImaging can be tough to read :).

This change is great, in terms of making GetPathForInstanceIndex more readable/comprehensible.  I have one nitpick: you added a few comments referring to "instancer rprims", which I think is carrying over terminology already present in the code.  In hydra terms, though, an rprim exclusively refers to geometry, and an instancer is a state object that tells geometry how many times to draw.  The usage of "ProtoRprim" and "rprimPath" and etc. in the instancing code is a historical artifact of when instancers couldn't be nested, and it would be nice to move away from that language (towards e.g. ProtoUsdPrim, usdPrimPath, etc).

Your questions above hit on some of the confusing bits of usdImaging, so I thought I'd post some context.

**Cache path vs Usd path**: usdImaging has four different prim tables, and consequently four different path domains.  We've tried to rename variables to indicate which domain they operate in.
- A "usd path" is a path you can pass to stage->GetPrimAtPath(), i.e. representing a real USD prim.
- A "cache path" is a key to a bundle of data in the usdImaging value cache.  Each hydra prim reads data out of the value cache, so cache paths are one to one with hydra objects.  Cache paths are *not* one-to-one with usd paths.  UsdSkel creates multiple hydra prims from the same USD prim; instancing has the potential to create multiple hydra prims from the same USD prim (if, for example, two point instancers refer to the same prototype); and instanced prims will always have a synthesized cache path different from their usd path (computed in GprimAdapter::_ResolveCachePath).  Instancing and skel aside, cache paths and usd paths should match.
- An "index path" is a key to an object in the HdRenderIndex, i.e. a hydra prim.  These are one-to-one with "cache path", but prefixed with the UsdImagingDelegate's "delegateID"; this is so you can embed USD files in heterogenous scene graphs.  In usdview, the delegateID is empty, but in other apps (like Maya) it's not.
- A "primInfo path" is a key to an object in the delegate's "hdPrimInfo" structure.  Every cache path is a valid primInfo path, but native instancing and UsdGeomSubset both add primInfo paths not backed by hydra objects, for dealing with scene change processing.

This is a mess, and eventually it would be lovely to only have "usd path" and "hydra path".

Your ConvertCachePathToUsdPath seems like it would work for prototypes (if you passed the right instance index), and probably for uninstanced rprims.  It won't for instanced instancers, since an instanced point instancer might have a path like /path/to/PI{1}, which isn't a USD path.  In general, the knowledge of which USD prim backs a hydra prim is diffuse: it's in hdPrimInfo for uninstanced objects, and in tables in the pointInstancerAdapter and instanceAdapter for instanced objects.  This is something we should fix.  What's your aim with ConvertCachePathToUsdPath?

**Instancers as prims that generate instances**: this is how hydra uses the term, so it sounds good to me.

**rprimPath and instanceContext**: I think your rename looks good.  rprimPath and instanceContext are used a few places internally, but they mostly aren't used since, as you point out, they are only filled in for native instances.  They seem extraneous to the API, but we'd probably need to change the instanceAdapter implementation of GetPathForInstanceIndex to truly deprecate them.

**instancerIndex**: You hit on one of the core problems in this area of code.  Hydra sees instance indices in two forms: as tuples, and a flattened index.  For example:
/World/PI_A has prototype /World/PI_A/protos/PI_B, indices [1,3]
/World/PI_A/protos/PI_B has prototype /World/PI_A/protos/PI_B/protos/cube, indices [2,4]
The result is that cube draws four times, with index tuples [(1,2), (1,4), (3,2), (3,4)].  The flattened index is 0,1,2,3 respectively.  It's possible to define a canonical mapping from integer to tuple (although I think we don't have it specifically documented anywhere), but the different representations are going to be useful in different contexts, and you'd want to be able to easily switch between them.

GetPathForInstanceIndex and PopulateSelection are designed, in large part, to translate from hydra picking results (rprim ID, flat instance index) to USD selection state (prim ID and instance ID) and back to hydra selection (rprim ID, instance index tuples).  There's some work to do in aligning these representations, but in general, you'd assume that if you selected a native instanced cube, you'd see the following transformation:

(/__Master_1.cube_id0, 2) -> (/Path/To/Instance2/cube, N/A); e.g. the result shouldn't care about whether the cube is instanced-by-reference or not.  If the path is a point instancer prototype, likewise, the result should have the absolute path to the point instancer and the correct point instancer index.  (Note: I don't think this is what it does right now?  It looks like it will return (/Path/To/Instance2) instead...)

There's not really a conclusion here, just a statement of issues :).  Anyway, I hope this clears some things up!

jilliene:
Filed as internal issue #USD-5346

elrond79:
> I have one nitpick: you added a few comments referring to "instancer rprims", which I think is carrying over terminology already present in the code. In hydra terms, though, an rprim exclusively refers to geometry, and an instancer is a state object that tells geometry how many times to draw. The usage of "ProtoRprim" and "rprimPath" and etc. in the instancing code is a historical artifact of when instancers couldn't be nested, and it would be nice to move away from that language (towards e.g. ProtoUsdPrim, usdPrimPath, etc).

Gotcha - This should be fixed in the latest push!

> * A "cache path" is a key to a bundle of data in the usdImaging value cache.  Each hydra prim reads data out of the value cache, so cache paths are one to one with hydra objects.  Cache paths are _not_ one-to-one with usd paths.  UsdSkel creates multiple hydra prims from the same USD prim; instancing has the potential to create multiple hydra prims from the same USD prim (if, for example, two point instancers refer to the same prototype); and instanced prims will always have a synthesized cache path different from their usd path (computed in GprimAdapter::_ResolveCachePath).  Instancing and skel aside, cache paths and usd paths should match.

Hmm - so this answers my question about whether we can assume that if it's an instance, the paths should match.  Currently, the answer is that if it's not an instance or skel, it should match.  However, if there's two cases, there may be more... perhaps we should formalize this, ie, provide some method that can tell us whether a given cache path is the same as the usd path?

> Your ConvertCachePathToUsdPath seems like it would work for prototypes (if you passed the right instance index), and probably for uninstanced rprims. It won't for instanced instancers, since an instanced point instancer might have a path like /path/to/PI{1}, which isn't a USD path. In general, the knowledge of which USD prim backs a hydra prim is diffuse: it's in hdPrimInfo for uninstanced objects, and in tables in the pointInstancerAdapter and instanceAdapter for instanced objects. This is something we should fix. What's your aim with ConvertCachePathToUsdPath?

Well, just a combination of wanting to get an understanding of the mapping, and also wanting an easy way to do the conversion.  The only place where I've actually had to do this back-conversion so far is in when translating picking results, though.
 
> **instancerIndex**: You hit on one of the core problems in this area of code. Hydra sees instance indices in two forms: as tuples, and a flattened index. For example:
> /World/PI_A has prototype /World/PI_A/protos/PI_B, indices [1,3]
> /World/PI_A/protos/PI_B has prototype /World/PI_A/protos/PI_B/protos/cube, indices [2,4]
> The result is that cube draws four times, with index tuples [(1,2), (1,4), (3,2), (3,4)]. The flattened index is 0,1,2,3 respectively. It's possible to define a canonical mapping from integer to tuple (although I think we don't have it specifically documented anywhere), but the different representations are going to be useful in different contexts, and you'd want to be able to easily switch between them.

Hmm... I thought I had a handle on this, and now I'm not so sure.  I'm assuming form this that the input index - what I've renamed to `protoIndex` (and was formerly `instanceIndex`) is what you're calling the "flattened index" - ie, something in [0,1,2,3].  What is returned in the `instancerIndex` (formerly, `absoluteInstanceIndex`), however? Are the possible values [1,3], or [0,1]?

> GetPathForInstanceIndex and PopulateSelection are designed, in large part, to translate from hydra picking results (rprim ID, flat instance index) to USD selection state (prim ID and instance ID) and back to hydra selection (rprim ID, instance index tuples). There's some work to do in aligning these representations, but in general, you'd assume that if you selected a native instanced cube, you'd see the following transformation:
> 
> (/__Master_1.cube_id0, 2) -> (/Path/To/Instance2/cube, N/A); e.g. the result shouldn't care about whether the cube is instanced-by-reference or not. If the path is a point instancer prototype, likewise, the result should have the absolute path to the point instancer and the correct point instancer index. (Note: I don't think this is what it does right now? It looks like it will return (/Path/To/Instance2) instead...)

Yeah, that's what I'm seeing - it returns `/Path/To/Instance2`

tcauchois:
Re: cachePath and usdPath: it would be great to both formalize that, and also make them match in more cases, since for simple instancing setups there's no need to synthesize a cache path.

GetPathForInstanceIndex(): the input index is the flattened index from hydra, correct.  Without digging I can't tell you for sure what the output index is; I know there's currently some issues round-tripping instance indices from GetPathForInstanceIndex() to PopulateSelection(), for some of the nested instancing combinations.

Issue #860, opened 2019-05-30
Indexed primvars use :vals suffix in ABC 


AlexSchwank:
### Description of Issue
Coming from Alembic, indexed primvars use a `:vals` suffix for the value attributes. 

For example, adding an additional UV Set to a cube in Maya and then export it to alembic, results in this ABC structure:
```
  ScalarProperty name=statistics;interpretation=;datatype=string;arraysize=1;numsamps=1
  ScalarProperty name=1.samples;interpretation=;datatype=uint32_t;arraysize=1;numsamps=1
  ScalarProperty name=.childBnds;interpretation=box;datatype=float64_t[6];arraysize=6;numsamps=1
Object name=/pCube1
  CompoundProperty name=.xform;schema=AbcGeom_Xform_v3
Object name=/pCube1/pCubeShape1
  CompoundProperty name=.geom;schema=AbcGeom_PolyMesh_v1
    ScalarProperty name=.selfBnds;interpretation=box;datatype=float64_t[6];arraysize=6;numsamps=1
    ArrayProperty name=.faceIndices;interpretation=;datatype=int32_t;arraysize=24;numsamps=1
    CompoundProperty name=uv;schema=
      ArrayProperty name=.indices;interpretation=;datatype=uint32_t;arraysize=24;numsamps=1
      ArrayProperty name=.vals;interpretation=vector;datatype=float32_t[2];arraysize=24;numsamps=1
    ArrayProperty name=P;interpretation=point;datatype=float32_t[3];arraysize=8;numsamps=1
    ArrayProperty name=N;interpretation=normal;datatype=float32_t[3];arraysize=24;numsamps=1
    CompoundProperty name=.arbGeomParams;schema=
      CompoundProperty name=secondaryUVsA;schema=
        ArrayProperty name=.indices;interpretation=;datatype=uint32_t;arraysize=24;numsamps=1
        ArrayProperty name=.vals;interpretation=vector;datatype=float32_t[2];arraysize=14;numsamps=1
    ArrayProperty name=.faceCounts;interpretation=;datatype=int32_t;arraysize=6;numsamps=1
```
`secondaryUVsA` is a compound attribute with two children: `.indices` and `.vals` . The same is the case for the `uv` attribute. Converting it to USD results in a stage with this:
```
        custom uint[] primvars:secondaryUVsA:indices
        uint[] primvars:secondaryUVsA:indices.timeSamples = {
            1: [ ... ],
        }
[...]

PR #864, opened 2019-06-04
[usdAbc] fixed reading indexed attributes 

AlexSchwank:
### Description of Change(s)

- #860 : removed `:vals` name suffix for indexed attributes
- #859 : added support for indexed attributes using indices of type `uint[]` by converting to `int[]`
- updated `testUsdAbcIndexedProperties` baseline USD file

### Fixes Issue(s)
- #860 
- #859

### DIFF
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -76,6 +76,8 @@ TF_DEFINE_PRIVATE_TOKENS(
     _tokens,
     (transform)
     ((xformOpTransform, "xformOp:transform"))
+    ((indicesSuffix, ":indices"))
+    ((valuesSuffix, ":vals"))
 );
 
 TF_DEFINE_ENV_SETTING(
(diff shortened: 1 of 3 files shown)


pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -2156,18 +2158,36 @@ _PrimReaderContext::AddOutOfSchemaProperty(
...
+            // If the property is indexed and we're adding the indices here
+            // then we want to make sure they are added as an IntArray
+            // as that's the only supported type for index arrays.
+            usdTypeName = SdfValueTypeNames->IntArray;
+        }
+        else if (TfStringEndsWith(name, _tokens->valuesSuffix)) {
+            // If the property is indexed and we're adding the values
+            // it might have a ":vals" suffix. We need to remove it
+            // to be compatible.
+            propName = TfToken(name.substr(0, name.size() - _tokens->valuesSuffix.size()));
```
spiffmon: In the comment you indicate that the ":vals" suffix may or may not be there... this calculation will break if they are not there.  More robust to do a find() of the suffix if the comment is accurate.

pxr/usd/plugin/usdAbc/testenv/testUsdAbcIndexedProperties/baseline/indexedTextureCoordinates.def.usda
```diff
@@ -53,8 +53,8 @@ def Mesh "Cube"
     uchar xform:ops.timeSamples = {
         0: 48,
     }
-    custom matrix4d xform:vals
-    matrix4d xform:vals.timeSamples = {
+    custom matrix4d xform
+    matrix4d xform.timeSamples = {
```
spiffmon: I guess it's independent of this change, but this change in output hilites a problem we should probaly catch and fix:  if alembic allows arbitrary properties to be indexed, we can't support that, since USD only allows indexing of primvars.  So if we find anything other than an arbGeom property that's indexed, we should flatten it on the Alembic side?

jilliene:
Filed as internal issue #USD-5348

pxr/usd/plugin/usdAbc/alembicReader.cpp
spiffmon: Apologies - brain malfunction; you’ve of course already determined it has the suffix.

pxr/usd/plugin/usdAbc/testenv/testUsdAbcIndexedProperties/baseline/indexedTextureCoordinates.def.usda
AlexSchwank: I think this points to two more problems:
 - in this case "vals" shouldn't be removed as there are no indices
 - the xform attribute isn't really used in this prim

cameronwhite:
I encountered similar issues and saw that PR #1635 already fixed the `:vals` suffix, but the `int[]` issue for indices still remains

@AlexSchwank - would you be able to rebase your changes? If you don't have time to work on this, I could also put together an updated PR

PR #868, opened 2019-06-07
Fixes for newer versions of pyside

elrond79:
### Description of Change(s)

Newer versions of PySide (ie, 5.11) have changes that break existing usdview code.
This fixes these issues, while still maintaining backwards compatibility with older versions.

Note that I've closed https://github.com/PixarAnimationStudios/USD/pull/783 in favor of this

### Fixes Issue(s)
- Usdview - reset view doesn't work with newer PySide (5.11)
- Usdview - debug flags window doesn't work with newer PySide (5.11)

### DIFF
--- a/pxr/usdImaging/lib/usdviewq/appController.py
+++ b/pxr/usdImaging/lib/usdviewq/appController.py
@@ -742,7 +742,7 @@ class AppController(QtCore.QObject):
 
             self._ui.actionFrame_Backwards.triggered.connect(self._retreatFrame)
 
-            self._ui.actionReset_View.triggered.connect(self._resetView)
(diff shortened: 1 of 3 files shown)


spiffmon:
Hi @elrond79 ,
   I think we'd rather have pyside/pyside2 mapping issues addressed in [qt.py](https://github.com/PixarAnimationStudios/USD/blob/61c6685d05d777a973539ba2a28f6f8deb5c042f/pxr/usdImaging/lib/usdviewq/qt.py)

Thanks!
@spiffmon

spiffmon:
(and sorry we missed this one!)

elrond79:
Moved into qt.py!

PR #869, opened 2019-06-07
Fixes for building with gcc-6 / c++14 with PXR_STRICT_BUILD_MODE

elrond79:
### Description of Change(s)

Were a few leftover warnings when I tried to build with gcc-6 / c++14... but not many, so I just fixed them up and bundled into a PR.

### Fixes Issue(s)
- Sign comparison warnings with gcc-6 / c++14
- Unused function warnings with gcc-6 / c++14

### DIFF
--- a/pxr/base/lib/tf/span.h
+++ b/pxr/base/lib/tf/span.h
@@ -196,12 +196,12 @@ public:
     /// has a range of [data()+offset, data()+size()). Otherwise, the new
     /// span has range [data()+offset, data()+offset+count).
     TfSpan<T> subspan(difference_type offset, difference_type count=-1) const {
-        TF_DEV_AXIOM(offset >= 0 && offset < _size);
+        TF_DEV_AXIOM(offset >= 0 && (index_type)offset < _size);
         if (count == -1) {
             return TfSpan<T>(_data + offset, _size - offset);
         } else {
             TF_DEV_AXIOM(count >= 0);
-            TF_DEV_AXIOM((offset+count) <= _size);
+            TF_DEV_AXIOM((index_type)(offset+count) <= _size);
             return TfSpan<T>(_data + offset, count);
         }
     }
--- a/pxr/base/lib/work/filter.h
+++ b/pxr/base/lib/work/filter.h
@@ -107,7 +107,7 @@ WorkParallelFilterN(size_t N,
          &accumVector,
          &starts](size_t begin, size_t end)
            {
-                for(int j = begin; j < end; ++j){
+                for(size_t j = begin; j < end; ++j){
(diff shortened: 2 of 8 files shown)


c64kernal:
Thanks again @elrond79  -- there are some suggested changes that I'll take care of on our end, but wanted to leave them in the code so you know and in case you had other thoughts.

pxr/base/lib/tf/span.h
```diff
@@ -196,12 +196,12 @@ class TfSpan
...
     /// span has range [data()+offset, data()+offset+count).
     TfSpan<T> subspan(difference_type offset, difference_type count=-1) const {
-        TF_DEV_AXIOM(offset >= 0 && offset < _size);
+        TF_DEV_AXIOM(offset >= 0 && (index_type)offset < _size);
         if (count == -1) {
             return TfSpan<T>(_data + offset, _size - offset);
         } else {
             TF_DEV_AXIOM(count >= 0);
-            TF_DEV_AXIOM((offset+count) <= _size);
+            TF_DEV_AXIOM((index_type)(offset+count) <= _size);
```
c64kernal: Here, I've converted offset and count separately.

pxr/base/lib/work/filter.h
```diff
@@ -107,7 +107,7 @@ WorkParallelFilterN(size_t N,
          &accumVector,
          &starts](size_t begin, size_t end)
            {
-                for(int j = begin; j < end; ++j){
+                for(size_t j = begin; j < end; ++j){
```
c64kernal: Thanks for fixing this, these are unused in pxr, so we'll be moving them out soonish.

pxr/usd/lib/pcp/testenv/testPcpIterator.cpp
```diff
@@ -215,7 +215,7 @@ _TestRandomAccessOperations(IteratorType first, IteratorType last)
 {
     TF_AXIOM(first != last);
 
-    size_t idx = 0;
+    decltype(first - last) idx = 0;
```
c64kernal: This one we thought to use std::iterator_traits<IteratorType>::difference_type instead...

c64kernal:
And for all the usdAbc UNUSED_FUNCTIONS, we're just going to remove them.  The ones in my previous change that I tagged this way was because we want to cause minimal difference to code that we integrated from elsewhere.

Hope all these make sense, and please let me know if you have any thoughts!  (Again no action needed on your end).  We'll likely get this merged in after we push 19.07 out.

Thanks again @elrond79

elrond79:
All that sounds fine to me, thanks!

jilliene:
Filed as internal issue #USD-5376

PR #873, opened 2019-06-10
fix for OpenImageIO >= 1.9.1

elrond79:
### Description of Change(s)

the non-ROI get_pixels overloads were removed in this version, and USD was
still using it, despite the fact that it was deprecated.

This adds a version switch, so that the ROI version is used if available

### Fixes Issue(s)
- Building against OpenImageIO >= 1.9.1 failed


jilliene:
Filed as internal issue #USD-5378

Issue #874, opened 2019-06-11
An unconfigured HdxColorizeTask never converges


marktucker:
### Description of Issue
When an HdxTaskController creates an HdxColorizeTask, but the colorize task is not configured with a specific aov buffer, the task will never converge, so the task controller will never converge, so any progressive rendering will continue forever.

### Steps to Reproduce
1. Create an HdxTaskController but with no aov buffer specified through SetViewportRenderOutput.
2. Note that the HdxColorizeTask is create, but with an empty _aovBufferPath, which results in a null _aovBuffer.
3. In the Execute method, the null _aovBuffer results in an early exit which does not set the _converged flag.
4. Any subsequent call to HdxTaskController::IsConverged will return fals because the colorize task can never converge.

This is admittedly not a standard use case, but it is part of our use case. And it makes sense to me that a colorize task that has not been asked to do anything should always be marked as converged, since it by definition has completed all its work.

jilliene:
Filed as internal issue #USD-5374

tcauchois:
Fixed by PR 875.

PR #875, opened 2019-06-11
Always mark unconfigured HdxColorizeTasks as converged on an Execute call

marktucker:
If an HdxColorizeTask has not been configured with an aov buffer, before
doing an early exit from the Execute method, set the _converged flag to
true so this task isn't left in an unconverged state forever.

### Fixes Issue(s)
#874


jilliene:
Filed as internal issue #USD-5379

c64kernal:
Nice catch @marktucker  -- thank you so much!

Issue #872, opened 2019-06-10
pxr.UsdGeom.Xformable.ClearXformOpOrder on an invalid prim crashes


marktucker:
### Description of Issue
Invoking ClearXformOpOrder on an invalid UsdGeom.Xformable object in python will cause a crash.

### Steps to Reproduce
1. start python
2. run:
from pxr import UsdGeom
p = UsdGeom.Xformable()
p.ClearXformOpOrder()
The process will crash.

There are about a dozen levels in the call stack where the proverbial null pointer check could be added, but I'm not sure which of those is the right level, in line with existing protections for python functions on invalid Usd objects (trading off between the number of functions to which a null pointer check must be added vs the additional runtime of performing redundant null pointer checks).

### Package Versions
19.05

spiffmon:
Thanks for reporting, @marktucker !  I think the desired fix here is to add similar __getattribute__ treatment as is given in [usd/wrapObject.cpp](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/lib/usd/wrapObject.cpp#L148) to wrapSchemaBase.cpp ... and also the one or two "property schemas" we have, like UsdGeomPrimvar.

marktucker:
I just did a quick test, and was able to implement this for UsdGeomXformable and UsdGeomPrimvar pretty easily... If you want to tell me which other classes fall into this category I can wrap them all into one PR? I also didn't whitelist any particular functions since I didn't see any of them making sense to call on "null" prims. Let me know if you think there are functions that should pass through the __getattribute__ filter.

spiffmon:
Hi @marktucker !  If you hit wrapSchemaBase.cpp in core usd, that should handle *all* derived generated schema classes.  UsdGeomPrimvar is the only "USD Object wrapper" schema I can recall that needs special treatment of its own.

I do think we want to pass at least GetPrim(), GetPath(), and the operator bool (though I'm not sure how to indicated that in this context.. my python-wrapping fu is not great).  @gitamohr might have suggestions of other methods it would be good to pass.

marktucker:
Not claiming the above pull request is complete, but I think it's easier to have something concrete to look at when discussing possible refinements. I ended up allowing all SchemaBase methods to be called on invalid SchemaBase objects, because they are safe, and make sense even if the underlying prim is null. But this change blocks all methods of all subclasses.

jilliene:
Filed as internal issue #USD-5373

[...]

PR #876, opened 2019-06-13
Prevent crashes when calling methods on invalid Usd.SchemaBase objects

marktucker:
In wrapSchemaBase, replace the __getattribute__ method as was already being
done in wrapObject. This overloaded __getattribute__ actually allows any of
the UsdSchemaBase APIs to be called (as all of the methods at this level can
still return useful information). But its purpose is to block calls to APIs
defined on subclasses.

Provided similar protection for UsdGeomPrimvar, which is an odd case that is
like a schema class, but operates on a single attribute and so is not derived
from SchemaBase.

Added some test cases, including an explicit test of issue #872, a crash when
calling UsdGeom.Xformable().ClearXformOpOrder(). TestUsdRiSplineAPI had to be
updated because a test in there was not expecting the RuntimeException from
an invalid object.

### Fixes Issue(s)
- #872

### DIFF
--- a/pxr/usd/lib/usd/CMakeLists.txt
+++ b/pxr/usd/lib/usd/CMakeLists.txt
@@ -189,6 +189,7 @@ pxr_test_scripts(
     testenv/testUsdPrims.py
     testenv/testUsdReferences.py
     testenv/testUsdRelationships.py
+    testenv/testUsdSchemaBasePy.py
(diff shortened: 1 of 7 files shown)


marktucker:
Thanks for the feedback, @spiffmon. Addressing this first round of comments with an additional commit.

jilliene:
Filed as internal issue #USD-5380

PR #880, opened 2019-06-19
FindRenderman.cmake: export RENDERMAN_BINARY_DIR, use it to find oslc

elrond79:
### Description of Change(s)

the building of the renderman-22 plugin made use of the RENDERMAN_LOCATION
var, which is not guaranteed to be set by FindRenderman.cmake; instead,
directly export RENDERMAN_BINARY_DIR, and use that to find oslc

### Fixes Issue(s)
- error building renderman plugin if RMANTREE or RENDERMAN_LOCATION supplied only via environment variable


jilliene:
Filed as internal issue #USD-5387

Issue #852, opened 2019-05-28
Incorrect behaviour of translucent blending


sirpalee:
### Description of Issue
At the moment translucent blending in dev doesn't behave as expected. This is due to an unusual blending logic in oitResolveImageShader and an incorrect OpenGL state setting.

More details in the comments below.

### Steps to Reproduce
1, either hardcode the material tag to translucent for the preview surface (in hydraMaterialAdapter.cpp) or merge in #846 and set PXR_PREVIEW_SURFACE_TRANSLUCENT_TAG to translucent.
2, open any of the attached files.

sirpalee:
The following example shows transparent planes over a plane, with opacity varying from left to right as 0.25, 0.5, 0.6, 0.7, 0.8, 0.9, 0.95, 0.96, 0.97, 0.98, 0.99, 1.0 and the last plane has an opaque material (opacity is one, and non connected, meaning it doesn't have the translucent material tag).

How it looks with the current dev branch:
![opacity_original](https://user-images.githubusercontent.com/582376/58449491-179bd980-814f-11e9-9f71-26f85a7498c2.png)

After making some changes to the OpenGL state and blending logic (see details below):
![opacity_updated](https://user-images.githubusercontent.com/582376/58449784-8a0cb980-814f-11e9-8d39-f7d59a5b9b51.png)

Reference render using Arnold and a uniform white skydome light:
![opacity_arnold](https://user-images.githubusercontent.com/582376/58449482-0fdc3500-814f-11e9-977d-b505dbdfa833.png)

As you can see, there isn't a smooth interpolation between different opacity values in the original code, and translucent 1.0 doesn't create the same look as non-translucent, opaque.

sirpalee:
The next example shows two, white, planes with 0.5 opacity and a red plane behind them, also with 0.5 opacity.

Original:
![opacity_overlap_original](https://user-images.githubusercontent.com/582376/58449879-cd672800-814f-11e9-8eaa-d44f6bd9ac13.png)

Updated:
![opacity_overlap_updated](https://user-images.githubusercontent.com/582376/58449887-d2c47280-814f-11e9-8226-f2179ffaeba0.png)

Arnold reference:
![opacity_overlap_arnold](https://user-images.githubusercontent.com/582376/58449890-d952ea00-814f-11e9-89c7-322f93305bc7.png)

As you can see, in the original example the red plane is completely hidden.

sirpalee:
[...]

PR #884, opened 2019-06-24
[Imaging] Improvements to the Order Independent Transparency feature in HdStream

sirpalee:
### Description of Change(s)
The following PR adds new features and fixes issues related to OIT for HdStream.
- Previously viewport resizing was slow because the OIT buffers were initialized with a default buffer source argument, leading to the creation of large CPU side buffers with the same value. This is now removed, and the new code is using the "Resize" function on the buffers. We don't need to set their default value since it's either going to be cleared, or guaranteed to be overwritten by the OIR Render Pass.
- Removed all the buffer clear for OIT, except the counter buffer. We don't need to clear the buffers before the render since the values of every other OIT buffer is going to be overwritten by the OIT render pass.
- Replaced the insert sort function with a select sorting for the oitResolveImage function. While benchmarking the shader, we found that select sorting is roughly the same performance as insert sorting with 8 samples, but it's significantly faster with more samples (about 2-3 times faster with 32-64 samples). These test results were seen on Centos, with a GTX 1060 (410.78) and the "powerplant" scene with all the faces being partially transparent. Alternatively, it's easy to make the sorting method selectable using an environment variable, if there are cases where select sorting is slower than insert sorting.
- The number of OIT samples per pixel is now a render delegate parameter and can be changed on the fly. To maintain consistency with parameter names across all the different bits of the code, I renamed all the parameter uses to oitNumSamples. Please let me know if you prefer the name being oitSamples instead.
- The PR also includes our fix for #852 . Please let me know if you would like to get it removed from this PR.
- We saw issues with OIT in Maya 2019 and 2018 on Centos, glClearNamedBuffer is not available in those contexts, so we added a simple function that falls back to glClearNamedBufferEXT if glClearNamedBuffer is not available.
- Added an early exit to RenderOutput (in hdx/renderPass.glslfx), if the alpha value of the fragment is small.

### Fixes Issues
- #852

### DIFF
--- a/pxr/imaging/lib/hd/renderDelegate.h
+++ b/pxr/imaging/lib/hd/renderDelegate.h
@@ -72,6 +72,15 @@ typedef TfHashMap<TfToken, VtValue, TfToken::HashFunctor> HdRenderSettingsMap;
 /// wants to export (e.g. to UI).
 ///
 struct HdRenderSettingDescriptor {
+    // Default constructor.
+    HdRenderSettingDescriptor() = default;
+    // Constructor that allows the use of emplace_back on standard
+    // containers.
+    HdRenderSettingDescriptor(std::string const& _name,
+                              TfToken const& _key,
+                              VtValue const& _defaultValue)
+        : name(_name), key(_key), defaultValue(_defaultValue)
+    { }
     // A human readable name.
     std::string name;
     // The key for HdRenderDelegate::SetRenderSetting/GetRenderSetting.
--- a/pxr/imaging/lib/hdSt/renderDelegate.cpp
+++ b/pxr/imaging/lib/hdSt/renderDelegate.cpp
@@ -57,6 +57,12 @@ PXR_NAMESPACE_OPEN_SCOPE
 TF_DEFINE_ENV_SETTING(HD_ENABLE_GPU_TINY_PRIM_CULLING, false,
                       "Enable tiny prim culling");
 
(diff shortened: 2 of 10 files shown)


jilliene:
Filed as internal issue #USD-5388

lumonix:
Thank you for this contribution @sirpalee 
This is great.

Since some of your changes overlap with internal OIT work (will start to land in dev soon) I'll take some of your contributions in manually via internal change-lists instead of merging this request.

We'll wait a bit longer on exposing settings (like the numSamples) until we have completed our on-going volume rendering work in Hydra.

sirpalee:
@lumonix sounds good to me!

I updated the PR with a few small changes:
- Fixed a debug message if the oit num samples parameter's type doesn't match.
- Missed a minimum value limit for the number of samples in the resolve task.
- Added a new constructor to HdRenderSettingDescriptor, so we can use .emplace_back on the HdRenderSettingDescriptorList. This way adding new render settings to the HdSt list is a bit more convenient, and less error-prone (when there is a mismatch in the preallocated vector, vs number of settings). I also added a default constructor, so existing code in other render delegates work as well.

poljere:
Thanks @sirpalee. We are going to close this PR now, we have internally incorporated some of your changes and you will see them in the coming days in dev, for instance: https://github.com/PixarAnimationStudios/USD/commit/42a7ebf5f5fed3654e8416287099f5c1d4a2900c

PR #885, opened 2019-06-24
cmake: added support for external libraries for pxr_python_bin

AlexSchwank:
### Description of Change(s)
 - cmake: added support for external libraries for pxr_python_bin

This is not yet needed in the vanilla pxr setup but it is in others like ours where we've split our USD build into separate builds.

### DIFF
--- a/cmake/macros/Public.cmake
+++ b/cmake/macros/Public.cmake
@@ -29,6 +29,7 @@ function(pxr_python_bin BIN_NAME)
     )
     set(multiValueArgs
         DEPENDENCIES
+        LIBRARIES
     )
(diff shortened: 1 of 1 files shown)


AlexSchwank:
Oh? cmake doesn't always like the change:
```
Utility target "usdchecker_script" must not be used as the target of a
  target_link_libraries call.
```
Is there anything special about the Windows cmake builds?

AlexSchwank:
https://cmake.org/cmake/help/v3.0/policy/CMP0039.html is causing this. I'll check for a better solution.

jilliene:
Filed as internal issue #USD-5389

PR #886, opened 2019-06-24
make PXR_VERSION a valid / comparable c/c++ integer

elrond79:
### Description of Change(s)

currently, PXR_VERSION is defined as "0197" - which isn't even a valid
c/c++ integer literal, as the '0' will cause it to be interpreted as an
octal value, but '9' isn't an octal digit.  Even if it didn't have the
leading 0, however, it still wouldn't be useful for comparing numerically,
because we would have, ie, 19.07 < 18.11:

197 < 1811

### Fixes Issue(s)
- PXR_VERSION macro not a valid integer literal


jilliene:
Filed as internal issue #USD-5390

c64kernal:
Great sug, thanks @elrond79  -- don't worry about the merge conflict, we'll take care of it.

PR #888, opened 2019-06-29
Adding new APIs to the render delegate for pause, resume, and getting status

marktucker:
Adding new APIs to the render delegate (exposed through the UsgImagingGLEngine)
to pause and resume a render (if supported). Also added an API to get a status
message from the render delegate that can be presented to the user to give some
indication of the progress of the render.

Added implementations of these APIs to the embree delegate, which uses matching
new methods on HdRenderThread to assist with the pause and resume operations.

Added a Pause menu item to usdview, and display the render status message in
the upper right hand corner below the render delegate name.


jilliene:
Filed as internal issue #USD-5407

PR #898, opened 2019-07-11
[imaging] fix another sign compare issue

elrond79:
### Description of Change(s)
slipped through because it was in a template that's not used by core
(but is by HdPrman)

### Fixes Issue(s)
- building renderman-22 plugin with PXR_STRICT_BUILD_MODE enabled


jilliene:
Filed as internal issue #USD-5430

PR #900, opened 2019-07-12
MESA AMD. Added VAO bind()/unbind() to HdxCompositor to stop error in Core Context.

DuncanHopkinsFoundry:
If running on a Core OpenGL context a VAO is needed before any vertex attrib binds can be done.
Errors seen on Mesa AMD Vega drivers under Ubuntu.


jilliene:
Filed as internal issue #USD-5432

PR #901, opened 2019-07-12
Core Context. Added USDVIEW_ENABLE_COREGL envvar to force GL Core context

DuncanHopkinsFoundry:
Added an environment variable, similar to MSAA, to force a GL Core Context in USDView.


jilliene:
Filed as internal issue #USD-5433

PR #902, opened 2019-07-12
MESA AMD. Fixed syntax arror with ';' after function body.

DuncanHopkinsFoundry:
MESA AMD Vega drivers failed with the extra ';' in GLSL! 
Seen on Ubuntu.


jilliene:
Filed as internal issue #USD-5434

PR #920, opened 2019-07-25
fix for OIT with older glew

elrond79:
### Description of Change(s)

...such as that available within maya.  Older glew will mean a different set
of gl functions are available.

### Fixes Issue(s)
- Crash in maya when OIT is used

### DIFF
--- a/pxr/imaging/lib/hdx/oitRenderTask.cpp
+++ b/pxr/imaging/lib/hdx/oitRenderTask.cpp
@@ -232,7 +232,7 @@ _GetScreenSize()
 
     GlfContextCaps const &caps = GlfContextCaps::GetInstance();
 
-    if (ARCH_LIKELY(caps.directStateAccessEnabled)) {
+    if (ARCH_LIKELY(caps.directStateAccessEnabled) && glGetTextureLevelParameteriv) {
         if (attachType == GL_TEXTURE) {
             glGetTextureLevelParameteriv(attachId, 0, GL_TEXTURE_WIDTH, &s[0]);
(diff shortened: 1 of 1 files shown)


lumonix:
Hey @elrond79 

Ideally we don't check for individual functions (though we do it in other places too).
Does your problem go away if we check for:

contextCaps gl version >= 4.5?
Or perhaps we need to do a glew version check (I've seen other problems with older glew)

Side note, I'm working on removing this GetScreenSize function all together.

Thank you for submitting this, we don't always catch problems with using newer gl functions on our local machines.

elrond79:
Our GL version is 4.6.  I was basing my change off a similar change here:

https://github.com/PixarAnimationStudios/USD/blob/7945fe1586bc2bc081dc82fa6c00afe7c1df0eb4/pxr/imaging/lib/hdx/oitRenderTask.cpp#L412

If you'd rather I do a check against glew version, I can do that... Though
I should note I'm not 100% certain this is caused by glew - I'm making that
assumption off of conversations with Pal where he had a similar issue, and
since we're on GL 4.6, I'm not sure what else would cause this.

Would be good to get some fix in soonish, though, as it basically prevents
newer USD from being used with Maya.

jilliene:
Filed as internal issue #USD-5470

lumonix:
Thanks for the info. 
And yes the code you highlight is the same work-around. So i guess this is ok too :). 

We run Maya by force pre-loading a newer version of GLEW, btw. 
Maya’s is very old.

sunyab:
Hi @elrond79, we think this should be resolved by commit 1e193cc so I'm going to close this PR out. Please let us know if this is still an issue, thanks!

PR #925, opened 2019-08-01
Fixed Alembic URL

tpoveda:
Fixed Alembic URL on BUILDING.md


jilliene:
Filed as internal issue #USD-5475

PR #926, opened 2019-08-01
[Usdview] Allow plugins to customize action items

hshakula:
### Description of Change(s)
As for now, there is no way for loaded usdview plugins to configure the appearance of CommandPlugin's item. I propose adding wrapper over QAction to let plugin users do it


jilliene:
Filed as internal issue #USD-5476

PR #927, opened 2019-08-02
Fixing texture size query in case of multisampled textures.

elrond79:
### Description of Change(s)

In the branch where directStateAccessEnabled is false, if you bind a
multisampled texture to GL_TEXTURE_2D, querying the width and height
will always return 1024 (the default).

This fix is most evident with https://github.com/PixarAnimationStudios/USD/pull/920 also applied (ie, in Maya), though technically it should be fixed regardless, so I made a separate PR.

### Fixes Issue(s)
- wrong OIT render size in Maya


jilliene:
Filed as internal issue #USD-5477

lumonix:
Thanks!
I'm in process of removing the GetScreenSize all together in favor of grabbing the size from AOVs.

PR #933, opened 2019-08-12
[usd] Register missing C++ unit tests.

asluk:
Some C++ tests were not being run by ctest due to missing registration.


jilliene:
Filed as internal issue #USD-5499

PR #936, opened 2019-08-14
Missing export for Windows build of classes based on UsdImagingGLEngine

kxl-adsk:
### Description of Change(s)
On windows classes derived from UsdImagingGLEngine don't have export properly set for protected method.

### Fixes Issue(s)
- Fix windows build of AL plugin and USD dev.


jilliene:
Filed as internal issue #USD-5500

PR #945, opened 2019-08-20
Build boost with more settings whether or not Maya is built

HamedSabri-adsk:
This change allows building boost library with more settings and therefore no need to pass this long argument to build_usd script.

boost,"--with-date_time --with-thread --with-system --with-filesystem"

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -578,12 +578,12 @@ def InstallBoost(context, force, buildArgs):
         if context.buildPython:
             b2_settings.append("--with-python")
 
-        if context.buildKatana or context.buildOIIO:
+        if context.buildKatana or context.buildOIIO or not context.buildMaya or context.buildMaya:
             b2_settings.append("--with-date_time")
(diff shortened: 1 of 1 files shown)


hshakula:
Consider simplifying ``` not context.buildMaya or context.buildMaya ``` to ``` True ``` 😉

jilliene:
Filed as internal issue #USD-5507

c64kernal:
Hi @HamedSabri-adsk -- we deliberately only build as little of boost as is needed for the selected components.

c64kernal:
After discussing it, it seems like the possible solutions are:
- Documenting what a particular product needs to build with (as we believe Autodesk is currently doing)
- Look into ways of reducing the dependencies on non-header-only boost libraries, which is a long term goal for the USD project in general.
Closing this one out for now, but happy to discuss more, if not having this will cause undue trouble.

HamedSabri-adsk:
@c64kernal Sure, it makes sense. Thank you.

PR #949, opened 2019-08-28
Adding a validity check on a UsdSkelAnimQuery object before using it

marktucker:
In this block of code, it is possible to make a call into an invalid UsdSkelAnimQuery object. This results in error messages being generated for no reason. I'm not certain that calling SetMightBeTimeVarying(false) if the anim query is invalid is the right thing to do.

This error was being triggered when calling BakeSkinning on a scene containing skeletons with a restTransforms attribute.


jilliene:
Filed as internal issue #USD-5520

Issue #952, opened 2019-08-29
Disappearing geometry and/or crash when running BakeSkinning


marktucker:
### Description of Issue
When running BakeSkinning on a stage loaded in usdview, USD exhibits a variety of bad behaviors depending on the render delegate and OS.

### Steps to Reproduce
1. Load the attached USD file into usdview:
[crowd.zip](https://github.com/PixarAnimationStudios/USD/files/3556433/crowd.zip)
2. In the interpreter window run:
UsdSkel.BakeSkinning(usdviewApi.stage.Traverse())
3. On Linux running Storm or Embree, the geometry all disappears. (switching to another renderer and back makes the geometry show up properly).
4. On Windows, this just crashes.

### System Information (OS, Hardware)
Various.

### Package Versions
Dev branch from aug 23, 2019: b85ddac2ad2d4be3ea2b49d4383b9dd95c7038ca

jilliene:
Filed as internal issue #USD-5519

PR #953, opened 2019-08-29
Fix crash in BakeSkinning

marktucker:
Fix crash in BakeSkinning caused by multiple threads populating a shared map object at the same time.

### Description of Change(s)
Pre-populate the shared map before giving the mutliple threads access to it.

### Fixes Issue(s)
- I believe this fixes the crash in #952, though the geometry still disappears after calling BakeSkinning.


jilliene:
Filed as internal issue #USD-5522

PR #955, opened 2019-09-01
[HdSt] Installing domeLight.glslfx.

sirpalee:
### Description of Change(s)
Domelight glslfx is not installed as part of HdSt, the PR addresses this issue.


poljere:
Good catch, thanks @sirpalee!

jilliene:
Filed as internal issue #USD-5527

Issue #950, opened 2019-08-28
usdMtlx fails to read a mtlx file with more than one bind input


juliencb2018:
### Description of Issue
usdMtlx is failing to properly read a materialX file that has more than one bind input.

### Steps to Reproduce
1. Input mtlx file with more than one bind input:

MultiBindInputs.mtlx

```
<?xml version="1.0"?>
<materialx version="1.36">                                                                                                                                                                                    
  <nodedef name="ND_layerShader" type="surfaceshader" node="layerShader">                                                                                                                                     
        <input name="weight_1" type="color3"/>                                                                                                                                                                
        <input name="weight_2" type="color3"/>                                                                                                                                                                
        <input name="weight_3" type="color3"/>                                                                                                                                                                
    </nodedef>                                                                                                                                                                                                
  <Gradient name="layered_layer1_gradient" type="color3">                                                                                                                                                     
  </Gradient>                                                                                                                                                                                                 
  <Gradient name="layered_layer2_gradient" type="color3">                                                                                                                                                     
  </Gradient>                                                                                                                                                                                                 
  <Gradient name="layered_layer3_gradient" type="color3">                                                                                                                                                     
  </Gradient>                                                                                                                                                                                                 
  <material name="layered">                                                                                                                                                                                   
    <shaderref name="layered_sr" node="layerShader">                                                                                                                                                          
      <bindinput name="weight_1" type="color3" output="o_layered_layer1_gradient" />                                                                                                                          
      <bindinput name="weight_2" type="color3" output="o_layered_layer2_gradient" />                                                                                                                          
      <bindinput name="weight_3" type="color3" output="o_layered_layer3_gradient" />                                                                                                                          
    </shaderref>                                                                                                                                                                                              
  </material>                                                                                                                                                                                                 
  <output name="o_layered_layer1_gradient" type="color3" nodename="layered_layer1_gradient" />                                                                                                                
  <output name="o_layered_layer2_gradient" type="color3" nodename="layered_layer2_gradient" />                                                                                                                
  <output name="o_layered_layer3_gradient" type="color3" nodename="layered_layer3_gradient" />                                                                                                                
</materialx>
```

2. python  calls to read and parse this file:

```
from pxr import Usd, UsdShade, Sdf
stage = Usd.Stage.CreateInMemory()
stage = stage.Open('MultiBindInputs.mtlx')
[...]

PR #956, opened 2019-09-04
Dev issue950

juliencb2018:
### Description of Change(s)
usdMtlx can now properly read a materialX file that has more than one bind input.
Also added debug statements.

### Fixes Issue(s)
Issue #950:
usdMtlx fails to read a mtlx file with more than one bind input 

-

### DIFF
--- a/pxr/usd/plugin/usdMtlx/CMakeLists.txt
+++ b/pxr/usd/plugin/usdMtlx/CMakeLists.txt
@@ -24,6 +24,7 @@ pxr_plugin(usdMtlx
         ${MATERIALX_INCLUDE_DIRS}
 
     CPPFILES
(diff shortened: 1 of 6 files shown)


jilliene:
Filed as internal issue #USD-5528

c64kernal:
Thanks for the fix @juliencb2018 -- we'll get this merged in soon.  Two small things we changed in reader.cpp is that we kept what look like extraneous parens to avoid some compilers warning about them (it looks like a missed equality check) and then we use the IsEmpty() method on token instead of comparing against an empty string.  Thanks again, look for the merge soon!

juliencb2018:
Great, thanks for letting us know!

phuchau1989 (approved):

PR #963, opened 2019-09-16
MESA AMD. Fixed syntax arror with ';' after function body.

rfanner:
### Description of Change(s)
MESA AMD Vega drivers failed with the extra ';' in GLSL. 

Seen on Ubuntu.
### Fixes Issue(s)
- Filed as Pixar-internal issue #USD-5434


jilliene:
Filed as internal issue #USD-5545

PR #964, opened 2019-09-16
MESA AMD. Disabled directStateAccess if glNamedBufferDataEXT function…

rfanner:
### Description of Change(s)
Disables directStateAccess on systems with NULL glNamedBufferDataEXT pointers.
This was seen when using MESA Radeon Vega drivers on Ubuntu.
### Fixes Issue(s)
- Filed as Pixar-internal issue #USD-5431

### DIFF
--- a/pxr/imaging/lib/glf/contextCaps.cpp
+++ b/pxr/imaging/lib/glf/contextCaps.cpp
@@ -277,7 +277,7 @@ GlfContextCaps::_LoadCaps()
     if (!TfGetEnvSetting(GLF_ENABLE_MULTI_DRAW_INDIRECT)) {
         multiDrawIndirectEnabled = false;
     }
-    if (!TfGetEnvSetting(GLF_ENABLE_DIRECT_STATE_ACCESS)) {
+    if (!TfGetEnvSetting(GLF_ENABLE_DIRECT_STATE_ACCESS) || NULL == glNamedBufferDataEXT) {
         directStateAccessEnabled = false;
     }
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-5546

davidgyu:
We're investigating whether a better fix might be for us switch our implementation to call the ARB extension methods, i.e. glNamedBufferDataARB.

It would be helpful to get some more information about your system. Could you attach the output of running glxinfo on your system?, e.g.
  % glxinfo > glxinfo_output.txt

This would help confirm which API version and extensions are supported.

Thanks!

DuncanHopkinsFoundry:
> It would be helpful to get some more information about your system. Could you attach the output of running glxinfo on your system?, e.g.
> % glxinfo > glxinfo_output.txt
Attached the the glxinfo file.

In summary the system is:
Ubuntu 18.04 LTS
Radeon Vega 64
MESA Open Source drivers that ship with Ubuntu/MESA (not the AMD AMDGPU proprietary ones)

[glxinfo_output.txt](https://github.com/PixarAnimationStudios/USD/files/3631606/glxinfo_output.txt)

davidgyu:
Thank you! That's very helpful. -David

davidgyu:
Thank you again. This information confirms that your implementation supports the core direct state access API even though it does not support EXT_direct_state_access.

So, we think a better fix is for us to update the implementation to eliminate calls to methods like glNamedBufferDataEXT altogether and instead use the core direct state access methods when available on system like yours which support OpenGL 4.5 and/or ARB_direct_state_access.

We're running that change through our tests now and we will commit that an close out this PR.

Thanks again for your help bringing this to our attention!

PR #965, opened 2019-09-16
Core Context. Added USDVIEW_ENABLE_COREGL envvar to force GL Core context

rfanner:
### Description of Change(s)
Added an environment variable, similar to MSAA, to force a GL Core Context in USDView.
### Fixes Issue(s)
- Filed earlier as Pixar-internal issue #USD-5433


jilliene:
Filed as internal issue #USD-5547

PR #966, opened 2019-09-16
MESA AMD. Added VAO bind()/unbind() to HdxCompositor to stop error in Core Context.

rfanner:
### Description of Change(s)
If running on a Core OpenGL context a VAO is needed before any vertex attrib binds can be done.
Errors seen on Mesa AMD Vega drivers under Ubuntu.
### Fixes Issue(s)
- Filed earlier as Pixar-internal issue #USD-5432

### DIFF
--- a/pxr/imaging/lib/hdx/compositor.cpp
+++ b/pxr/imaging/lib/hdx/compositor.cpp
@@ -54,7 +54,7 @@ namespace {
 
 HdxCompositor::HdxCompositor()
     : _colorTexture(0), _depthTexture(0)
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-5548

FlorianZ:
We are looking at some older issues on our backlog and decided to close this one, because the code has changed so much. We don't know for sure that this has been addressed since, but even if it's an outstanding bug still, we'd have to approach it from first principals. Sorry, and thanks for understanding.

PR #967, opened 2019-09-16
Selection. Allow OpenGL render delegates to support selection.

rfanner:
### Description of Change(s)
Removed limit on only hdSt being the only allowed render delegate for selection. Code now checks virtual IsOpenGL() method when allowing selection to happen.
Also added [Get|Set]IsRenderPassId() methods to render pass to check if it is used by the selection code.

We appreciate that there are likely changes planned to generalise support for different graphics back-ends and that this code may well overlap/clash with such future changes. In the meantime, however, this minor tweak makes it possible to support selection for other GL render delegates to support selection.
### Fixes Issue(s)
-


jilliene:
Filed as internal issue #USD-5549

PR #971, opened 2019-09-19
If Embree fails to validate the requested AOVs, mark all buffers as converged

marktucker:
If Embree fails to validate the requested AOVs, mark all buffers as 
converged so the render will exit. Also raise a runtime error indicating
that the render will not complete.

### Fixes Issue(s)
- After invoking the Embree render delegate with unsupported buffer formats (asking for 16bit float color values, for example), the delegate would never reach a converged state even though it wouldn't actually be doing any rendering.


jilliene:
Filed as internal issue #USD-5568

PR #980, opened 2019-09-23
Adds USDA_FILESIZE_WARNING setting

ablev-dwa:
### Description of Change(s)

Adds an optional filesize check to textFileFormat that will print a warning if a text layer is loaded that is larger than the specified size in megabytes. This has been really useful for early detection of layers that are accidentally delivered as ascii instead of crate.

The default is 0, which disables the check and maintains the existing behavior of textFileFormat.

### DIFF
--- a/pxr/usd/lib/sdf/textFileFormat.cpp
+++ b/pxr/usd/lib/sdf/textFileFormat.cpp
@@ -34,6 +34,7 @@
 
 #include "pxr/base/trace/trace.h"
 #include "pxr/base/tf/atomicOfstreamWrapper.h"
+#include "pxr/base/tf/envSetting.h"
 #include "pxr/base/tf/fileUtils.h"
 #include "pxr/base/tf/registryManager.h"
 #include "pxr/base/tf/staticData.h"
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-5582.

spiffmon:
Thanks for this, @ablev-dwa!  We’d have made great use of this several times already.

c64kernal:
Thanks again @ablev-dwa -- this is a great idea. We'll merge it in soon, but I wanted to mention that we had to make a few changes, in particular since this is a lower level library than USD, we had to remove mentions of USD-specific things (e.g., crate) and we changed the env var name to SDF_TEXTFILE_SIZE_WARNING_MB -- I hope that all sounds okay to you. Let us know if you have any feedback on the change once it lands in dev.  Thanks again!

ablev-dwa:
> we had to remove mentions of USD-specific things (e.g., crate) and we changed the env var name to SDF_TEXTFILE_SIZE_WARNING_MB

Ah, duh, of course 😃  If I would have put a little more thought into it, I would have done the same. All of your changes are fine by me, thanks for sorting it out on your end.

PR #982, opened 2019-09-25
Fix out of range access in Hd_SortedIds

hshakula:
### Description of Change
Validate _afterLastDeletePoint before accessing an array element pointed by it.

### Fixes Issues
- Debug assert on Windows
- Undefined behavior in release build

### How to Reproduce
```
Hd_SortedIds sortedIds;
sortedIds.Insert(SdfPath("/Test2"));
sortedIds.Insert(SdfPath("/Test1"));
sortedIds.Remove(SdfPath("/Test2"));
sortedIds.Remove(SdfPath("/Test1")); // Crash
```

### DIFF
--- a/pxr/imaging/lib/hd/sortedIds.cpp
+++ b/pxr/imaging/lib/hd/sortedIds.cpp
@@ -147,9 +147,14 @@ Hd_SortedIds::Remove(const SdfPath &id)
             if (idToRemove != lastElement) {
                 std::iter_swap(idToRemove, lastElement);
 
+                if (std::distance(idToRemove, lastElement) == 1) {
+                    // idToRemove points to the last element after pop_back()
+                    _afterLastDeletePoint = INVALID_DELETE_POINT;
+                } else {
+                    _afterLastDeletePoint = idToRemove - _ids.begin();
+                    ++_afterLastDeletePoint;
+                }
                 _ids.pop_back();
-                _afterLastDeletePoint = idToRemove - _ids.begin();
-                ++_afterLastDeletePoint;
 
                 // As we've moved an element from the end into the middle
                 // the list is now only sorted up to the place where the element


hshakula:
Here are some real-world examples that gave me the ability to reproduce the problem
[files](https://github.com/PixarAnimationStudios/USD/files/3653459/examples.zip)

Just open it in usdview and close to trigger the bug

c64kernal:
Thanks for the PR @hshakula -- have you submitted a CLA to us before? I can't find a record of one, you can find the instructions here: https://graphics.pixar.com/usd/docs/Contributing-to-USD.html

Thanks again!

hshakula:
Done, my CLA has been received

jtran56:
Filed as internal issue #USD-982.

c64kernal:
Thanks for the issue @hshakula (and the CLA :)).  The trouble here is that the existing code should be handling this, but clearly something's wrong, we're concerned that this fix is hiding some bigger issue with the code. Unfortunately, it will take us some time to debug this, or if you want to keep poking at it, please feel free to!

hshakula:
Actually I have debugged and completely figured out how this small piece of code works before committing the fix. Here is a thing I can tell you about it.

Main point: the current code has incomplete handling of the removal of the second to last Id.
[Here](https://github.com/PixarAnimationStudios/USD/blob/50c610e26ed73a3913919586f54f4e8e32e5643e/pxr/imaging/lib/hd/sortedIds.cpp#L151-L153) you can see how it's calculated:
```
_ids.pop_back();
_afterLastDeletePoint = idToRemove - _ids.begin();
++_afterLastDeletePoint;
```

In that particular case (removal of the second to last Id), before incrementing, `_afterLastDeletePoint` will always point to the last element (due to pop_back) and after incrementing it simply goes out of valid index range.

So instead of handling it locally (just after calculation), I decided to handle it globally (just when it used). Actually both ways will fix the current issue

c64kernal:
Got it -- thanks so much for the thorough investigation @hshakula -- I ran this by some other folks here and the preference is to fix it as soon as it goes bad, that way the state stays consistent throughout and is a bit more future-proof.  Would you mind changing it so that _afterLastDeletePoint gets set to INVALID_DELETE_POINT right after it becomes invalid?

hshakula:
If you don't mind I set _afterLastDeletePoint to INVALID_DELETE_POINT just before it might become invalid

c64kernal:
It's great, thanks again @hshakula -- we'll get it merged in on the next push up.

hshakula:
I've decided to validate these changes on windows and with the latest changes I got debug assertion hit "vector iterators incompatible". This is because after `_ids.pop_back()` `lastElement` is no more a valid iterator. I'm going to fix it by swapping the order of calculation `_afterLastDeletePoint ` and pop of the last element

PR #984, opened 2019-09-26
Avoid plugInfo symlink collisions

jjeziorski:
### Description of Change(s)
Improve support for plugins accessed via symlinked paths.

### Fixes Issue(s)
- Avoids collisions for plugins using symlinks


jtran56:
Filed as internal issue #USD-5592.

PR #991, opened 2019-10-04
 Fix TBB library linkage in hdEmbree

hshakula:
Same motivation and changes as in https://github.com/PixarAnimationStudios/USD/commit/a29529b44b608a4fa2dffd77826444614860ff37


jtran56:
Filed as internal issue #USD-5609.

PR #1002, opened 2019-10-17
[hdx] we shouldn't be checking an uninitialized value

elrond79:
### Description of Change(s)
I was checking the changes in 1e193cce4e2568ac66a84b1a82e18493929035b7 to verify they made https://github.com/PixarAnimationStudios/USD/pull/920 obsolete (which they did!), but happened to stumble across this.

Perhaps this was supposed to be checking something else, or there is some other step which is missing?

### Fixes Issue(s)
- OIT with no AOV results in checking against an uninitialized value.

### DIFF
--- a/pxr/imaging/lib/hdx/oitResolveTask.cpp
+++ b/pxr/imaging/lib/hdx/oitResolveTask.cpp
@@ -265,9 +265,7 @@ HdxOitResolveTask::Prepare(HdTaskContext* ctx,
     } else {
         // Without AOVs we don't know the window / screen size.
         const int oitScreenSizeFallback = 2048;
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-5637

lumonix:
There appear to be a few bugs in screenSize usage.
We compare against _screenSize, but never update that value.
The intention was to update _screenSize after we have resized.

Then compare against _screenSize (not: screenSize).

elrond79:
This was addressed in 62c22dcde39cb41ccfedab7b0cd2fff364afbd35.

Closing this PR...

PR #1012, opened 2019-10-24
Fix crash when printing an expired stage python object

fabal:
### Description of Change(s)
Test weak pointer's validity in `__repr__`.

### Fixes Issue(s)
- Crash when printing an expired stage python object


spitzak:
Might want to return "None" as a more accurate representation.

jilliene:
Filed as internal issue #USD-5661

sunyab:
FYI we modified this to return "invalid null stage" since that's consistent with the message from other invalid USD objects.

Issue #1014, opened 2019-10-25
[katana] Nested instancing breaks `geometry.instanceSource` paths


nrusch:
### Description of Issue

When multiple levels of instancing are present in a prim's hierarchy, and the prim is read in by a `PxrUsdIn` node in `as sources and instances` mode, the `geometry.instanceSource` paths generated for the second (lower) level of instancing incorrectly point to a `Masters` group under the parent instance source (which has no child groups).

I'm attaching a tarball containing the files necessary for reproduction.

### Steps to Reproduce
1. Download and extract [the attached tarball (nestedInstanceBug.tar.gz)](https://github.com/PixarAnimationStudios/USD/files/3774037/nestedInstanceBug.tar.gz)
2. Read in `threeMixers.usda` in `as sources and instances` mode, and observe that instancing works correctly.
3. Replace the stage path with `nineMixers.usda`, and observe that the instances no longer compose properly.

More specifically, if you look at the "outer" `instance source` location (at `/root/world/geo/Masters/master/variants/m0`), you will see that its children are themselves instances, which is what one would expect. However, their `geometry.instanceSource` paths incorrectly point to paths mounted under the parent `m0` location, when the `instance source` location for the nested instances is actually built under `/root/world/geo/Masters/Mixer/variants/m0`.

### Package Versions

USD 0.19.11, Katana 3.1v7

jilliene:
Filed as internal issue #USD-5665

emeroo:
Hi Nathan,

When we build out masters, we create a stub hierarchy to put them in and then re-invoke PxrUsdIn with a different "location" and "isolatePath" in order to get it to read just the prim we care about. This is causing usdInArgs->GetRootLocation() to no longer be the "original" root location and breaks computation of the master's path in the Katana scene (e.g., instead of "/root/world/geo/Masters/foo", you get something like "/root/world/geo/Masters/m0/nested/geo/Masters/foo" for a nested instance).

I would argue that the change proposed in https://github.com/PixarAnimationStudios/USD/pull/1015 shouldn't just apply to the case where "masterParentPath" is empty, but rather all cases. I might also suggest using the "location" stored in the UsdIn op args before falling back on interface.getRootLocationPath().

I am using your repro scene to test out a reworked version of PR 1015 that has this chunk of logic instead:

`
                if (!masterPath.empty())
                {
                    interface.setAttr(
                            "type", FnKat::StringAttribute("instance"));

[...]

PR #1015, opened 2019-10-25
[katana] Fix broken geometry.instanceSource attrs with nested instancing

nrusch:
### Description of Change(s)

Fixes the computation of `geometry.instanceSource` paths for nested instances (that is, `instance` locations which are themselves nested inside other `instance source` locations).

### Fixes Issue(s)
- #1014


jilliene:
Filed as internal issue #USD-5666

Issue #1017, opened 2019-10-29
HDST_DUMP_FAILING_SHADER_SOURCE debug flag does not dump attached shader source on link failure


danlea:
The HDST_DUMP_FAILING_SHADER_SOURCE debug flag is intended to dump the source for all attached shaders on link failure: https://github.com/PixarAnimationStudios/USD/commit/f553a1e328da3a6d6906f364d95ab0050bacd568

However, the debug information never makes it to stdout: `_DebugLinkSource()` returns a string that is not used (the function is being called as if it were `_DumpLinkSource()`).

https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/pxr/imaging/lib/hdSt/glslProgram.cpp#L377-L379

The trivial fix is simply to direct the return value to stdout.

jilliene:
Filed as internal issue #USD-5668

PR #1020, opened 2019-10-31
HDST_DUMP_FAILING_SHADER_SOURCE debug flag does not dump attached shader source on link failure.

rfanner:
### Description of Change(s)
Dump shader source to std::cout on link failure if relevant debug flag is set.
### Fixes Issue(s)
- Issue #1017


jilliene:
Filed as internal issue #USD-5673

c64kernal:
Thanks for the fix!

PR #1042, opened 2019-11-23
Update build script to support VS2019 building

meshula:
### Description of Change(s)

Add detection for VS2019, and logic for the new syntax in cmake for VS2019 generators. The boost toolset is detected automatically for vs2019, so it is not necessary to specify the toolset in the build script.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1041


jilliene:
Filed as internal issue #USD-5712

c64kernal:
This is great, thanks so much @meshula

PR #1043, opened 2019-11-25
Fix a bug in UdsGeomImageable::MakeVisible

marktucker:
Fix a bug in UdsGeomImageable::MakeVisible, in which the recursive setting
of the visibility attribute to "inherited" would only be done to the invisible
ancestor prim closest to the root. If other ancestors were also marked as
invisible (redundantly, but legally) the original prim would not end up
being visible.

### Description of Change(s)
Reversed the order of a conditional inside the recursive _MakeVisible function so
that the setting of the visibility attribute to "inherited" on ancestors is not short
circuited after the first ancestor has this attribute changed.

### Fixes Issue(s)
The provided test case will fail without this change. The prim will not be visible
after calling MakeVisible on it.

### DIFF
--- a/pxr/usd/usdGeom/CMakeLists.txt
+++ b/pxr/usd/usdGeom/CMakeLists.txt
@@ -116,6 +116,7 @@ pxr_test_scripts(
     testenv/testUsdGeomConsts.py
     testenv/testUsdGeomCurves.py
     testenv/testUsdGeomExtentTransform.py
+    testenv/testUsdGeomImageable.py
     testenv/testUsdGeomMesh.py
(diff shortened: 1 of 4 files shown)


jilliene:
Filed as internal issue #USD-5713

spiffmon:
Thank you for catching and fixing this, @marktucker !  Looks great - the test imports could be cleaned up; I think all you really need is Usd, UsdGeom, and unittest ?

c64kernal:
Thanks guys! I'll take care of cleaning up the imports in the test (THANK YOU for adding a test!).

Issue #1044, opened 2019-11-25
Having non-animated UsdGeomSubsets breaks animated topology through Hydra


marktucker:
### Description of Issue
If you have a mesh with animated topology, Hydra updates the topology on each frame as expected. If you add a UsdGeomSubset child to that mesh, and that subset has a non-animated definition, the update of the mesh topology through Hydra no longer works.

### Steps to Reproduce
1. Load the attached animated_topology_no_subsets.usda file into usdview.
[animated_topology_subsets.zip](https://github.com/PixarAnimationStudios/USD/files/3888840/animated_topology_subsets.zip)
2. Advancing from frame 1 to frame 2 you can see the topology changes from triangles to quads.
3. Now load animated_topology_subsets.usda, which is exactly the same mesh but with non-animated subset children.
4. Advance from frame 1 to frame 2. The geometry doesn't change in the viewport.

### Package Versions
19.11

jilliene:
Filed as internal issue #USD-5711

PR #1045, opened 2019-11-25
Fix a bug where time-invariant geometry subsets would mask time-varying topology from Hydra.

marktucker:
### Description of Change(s)
If a mesh is already known to have time-varying topology, don't bother looking at geometry subsets for animation.

The _IsVarying method will actually clear the topology dirty bit, hiding the time varying topology if there are non-animated subsets.

### Fixes Issue(s)
- #1044


jilliene:
Filed as internal issue #USD-5714

PR #1050, opened 2019-11-29
[build_usd] fix for building usd from a symlinked directory

elrond79:
[build_usd] fix for building usd from a symlinked directory

Currently (on MacOS anyway, haven't tested on linux), invoking
build_usd.py from inside a symlinked directory results in usd installing
in the incorrect location - the dependency install dir (if you've
specified a separate one, with --inst).

This fixes this by explicitly storing which dependency we're building on
the context, instead of trying to implicitly figure it out by doing a
text comparison of the current directory.

We could attempt to fix this by applying path regularization - at minimum, you'd
need to do:

```
   def regularize_path(path):
       return os.normcase(os.normpath(os.realpath(os.abspath(path))))
    if regularize_path(srcDir) == regularize_path(context.usdSrcDir)
```

However, there might still be contingencies these don't account for - comparing
directories via string comparison is a fairly unreliable technique.  A more
reliable method would be to compare, ie, inodes, but this makes it less cross-
platform... and even that is still an indirect way of answering the question
we really want to answer: what project / dependency are we currently building?

This is information that the build system itself obviously knows, so I thought
it would make more sense to just pass it through directly.

One note on why dealing correctly with symlinks might be important: I came
across this bug after upgrading MacOS to catalina... which puts restrictions on
itmes you can put at root level.  The only way around it is to use "synthetic
firmlinks", which create symlinks:

    https://apple.stackexchange.com/questions/371908/how-to-make-root-volume-writeable-again-in-catalina

...so we may be seeing a lot more usage of symlinks from MacOS users.

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -260,7 +260,7 @@ def RunCMake(context, force, extraArgs = None):
     # Create a directory for out-of-source builds in the build directory
     # using the name of the current working directory.
     srcDir = os.getcwd()
-    instDir = (context.usdInstDir if srcDir == context.usdSrcDir
+    instDir = (context.usdInstDir if context.currentDependency.name == "USD"
                else context.instDir)
     buildDir = os.path.join(context.buildDir, os.path.split(srcDir)[1])
     if force and os.path.isdir(buildDir):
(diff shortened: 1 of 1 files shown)


elrond79:
Just looked at the build failure, and I don't think it's related to my change - other PRs are failing with the same cmake error (unable to find boost), ie:

https://github.com/PixarAnimationStudios/USD/pull/1042/checks?check_run_id=317214603
https://github.com/PixarAnimationStudios/USD/pull/1045/checks?check_run_id=320134386

jilliene:
Filed as internal issue #USD-5727

Issue #1058, opened 2019-12-12
Hydra mesh adapter doesn't convey time sampled values to render delegate


marktucker:
### Description of Issue
The hydra mesh adapter ignores time sampled indices attribute values on geometry subsets.

### Steps to Reproduce
1. Load the attached USDA file into usdview:
[time_sampled_subset.zip](https://github.com/PixarAnimationStudios/USD/files/3956567/time_sampled_subset.zip)
2. Switch to a render delegate that supports material assignments to geometry subsets.
3. Note that the whole sphere is green rather than being half red, because the geometry subset is reported to the render delegate as being empty because the subset only has a time sampled value, not a default value.

### Package Versions
19.11

jilliene:
Filed as internal issue #USD-5750

PR #1059, opened 2019-12-12
Use the current UsdTimeCode to get geometry subset indices

marktucker:
When getting the indices attribute from a UsdGeomSubset, pass in the current
UsdTimeCode, otherwise the default value will always be returned (or no value
if a default value isn't explicitly authored).

### Description of Change(s)
Add the current time code to the call that loads the indices attribute value.

### Fixes Issue(s)
- #1058


c64kernal:
Nice find, @marktucker -- thank you!

jilliene:
Filed as internal issue #USD-5755

PR #1061, opened 2019-12-13
Fixed usdview warnings with valid UsdGeomMesh face-varying options

barfowl:
usdview emits warnings when UsdGeomMesh's FaceVaryingLinearInterpolation attribute is set with the valid values "cornersOnly" or "cornersPlus2", i.e.:

        Warning: in GetOptions at line 142 of src/pxr/imaging/lib/pxOsd/refinerFactory.cpp -- Unknown face-varying boundary interpolation rule (cornersOnly) (...)

### Description of Change(s)

This change adds the two missing conditions to the conversion of attribute values to corresponding OpenSubdiv options.


jilliene:
Filed as internal issue #USD-5756

PR #1065, opened 2019-12-18
[tf] add an option to treat warnings as errors

elrond79:
### Description of Change(s)

I was recently trying to debug a problem that caused a TF_WARN to be issued.  However, there was no easy way to see what line of my code was triggering that, as TF_LOG_STACK_TRACE_ON_ERROR doesn't apply for warnings.  So I made a new TF_DEBUG signal, TF_TREAT_WARNING_AS_ERROR, to allow me to get a stack trace for the warning.

### Fixes Issue(s)
- None, but might be useful for other people debugging


jilliene:
Filed as internal issue #USD-5757

superfunc:
Hah @elrond79 fwiw I wrote a very similar macro at WDAS, so +1 on the utility of this.

c64kernal:
Hey @elrond79 -- thanks so much for this PR. We discussed it a bit internally, and @gitamohr suggested that instead of this, could we simply add a TF_LOG_STACK_TRACE_ON_WARNING? That approach will get you much closer to the results that we think you're looking for, because simply turning warnings into errors might give you some surprising results. The argument that won me over to this line of thinking is that errors will raise python exceptions. So by treating warnings as errors, your program might never actually produce the same warnings as you would get under normal running conditions (i.e., treating warnings as warnings).  Whereas we think that with TF_LOG_STACK_TRACE_ON_WARNING, you'd get the stack traces you're looking for without modifying the runtime behavior.  What do you think?

elrond79:
I'd be fine with that... I probably won't bother implementing it, now, though.  Should I close this PR, then?

gitamohr:
Sure -- I'll add TF_LOG_STACK_TRACE_ON_WARNING.

elrond79:
Done!

Issue #1030, opened 2019-11-12
Build problems with clang-cl (clang on Windows)


e4lam:
I have a clang-cl project that uses a pre-built USD library (with MSVC, but that's fine because they're ABI compatible). I had to make a few source header adjustments that clang complained about:

- `pxr/base/lib/tf/preprocessorUtils.h`: Replace uses of `ARCH_OS_WINDOWS` with `ARCH_COMPILER_MSVC` because clang-cl's preprocessor is better. This fixes many compiler errors with the use of boost cpp.
- `pxr/usd/lib/sdf/fileFormat.h`: Remove `SDF_API` from "= delete" methods because it doesn't make sense to export deleted methods. I'm surprised that the other compilers allow this.
- `pxr/usd/lib/ar/asset.h`: Remove `AR_API` from "= delete" methods for same reason as previous.

I hope someone can make these fixes. Thank you!

jilliene:
Filed as internal issue #USD-5691

e4lam:
I found another problem, this time with the definitions of ARCH_CONSTRUCTOR and ARCH_DESTRUCTOR in `pxr/base/lib/arch/attributes.h` where they need to be defined in the section used by `defined(ARCH_COMPILER_CLANG)`, however, it tests for `defined(ARCH_OS_WINDOWS)` first. I just moved that whole `#elif defined(ARCH_COMPILER_GCC) || defined(ARCH_COMPILER_CLANG)` so that it's tested before the `#elif defined(ARCH_OS_WINDOWS)` and it fixed the problem. I think that perhaps the right change here is to move the GCC/CLANG section up and also change `#elif defined(ARCH_OS_WINDOWS)`  to `#elif defined(ARCH_COMPILER_MSVC)` to be more precise.

spitzak:
People do use gcc and other non-msvc compilers to compile code for Windows, so it is likely ARCH_OS_WINDOWS is the correct test for some of these.

e4lam:
@spitzak I'm certainly *not* advocating that ARCH_OS_WINDOWS be replaced with with ARCH_COMPILER_MSVC everywhere, but only in these particular cases where it seems to me that they were intended for MSVC. The particular issue in `pxr/base/lib/arch/attributes.h` deals with the understanding of particular MSVC specific pragma's and `__declspec(allocate())`. clang-cl is usually pretty good in warning about things that it doesn't understand/implemented but for some reason is just silently not doing the right thing in this case.

[...]

PR #1079, opened 2020-01-08
Fixes for calling into a USD library built with MSVC, from code built with clang-cl

marktucker:
### Description of Change(s)

Submitting these changes on behalf of @e4lam.

From a clang-cl project that uses a pre-built USD library (with MSVC, but that's fine because they're ABI compatible). A few source header adjustments that clang complained about:

* pxr/base/tf/preprocessorUtils.h: Replace uses of ARCH_OS_WINDOWS with ARCH_COMPILER_MSVC because clang-cl's preprocessor is better. This fixes many compiler errors with the use of boost cpp.
* pxr/usd/sdf/fileFormat.h: Remove SDF_API from "= delete" methods because it doesn't make sense to export deleted methods. I'm surprised that the other compilers allow this.
* pxr/usd/ar/asset.h: Remove AR_API from "= delete" methods for same reason as previous.
* pxr/base/arch/attributes.h: Fix ARCH_CONSTRUCTOR and ARCH_DESTRUCTOR. This issue only showed up when trying to use AR_DEFINE_RESOLVER() in client code and then noticing that it had failed to register the resolver because the ctor (and presumably the dtor as well) never gets called when the plugin library is loaded.

### Fixes Issue(s)
- #1030

### DIFF
--- a/pxr/base/arch/attributes.h
+++ b/pxr/base/arch/attributes.h
@@ -254,6 +254,17 @@ struct Arch_ConstructorEntry {
     };                                                                         \
     static void _name(__VA_ARGS__)
 
+#elif defined(ARCH_COMPILER_GCC) || defined(ARCH_COMPILER_CLANG)
+
+// The used attribute is required to prevent these apparently unused functions
+// from being removed by the linker.
+#   define ARCH_CONSTRUCTOR(_name, _priority, ...) \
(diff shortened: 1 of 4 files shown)


jilliene:
Filed as internal issue #USD-5792

e4lam:
Thank you, everyone!

PR #1081, opened 2020-01-09
gcc: disable placement-new warnings for boost<1.62

elrond79:
### Description of Change(s)

This is an alternative solution to the same problem addressed by:

Locally disable placement-new warnings in later versions of gcc.
898b8c1

After discussion (#1074), we decided that the best solution would be to globally disable
this for the affected combo of gcc + boost versions, since newer
versions of boost are not affected.

### Fixes Issue(s)
- Building with PXR_STRICT_BUILD_MODE + gcc-6 + boost-1.61

### DIFF
--- a/cmake/defaults/gccdefaults.cmake
+++ b/cmake/defaults/gccdefaults.cmake
@@ -24,4 +24,13 @@
 
 include(gccclangshareddefaults)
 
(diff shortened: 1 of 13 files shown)


jilliene:
Filed as internal issue #USD-5793

c64kernal:
Looks great @elrond79 -- thanks very much!

PR #1090, opened 2020-01-16
Fix mismatch between OpenSubdiv and USD preferred glew libs

meshula:
This works around the problem of glew libraries are inconsistently selected between OpenSubdiv and USD. OpenSubdiv prioritizes glews, and USD prioritizes glew. If both are installed on a system, hilarity ensues (symbols in glew cannot be resolved.)


c64kernal:
Thanks for this @meshula -- I think the Travis failure is legit? I haven't looked at it deeply, but it looks related to the PR.

meshula:
Yep, looks like the naming convention for glew is not what I thought it was under gcc.

jilliene:
Filed as internal issue #USD-5810

c64kernal:
GLEW is gone from both OpenSubdiv and USD -- happiness. Closing.

PR #1094, opened 2020-01-18
[usdedit] Don't keep a file descriptor open for the temporary usda.

asluk:
### Description of Change(s)

[usdedit] Don't keep a file descriptor open for the temporary usda.

It is not needed, and prevents usdcat from writing to it in Windows.
Also ensure write permissions when removing the temporary usda,
otherwise usdedit -n will be unable to remove it on Windows.

### Fixes Issue(s)

`usdedit extras\usd\tutorials\convertingLayerFormats\Sphere.usd`
The process cannot access the file because it is being used by another process.
Error: Failed to open file extras\usd\tutorials\convertingLayerFormats\Sphere.usd, exiting.


jilliene:
Filed as internal issue #USD-5811

asluk:
This would be another one that would be nice to get into 20.05, as it gets usdedit working on Windows.

c64kernal:
Thanks for the fix, @asluk

Issue #1092, opened 2020-01-18
[usdGenSchema] Header guard for schemaClass.h should have corresponding #endif in same file rather than being encoded in usdGenSchema.py logic


moddyz:
### Description of Issue

The `#ifndef ... #define` header guard started in [codegenTemplates/schemaClass.h](https://github.com/PixarAnimationStudios/USD/blob/v20.02-rc1/pxr/usd/usd/codegenTemplates/schemaClass.h#L24) should have a corresponding `#endif` at the end of the same file.  Currently, that `#endif` is added by [usdGenSchema.py](https://github.com/PixarAnimationStudios/USD/blob/v20.02-rc1/pxr/usd/usd/usdGenSchema.py#L932).

The use case is we prefer using `#pragma once` as the header guarding mechanism throughout our codebase.  It would be useful to be able to make this customisation purely through `codegenTemplates`.

### System Information (OS, Hardware)
ubuntu-18.04

### Package Versions
usd-19.11

jilliene:
Filed as internal issue #USD-5806

spiffmon:
Hi @moddyz !
   Moving the "terminator" logic entirely out of usdGenSchema and into schemaClass.h templates would require the addition of a new line/token to robustly bracket the custom section... and I think you might also lose the ability to add free functions. But most importantly, the implication is that it would cause a diff to every single generated schema header, which is a pretty high overhead to accommodate this need.
   However, we would support a PR that adds a command-line option to usdGenSchema , --headerTerminatorString so that you are not stuck with the hardcoded string that's in there, now, when generating your local schemas.  You may want to add a companion argument that customizes the message that accompanies the ` --(BEGIN CUSTOM CODE)--` so that your users aren't still admonished to "Close the include guard with #endif"

moddyz:
@spiffmon sounds like a solid plan 👍

PR #1097, opened 2020-01-25
[usdGenSchema] Added --headerTerminatorString optional argument.

moddyz:
### Description of Change(s)

Added `--headerTerminatorString` as an optional command-line argument to `usdGenSchema`.

This enables developers to customize the otherwise hard-coded `#endif` string which terminates the generated header files.  The use-case outlined in #1092 was for supporting developers intending to use `#pragma once` as the header guarding mechanism.

The default value of the `--headerTerminatorString` argument is `#endif` to preserve old behavior.  

### Fixes Issue(s)
- #1092

### DIFF
--- a/pxr/usd/usd/usdGenSchema.py
+++ b/pxr/usd/usd/usdGenSchema.py
@@ -877,7 +877,7 @@ def GatherTokens(classes, libName, libTokens):
 
 def GenerateCode(templatePath, codeGenPath, tokenData, classes, validate,
                  namespaceOpen, namespaceClose, namespaceUsing,
-                 useExportAPI, env):
+                 useExportAPI, env, headerTerminatorString):
     #
     # Load Templates
     #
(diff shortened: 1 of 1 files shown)


moddyz:
@spiffmon I know we discussed on @1092 that it would be also useful to a companion argument which can customize the "Close the include guard with #endif" comment.  

It looks like those comments are encoded in the [schemaClass.h template](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usd/codegenTemplates/schemaClass.h#L401)  already so the argument won't be necessary!

spiffmon:
Ah, of course, @moddyz - thanks!

spiffmon:
Change looks great, @moddyz - we'll try to get it in soon!

jilliene:
Filed as internal issue #USD-5820

PR #1098, opened 2020-01-28
[usdImaging] Fix default values for displayColor, displayOpacity, and widths

marsupial:
Changes the usdImaging delegate to deliver a proper single value instead of current return of GfVec3f[2] or float[2] when the ExtractXXX function fails for displayColor, displayOpacity, and widths.

.


jilliene:
Filed as internal issue #USD-5821

c64kernal:
Thanks for the great find, @marsupial  -- mind pulling against dev?

PR #1099, opened 2020-01-28
[usdAbc] Pass Sdf arguments to Alembic plugin.

marsupial:
# Description of Change(s)
And adds two arguments to the reader:
 **_abcReRoot_**  Read the entire hierarchy in under a single parent held in argument's value 
 **_abcLayers_** Comma separated list of secondary files to load as a layered Alembic.

### Fixes Issue(s)
- `<defaultPrim>` is a bit cumbersome for Alembics prepared without any concept of it


spiffmon:
This sounds interesting, @marsupial, and will try to review soon. But up front, can I ask you to add new tests and test files for the new behaviors?

marsupial:
I knew that was coming, so yeah, on its way

jilliene:
Filed as internal issue #USD-5822

pxr/usd/plugin/usdAbc/alembicData.h
```diff
@@ -43,7 +44,7 @@ class UsdAbc_AlembicData : public SdfAbstractData {
     /// Returns a new \c UsdAbc_AlembicData object.  Outside a successful
     /// \c Open() and \c Close() pairing, the data acts as if it contains
     /// a pseudo-root prim spec at the absolute root path.
-    static UsdAbc_AlembicDataRefPtr New();
+    static UsdAbc_AlembicDataRefPtr New(const SdfFileFormat::FileFormatArguments* = nullptr);
```
spiffmon: It introduces a couple of layers of extra complexity to change the higher-level pattern of passing FileFormatArguments by const-ref to by pointer, and I'm curious why we'd want to do it?  You're already paying for heap-allocating a copy of the std::map (which is generally going to be empty or very small), so why not eliminate all of that and the checks-for-nullptr, and just plumb the const-ref all the way down and let the member-var hold a copy?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -43,16 +43,7 @@
...
-#ifdef PXR_MULTIVERSE_SUPPORT_ENABLED
-#include <Alembic/AbcCoreGit/All.h>
-#endif // PXR_MULTIVERSE_SUPPORT_ENABLED
-
-#ifdef PXR_HDF5_SUPPORT_ENABLED
-#include <Alembic/AbcCoreHDF5/All.h>
-#endif // PXR_HDF5_SUPPORT_ENABLED
-
-#include <Alembic/AbcCoreOgawa/All.h>
+#include <Alembic/AbcCoreFactory/IFactory.h>
```
spiffmon: This is a nice simplification - thanks!

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -894,22 +878,58 @@ _ReaderContext::_ReaderContext() :
...
 {
     Close();
 
-    IArchive archive;
+    std::vector<std::string> layeredABC;
+    if (args) {
+        const auto abcLayers = args->find("abcLayers");
+        if (abcLayers != args->end()) {
+            for (auto&& l : TfStringSplit(abcLayers->second, ",")) {
+                layeredABC.emplace_back(std::move(l));
```
spiffmon: I know Alembic doesn't do any asset resolution, by-design... but would it not be valuable in the USD universe to attempt to anchor (to "filePath") and resolve each of these paths using the installed ArAssetResolver before passing the paths down to Alembic?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -894,22 +878,58 @@ _ReaderContext::_ReaderContext() :
...
-_ReaderContext::Open(const std::string& filePath, std::string* errorLog)
+_ReaderContext::Open(const std::string& filePath, std::string* errorLog,
+                     const SdfFileFormat::FileFormatArguments* args)
 {
     Close();
 
-    IArchive archive;
+    std::vector<std::string> layeredABC;
+    if (args) {
+        const auto abcLayers = args->find("abcLayers");
```
spiffmon: Although you could, I won't ask you to add the string-literals for the arg-strings into alembicUtils.h as Tokens (both because you only use each once, and because, since this is a plugin, clients can't really use the header files anyways).

**But**, please do add a new section to overview.dox that both names the new args, and gives example authored asset paths that use them.  And for bonus points, refer people to SdfLayer::CreateIdentifier() as a helper for building up such asset paths.  That will show up [here](http://graphics.pixar.com/usd/docs/api/usdabc_page_front.html).

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -894,22 +878,58 @@ _ReaderContext::_ReaderContext() :
...
+                layeredABC.emplace_back(std::move(l));
+            }
+        }
+    }
+    layeredABC.emplace_back(filePath);
+
+#if PXR_HDF5_SUPPORT_ENABLED && !H5_HAVE_THREADSAFE
+    // HDF5 may not be thread-safe.
+    using lock_guard = std::lock_guard<std::recursive_mutex>;
+    std::unique_ptr<std::lock_guard<std::recursive_mutex>> hfd5Lock(new lock_guard(*_hdf5));
```
spiffmon: Does the lock really need to be held for the call to getArchive() below?  If so, that's disappointing that Alembic doesn't take care of it itself, and _might_ lead to some performance regression in multithreaded stage-opening since we're now holding the lock while checking what the archive type is, but it might be pretty minor and the simplification is probably worth it.

If **not**, then could we start out with the unique_ptr as empty and reset it to a new lock_guard below only if we need to?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -973,8 +993,24 @@ _ReaderContext::Open(const std::string& filePath, std::string* errorLog)
...
+
     // Fill rest of the cache.
-    _ReadPrimChildren(*this, root, rootPath, *_pseudoRoot);
+    if (!abcReRoot.IsEmpty()) {
+        SdfPath compPath = rootPath.AppendChild(abcReRoot);
+        auto& xform = _prims[compPath];
+        xform.typeName = UsdAbcPrimTypeNames->Xform;
+        xform.specifier = SdfSpecifierDef;
+        _ReadPrimChildren(*this, root, compPath, xform);
+        _pseudoRoot->children.emplace_back(std::move(abcReRoot));
```
spiffmon: Will _ReadPrimChildren() set the layer's _defaultPrim_ metadata to abcReRoot?  If not, it would be really nice to make it so, here, now that we're done.

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -973,8 +993,24 @@ _ReaderContext::Open(const std::string& filePath, std::string* errorLog)
         _SetupInstancing(instances, promotable, &usedRootNames);
     }
 
+    // Re-root so the <defaultPrim> is actually the archive!
+    TfToken abcReRoot;
+    if (args) {
+        const auto reRoot = args->find("abcReRoot");
+        if (reRoot != args->end())
+            abcReRoot = TfToken(reRoot->second);
```
spiffmon: You should verify TfIsValidIdentifier() on abcReRoot, and discard and warn if not.

pxr/usd/plugin/usdAbc/alembicReader.cpp
marsupial: I've looked into this, but it's currently a bit problematic on the Alembic side to pass layers as relative (particularly in a multi-threaded env).  Second, there may be cases where the layer is in a totally separate location, unrelated to the -base layer-.

Or are you suggesting to call the resolver within usdAbc.so ?

spiffmon: I wasn't suggesting that you pass relative paths to Alembic.  Instead, use ArGetResolver().Resolve() (etc) to resolve the paths to full paths **in USD-land** before passing them on to Alembic.

But if that seems undesirable, then just please document in overview.dox that you **must** only supply full paths in the URI, which is a bit limiting for fancy resolver setups, but probably OK since this is for Alembic backwards compatibility.

Thanks!

marsupial: Ah OK, then yes (i.e. this is currently already being done on the resolver side here), I'll add the note.

pxr/usd/plugin/usdAbc/alembicData.h
marsupial: No idea why I chose that initially other than a quick way to enable disable the custom behavior.
Easy enough to change.

pxr/usd/plugin/usdAbc/alembicReader.cpp
marsupial: I don't believe that is the current behavior, as _defaultPrim_ is handled later (d332f8d).

pxr/usd/plugin/usdAbc/alembicReader.cpp
marsupial: This isn't exactly a code-path that's going to be exercised a lot here, but the problem is IFactory::getArchive may use HDF5 API's of which I'm not entirely clear are thread-safe or not.

As HDF5 support is optional in both USD and Alembic, I'd vote it's up to the client to make sure it's configured in the most optimal way (i.e. when **H5_HAVE_THREADSAFE=1** or **PXR_HDF5_SUPPORT_ENABLED=0**, there's no lock at all).

@lamiller0 suggested if you want to remove the lock entirely he'd be willing to handle it on the Alembic side...in the future...with a concrete case...whenever anyone encounters a problem.

marsupial:
Think I got to all the reviews minus the HDF5 threading issue, but if I missed any let me know

pxr/usd/plugin/usdAbc/alembicReader.cpp
spiffmon: Ok, I’m fine leaving the pattern as-is, and taking @lamiller0 up on his kind offer only if it proves problematic. We don’t even have the datasets to be able to profile in vtune, so we’d rely on you or other heavy Alembic users to keep an eye out for stage-opening performance!

Issue #1105, opened 2020-02-06
[usdGenSchema] Docstrings in generated tokens.h are incorrect if schema attribute specification defines differentiating apiName metadatum


moddyz:
### Description of Issue

Running `usdGenSchema` on a `schema.usda`, with an attribute specification which defines a differentiating apiName meta-datum, produces a `tokens.h` with incorrect docstrings referring to get attribute functions that do not exist.

The suspicion is that the `AttrInfo.name` is used instead of `AttrInfo.apiName` when writing the docstring:
```
https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usd/usdGenSchema.py#L844
```

### Steps to Reproduce
1. Run `usdGenSchema` on the following `schema.usda`:
```
#usda 1.0
(
    subLayers = [
        @usd/schema.usda@,
    ]
)

over "GLOBAL" (
    customData = {
        string libraryName       = "TestLibrary"
        string libraryPath       = "."
    }
) {
}

class "TestSchemaAPI" (
    inherits = </APISchemaBase>
    customData = {
        token apiSchemaType = "singleApply"
    }
)
{
    uniform token foo:bar = "a" (
        allowedTokens = ["a", "b", "c"]
        customData = {
            string apiName = "bar"
        }
    )
}
```
2. The generated `tokens.h` will encode the following docstrings for the token `a`:
```cpp
...
    /// \brief "a"
    /// 
    /// Possible value for TestLibraryTestSchemaAPI::GetFooBarAttr(), Default value for TestLibraryTestSchemaAPI::GetFooBarAttr()
    const TfToken a;
...
```
3. The actual get attribute function is rather `GetBarAttr()`, generated in `testSchemaAPI.h`:
```
    TESTLIBRARY_API
    UsdAttribute GetBarAttr() const;
```

### System Information (OS, Hardware)

Ubuntu-18.04

### Package Versions

usd-20.02

jilliene:
Filed as internal issue #USD-5834

PR #1107, opened 2020-02-07
[usdGenSchema] Fix incorrect get attribute docstrings being written in tokens.h

moddyz:
### Description of Change(s)

1. in `usdGenSchema.py`, use `AttrInfo.apiName` instead of `name` to write the get attribute docstrings in `token.h`.
2. With the change from above, ran `usdGenSchema` against `schema.usda` for each schema library.

### Fixes Issue(s)
- #1105


jilliene:
Filed as internal issue #USD-5841

moddyz:
Closing, as this was filed under the wrong account.

PR #1108, opened 2020-02-07
Balance HdxPrman_InteractiveContext::_Initialize and HdxPrman_InteractiveContext::End a bit better.

marsupial:
### Description of Change(s)
Balance HdxPrman_InteractiveContext::_Initialize and HdxPrman_InteractiveContext::End a bit better.
 
### Fixes Issue(s)
https://github.com/Autodesk/maya-usd loads up each delegate once to get their settings, and Prman can intermittently segfault on destruction.


jilliene:
Filed as internal issue #USD-5842

PR #1114, opened 2020-02-12
[usdAbc] Only deliver interesting timeSamples for constant properties.

marsupial:
### Description of Change(s)
Only deliver the number of _unique_ samples in an Alembic archive.
### Fixes Issue(s)
By delivering all the samples, there's a larger memory foot-print on larger scenes.
More importantly, those samples get pushed to Hydra as changes:
  **pxr/usd/plugin/usdAbc/testenv/testUsdAbcIsConstant/testUsdAbcIsConstant.abc**


jilliene:
Filed as internal issue #USD-5866

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -1689,6 +1689,16 @@ _ReaderContext::_HasValue(
...
+template <typename T> size_t
+getNumSamples(const T& object) {
+    size_t nSamples = object.getNumSamples();
+    if (!object.isConstant())
+        return nSamples;
+
+    return std::min(nSamples, size_t(1));
+}
+
 /// Fill sample times from an object with getTimeSampling() and
```
spiffmon: I am not a super-expert in this code, but I don't understand why you do not need to apply the same treatment to _GetSampleTimes() ?

pxr/usd/plugin/usdAbc/testenv/testUsdAbcIsConstant.py
```diff
@@ -0,0 +1,44 @@
...
+
+class TestUsdAbcIsConstant(unittest.TestCase):
+    def test_Read(self):
+      testFile = 'testUsdAbcIsConstant.abc'
+      stage = Usd.Stage.Open(testFile)
+      self.assertTrue(stage)
+
+      animPoints = UsdGeom.Mesh.Get(stage, "/sphere_object1")
+      self.assertTrue(animPoints)
+      self.assertEqual(animPoints.GetPrim().GetProperty('points').GetNumTimeSamples(), 240)
```
spiffmon: `UsdGeom.Tokens.points`

pxr/usd/plugin/usdAbc/alembicReader.cpp
marsupial: A bit confused, **_GetSampleTimes** will call though to **getNumSamples**.
In the case where **object.isConstant()** is true, it doesn't matter which sample you query as they'll all be valid (and the same), I just choose the lowest one.

spiffmon: Ah, so sorry, @marsupial - I let my offline editor confuse me, and thought there was a different implementation of _GetSampleTimes() that you hadn't updated.  I was wrong!

pxr/usd/plugin/usdAbc/testenv/testUsdAbcIsConstant.py
spiffmon: Oops - looks like you applied the fix to a subsequent occurrence a few lines down, but didn't change line 35?

Thanks!

marsupial: Indeed...corrected

PR #1115, opened 2020-02-13
[usdAbc] Relax type-checking on position property and let any float[3] type through.

marsupial:
### Description of Change(s)
Older Alembics could have been written with .P as a Vec3f or even float[3].

### Fixes Issue(s)
Resurrecting old/archived assets should just work with least amount of steps.


jilliene:
Filed as internal issue #USD-5867

c64kernal:
@marsupial -- wondering why we didn't update _ReadCurves and _ReadPoints? Just an oversight, or is there a reason to keep the strict matching for those?

marsupial:
No other reason than the asset that triggered this was a mesh only.
Can add the same treatment to those two if you like.

c64kernal:
Okay great, thanks @marsupial  -- no worries, we can take care of it.

PR #1119, opened 2020-02-21
Noncopyable

meshula:
### Fixes Issue(s)
-

### DIFF
--- a/pxr/imaging/glf/simpleShadowArray.cpp
+++ b/pxr/imaging/glf/simpleShadowArray.cpp
@@ -460,7 +460,7 @@ GlfSimpleShadowArray::_AllocBindlessTextures()
 
         TF_DEBUG(GLF_DEBUG_SHADOW_TEXTURES).Msg(
             "Created bindless shadow map texture of size %dx%d "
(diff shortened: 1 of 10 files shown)


pxr/imaging/glf/simpleShadowArray.cpp
```diff
@@ -460,7 +460,7 @@ GlfSimpleShadowArray::_AllocBindlessTextures()
 
         TF_DEBUG(GLF_DEBUG_SHADOW_TEXTURES).Msg(
             "Created bindless shadow map texture of size %dx%d "
-            "(id %#x, handle %#lx)\n" , size[0], size[1], id, handle);
+            "(id %#x, handle %#llx)\n" , size[0], size[1], id, handle);
```
c64kernal: I don't think this line will compile as-is on all platforms. I'll omit it from this merge for now.

jilliene:
Filed as internal issue #USD-5877

pxr/imaging/glf/simpleShadowArray.cpp
meshula: sounds good

PR #1122, opened 2020-02-27
Expose VtValue holding a TfEnum as a valid render parameter.

marsupial:
### Description of Change(s)
Expose VtValue holding a TfEnum as a valid render parameter.

### Fixes Issue(s)
Addresses a deficiency of UI for certain parameters that can be set via string or int (and ideally displayed as a list of strings).


marsupial:
Maya-Hydra implementation:
https://github.com/Autodesk/maya-usd/pull/312

jilliene:
Filed as internal issue #USD-5889

marsupial:
It's obviously been a while since I've been here (but we have been using this internally for quite some time and does some nice UI stuff).  So the following is my best attempt at an answer:

If the request is just to rename **allEnumerators** to **allValues**?  If so, more than happy to do that and you can ignore the following:

The core difference as I see it is **allValues** exists on the base-enum-type, where **allEnumerators** exists on a concrete value to inspect other possible values.

Attempting and failing to reach **allValues** on the enums we push out, makes me think that this property is needed (these enum's are runtime driven).  To get the **allValues**  property we'd have to use **TfPyWrapEnum** which would mean we'd have to make a separate enum in our delegate which would have to be kept in sync with the renderers core (vs the current implementation where we can just query a parameter for the available enum values).  The use of that class would also make python/boost a direct dependency of out delegate which is something I'd like to avoid.

Feel free to school me on a simple way I'm missing though!

sunyab:
Oh interesting, I think I understand now after looking at the PR in the Autodesk repo, but let me restate just to make sure I'm getting it:

Usually, any enum that is registered with the TfEnum system and also wrapped to Python using TfPyWrapEnum will get a `.allValues` property on both the enum class and on each of the enumerants. Picking some random enum for example, both `Usd.LoadPolicy.allValues` and `Usd.LoadWithDescendants.allValues` provides the list of all registered enumerants. However, you have enums that are registered with TfEnum, but are *not* wrapped to Python using TfPyWrapEnum, and you want to be able to get all values for those enums when you have them in Python. And that's why you needed to add a new property here.

In that case, then yes, I think I'd suggest renaming `allEnumerants` to `allValues` to be consistent. However, I am not sure if that would change the behavior for the case where the enum has been wrapped using TfPyWrapEnum. That would be something good to check on when there's time.

PR #1130, opened 2020-03-04
Trying Azure Pipelines and GitHub Actions

aloysbaillet:
### Description of Change(s)
This is *just* an example of how to build USD and run most unittests using `ASWF`'s docker images. 

This example uses the `aswf:ci-usd:2019` docker image which is purposefully made to *build* USD (it does not contain USD despite the name...) and includes most USD dependencies already installed in `/usr/local` (I believe that Alembic is not quite working yet though).

The script uses ccache to speed up most builds, which means a full build+test can take as low as 12 minutes if no code has changed (see https://github.com/aloysbaillet/USD/runs/484542692?check_suite_focus=true for an example).

One caveat is that the GitHub actions run as `root` which breaks a few USD Tf tests. We will have to change that with actions as I don't think we should run builds and tests as root in general...
Azure build can run more tests, but still a few have been excluded as we don't have good GPU integration yet. I'm expecting that we should be able to run the full test suite automatically once we have GPU machines available, which is a high priority goal for the ASWF CI working group.

Another big caveat is that this is linux only for now... Making these docker images available on other platform is also a ASWF goal, but we're not there yet. That said the mac and windows builds could keep using the `build_usd.py` on GitHub Actions for a nice github experience... 

N.B. GitHub Actions do not have `ctest` support so there's no nice test report, Azure has that feature as seen there: https://dev.azure.com/aloysbaillet/ASWF/_build/results?buildId=836&view=ms.vss-test-web.build-test-results-tab


jilliene:
Filed as internal issue #USD-5902

c64kernal:
Thanks @aloysbaillet -- super timely! We were just talking about finally rolling out Azure possibly for the next release. We are going to base it off of the work in this branch: https://github.com/c64kernal/USD/tree/dev_az

aloysbaillet:
Interesting timing indeed :-)
It might make sense then to add another job in your existing azure pipeline file that does the build+test combo from this draft. I'm happy to help with that if you think it's worth it? I'll create a PR from your `dev_az` branch in that case.

c64kernal:
Thanks @aloysbaillet -- that sounds great, but give us a bit of time to actually get it merged in and up and running in dev, and then we'd love any proposals you have against that in dev, if that's okay with you?

aloysbaillet:
Of course! No worries at all.
Please let me know when you have migrated to Azure, and don't hesitate to ask questions to the ASWF CI people as there's a growing pool of knowledge there to benefit from!

I'm going to ask though, have you considered GitHub Actions? A few projects are quite happy to have a "single platform" CI system which makes a few things easier and still provides all platforms and self-hosted pools for GPU. I'm quite neutral myself, and I know GHA still has a few rough edges, but as a recent "fork" of azure pipelines it has some benefits and is improving at a fast pace.

aloysbaillet:
Closing in favour of #1253

PR #1132, opened 2020-03-04
[extras] Don't build imaging extras if imaging is disabled.

asluk:
### Description of Change(s)

Don't try to build hdTiny if imaging is disabled.

### Fixes Issue(s)
--no-imaging fails when building hdTiny.


jilliene:
Filed as internal issue #USD-5903

c64kernal:
Nice catch, thanks @asluk!

PR #1137, opened 2020-03-11
[ar, sdf] Allow ArResolvers to declare whether they support implicit …

asluk:
[ar, sdf] Allow ArResolvers to declare whether they support implicit folder creation or not.

See https://groups.google.com/d/msg/usd-interest/9nyQ1xqJLFU/R8UnMeojEQAJ

### DIFF
--- a/pxr/usd/ar/defaultResolver.cpp
+++ b/pxr/usd/ar/defaultResolver.cpp
@@ -310,6 +310,13 @@ ArDefaultResolver::CanCreateNewLayerWithIdentifier(
     return true;
 }
 
+bool
+ArDefaultResolver::CanCreateNewFoldersIfNeeded(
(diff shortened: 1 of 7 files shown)


jilliene:
Filed as internal issue #USD-5931

spiffmon:
Hi @asluk ! 
   We think that when we tackle Ar 2.0, a goal will be making the ArResolver interface _less_ aware of filesystem concerns, whereas this change goes in the other direction.  We do want to spend a little time seeing if we can get you an intermediate-term solution - could you file an Issue for the problem discussed on the interest forum?

Thanks!
@spiffmon

asluk:
Filed https://github.com/PixarAnimationStudios/USD/issues/1148

PR #1143, opened 2020-03-14
[hdx] Report GL errors during teardown only if GL operations were actually performed.

asluk:
This isn't urgent per se, but we've noticed cases where errors get reported during hdx teardown even when gl operations are not performed. It's arguable that this change is hiding real problems, but reporting them during teardown hasn't been really helpful to developers and certainly not users, and has sometimes led to drastic measures like disabling tf diagnostics entirely.

Figured I'd float it by you all before 20.05 enters codefreeze. Thanks!

### DIFF
--- a/pxr/imaging/hdx/colorChannelTask.cpp
+++ b/pxr/imaging/hdx/colorChannelTask.cpp
@@ -70,12 +70,16 @@ HdxColorChannelTask::HdxColorChannelTask(HdSceneDelegate* delegate,
 
 HdxColorChannelTask::~HdxColorChannelTask()
 {
+    bool performedGLOperation = false;
+
     if (_texture != 0) {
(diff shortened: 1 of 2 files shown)


c64kernal:
Thanks @asluk -- we should probably hold off on this, a whole lot of work is happening in this area as we try to flesh out Hgi with multiple graphics platform support (e.g., hdx color tasks will not have direct gl calls anymore).

jilliene:
Filed as internal issue #USD-5933

spitzak:
If there were "old" GL errors, and your code did something that did not error, it would still print the "old" errors, which I think you are trying to avoid. It probably would be better to clear the GL errors before you run your code.

asluk:
Good point;thanks Bill!

Issue #1116, opened 2020-02-13
Reverse timings in clips ignored when calling UsdStage.Flatten()


benmalartre:
### Reverse Timings in clips are ignored when calling UsdStage.Flatten()

**Steps to reproduce:**

-reference an animated usd file (ex: [Maneki USD Animated Model](https://j-cube.jp/solutions/multiverse/assets/)) and author a loop with forward and reverse timing in a anim_clips.usda:

```
#usda 1.0
(
    endTimeCode = 48
    startTimeCode = 1
    defaultPrim = "maneki"
)
over "maneki" (
  prepend references = @./maneki_anim.usd@
    clips = {
      dictionary default = {
        string primPath = "/manekineko"
        asset[] assetPaths = [@./maneki_anim.usd@]
        double2[] active = [(1, 0), (24, 0), (48, 0)]
        double2[] times = [(1, 1), (24, 24), (48, 1)]
      }
    }
)
{
}
```
-open in usdview : great it works!
-in usdview, File>Save Flattened As, create a new usd file : anim_clips_flat.usd
-open in usdview : reverse timings have disappeared !

**What I have found**
It seems that the problem come from [here](https://github.com/PixarAnimationStudios/USD/blob/master/pxr/usd/usd/clip.cpp#L1286-L1347).

I can have it working by modifying [this part](https://github.com/PixarAnimationStudios/USD/blob/master/pxr/usd/usd/clip.cpp#L1300-L1329) to this:

```
    for (InternalTime t: timeSamplesInClip) {
        for (size_t i = 0; i < times.size() - 1; ++i) {
            const TimeMapping& m1 = times[i];
            const TimeMapping& m2 = times[i+1];

            // Ignore time mappings whose external time domain does not 
            // intersect the times at which this clip is active.
            InternalTime minTime = 
                (m1.internalTime <= m2.internalTime) ? 
                    m1.internalTime : m2.internalTime;
            InternalTime maxTime = 
                (m1.internalTime <= m2.internalTime) ? 
                    m2.internalTime : m1.internalTime;

            const GfInterval mappingInterval(minTime, maxTime);
            if (!mappingInterval.Intersects(clipTimeInterval)) {
                continue;
            }
            
            if (minTime <= t && t <= maxTime) {
                if (minTime == maxTime) {
                    timeSamples.insert(m1.externalTime);
                    timeSamples.insert(m2.externalTime);
                }
                else {
[...]

PR #1144, opened 2020-03-16
#1116 reverse timing authored in clips was ignored when calling UsdStage.flatten()

benmalartre:
### Description of Change(s)
file pxr/usd/usd/clip.cpp
function Usd_Clip::ListTimeSamplesForPath(const SdfPath& path)

Now uses minTime and maxTime variables and a bit of logic
instead of using sequentialTime variables.

### Fixes Issue(s)
#1116 reverse timing authored in clips was ignored when flatten UsdStage.


jilliene:
Filed as internal issue #USD-5934

sunyab:
Hi @benmalartre, I'm going to close this PR per my comments in #1116. Thanks again!

PR #1156, opened 2020-04-04
Provide a virtual destructor for PcpDynamicFileFormatInterface

marktucker:
Provide a virtual destructor for PcpDynamicFileFormatInterface to prevent
build errors with some compiler/linker combinations when building plugins
containing subclasses of this class. In particular, gcc compilers with
-Werror=non-virtual-dtor would fail without this.

### Description of Change(s)
Add an empty virtual destructor to the abstract PcpDynamicFileFormatInterface base class.


jilliene:
Filed as internal issue #USD-5972

PR #1161, opened 2020-04-09
[usdMtlx] Use Export in test instead of ExportToString

rstelzleni:
Updating this test based on conversations when we reviewed some python 3 related changes. This switches to using a layer export instead of opening a file and writing to it using ExportToString. Each baseline needed to have a new line removed from the end, but they are identical aside from that.


jilliene:
Filed as internal issue #USD-5984

PR #1164, opened 2020-04-16
Improve performance of hydra updates that involve removing point instancers

marktucker:
When removing a prim that is of interest to a point instancer, eliminate
all code that required iterating through all point instancers. This required
creating some reverse lookup maps equivalent to the _instancerData map to
let us quickly find all instancers with a particular prototype, or with a
particular parent point instancer.

### Description of Change(s)
Adding some reverse lookup maps to the point instancer adapter that allow for the elimination of a loop over all point instancers when removing a point instancer-related prim.

### Fixes Issue(s)
- I can't submit a test file, but in one example I had a scene with thousands of point instancers, and I would add a layer that marked 95% of them invisible. Hydra would then spend 8 minutes sitting in the UsdImagingPointInstancerAdapter::_ProcessPrimRemoval function. After this change, this updates take 4 seconds.


jilliene:
Filed as internal issue #USD-6002

marktucker:
After speaking to @c64kernal, this is the wrong way to go about addressing the underlying problem here. Closing.

Issue #1168, opened 2020-04-19
UsdFlattenLayerStack doesn't resolve time sampled asset paths


marktucker:
### Description of Issue
The UsdFlattenLayerStack method updates asset paths in the flattened layer (since the locations of the original layers may be different from the location of the new layer). By default relative asset paths are converted to full paths. But this asset path updating doesn't happen for time sampled asset paths (or arrays of asset paths).

I have a fix for this which I will submit as a PR shortly.

### Steps to Reproduce
1. From the attached archive, run the flattentest.py script.
[flatten_test.zip](https://github.com/PixarAnimationStudios/USD/files/4498149/flatten_test.zip)
2. Notice in flattened_default.usda, the asset paths for the texture map on the cube and the field path on the volume have been update to use full paths. In flattened.usda, where the texture map and field paths are time sampled, the asset paths have not been updated to full paths, and are no longer valid.

### Package Versions
Tested in 20.05

jilliene:
Filed as internal issue #USD-6001

sunyab:
Closing, this should be fixed in the 20.08 release. Thanks!

PR #1169, opened 2020-04-19
Fix UsdFlattenLayerStack to handle time sampled asset paths

marktucker:
### Description of Change(s)

Add code to handle time sampled asset paths or arrays of asset paths when
flattening a layer stack. This complements the existing code for fixing
single asset paths or arrays of asset paths.

### Fixes Issue(s)
- #1168

### DIFF
--- a/pxr/usd/usd/flattenUtils.cpp
+++ b/pxr/usd/usd/flattenUtils.cpp
@@ -342,6 +342,38 @@ _FixAssetPaths(const SdfLayerHandle &sourceLayer,
         val->Swap(a);
         return;
     }
+    else if (val->IsHolding<SdfTimeSampleMap>()) {
+        // Quick test that the first entry of the time sample map is holding
+        // either an asset path or an array of asset paths.
+        const SdfTimeSampleMap &tsmc = val->UncheckedGet<SdfTimeSampleMap>();
+        if (!tsmc.empty() &&
+            (tsmc.begin()->second.IsHolding<SdfAssetPath>() ||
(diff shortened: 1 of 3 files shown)


jilliene:
Filed as internal issue #USD-6003

c64kernal:
Thanks for this @marktucker -- would it be possible to get a test for this? You can potentially augment an existing test hopefully to make it easier. You don't have to amend this PR, we're getting ready to merge it, so if you're up for adding the test, getting a different PR for that would be awesome.

Thanks again! Really appreciate it.

marktucker:
Yep, I think I can turn the sample files from #1168 into a test. I'll put the test in a separate PR.

c64kernal:
That's perfect, thank you so much Mark!

spiffmon:
Hi @marktucker ,
   Apologies for slowness on getting you review comments.  Here they are:

1. I think we can iterate references in a std::map in C++-11, so why introduce a first use of TF_FOR_ALL in this file when a range-based for should do?  We are (long term) trying to deprecate TF_FOR_ALL.
2. We now have two different chunks of duplicated code in this function.. granted it's all reasonably colocated, but could we make two nice little lambda's to share?
3. Not sure how big an actual performance difference it would make, but I'd invert the loop ordering of timeSamples and value-type-checks, which would avoid an IsHolding() call for each timeSample.... if the timeSamples aren't of uniform type, we have much bigger problems.

With these changes, it might be nice to also fold the unit test into the same PR?

Cheers,
@spiffmon

marktucker:
Thanks for the suggestions. I have implemented these changes, and added the unit test in a new commit on this branch.

PR #1173, opened 2020-04-20
Multithread calls to RefreshUsdObject.

marktucker:
The goal of this commit is to multithread calls to TrackVariability in the face of a large number of changes being processed by Hydra at once. Basically trying to fulfill the comment in _RefreshUsdObject which says "PERFORMANCE: We could execute this in parallel, for large numbers of prims."

This addresses performance issues when hiding or display large number of prims. The ApplyPendingChanges call would consume the vast bulk of the time calling TrackVariability in a single thread. Hiding half your prims could take several times longer than displaying them all in the first place (depending on the number of CPUs).

### Description of Change(s)
The goal was to execute TrackVariability (invoked by RefreshUsdObject) in parallel during an update. The implementation was complicated by the intermingling of _ResyncUsdPrim calls which I believe are not thread safe with the RefreshUsdObject calls. Also, in some cases we get a large number of small changes, where multithreading within RefreshUsdObject wouldn't help. So I moved the multithreading up to ApplyPendingUpdates method. _RefreshUsdObject now just collects a bunch of work to be done, and _ResyncUsdPrim calls are similarly queued up. Then the actual work of RefreshUsdObject is done in multiple threads, followed by all the ResyncUsdPrim calls in a single thread. This is managed by a new _RefreshWorker class.

Obviously the ordering of operations is not maintained faithfully but I believe the overall work done by the single and multithreaded versions should be the same.

One obvious downside to all this is that it adds useless overhead for small numbers of changes. I'm not sure how to address this without branching the code. And this question is complicated by the fact that you can't in general tell how many changes there will be until you've already gone through the changes.

### DIFF
--- a/pxr/usdImaging/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/usdImaging/delegate.cpp
@@ -529,14 +529,6 @@ public:
         // RefreshObject doesn't apply changes to removed objects.
         _proxy->_ProcessRemovals();
     }
-
-    void Clear()
-    {
-        _tasks.clear();
-        _usdPaths.clear();
-        _changedInfoFields.clear();
-        _deferredResyncs.clear();
-    }
 };
 
 void 
@@ -909,7 +901,6 @@ UsdImagingDelegate::_ExecuteWorkForObjectRefresh(_RefreshWorker* worker)
     worker->EnableValueCacheMutations();
 
     worker->ResyncUsdPrims();
-    worker->Clear();
 }
 
 void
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-6013

marktucker:
This change should not be used because (among other issues) HdChangeTracker:MakeDirty is not thread safe. Thanks to @c64kernal and @tcauchois for pointing this out. This PR may be revived at some point in the future with some additional work, but closing it for now.

PR #1181, opened 2020-04-27
Resolve UDIM tiles path

fabal:
### Description of Change(s)
Resolve UDIM tiles path instead of using the original asset path.

### Fixes Issue(s)
- UDIM textures are not loaded when they are referenced using URLs instead of file paths.

### DIFF
--- a/pxr/usdImaging/usdImaging/textureUtils.cpp
+++ b/pxr/usdImaging/usdImaging/textureUtils.cpp
@@ -58,7 +58,7 @@ UsdImaging_GetUdimTiles(
                 layerHandle, TfStringPrintf(formatString.c_str(), t))
             : TfStringPrintf(formatString.c_str(), t);
         if (!resolver.Resolve(path).empty()) {
-            ret.emplace_back(t - startTile, TfToken(path));
+            ret.emplace_back(t - startTile, TfToken(resolver.Resolve(path)));
         }
     }
     ret.shrink_to_fit();


pxr/usdImaging/usdImaging/textureUtils.cpp
```diff
@@ -58,7 +58,7 @@ UsdImaging_GetUdimTiles(
                 layerHandle, TfStringPrintf(formatString.c_str(), t))
             : TfStringPrintf(formatString.c_str(), t);
         if (!resolver.Resolve(path).empty()) {
-            ret.emplace_back(t - startTile, TfToken(path));
+            ret.emplace_back(t - startTile, TfToken(resolver.Resolve(path)));
```
fabal, line 61: I've called `Resolve` again here as resolution happens in the context of a resolver scoped cache.

https://github.com/PixarAnimationStudios/USD/blob/0f4dfeee951f32934f08d4e6ee8efcd9cfda2f68/pxr/usdImaging/usdImaging/textureUtils.cpp#L47

spiffmon: Hi @fabal - thanks for this!  As it turns out, we are in the midst of a complete rewrite of Storm's texture manager, and in changes that I believe will be pushed out to the github dev branch shortly, we have also "resolved" this UDIM issue as part of that work!

jilliene:
Filed as internal issue #USD-6026

c64kernal:
Hi @fabal -- we think the new texture system just checked into dev should address this. Please have a look and see if there's anything else missing for you. Closing this PR. Thanks!

PR #1191, opened 2020-04-30
[usdAbc] Interpret Maya locator property properly.

marsupial:
### Description of Change(s)
Add _ReadLocator convertor to Alembic XForm objects
### Fixes Issue(s)
Maya exports additional locator information as float64_t[3][2].
Dropping this info leads to some degree of data loss, and winds up spamming the console about out-of-schema properties obscuring more interesting output.


jilliene:
Filed as internal issue #USD-6039

PR #1192, opened 2020-04-30
[usdview] Fix _reopenStage for cases when launched with --norender

marsupial:
### Description of Change(s)
Guard `self._stageView` usage and remove redundant `_stepSizeChanged` call.

### Fixes Issue(s)
Triggering a stage reload from **usdview** without an active renderer will currently fail.


jilliene:
Filed as internal issue #USD-6040

PR #1209, opened 2020-05-17
Add ExpandUSDZ Python context

dgovil:
### Description of Change(s)

This PR adds a `ExpandedUSDZ` Python Context Manager.  This context can also be used without creating a context for flexibility (similar to a File Open context).

The context expands a given `.usdz` file into a temporary directory where the files can be modified. These can then optionally be repackaged up into a usdz, either overwriting the original or in to a new location. The context then can optionally cleanup any temporary files.

This is a common task that we encounter with dealing with `.usdz` files where the files need some changes, and we see this code duplicated around the place. I figure it makes sense to have something like this as a convenience inside the main USD repo itself so people aren't writing it themselves each time.

I have some basic unit tests as well.

### Fixes Issue(s)
- None


jtran56:
Filed as internal issue #USD-6080.

PR #1212, opened 2020-05-19
[glf] Support uint16 type for GL tetxures and OpenImageIO.

marsupial:
### Description of Change(s)
Allow usage of GL_UNSIGNED_SHORT for texture creation, and support proper type conversion for OIIO writing of a GlfImage.

Make sure to use sized OIIO constants, as OpenGL's constants are sized (https://www.khronos.org/opengl/wiki/OpenGL_Type)
### Fixes Issue(s)
GLF/Hydra support 16-bit textures for `half` and `unsigned int` already, but `uint16` textures will be downgraded to 8-bit.

### DIFF
--- a/pxr/imaging/glf/baseTextureData.cpp
+++ b/pxr/imaging/glf/baseTextureData.cpp
@@ -41,6 +41,7 @@ GlfBaseTextureData::_GLInternalFormatFromImageData(
 {
     int numElements = GlfGetNumElements(format);
     switch (type) {
+    case GL_UNSIGNED_SHORT:
     case GL_UNSIGNED_INT:
         switch (numElements) {
         case 1:
--- a/pxr/imaging/glf/oiioImage.cpp
+++ b/pxr/imaging/glf/oiioImage.cpp
@@ -134,8 +134,10 @@ static GLenum
 _GLTypeFromImageData(TypeDesc typedesc)
 {
     switch (typedesc.basetype) {
-    case TypeDesc::UINT:
+    case TypeDesc::UINT32:
         return GL_UNSIGNED_INT;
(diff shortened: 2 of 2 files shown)


jtran56:
Filed as internal issue #USD-6078.

c64kernal:
Thanks for the PR, @marsupial -- we accepted this PR despite the fact that we don't internally exercise these code paths. This means that we are likely to inadvertently break it again. Please consider submitting another PR with a test case that covers your use-case for this. Also we resolved the conflict, we tried to be careful, but again there's a chance that introduced  some subtle issue. A test in a new PR would be really awesome. (Note that if the test requires imagediffs, we currently don't support that in the open source repo but we do support it internally, so even if it's a test that doesn't run meaningfully in open source, we can adapt it internally to verify correct results). Thanks again!

marsupial:
Is there a template to follow for these `imagediffs` ?
(We're starting to getting more  testing on our delegate but it would be nice to also test Storm elsewhere and don't want to be re-inventing the wheel).

c64kernal:
Here's an example of a test that internally we actually do imagediffs against that only run in open source to produce the images, but we don't actually diff against the baselines (internally we do though):

pxr/usdImaging/bin/testusdview/testenv/testUsdviewSkinning/

I'm not necessarily suggesting that you add a testusdview test though, you might be able to get away with something more direct inside of pxr/imaging/plugin/glfOiio/testenv (which would be the first one).

Come to think of it, we actually have a bunch of tests that aren't exposed to Open Source (yet) -- and in fact there are some tests that we should move to glfOiio as well.

So all of that is to say, maybe a testUsdview test is the easiest thing after all...

PR #1213, opened 2020-05-19
Fix potential crash in BakeSkinning.

cameronwhite:
The crash occurs in `_ExtendWorldTransformTimeSamples()` if the skinned prim or its parent has time-sampled transforms. Since multiple skinning adapters can share the same skel adapter, this could result in modifying the same list of time samples from multiple threads.

This change just removes the parallel for loop over the skinning adapters, but it would be possible to achieve some parallelism if extra preprocessing work is done.

To reproduce, load the attached file ([agents.usda.zip](https://github.com/PixarAnimationStudios/USD/files/4652567/agents.usda.zip)) with `usdview --numThreads 24`, and then run `UsdSkel.BakeSkinning(usdviewApi.stage.Traverse())` in the interpreter

### DIFF
--- a/pxr/usd/usdSkel/bakeSkinning.cpp
+++ b/pxr/usd/usdSkel/bakeSkinning.cpp
@@ -1885,16 +1885,16 @@ _ComputeTimeSamples(
 
     // Extend the time samples of each skel adapter with the time samples
     // of each skinning adapter.
-    WorkParallelForN(
-        skinningAdapters.size(),
-        [&](size_t start, size_t end) {   
-            for (size_t i = start; i < end; ++i) {
-                skinningAdapters[i]->ExtendTimeSamples(
-                    interval,
-                    &skelTimesMap[skinningAdapters[i]->GetSkelAdapter()]);
-            }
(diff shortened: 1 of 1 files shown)


spiffmon:
Thanks so much for this contribution, @cameronwhite !  For us to be able to consider it, we need to have a CLA in place for you, as described in: http://graphics.pixar.com/usd/docs/Contributing-to-USD.html

Thank you!
--spiff

cameronwhite:
I wasn't sure if I was already covered by the SideFX CLA, but if not I can fill that out!

spiffmon:
Ah, so if it's work done for SideFx, then I believe we need an email from
Cristin requesting that we add you to the SideFX CLA, rather than you
sending in a personal CLA.  Thanks!

--spiff

cameronwhite:
Thanks, that email has been sent!

spiffmon:
Thanks, @cameronwhite - we received it!

jtran56:
Filed as internal issue #USD-6083.

PR #1214, opened 2020-05-20
OIIO as plugin for glf

ablev-dwa:
Signed-off-by: Alan Blevins <alan.blevins@dreamworks.com>

### Description of Change(s)
Restructures OpenImageIO support as a plugin to Glf


jtran56:
Filed as internal issue #USD-6079.

Issue #1222, opened 2020-05-22
Crash in Hydra when updating a primvar value on a mesh


marktucker:
### Description of Issue
Sometimes performing an update of a primvar on a mesh can cause a crash within Hydra, which tries to access the interpolation metadata of a primvar that doesn't exist.

### Steps to Reproduce
1. From the attached archive, open primvar1.usda into usdview:
[primvarchange.zip](https://github.com/PixarAnimationStudios/USD/files/4670757/primvarchange.zip)
2. Run the following in the interpreter window:
l = Sdf.Layer.FindOrOpen('./primvar2.usda')
usdviewApi.stage.GetRootLayer().TransferContent(l)
3. usdview will crash

### Package Versions
USD 20.05

jtran56:
Filed as internal issue #USD-6088.

PR #1223, opened 2020-05-22
Fix crash when a updating a primvar

marktucker:
Fix crash when a primvar gets removed from a primitive. The call to get the
interpolation of the primvar would crash because the primvar doesn't exist.

### Description of Change(s)
Make sure the primvar exists before trying to fetch its interpolation metadata.
Avoid calling _UsdToHdInterpolation on an empty string in this scenario, as doing so would result in TF_CODING_ERROR being triggered.

### Fixes Issue(s)
- #1222


jtran56:
Filed as internal issue #USD-6089.

PR #1229, opened 2020-06-01
Fix Windows junction handling in Tf layer

wadetb:
### Description of Change(s)

The Tf layer TfIsDir, TfIsFile and TfIsLink were acting as though any file with FILE_FLAG_REPARSE_POINT is a symlink.

On Windows, reparse points can have all sorts of purposes, such as folder junctions (like mounts), or hierarchal storage systems like OneDrive.

This change ensures that a file with the reparse point flag is actually a symlink, by attempting to parse its contents, before treating it as such.

The cost is an extra GetFileAttributes call in the common codepaths of TfIsDir and TfIsFile when followSymLinks is false (the default), and attempting to parse the symlink when one is detected.

### Fixes Issue(s)

Saving a file to any folder which is a junction or a subfolder of a junction will fail.

```python
from pxr import Sdf
fn = r'c:\normal-dir\test.usda'
root_layer = Sdf.Layer.CreateNew(fn)
root_layer.Save()
fn = r'c:\junction-to-normal-dir\test.usda'
root_layer = Sdf.Layer.CreateNew(fn)
root_layer.Save()
```

The latter fails with:

```
    root_layer = Sdf.Layer.CreateNew(fn)
pxr.Tf.ErrorException:
        Error in 'pxrInternal_v0_20__pxrReserved__::SdfLayer::_WriteToFile' at line 4247 in file C:\Users\Wade\USD\pxr\usd\sdf\layer.cpp : 'Cannot create destination directory c:\junction\'
```

### DIFF
--- a/pxr/base/tf/testenv/fileUtils.cpp
+++ b/pxr/base/tf/testenv/fileUtils.cpp
@@ -234,6 +234,28 @@ TestTfIsFile()
     return true;
 }
 
+#if defined(ARCH_OS_WINDOWS)
+static bool
+TestTfJunction()
+{
+    cout << "Testing Windows junctions" << endl;
+
+    TF_AXIOM(TfMakeDir("junction-target"));
+    TF_AXIOM(system("mklink /j junction junction-target") == 0);
+    TF_AXIOM(TfTouchFile("junction/test-file"));
+    TF_AXIOM(!TfIsLink("junction"));
+    TF_AXIOM(TfIsDir("junction", false));
+    TF_AXIOM(TfIsDir("junction", true));
+    TF_AXIOM(TfIsFile("junction/test-file", false));
+    TF_AXIOM(TfIsFile("junction/test-file", true));
+    TF_AXIOM(TfDeleteFile("junction-target/test-file"));
+    (void)ArchRmDir("junction");
+    (void)ArchRmDir("junction-target");
+
+    return true;
+}
+#endif
+
 static bool
 TestTfIsWritable()
 {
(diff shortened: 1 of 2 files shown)


spiffmon:
Thank you, @wadetb !  I realize it may be difficult, but is it possible to have a Tf unit test for this functionality (even if it only provides meaningful testing on Windows)?

spiffmon:
... and sorry, could you please file a CLA, available here: http://graphics.pixar.com/usd/docs/Contributing-to-USD.html

Thanks so much!
@spiffmon

wadetb:
Sure, will do on both counts.

jtran56:
Filed as internal issue #USD-6099.

pxr/base/tf/testenv/fileUtils.cpp
```diff
@@ -234,6 +234,28 @@ TestTfIsFile()
     return true;
 }
 
+#if defined(ARCH_OS_WINDOWS)
+static bool
+TestTfJunction()
+{
+    cout << "Testing Windows junctions" << endl;
+
+    TF_AXIOM(TfMakeDir("junction_target"));
```
BrianHarris: This should be junction-target

wadetb: Good catch, thx

wadetb: Per https://dev.azure.com/PixarAnimationStudios/USD/_build/results?buildId=67&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=61b979ba-27c9-588c-740e-a675d85b471e, does the CI run the tests on Windows? I've been running the test exe manually as the recommended ctest command line is reports no tests found.

sunyab:
Closing this PR out, I believe this was addressed by #1378.

PR #1237, opened 2020-06-08
Colorize Task improvements

ablev-dwa:
### Description of Change(s)
Primary fix is to choose a colorizeTask based on the data type in the buffer, if no specific colorizeTask is found. This also changes a few of the colorizeTask methods.

### Fixes Issue(s)
Primary fix is to be able to display any AOV that a renderDelegate can produce (as long as it turns into VecNf for N=1..4).

Also some improvements to existing colorizeTask to make it easier to distinguish integer id's for prims and instances, and improves display for UDIM tiles.

### DIFF
--- a/pxr/imaging/hdx/colorizeTask.cpp
+++ b/pxr/imaging/hdx/colorizeTask.cpp
@@ -149,10 +149,15 @@ static void _colorizeST(
     {
         for (size_t i=begin; i<end; ++i) {
             float s = buffer[i*2+0];
+            float fs = floorf(s);
+            if (s == fs && fs > 0) --fs;
+            dest[i*4+0] = (uint8_t)((s - fs) * 255.0f);
             float t = buffer[i*2+1];
-            dest[i*4+0] = (uint8_t)(s * 255.0f);
-            dest[i*4+1] = (uint8_t)(t * 255.0f);
-            dest[i*4+2] = (uint8_t)(floorf(s)+floorf(t)*10); // put udim tile in blue
+            float ft = floorf(t);
+            if (t == ft && ft > 0) --ft;
+            dest[i*4+1] = (uint8_t)((t - ft) * 255.0f);
+            dest[i*4+2] = (uint8_t)((unsigned)(10 * ft + fs) * 177u); // put udim tile in blue
+            dest[i*4+3] = 255;
         }
     });
 }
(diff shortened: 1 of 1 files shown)


pxr/imaging/hdx/colorizeTask.cpp
```diff
@@ -419,6 +487,15 @@ static _Colorizer _colorizerTable[] = {
     { HdAovTokens->primId, HdFormatInt32, _colorizeId },
     { HdAovTokens->elementId, HdFormatInt32, _colorizeId },
     { HdAovTokens->instanceId, HdFormatInt32, _colorizeId },
+    { pxr::TfToken("primvars:st"), HdFormatFloat32Vec3, _colorizePrimvar }, // match _colorizeST for 3-channel st
+    // fallback converters
+    { HdPrimvarRoleTokens->none, HdFormatFloat32, _colorizeCameraDepth },
+    { HdPrimvarRoleTokens->none, HdFormatFloat32Vec2, _colorizeST },
+    { HdPrimvarRoleTokens->none, HdFormatFloat32Vec3, _float32Vec3ToDisplay },
+    { HdPrimvarRoleTokens->none, HdFormatFloat32Vec4, _float32ToDisplay },
```
hshakula, line 495: Here you are going to convert from linear to sRGB all unknown AOVs with HdFormatFloat32Vec{3/4} format. Have you considered non-color AOVs (e.g. world coordinate, velocity)?

spitzak: This is just a fallback for when it does not recognize the name of the AOV. At least for our examples most of the AOV outputs are a color (such as diffuse or specular contribution, or the lighting from a particular light).

spitzak: I suppose a better fix would be to add some indicator for the conversion wanted to the RenderBuffer by the RenderDelegate.

jtran56:
Filed as internal issue #USD-6119.

c64kernal:
Hi @ablev-dwa and @spitzak  -- are you guys using the colorize task in your own task controller somehow or are you just relying on the HdxTaskController in release? In dev colorizeTask.cpp is no longer being used and we will hopefully be removing it from the open source distribution (we're still using it internally for other projects -- at least for now). If you guys are still using it, we will keep it around. If you do get a chance to check out dev and seeing how well it works in your environment, it would be awesome to get your feedback before we get close to the 20.08 release which is creeping up on us. If dev works better for you, we'll close out this PR. Thanks!

spitzak:
We are just using whatever the default behavior for usdview to convert
images to the final textures. If this is changed all that is needed is
similar changes to the new code so it will attempt to draw any aov even if
it's name is not recognized. If it is possible to give the render delegate
more control over the conversion to a displayed image that would be great,
but it is still nice to use the same code as other renderers as it makes it
easier to compare the outputs.

ablev-dwa:
As Bill mentioned, we are only using the stock HdxTaskController. Since colorizeTask is no longer used there, there's no need to preserve it on our account, and we'll figure out the implications for AOV drawing once we deploy 20.08. It does appear that arbitrary AOV names are supported in my tests on the dev stream so I think we can build on those changes.

c64kernal:
Got it, thanks a lot for the explanations, guys. If you run into any other issues in dev, let us know!

PR #1239, opened 2020-06-10
[build] Fix Draco's check files

hshakula:
### Description of Change(s)
Replace Draco's dependency `filesToCheck` with the valid file path.  Here is [Draco Installation Log](https://github.com/PixarAnimationStudios/USD/files/4757677/dracoInstallationLog.txt), as you might see there is no `include/Draco/src/draco` directory.

### Fixes Issue(s)
- Redundant Draco dependency build calls


jtran56:
Filed as internal issue #USD-6118.

c64kernal:
Thanks for fixing this!

PR #1242, opened 2020-06-11
UsdGeomProcedural schema

ablev-dwa:
### Description of Change(s)

Adds a new schema, Procedural. This is useful for representing procedural geometry that is only generated at render time and does not otherwise have a USD schema defined to represent it on a USD stage. 

A fully supported procedural geometry in USD (hair, etc) would have one or more thoughtfully-defined schemas, Hydra adapters, and supporting pipeline mechanics. But that infrastructure may not be available in some cases. The Procedural schema is intended to back-fill support for rendering engineers, TDs, and pipeline developers that need a generic solution for passing procedural geometry though USD in the absence of a first-class schema.

An example usage of the schema might look like:
```
def Procedural "hair_0"
{
    uniform token procedural:class = "DwaHair_v2"
    
    custom asset procedural:hair_anim = @${workspace_models}/char/hiccup/hair_0/hair_0.hair@
    custom asset procedural:hair_ref = @${workspace_models}/char/hiccup/hair_0/hair_0_ref.hair@
    custom int procedural:refine_level = 2
    custom bool procedural:use_reference_positions = 1
    
    uniform token purpose = "render"
    float3[] extent = [(-0.9583198, -0.36994034, 0.8147524), (0.9581408, 0.36996114, 3.637069)]
    matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) )
    uniform token[] xformOpOrder = ["xformOp:transform"]
}
```

**procedural:class** is the only attribute defined by the schema, and this specifies the class / type / dso of the procedural geometry to create at render time.

Attributes that should be passed through to the renderer procedural are created and queried using CreateProceduralAttr / GetProceduralAttr, and are prefixed with the procedural: namespace.

Procedural is Boundable so that extents may be authored in the pipeline.

There are some downsides to Procedural. It has no inherent visualization beyond a bounding box that could be driven by extents, so a pipeline solution is necessary to support display in Storm - ModelAPI drawMode, a proxy geometry, etc. We do not have Hydra support for Procedural either, although that may change as our render delegate development advances. And finally, this concept may not completely align with USD philosophy - a fully-specified schema is more appropriate for representing novel geometry types.

In spite of all of the flaws, we still think this is a worthwhile contribution due to a common need for this, or something like it, across studios. But if it's not a good fit, we're happy to host it in the dwa_usd_plugins repo in case anyone else would find it useful.


slooper:
Great idea! ;)

jtran56:
Filed as internal issue #USD-6121.

alelordelo:
Hello @ablev-dwa, I am very interested on this!

 Is it possible currently to export Procedural controllers?

I opened up a new issue bellow:
https://github.com/PixarAnimationStudios/USD/issues/1541

nvmkuruc:
@ablev-dwa / @jesschimein Is this necessary now that there's a `UsdProc` schema domain?

jesschimein:
Hi!  Thank you for taking the time to think through and submit this PR!  As Matt noted above, we have UsdProc now and would love to focus on improving it to cover your use case.   I'm going to close this issue, but we'd welcome suggestions/contributions to UsdProc!

PR #1246, opened 2020-06-16
[usdAbc] Fix conversion of UsdGeomMesh subd interpolation options

barfowl:
### Description of Change(s)
These changes correct errors in the conversion of mesh subdivision interpolation options between UsdGeomMesh and AbcGeom's ISubD and OSubD schemas:
- the Abc reader was modified to use the correct name for Alembic's face-varying interpolation property (".faceVaryingInterpolateBoundary").
- the Abc writer was modified to include two missing legal values values when converting faceVaryingLinearInterpolation, and to write the correct default value for interpolateBoundary.
- the existing test for subdivision conversion between USD and Abc was updated to compare boundary and face-varying interpolation values after round trip conversion.

### Fixes Issue(s)
- conversion of faceVaryingLinearInterpolation from Abc ignores original values and results in the USD default.
- conversion of faceVaryingLinearInterpolation to Abc generates errors for two legal USD values.
- conversion of interpolateBoundary to Abc does not write the correct value for the USD default.

### DIFF
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -3285,7 +3285,7 @@ _ReadSubD(_PrimReaderContext* context)
         UsdGeomTokens->faceVaryingLinearInterpolation,
         SdfValueTypeNames->Token,
         _CopyFaceVaryingInterpolateBoundary(
-            context->ExtractSchema(".faceVaryingLinearInterpolation")));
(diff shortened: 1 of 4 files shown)


barfowl:
According to Azure the macOS job completed successfully after 34m.  After over 4hrs now of "In progress ..." the confirmation of that success back to GitHub appears to be hanging.

jtran56:
Filed as internal issue #USD-6167.

hughes123b:
[success   confirmed]()

PR #1253, opened 2020-06-26
Added build and tests on aswf docker image

aloysbaillet:
### Description of Change(s)
Added a GitHub action to build USD with VFX platforms 2019 (gcc-6.3.1 and python-2.7), 2020 (gcc-6.3.1 and python-3.7) and 2021 (gcc-9.3.1 and python-3.7).

Most tests (544 in total) are running fine, for now these tests are skipped as they require a GPU: `testusdview.*|testUsdRecord.*|testUsdAppUtilsFrameRecorder`.
As GPU support gets added to GitHub actions (currently in progress for OpenColorIO) we should be able to run these tests as well.

Thanks to `ccache` we also get very speedy rebuilds at around 10 minutes for the 3 builds and all tests! See this for example: https://github.com/aloysbaillet/USD/actions/runs/147985444

The existing Azure Pipeline jobs could also be converted trivially to run in GitHub actions, but that could be done independently.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,90 @@
+on: push
+
+jobs:
+  build:
+    name: Build and Test
+    strategy:
+      matrix:
+        vfx: [2019,2020,2021]
+        include:
+          # GCC-6.3.1 Python 2.7
+          - vfx: 2019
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-6195.

PR #1258, opened 2020-07-07
UsdSkelCache instancing support

ablev-dwa:
### Description of Change(s)

Small changes to allow UsdSkelCache to operate on an instanceable SkelRoot. There are matching changes to UsdSkelBakeSkinning to prevent it from attempting to bake skinning on an instance proxy below an instanceable SkelRoot, with notes to implement this later (or not? this could be a pandora's box of managing de-instancing).

Additional changes to UsdSkelImaging stop the Hydra adapter from populating if a Skeleton or SkelRoot is an instance, instance proxy, or master. UsdSkelImaging currently doesn't support instancing with UsdSkel, so this just stops it from trying, failing, and flooding error messages into the terminal.

### DIFF
--- a/pxr/usd/usdSkel/bakeSkinning.cpp
+++ b/pxr/usd/usdSkel/bakeSkinning.cpp
@@ -2259,9 +2259,16 @@ UsdSkelBakeSkinning(const UsdSkelCache& skelCache,
     // Get the stage from the first valid binding.
     UsdStagePtr stage;
     for (const auto& binding : parms.bindings) {
-        if (binding.GetSkeleton()) {
+        if (!stage && binding.GetSkeleton()) {
             stage = binding.GetSkeleton().GetPrim().GetStage();
-            break;
+        }
+        for (const auto& query: binding.GetSkinningTargets()) {
+            if (query.GetPrim().IsInstance() || query.GetPrim().IsInstanceProxy()) {
+                // TODO: support BakeSkinning for instances
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkinningQuery <%s>\n",
+                        query.GetPrim().GetPath().GetText());
+                return false;
+            }
         }
     }
     if (!stage) {
@@ -2424,6 +2431,12 @@ UsdSkelBakeSkinning(const UsdPrimRange& range, const GfInterval& interval)
     for (auto it = range.begin(); it != range.end(); ++it) {
         if (it->IsA<UsdSkelRoot>()) {
 
+            if (it->IsInstance() || it->IsInstanceProxy()) {
+                // TODO: support BakeSkinning for instances
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkelRoot <%s>\n",
+                        it->GetPath().GetText());
+            }
+
             TF_DEBUG(USDSKEL_BAKESKINNING).Msg(
                 "[UsdSkelBakeSkinning] Populating cache for <%s>\n",
                 it->GetPath().GetText());
@@ -2455,6 +2468,13 @@ UsdSkelBakeSkinning(const UsdPrimRange& range, const GfInterval& interval)
 bool
 UsdSkelBakeSkinning(const UsdSkelRoot& skelRoot, const GfInterval& interval)
 {
+    if (skelRoot.GetPrim().IsInstance() || skelRoot.GetPrim().IsInstanceProxy()) {
+        // TODO: support BakeSkinning for instances
+        TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkelRoot <%s>\n",
+                skelRoot.GetPrim().GetPath().GetText());
+        return false;
+    }
+
     UsdSkelBakeSkinningParms parms;
     // Backwards-compatibility: do not save during skinning.
     parms.saveLayers = false;
--- a/pxr/usd/usdSkel/bakeSkinning.cpp
+++ b/pxr/usd/usdSkel/bakeSkinning.cpp
@@ -2265,7 +2265,8 @@ UsdSkelBakeSkinning(const UsdSkelCache& skelCache,
         for (const auto& query: binding.GetSkinningTargets()) {
             if (query.GetPrim().IsInstance() || query.GetPrim().IsInstanceProxy()) {
                 // TODO: support BakeSkinning for instances
-                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkinningQuery <%s>\n",
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning "
+                        "for instanced SkinningQuery <%s>\n",
                         query.GetPrim().GetPath().GetText());
                 return false;
             }
@@ -2433,8 +2434,10 @@ UsdSkelBakeSkinning(const UsdPrimRange& range, const GfInterval& interval)
 
             if (it->IsInstance() || it->IsInstanceProxy()) {
                 // TODO: support BakeSkinning for instances
-                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkelRoot <%s>\n",
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for "
+                        "instanced SkelRoot <%s>\n",
                         it->GetPath().GetText());
+                return false;
             }
 
             TF_DEBUG(USDSKEL_BAKESKINNING).Msg(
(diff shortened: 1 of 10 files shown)


jtran56:
Filed as internal issue #USD-6206.

pxr/usd/usdSkel/bakeSkinning.cpp
```diff
@@ -2259,9 +2259,16 @@ UsdSkelBakeSkinning(const UsdSkelCache& skelCache,
...
     for (const auto& binding : parms.bindings) {
-        if (binding.GetSkeleton()) {
+        if (!stage && binding.GetSkeleton()) {
             stage = binding.GetSkeleton().GetPrim().GetStage();
-            break;
+        }
+        for (const auto& query: binding.GetSkinningTargets()) {
+            if (query.GetPrim().IsInstance() || query.GetPrim().IsInstanceProxy()) {
+                // TODO: support BakeSkinning for instances
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkinningQuery <%s>\n",
```
c64kernal: Please break line up so that it fits in 80-columns

pxr/usd/usdSkel/bakeSkinning.cpp
```diff
@@ -2424,6 +2431,12 @@ UsdSkelBakeSkinning(const UsdPrimRange& range, const GfInterval& interval)
     for (auto it = range.begin(); it != range.end(); ++it) {
         if (it->IsA<UsdSkelRoot>()) {
 
+            if (it->IsInstance() || it->IsInstanceProxy()) {
+                // TODO: support BakeSkinning for instances
+                TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkelRoot <%s>\n",
```
c64kernal: Same here -- we try to have a limit of 80 characters per line.

pxr/usd/usdSkel/bakeSkinning.cpp
```diff
@@ -2455,6 +2468,13 @@ UsdSkelBakeSkinning(const UsdPrimRange& range, const GfInterval& interval)
 bool
 UsdSkelBakeSkinning(const UsdSkelRoot& skelRoot, const GfInterval& interval)
 {
+    if (skelRoot.GetPrim().IsInstance() || skelRoot.GetPrim().IsInstanceProxy()) {
+        // TODO: support BakeSkinning for instances
+        TF_WARN("[UsdSkelBakeSkinning] Cannot bake skinning for instanced SkelRoot <%s>\n",
```
c64kernal: 80 column limit here too

pxr/usd/usdSkel/cache.cpp
```diff
@@ -138,11 +138,8 @@ UsdSkelCache::ComputeSkelBindings(const UsdSkelRoot& skelRoot,
     // This is done to handle inherited skel:skeleton bindings.
 
     std::vector<UsdSkelSkeleton> skelStack(1);
-    
-    // TODO: Consider traversing instance proxies at this point.
-    // But when doing so, must ensure that UsdSkelBakeSkinning, et.al.,
-    // take instancing into account.
-    const auto range = UsdPrimRange::PreAndPostVisit(skelRoot.GetPrim());
+
+    const auto range = UsdPrimRange::PreAndPostVisit(skelRoot.GetPrim(), UsdTraverseInstanceProxies());
```
c64kernal: Hey, @ablev-dwa -- we were thinking we should actually make this optional, e.g., changing the signature of this function to something like:

`ComputeSkelBindings(const UsdSkelRoot& skelRoot, std::vector<UsdSkelBinding>* bindings, bool traverseInstances=false)`

Note defaulting to false we think is important because there may be client code out there that would not expect to suddenly be traversing instance proxies, and also for clients (where for performance reasons) would not want to traverse the proxies.

Making this change might propagate to other parts of this change as well.

pxr/usd/usdSkel/cacheImpl.cpp
```diff
@@ -221,13 +221,9 @@ UsdSkel_CacheImpl::ReadScope::Populate(const UsdSkelRoot& root)
...
     std::vector<std::pair<_SkinningQueryKey,UsdPrim> > stack(1);
 
-    // TODO: Consider traversing instance proxies at this point.
-    // But when doing so, must ensure that UsdSkelBakeSkinning, et.al.,
-    // take instancing into account.
     const UsdPrimRange range =
         UsdPrimRange::PreAndPostVisit(root.GetPrim(),
-                                      UsdPrimDefaultPredicate);
-                                      // UsdTraverseInstanceProxies());
+                                      UsdTraverseInstanceProxies());
```
c64kernal: On line 259 below, in the case where we do traverse the proxies, we should update the key from 

`key.skel = skel.GetPrim()` 

to something like:
```
key.skel = skel.GetPrim();
if (key.skel.IsInstanceProxy()) {
   key.skel = key.skel.GetMaster();
}
```

pxr/usd/usdSkel/testenv/testUsdSkelCache.py
```diff
@@ -163,9 +159,9 @@ def test_InheritedSkeletonBinding(self):
...
         allBindings = cache.ComputeSkelBindings(root)
-        # Expecting two resolved bindings. This should *not* include bindings
-        # for any inactive skels.
-        self.assertEqual(len(allBindings), 2)
+        # Expecting three resolved bindings. This should *not* include bindings
+        # for any inactive skels, but does include instances
+        self.assertEqual(len(allBindings), 3)
 
         self.assertEqual(binding1.GetSkeleton().GetPrim(),
                          allBindings[0].GetSkeleton().GetPrim())
```
c64kernal: Thank you for updating the test!

pxr/usdImaging/usdSkelImaging/skelRootAdapter.cpp
```diff
@@ -63,6 +63,10 @@ UsdSkelImagingSkelRootAdapter::Populate(
     if(!TF_VERIFY(prim.IsA<UsdSkelRoot>())) {
         return SdfPath();
     }
+    if(prim.IsInstance() || prim.IsInstanceProxy() || prim.IsInMaster()) {
```
c64kernal: It might be simpler and faster to just check if instancerContext is non-null and bail out then.

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
```diff
@@ -141,6 +141,10 @@ UsdSkelImagingSkeletonAdapter::Populate(
     if(!TF_VERIFY(prim.IsA<UsdSkelSkeleton>())) {
         return SdfPath();
     }
+    if(prim.IsInstance() || prim.IsInstanceProxy() || prim.IsInMaster()) {
```
c64kernal: Same with this one, we can probably just check instancerContext?

c64kernal (commented):
Thanks so much for this great PR, @ablev-dwa.  I left review notes from me, @spiffmon, @tcauchois, and @sgustafso. Looking forward to the next iteration!

Thanks again,
George.

ablev-dwa:
Thanks @c64kernal I'll have an update for you next week.

pxr/usd/usdSkel/cache.cpp
sgustafso: Just to make things a little bit more flexible, I think instead of a bool traverseInstances, I would have this take a UsdPrimFlagsPredicate.

c64kernal:
Just a heads up @ablev-dwa -- we have some code in flight that might cause some conflicts in the files affected here. We can help resolve the conflicts, but just wanted to give you a heads up.

ablev-dwa:
No problem! Sorry I don't have an update you, I haven't been able to test my changes from these notes with Moonray yet due to some other issues taking precedence.

c64kernal:
Got it, no worries at all! Thanks Alan!

ablev-dwa:
Sorry for the delay, I've updated the PR. After thinking about the comments, it seemed to make more sense to have instancing support as a property of the cache itself, rather than piecemeal per SkelRoot, so I added an argument to the constructor. It also defaults to False, so the existing behavior of SkelCache is unchanged.

c64kernal:
Thanks so much @ablev-dwa -- looks great! we'll get this one merged into dev asap.

sgustafso:
Hey Alan, just to give you a heads up, we're close to adding a change that will allow UsdImaging to render instanced skels (albeit in an inefficient manner). As part of that, we've ended up making minor changes on top of your change to get this to play nice with UsdImaging. Specifically, we've stripped out the `_includeInstances` member, in favor of a `Usd_PrimFlagsPredicate` arg that is passed to UsdSkelCache::Populate, UsdSkelCache::ComputeSkelBinding, etc.
The reason for that change is that it allows us to utilize the predicate from the scene delegate (UsdImagingDelegate::_GetDisplayPredicate), and makes it a little easier to update state correctly when that predicate changes.

ablev-dwa:
Thanks @sgustafso ! That makes sense.

Issue #1264, opened 2020-07-13
GetExternalReferences and UpdateExternalReferences ignore value clips


marktucker:
### Description of Issue
The GextExternalReferences and UpdateExternalReferences APIs don't pay any attention to value clips. I have a need to treat value clip files the same way as sublayers/references/payloads, and so I need to create versions of these methods that incorporate the value clip information. I'll be copy/pasting these existing functions to create new functions with this additional functionality. I can either:
1. Leave these functions outside USD if you don't think they are generally useful
2. Make a pull request to incorporate these new functions into USD with new names
3. Make a pull request that incorporates this functionality into the existing methods, either optionally or always

I'm also ignoring the UsdUtilsExtractExternalReferences method in this discussion. I don't particularly need that method, but if you go with option 3 it might make sense to include this method in the discussion as well?

marktucker:
Nevermind. Turns out not to be the core of the problem I'm dealing with, and the more I think about it, the more I think clip files should be handled like other SdfAssetPath attributes/metadata, rather than dealing with them like payloads/references/sublayers.

spiffmon:
It might make sense to keep this closed, but as I was investigating, I noticed some disturbing things...
1. First, as you are aware, those two SdfLayer functions **only** process composition arcs, though neither their names nor documentation really specify that.  The UsdUtils functions handle **all** external references, which is much more useful.  @sunyab suspects those functions may have predated general asset-valued attributes and metadata!
2. UpdateExternalReference() is broken - if a layer both sublayers and references the same external layer (which is completely legit), the references won't be updated.

We can see that the Solaris bridge uses these functions, so at minimum we should fix what is broken, or deprecate the functions altogether in favor of Usd-level or UsdUtils-level functionality... the other thing that is a bit concerning is that it's unclear exactly how to update clip template metadata in these scenarios?

Btw, 20.08 now has [ComputeClipAssetPaths](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usd/clipsAPI.h#L237-L247) so that you don't need to replicate logic for expanding clip templates, though it operates on prims, not layers.

marktucker:
Yeah, we rely on those functions in some very important areas of the code. I haven't seen them fail, but maybe we just haven't seen the specific arrangement of composition you're referring to. We do have our own code for traversing a layer and updating any SdfAssetPaths it finds in attributes or metadata because we know UpdateExternalReferences doesn't provide that capability.

[...]

PR #1266, opened 2020-07-14
Fix UsdUtilsFlattenLayerStack to apply the resolveAssetPathFn to clip manifest and asset paths

marktucker:
### Description of Change(s)
In the _FixAssetPaths function, check for a clip dictionary. If found, update the manifestAssetPath and assetPaths data, also using _FixAssetPaths. Updated the FlattenLayerStack test to validate that the paths are updated. Because the clip times and asset paths are updated independently, this could be made more efficient at the expense of greater code rearrangement and complexity. But, as validated by the test, updating the clip info twice (in _ApplyLayerOffset and again in _FixAssetPaths) does work properly.

Note that this change does not apply the resolve function to the templateAssetPath metadata, because it's not totally clear how that kind of patterned path should be resolved.

### Fixes Issue(s)
- Brought up in #1264


hughes123b (approved):

jtran56:
Filed as internal issue #USD-6219.

PR #1267, opened 2020-07-17
Fix crash while destroying HdSt_TextureHandleRegistry

marktucker:
### Description of Change(s)
Move the _textureToHandlesMap to be the last data member so it is the
first to get destructed by the default destructor implementation. The
_textureToHandlesMap must be destroyed before the _textureObjectRegistry
because the destruction of the texture objects involves accessing the
_textureObjectRegistry.

This bug is unlikely to show up in builds that don't immediately write over the memory
of deleted objects. But it will crash every time on a Windows debug build when you shut
down usdview after viewing a scene with textures.


jtran56:
Filed as internal issue #USD-6215.

PR #1275, opened 2020-07-24
fix errors with PXR_OVERRIDE_PLUGINPATH_NAME and tests

pmolodo:
### Description of Change(s)

Don't hardcode PXR_PLUGINPATH_NAME in tests (turn it into a variable instead)

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1274


jtran56:
Filed as internal issue #USD-6254

PR #1280, opened 2020-07-30
Sdr dynamic array support

ablev-dwa:
### Description of Change(s)

In usd-0.20.5, SdrShaderProperty::GetTypeAsSdfType() doesn't return an array type for properties that are flagged in the metadata by "isDynamicArray". This can cause a error diagnostic in Ndr if the property has a default value that is an array.

If the code is modified to return Sdf array types for these properties, then it has to deal with the case that the property specifies both a fixed array count and the dynamic array flag. Properties like this can have a useful interpretation in Sdf : for example, a fixed-size float array of size 2 will be represented by Sdf type float2; with the dynamic array flag set this can become float2[]. If the fixed size is > 4, the type can collapse to float[].

Most of the new code is required to correctly conform default values for types like float2[].

There is an additional issue in _ConformDefaultValue(), where it calls the function IsPropertyAnAssetIdentifier(). This checks the "Widget" metadata entry : I have replaced it with a call to _IsAssetIdentifier(), which checks "IsAssetIdentifier" metadata, for consistency.

### DIFF
--- a/pxr/usd/sdr/shaderProperty.cpp
+++ b/pxr/usd/sdr/shaderProperty.cpp
@@ -144,8 +144,11 @@ namespace {
 
     // Helper to convert array types to Sdf types. Shouldn't be used directly;
     // use `_GetTypeAsSdfType()` instead
+    // If arraySize > 0 and isDynamic is true, tries to generate a dynamic array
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-6260

PR #1281, opened 2020-08-01
Fix TfHashMultimap container type for windows/macos

kxl-adsk:
The TfHashMultimap, when compiled under a non-GNU compiler, will not accept more than one item for a given key. The GCC section is using __gnu_cxx::hash_multimap at line 159 while the non-GCC is using std::unordered_map at line 361 of pxr\base\tf\hashmap.h.


jtran56:
Filed as internal issue #USD-6263

PR #1283, opened 2020-08-05
[usdUtils] don't recreate st and pref tokens every time

pmolodo:
### Description of Change(s)

This is admittedly a pretty minor performance fix - but it bothered me that we were always dynamically creating the "st" and "pref" tokens in UsdUtils.  Sort of goes against the point of the whole TfToken system.  This PR fixes this by making static tokens for these, and returning those.


jtran56:
Filed as internal issue #USD-6265

PR #1284, opened 2020-08-06
Fix extent computation plugins for simple primitive types (sphere/cube/cone/etc)

marktucker:
The extent computation functions for these basic primitive types were not passing
the time argument to the UsdAttribute.Get calls.

### Description of Change(s)
Use the timecode argument passed to extent computation plugins when fetching
attribute values from the primitive. Otherwise the default value for the
attribute would always be used. Implementations for other primitive types
already use the timecode this way.


jtran56:
Filed as internal issue #USD-6273

PR #1289, opened 2020-08-12
Fix for invalid value clip template asset path in dependency compute

timthehoff:
### Description of Change(s)
Adds a check for an empty clips directory before resolving and globbing for files.
This catches an error when an invalid templateAssetPath is specified in a Value Clip.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1288


jtran56:
Filed as internal issue #USD-6283

PR #1297, opened 2020-08-19
PrimCompositionQuery garbage collection fix

fabal:
### Description of Change(s)
When a `PrimCompositionQuery` python is created then garbage collected, the underlying c++ objects are not properly destructed. Among other things, this prevents unused layers to be discarded.

```python
from pxr import Sdf, Usd

def func():
    Usd.PrimCompositionQuery(s.GetPseudoRoot())

func()

assert not Sdf.Layer.GetLoadedLayers()
```

### Fixes Issue(s)
- Layers are not destroyed upon the garbage collection of a python `PrimCompositionQuery` object


jtran56:
Filed as internal issue #USD-6301

Issue #1317, opened 2020-09-10
Usdview: Poor framerate when moving free camera.


abau171:
### Description of Issue

In Usdview, any user manipulation of the free camera (tumble/pan/etc.) causes significant drop in apparent framerate despite the fps indicator showing the monitor's refresh rate. When animation is playing, the USD file's target framerate is very noticably reduced when the camera is manipulated.

On further inspection, render delegates are asked to render multiple frames without an actual camera update (eg. tumbling the camera results in a valid redraw followed by 2 unnecessary redraws) which explains why the discrepancy between apparent framerate and fps indicator. Camera updates also invoke unnecessary UI refreshing each frame which takes a chunk out of render time.

I have a [fix](https://github.com/abau171/USD/commit/1e0b9202afb5a4c84dd09eca82ee9a06f5809619) ready to submit pending CLA approval.

### Steps to Reproduce
1. Open any USD file in Usdview (light enough to support rendering at your monitor's refresh rate).
2. Tumble or pan the free camera and notice the movement is not as visibly smooth as it should be (despite fps indicator).
3. (Optional) With animated USD file, play the animation and tumble or pan the free camera. With a still camera, the target framerate (eg. 24fps) is hit, but with camera manipulation the framerate is clearly reduced.

### System Information (OS, Hardware)
Ubuntu 18.04.5 LTS
Python 2.7.17
PySide2 5.15.0

jtran56:
Filed as internal issue #USD-6334

PR #1321, opened 2020-09-11
[usdview] Fixed poor framerate when moving free camera.

abau171:
### Description of Change(s)
- Replaced all uses of updateGL() in StageView with plain update(), which aggregates updates to avoid spurious rendering.
- Added ability to "abort" a setting change with AbortSettingChange exception so it does not emit signals causing unnecessary work, especially during rendering (see commit message for more details).

Both of these lead to much smoother rendering when manipulating the free camera, especially with heavier scenes or during playback.

For reference, some changes from the original commit that at least Spiff looked at:
- Removed unconditional AbortSettingChange.
- Replaced updateGL() override in StageView with an update() override. Both update() and updateForPlayback() now call the plain super.update(). This was done because the updateGL() override wasn't actually being called since everything switched to update(), so moving the camera caused extra rendering during playback which slowed playback. This is fixed if everything just uses update().

### Fixes Issue(s)
- Usdview: Poor framerate when moving free camera. #1317

### DIFF
--- a/pxr/usdImaging/usdviewq/stageView.py
+++ b/pxr/usdImaging/usdviewq/stageView.py
@@ -767,7 +767,7 @@ class StageView(QtOpenGL.QGLWidget):
         self._overrideNear = value
         self.switchToFreeCamera()
         self._dataModel.viewSettings.freeCamera.overrideNear = value
-        self.updateGL()
+        self.update()
 
     @property
     def overrideFar(self):
@@ -780,7 +780,7 @@ class StageView(QtOpenGL.QGLWidget):
         self._overrideFar = value
         self.switchToFreeCamera()
         self._dataModel.viewSettings.freeCamera.overrideFar = value
-        self.updateGL()
+        self.update()
 
     @property
     def allSceneCameras(self):
@@ -976,7 +976,7 @@ class StageView(QtOpenGL.QGLWidget):
         if self._renderer:
             if self._renderer.SetRendererPlugin(plugId):
                 self._handleRendererChanged(plugId)
-                self.updateGL()
+                self.update()
                 return True
             else:
                 return False
@@ -992,7 +992,7 @@ class StageView(QtOpenGL.QGLWidget):
         if self._renderer:
             if self._renderer.SetRendererAov(aov):
                 self._rendererAovName = aov
-                self.updateGL()
+                self.update()
                 return True
             else:
                 return False
@@ -1013,7 +1013,7 @@ class StageView(QtOpenGL.QGLWidget):
     def SetRendererSetting(self, name, value):
         if self._renderer:
             self._renderer.SetRendererSetting(name, value)
-            self.updateGL()
+            self.update()
 
     def SetRendererPaused(self, paused):
         if self._renderer and (not self._renderer.IsConverged()):
@@ -1021,7 +1021,7 @@ class StageView(QtOpenGL.QGLWidget):
                 self._renderPauseState = self._renderer.PauseRenderer()
             else:
                 self._renderPauseState = not self._renderer.ResumeRenderer()
-            self.updateGL()
+            self.update()
 
     def IsPauseRendererSupported(self):
         if self._renderer:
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #USD-6338

spiffmon:
Just wanted to register a slight concern with what we're doing here - hopefully it's baseless, though it would be great (as a followup, though would have been better if we'd had one before) to add a performance test for playback.  The concern is that it was actually intentional that updateForPlayback() do an immediate draw, because the appController was trying to keep tight reins on the timing of image-generation.  That _is_ the source of the performance bug being addressed here, but it does mean there will be some pause between when updateForPlayback is called, and when the viewport actually gets refreshed, when we reenter the event loop.  That pause will likely be negligible, probably even with scenes that are light enough to hit the requested playback rate.  But I wanted to note it here, as we're reviewing the change internally.

abau171:
@spiffmon That's a valid concern; Qt could get up to anything between the update() call and the actual draw, probably even some arbitrary, expensive callbacks. I think it's worth noting that the changes from updateGL() to update() aren't actually needed with vanilla Usdview (no custom plugins) thanks to the FOV abort. A plugin could still schedule some work at a bad time though (eg. if it takes control of the free camera for some period of time).

spiffmon:
We had just discovered, independently, that changing from the vfx2018 platform to vfx2019 causes updateGl() to become more aggressive/immediate than I guess it was before, and so the updateGl() -> update() aspects of your change actually fix some serious performance regressions - we had just started considering that same approach!  So, I think we do want to keep the change, just being on the alert for other consequences.

spiffmon:
Hey @abau171 ,
   Our internal testing showed up some issues with both halves of this change (we're trying to enable image-based testing in the public repo so that the testusdview image-based tests are available to you, but we're not there yet).  

    The change to viewSettingsDataModel showed up an incompleteness in the migration from FreeCamera being an embedded part of StageView to being a free agent, and your change tickles the problem.  The issue is that refreshing the viewport should be entirely signal driven (in response to FreeCamera changes), but it is not.  If you start by removing the call to self.update() (updateGl prior to your change) in mouseMoveEvent [here](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdviewq/stageView.py#L2046), it should become clear what needs to happen.  Part of the problem is that ViewSettingsDataModel._updateFOV() is not the best method to have the FreeCamera.signalFrustumChanged signal connected to, because it is the only signal that gets emitted, for any FreeCamera change.

   So I think what needs to happen to address the issue you are trying to address here is to act instead on the comment in appController that discusses finer-grained notification for viewSettings, and maybe add either a distinct signal, or a parameter to signalViewSettingsChanged, so that if all that happened is that the FreeCamera changed, the appController response is to simply update() the StageView viewport (if it exists)?

@tallytalwar (who discovered and analyzed these issues) kindly provided a repro that doesn't even require making the change I suggested above (though I think that change is one that should be made as part of completing the FreeCamera's separation from StageView in a proper MVC fashion):

```
ac = usdviewApi._UsdviewApi__appController
ac._stageView.switchToFreeCamera()
freeCamera = ac._stageView._dataModel.viewSettings.freeCamera
freeCamera.Tumble(90, 0) # does not trigger an update of the view
ac._stageView.update() # an explicit update is required, which is a regression

```

    Another test that failed, due to the updateGl -> update change, is testusdviewLights.  Here's a usdview repro (again thanks to @tallytalwar !)  that demonstrates the problem; this one is subtler and we're not sure what the right solution is:

```
ac = usdviewApi._UsdviewApi__appController
def s(a, d):
    ac._ui.actionAmbient_Only.setChecked(a)
    ac._ambientOnlyClicked(a)
    ac._ui.actionDomeLight.setChecked(d)
    ac._onDomeLightClicked(d)
# Disable both lights
s(False, False)
# updateView called from ambientOnlyClicked
# updateView called from onDomeLightClicked
# single paintGL (all good)
# note the viewport

# enable both lights, same updateView and paintGL calls from above
s(True, True)

# disable both lights again, same updateView and paintGL calls from above
s(False, False)

# note the difference compared to first disable, ambient seems to be not off!
# ac._stageView.glDraw() # paintGL() via glDraw() fixes the render OR
# ac._stageView.update() # another update fixes the render

# Also note that as expected disabling/enabling these lights individually works, since an updateView, glDraw, paintGL is called for both.

```

Btw, no pressure, but our code complete data for 20.11, the next public release, is next Tuesday, so if you're really eager to get this in, we'd need the change by early Monday so we have time to bring it in and test.

Thanks, @abau171 !

PR #1323, opened 2020-09-14
Fix wrong usd tool executable finding on Windows

davidlatwe:
### Description of Change(s)

Refactored executable finding function for returning correct `usddiff`, `usdcat`, .. executables on Windows.

<details><summary>Outdated description</summary>

**The following infomation is misunderstood, see comments below**

Since usd bin tools already (for a long while) being using this ..

https://github.com/PixarAnimationStudios/USD/blob/71b4baace2044ea4400ba802e91667f9ebe342f0/cmake/macros/Public.cmake#L55

https://github.com/PixarAnimationStudios/USD/blob/71b4baace2044ea4400ba802e91667f9ebe342f0/pxr/usd/bin/usddiff/CMakeLists.txt#L4-L8

.. to wrap python script into executable, this part of code in `usddiff.py` and `usdedit.py` won't run on Windows and leads to wrong finding.

```python
# in _findExe
cmd = find_executable(name)
...
# already returned at this point, and it returns `usddiff`, not `usddiff.cmd`
if Windows:
    ...
```

</details>

For example, when looking for `usddiff` executable, it returns `usddiff` instead of `usddiff.cmd` and end up with this error :
```
OSError: [WinError 193] %1 is not a valid Win32 application
```

This PR fixed this.

### DIFF
--- a/pxr/usd/bin/usddiff/usddiff.py
+++ b/pxr/usd/bin/usddiff/usddiff.py
@@ -67,7 +67,7 @@ def __findExe_Win(name):
     if cmd and os.access(cmd + '.cmd', os.X_OK):
         return cmd + '.cmd'
 
-    # find_executable under Windows only returns *.EXE files (Python 3.7+)
+    # find_executable under Windows only returns *.EXE files
     # so we need to traverse PATH.
     for path in os.environ['PATH'].split(os.pathsep):
         base = os.path.join(path, name)
--- a/pxr/usd/bin/usddiff/usddiff.py
+++ b/pxr/usd/bin/usddiff/usddiff.py
@@ -51,30 +51,27 @@ def _generateCatCommand(usdcatCmd, inPath, outPath, flatten=None, fmt=None):
 
     return command
 
-def _findExe(name):
+def __findExe(name):
     from distutils.spawn import find_executable
     cmd = find_executable(name)
-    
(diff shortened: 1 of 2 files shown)


davidlatwe:
To be a bit clearly, this PR aims to remove dead code plus making the build tests 100% pass on Windows. ☺️

jtran56:
Filed as internal issue #USD-6343

davidlatwe:
Turns out the error that I encountered was because the Python I am using is from **Anaconda**, and the `find_executable` in their implementation is not restricted to only find `.exe` files, until `Python3.7+`.

In Anaconda `Python2.7,<=3.6`, it will pick up `.exe`, `.bat` or the one that has no extension. So `usdcat` will be picked up, not `usdcat.cmd`, and the error
```
OSError: [WinError 193] %1 is not a valid Win32 application
```
pops up in subprocess call. (the `%1` is actually the file path of, e.g. `usdcat`)

So, this PR only fixes for other who also use Anaconda Python 2.7 to build USD, if any.. :/
Hopes this make sense.

> just built USD in Python 2.7 & 3.7 with this change, tests passed in both build.

sunyab:
I'm going to close this PR out given the sunsetting of Python 2.7 support as of USD 22.08. Thanks!

PR #1325, opened 2020-09-15
[build_usd.py] add --toolset option

neilo-df:
### Description of Change(s)

This change adds the `--toolset` option to the build which corresponds to passing the cmake parameter `-T <toolset>` into the build, or the equivalent b2 flag `toolset=<toolset>` for any boost components that need to be built.

While the support for `--toolset` can be applied generally to any cmake generator supporting toolsets, the work here was motivated by Windows, where it is possible to have multiple versions of Visual Studio with one or more compiler toolsets.  With this change the caller can configure exactly which generator and toolset will be used.

### Fixes Issue(s)
- prior to this change there was no convenient way to compile USD with an older/alternate toolset on Windows; for example using Visual Studio 2019 but compiling for toolset v141 or earlier,

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -350,7 +350,11 @@ def RunCMake(context, force, extraArgs = None):
 
     if IsVisualStudio2019OrGreater():
         generator = generator + " -A x64"
-                
+
+    toolset = context.cmakeToolset
+    if toolset is not None:
+        toolset = '-T "{toolset}"'.format(toolset=toolset)
+
     # On MacOS, enable the use of @rpath for relocatable builds.
     osx_rpath = None
     if MacOS():
@@ -369,6 +373,7 @@ def RunCMake(context, force, extraArgs = None):
             '-DCMAKE_BUILD_TYPE={config} '
             '{osx_rpath} '
             '{generator} '
+            '{toolset} '
             '{extraArgs} '
             '"{srcDir}"'
             .format(instDir=instDir,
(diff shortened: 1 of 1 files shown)


spitzak:
I think it would be better to just make a way to pass arbitrary extra parameters to cmake from the build_usd.py script.

neilo-df:
Hi Bill, yes, I considered using the [`--build-args` option](https://github.com/PixarAnimationStudios/USD/blob/71b4baace2044ea4400ba802e91667f9ebe342f0/build_scripts/build_usd.py#L1518-L1524) and of course I tried that first, but it isn't quite that simple.

What I found is you need to modify `build_usd.py` anyway because [on Windows the boost build hardcodes the toolset](https://github.com/PixarAnimationStudios/USD/blob/71b4baace2044ea4400ba802e91667f9ebe342f0/build_scripts/build_usd.py#L726-L734).  What that logic says is, "use whatever is the newest toolset corresponding to the newest version of Visual Studio found on this machine."   The heuristic is wrong in the case where you use Visual Studio 2019 but need to target an older toolset like v141. The heuristic also ignores `--generator` which in principle can select from multiple versions of Visual Studio if more than one is available.

So following down that path was turning into a bunch more logic in the way Boost is built, and little additional clarity to the caller that they would need to match "--toolset=msvc-14.1" specifically for Boost, and "-T v141" specifically for things built with cmake.

IMHO this way seemed cleaner under the hood and offers more explicit parameterization for the caller.

meshula:
Related ~ Allow user to specify boost version  https://github.com/PixarAnimationStudios/USD/pull/347

jtran56:
Filed as internal issue #USD-6344

sunyab:
I haven't looked deeply at this PR yet but my initial reaction is that given the first-class status we've given to CMake's "generator" argument, adding "toolset" as well is consistent and avoids the issues that @neilo-df mentioned.

Issue #1329, opened 2020-09-17
UDIM resolution broken for symlinks to non-UDIM paths


marsupial:
### Description of Issue
New[ish] code to share UDIM textures when they are symlinks breaks down if they are symlinks to non-UDIM textures.
_ResolveAssetAttribute in pxr/usdImaging/materialParamUtils.cpp attempts to deduce the following:
```
/symlink/a_<UDIM>.exr => /realpath/realfile_1001.exr
/symlink/b_<UDIM>.exr => /realpath/realfile_1001.exr
```
This seems to allow sharing a common key for **/symlink/a_<UDIM>.exr**  and **/symlink/b_<UDIM>.exr**, but will break down when the symlinks point to a file that is not a UDIM tile (i.e. **/realpath/realfile_some_other_convention.exr**)

It occurs to me that this may actually be fundamentally flawed and even the solution in #1330 may be insufficient, given that there could be two UDIM paths that resolve to the same first-tile, but diverge on subsequent tiles:
```
/symlink/a_1001.exr => /realpath/realfile_1001.exr
/symlink/a_1002.exr => /realpath/another_1002.exr

/symlink/b_1001.exr => /realpath/realfile_1001.exr
/symlink/b_1002.exr => /realpath/totallydifferent_1002.exr
```
I think any resolution/transformation via symlinks in this case would make **/realpath/realfile_1001.exr** the key, for both **a_\<UDIM\>.exr** and **b_\<UDIM\>.exr** which would give incorrect results.

spitzak:
Changing the paths also makes error messages hard for users to figure out.
It would be better if the code that actually read the image did the readlink calls and determines if the image is one it has already loaded and if so reuse it.

marsupial:
I'm leaning towards @spitzak's conclusion too...all of this magic seems better suited at the ArResolver level or in the render itself.

Should #1330 be closed or changes to just remove the symlink resolution?

jtran56:
Filed as internal issue #USD-6351

sunyab:
I spent a little bit of time looking at this and I agree with you both. I've made a change internally that just removes the symlink processing, so the UDIM pattern matching is done on the authored paths and not the paths of the symlinked files. This should go out in the next push. Thanks!

PR #1330, opened 2020-09-17
Fix UDIM resolution for symlinks pointing to non-UDIM textures.

marsupial:
### Description of Change(s)
Checks that the resolved path returned, has a valid UDIM tile name.

As I was writing up the description for #1329, it occurred to me that my solution solves our problem, but the whole idea of resolving the sym-links might just be totally broken, given that there could be two UDIM paths that resolve to the same first-tile, but diverge on subsequent tiles:
```
/symlink/a_1001.exr => /realpath/realfile_1001.exr
/symlink/a_1002.exr => /realpath/another_1002.exr

/symlink/b_1001.exr => /realpath/realfile_1001.exr
/symlink/b_1002.exr => /realpath/totallydifferent_1002.exr
```
### Fixes Issue(s)
#1329

### DIFF
--- a/pxr/usdImaging/usdImaging/materialParamUtils.cpp
+++ b/pxr/usdImaging/usdImaging/materialParamUtils.cpp
@@ -190,6 +190,16 @@ _ResolveAssetAttribute(
     const std::string::size_type prefixLength =
         firstTilePath.size() - suffix.size() - UDIM_TILE_NUMBER_LENGTH;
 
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-6352

sunyab:
Per #1329, I'm closing this PR out in favor of a pending change that just removes the symlink processing altogether. Thanks!

PR #1335, opened 2020-09-22
Minor documentation fix - non-existent Car2.usda in instancing.dox

pmolodo:
### Description of Change(s)

Just happened to notice that the instancing documentation referred to "Car2.usda", which presumably should be "Car.usd", since "Car2.usda" isn't referenced anywhere else in the doc.


jilliene:
Filed as internal issue #USD-6370

Issue #1319, opened 2020-09-10
Add BlockVariantSelection to VariantSets


chrisrydalch:
There is currently no API to set a VariantSet's Selection to be an empty string, in order to effectively block weaker variant selections. Both SetVariantSelection() and ClearVariantSelection() usd the same Sdf APIs, which interpret an empty string as nothing, and don't author any specs into the layer.

After a discussion on [usd-interest](https://groups.google.com/g/usd-interest/c/F96G1S8AWUA), adding a request for a BlockVariantSelection() method.

Thanks!

jtran56:
Filed as internal issue #USD-6336

chrisrydalch:
![screenshot-area-2020-09-10-004635](https://user-images.githubusercontent.com/37886383/93689088-312a0880-fa99-11ea-80be-cb514931c5f1.png)

BTW, we'll have a PR for this, probably some time this coming week. 

Thanks!

sunyab:
PR #1340 was merged in commit fff84622 in the dev branch, marking as pending release.

tallytalwar:
Thanks, this was included in the 21.02 release.

PR #1340, opened 2020-09-25
Add BlockVariantSelection to VariantSets

chrisrydalch:
### Description of Change(s)

Adds the ability to block selections to variantsets with an empty string, overriding weaker variant selections. This is different from the existing clear selection methods, which removes the variant selections from the layer, but the selection from weaker layers can still take effect. Also adds unit tests for both variant selection blocking and clearing.

### Fixes Issue(s)
- #1319


jilliene:
Filed as internal issue #USD-6374

sunyab:
This was merged in commit fff84622, closing this out. Thanks!

PR #1360, opened 2020-10-16
Fix Python binding for CoordSysAPI.HasLocalBindings

ablev-dwa:
### Description of Change(s)
Fixes the Python binding for CoordSysAPI.HasLocalBindings to use UsdShadeCoordSysAPI::HasLocalBindings instead of UsdShadeCoordSysAPI::Bind. Probably just a copy+paste oversight that made it to production.

### Fixes Issue(s)
In Python, UsdShade.CoordSysAPI.HasLocalBindings() does not work as expected.


jilliene:
Filed as internal issue #USD-6419

Issue #1377, opened 2020-10-31
USD handles Windows mount points/junctions incorrectly


marktucker:
### Description of Issue
When trying to write files to locations that are under Windows mount points (junctions), the USD path handling code in Tf and Arch has issues that cause failures.

### Steps to Reproduce
1. Open a command prompt as an administrator.
2. Assuming a machine has two drives, C and D, create a folder D:\bar.
3. From C:\, run 'mklink /J foo D:\bar\sub'
4. Start python and run:
from pxr import Usd
s = Usd.Stage.CreateInMemory()
s.Export('C:/foo/sub/out.usda')
Produced this error: Insufficient permissions to write to destination directory 'C:\sub'
s.Export('C:/foo/out.usda')
Produced this error: Cannot create path to write 'c:/foo/sub/out.usda'

Note that these are two different errors, indicating two different bugs.

### System Information (OS, Hardware)
Windows

jilliene:
Filed as internal issue #USD-6462

sunyab:
Closing old issue, this was fixed by #1378.

PR #1378, opened 2020-10-31
Fixes to handling of Windows symbolic links and mount points

marktucker:
### Description of Change(s)
Two fixes to low level handling of symbolic links and mount points on Windows. One handles Windows "Junction" mount points the same way symbolic links are currently handled (junctions are able to span between local drives, whereas symbolic links can only point to the same drive). The other change causes Tf_HasAttribute in tf/fileUtils.cpp to treat paths with trailing slashes the same way lstat handles them on Linux, which is that is resolves the link (even though the same path without the trailing slash would cause lstat to return information about the link itself, not the destination of the link).

I'm not sure how to submit a regression test for this fix since any such test would require setting up a mount point between two drives on a Windows machine (which requires administrative priveleges and knowing that two separate drives exist onthe machine).

### Fixes Issue(s)
- #1377

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -1178,6 +1178,24 @@ std::string ArchReadLink(const char* path)
 
             return str;
         }
+        else if (reparse->ReparseTag == IO_REPARSE_TAG_MOUNT_POINT) {
+            const size_t length =
+                reparse->MountPointReparseBuffer.PrintNameLength /
+                                                                sizeof(WCHAR);
+            std::unique_ptr<WCHAR[]> reparsePath(new WCHAR[length + 1]);
+            wcsncpy(reparsePath.get(),
+              &reparse->MountPointReparseBuffer.PathBuffer[
+              reparse->MountPointReparseBuffer.PrintNameOffset / sizeof(WCHAR)
+              ], length);
+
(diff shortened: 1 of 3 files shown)


gonewest818:
See also #1229

marktucker:
I wasn't aware of that commit. But I see several issues with it. For one, it doesn't handle the "trailing slash on a directory should resolve the symlink" issue. It also requires running additional tests (which aren't necessary in this PR). In general I think #1229 adds more complications to the logic to make the higher level functions work on Windows in more cases, whereas this PR attempts to make the existing lowest level Windows methods simply match the Linux low level method behavior, which corrects the high level behavior without any additional logic.

The test code looks like something I could adopt into this PR. It might be worth making this a platform-agnostic test too, where only the "system" command differs between Windows and other platforms... I'll try adding such a test.

jilliene:
Filed as internal issue #USD-6463

DougRogers-DigitalFish:
I think the low level functions mentioned in #1229 should be updated as well.  They are used in other parts for the code.
ArchReadLink should be updated as well, but think there is a more straightforward way to implement it, though. Forcing the symbolic evaluation seems seems fine if Linux systems work that way.   I need to do some more testing, but I should have something to evaluate very soon.

Issue #1364, opened 2020-10-20
Animated extents don't cause animated bounding boxes in Bounds draw mode


marktucker:
### Description of Issue
As discussed in https://groups.google.com/u/1/g/usd-interest/c/oqt13R4wDdo/m/zkScXGwkCQAJ, even if a prim has an animated extent or extentsHint attribute, the bounds draw mode does not animate the bounding box. But in this limited case (ignoring more expensive cases where the bounding box must be calculated from all child prims), it seems wirth while to animate the bounding box created by the draw mode adapter.

### Steps to Reproduce
1. Open the attached animated_extents_bbox.usda file in usdview.
[animated_extents.zip](https://github.com/PixarAnimationStudios/USD/files/5408848/animated_extents.zip)
2. Unload the /payload payload (you may need to enable the drawing of bounding boxes for unloaded payloads).
3. Scrub the play bar. Note that the bounding boxes do not change size, even though the sphere prim has an animated extent attribute, and the unloaded payload has an animated extentsHint attribute.

PR to address this will be incoming shortly.

jilliene:
Filed as internal issue #USD-6429

PR #1380, opened 2020-11-02
Regression test for displaying animated bounding boxes in the bounds draw mode

marktucker:
Regression test for displaying animated bounding boxes in the bounds draw
mode when a prim has an animated extent or extentsHint attribute.

### Fixes Issue(s)
- Tests issue reported in #1364 and fixed by #1365


jilliene:
Filed as internal issue #USD-6465

c64kernal:
This is really great, thanks so much @marktucker

PR #1389, opened 2020-11-10
File utilities fixes for Windows junctions and symbolic links

neilo-df:
### Description of Change(s)

This draft pull request addresses the concerns regarding Windows-specific filesystem support raised in #1377  as well as #1229.  The intention is not to create confusion with multiple pull requests but rather, hopefully to consolidate and cover the issues in a unified way.

* in `pxr/base/tf/fileUtils.cpp`: The low level file functions `TfIsDir`, `TfIsLink` and `TfIsFile` have been updated to support junctions and symbolic links in Windows.  A new function `getFileType` is added to identify all Windows file and directory types, and this is now what the low level functions call to determine what what type of object it is.

* in `pxr/base/arch/fileSystem.cpp`: The function `ArchReadLink` is updated to fetch a symbolic link properly in Windows and return the handle to the target, which is the desired behavior.

* updated unit tests to add a Windows junction test and symbolic link test.

### Fixes Issue(s)
- This PR should supercede #1229 , per request from @wadetb we're helping on this issue
- This PR addresses #1377 , except for the trailing slash behavior, which was implemented in #1378 

@dougrogers-digitalfish
@wadetb
@marktucker

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -1109,78 +1109,28 @@ int ArchFileAccess(const char* path, int mode)
     return result ? 0 : -1;
 }
 
-// https://msdn.microsoft.com/en-us/library/windows/hardware/ff552012.aspx
-
-#define MAX_REPARSE_DATA_SIZE  (16 * 1024)
-
-typedef struct _REPARSE_DATA_BUFFER {
-    ULONG   ReparseTag;
-    USHORT  ReparseDataLength;
-    USHORT  Reserved;
-    union {
-        struct {
-            USHORT  SubstituteNameOffset;
-            USHORT  SubstituteNameLength;
-            USHORT  PrintNameOffset;
-            USHORT  PrintNameLength;
-            ULONG   Flags;
-            WCHAR   PathBuffer[1];
-        } SymbolicLinkReparseBuffer;
-        struct {
-            USHORT  SubstituteNameOffset;
-            USHORT  SubstituteNameLength;
-            USHORT  PrintNameOffset;
-            USHORT  PrintNameLength;
-            WCHAR   PathBuffer[1];
-        } MountPointReparseBuffer;
-        struct {
-            UCHAR  DataBuffer[1];
-        } GenericReparseBuffer;
-    };
-} REPARSE_DATA_BUFFER, *PREPARSE_DATA_BUFFER;
 
 std::string ArchReadLink(const char* path)
 {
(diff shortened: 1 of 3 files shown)


marktucker:
I'm not clear why you are intentionally making TfIsDir return different values under Windows than under Linux... Is there a python/Usd-level test case you can provide demonstrating bad behavior that this Windows/Linux discrepancy fixes? Having to execute different unit test cases for TfIsDir seems like a red flag to me.

Your ReadArchLink implementation certainly seems cleaner, especially if it covers more exotic reparse point types. Though it looks like it changes the behavior significantly by returning a non-empty string for any file that exists, not just reparse points. Am I misreading that?

Finally, it looks like Tf_HasAttribute is now used only by TfPathExists, so it probably would make sense to change TfPathExists to use getFileType as well, and eliminate Tf_HasAttribute completely?

DougRogers-DigitalFish:
Thank you for looking at this.

That is a fair comment and I agree that the tests should be the same on all platforms.   The question is: "Should a junction or symbolic link to a directory return that it is a directory when the resolve flag is not set."   This implementation return true for this case.   This could be changed to check the input path for a separator at the end of the path and only set the directory flag if it is present.  Do you concur?

TfPathExists can be updated as well.  Good catch.

marktucker:
> This could be changed to check the input path for a separator at the end of the path and only set the directory flag if it is present. Do you concur?

This is exactly what #1378 does, so yes, I concur :) I still question the need for the relatively major changes you've made here. AFAICT #1378 accomplishes exactly the same outward-facing results (and I'm a pragmatist, so this is really all that matters to me), but with only two lines of code. But of course it's for the Pixar folks to make this decision. And this is only for the fileUtils change. The conflicting ArchReadLink changes are a separate issue.

c64kernal:
Thanks all. Just wanted to note that we're getting very close to merging #1378 and once that's done and in the dev branch, please feel free to suggest any additional fixes on top of that one. Hope that sounds okay to everyone.

DougRogers-DigitalFish:
@marktucker The low level functions are called elsewhere in the code base so I think they need to be updated in addition to the ArchReadLink changes.

@c64kernal This change is a superset of 1378.  It is a simpler implementation of ArchReadLink and updates the file utilities as well.

DougRogers-DigitalFish:
@marktucker  I understand now what you are referring to.  Are you saying that since the low level functions call ArchReadLink m , that is the only function that needs to be updated?  That could be the case.

@c64kernal It is definitely worth merging  #1378.   We will post an update after this lands.

jilliene:
Filed as internal issue #USD-6480

Issue #1391, opened 2020-11-13
Dome light on near-horizontal surfaces and texture wrapping


tdelame:
### Description of Issue
Dome light on near-horizontal surfaces create some artefacts because:
- GetTexCoords() in `domeLight.glslfx` should apply an offset of (0.5, 0.5) to compute samples at pixel centers. Otherwise, samples are computed at pixel corners and the bottom line of the irradiance map is incorrect (see below):
![buggy_irradiance](https://user-images.githubusercontent.com/12368589/99086037-4454e280-25c9-11eb-89ed-2c78ab25e231.png)
Applying the offset produces the following irradiance map:
![fixed_irradiance](https://user-images.githubusercontent.com/12368589/99086308-a01f6b80-25c9-11eb-914c-17587d5cb16e.png)
- texture samplers should have HdWrapClamp mode for T and not HdWrapRepeat. Otherwise when the surface is near-horizontal, there can be a discontinuity producing this image
![buggy_render](https://user-images.githubusercontent.com/12368589/99086483-e2e14380-25c9-11eb-87dc-638db65acf7d.png)
instead of this one
![fixed_render](https://user-images.githubusercontent.com/12368589/99086641-158b3c00-25ca-11eb-87ea-81c52737a558.png)
- normals should be clamped in [-1,1] after being transformed to world coordinates before fetching texels in dome light textures, otherwise there can be artefacts on near-horizontal surfaces depending on the view point

### Steps to Reproduce
1. Open the following file (a floor tile from the Kitchen set with a UsdPreviewSurface shader applied to it) and render it
```
#usda 1.0

def Xform "root" {
  matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) )
  uniform token[] xformOpOrder = ["xformOp:transform"]

  def Mesh "object" {
    float3[] extent = [(-9.618811, -16.597124, -0.9819515), (24.007746, 17.029436, 0.6106356)]
[...]

PR #1392, opened 2020-11-13
Dome light preview surface

tdelame:
### Description of Change(s)
Fix artefact on near-horizontal surfaces when using dome lights in previewSurface shader

### Fixes Issue(s)
- #1391


jilliene:
Filed as internal issue #USD-6483

PR #1393, opened 2020-11-14
Add version for macOS

makayla-k:
### Description of Change(s)
While building USD on MacOs I ran into errors that I had to debug for several hours. Posting my errors in the google forum I came across others who had the same issue. Attempting to fix theses issues with the help of the forum I found out that the error was with my MacOs operating system version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,7 @@ Supported Platforms
 USD is currently supported on Linux platforms and has been built and tested
 on CentOS 7 and RHEL 7.
 
-We are actively working on porting USD to both Windows and Mac platforms. 
-Support for both platforms should be considered experimental at this time.
-Currently, the tree will build on Mac (macOS version 10.14 and above) and Windows, 
-but only limited testing has been done on these platforms.
+We are actively working on porting USD to both Windows and Mac platforms (Please see [Versions.md](VERSIONS.md) for supported software versions). Support for both platforms should be considered experimental at this time. Currently, the tree will build on Mac and Windows, but only limited testing has been done on these platforms.
 
 Dependencies
 ------------
--- a/README.md
+++ b/README.md
@@ -40,8 +40,8 @@ on CentOS 7 and RHEL 7.
 
 We are actively working on porting USD to both Windows and Mac platforms. 
 Support for both platforms should be considered experimental at this time.
-Currently, the tree will build on Mac and Windows, but only limited testing
-has been done on these platforms.
+Currently, the tree will build on Mac (macOS version 10.14 and above) and Windows, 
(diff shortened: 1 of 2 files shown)


jtran56:
Filed as internal issue #USD-6486

c64kernal:
Thank you so much @makayla-k! How do you feel about replacing this text with a reference to the VERSIONS.md file instead? e.g., "Currently, the tree will build on macOS and Windows, but only limited testing has been done on these platforms. **Please see VERSIONS.md for specific versions tested.**" Would that have helped you find that file? And does that file contain enough information? (e.g., is the macOS version number itself important or does knowing the version of Xcode for example would have been enough to help you?)

makayla-k:
Hi @c64kernal! The file unfortunately does not include the information of what version to run on macOS (10.14 & up). Specifically Xcode was not the issue. When running on versions lower than 10.14 I got an error "Failed to run 'cmake --build . --config Release --target install -- -j4' " and the [group forum](https://groups.google.com/u/2/g/usd-interest/c/6bJAsuXYCPA/m/77RmvrH3BwAJ) pointed out that USD would only build on macOS version 10.14 or above. Jerry (from thread) cited the following reason: " MTLLanguageVersion2_1 is only available for macOS 10.14 onwards".

Given that it is an operating system version issue does it make sense to leave it on the README.md or does it make sense to add it to the VERSIONS.md?

makayla-k:
Hi @c64kernal  I wanted to follow up on this. As I said in my previous comment the issue was in fact the MacOS version so, I believe it would be helpful to add this to the README.md. There are [multiple](https://groups.google.com/u/1/g/usd-interest/c/OWsa6O26Eik/m/SMx1VeUjAwAJ) [people](https://groups.google.com/u/1/g/usd-interest/c/6bJAsuXYCPA/m/77RmvrH3BwAJ) that have faced the same issue when building the codebase on Mac. I believe merging this pull request would save significant time for other people who are also trying to build on Mac.

c64kernal:
Thanks @makayla-k -- would you mind amending the PR so that the README.md (in the location that you've edited has a reference to the VERSIONS.md file?) (e.g., "Please see VERSIONS.md for supported software versions) and then add a row at the very top (above C++ compiler) in  VERISONS.md called "OS". You can just fill in the macOS column for now and we will take care of updating it for the rest. We'd really like to keep all the version information in one place so that it is easier to keep it up to date. Thanks.

makayla-k:
Hello @c64kernal I took your recommendation and referenced the VERSIONS.md file and added a row at the top for the supported OS versions.

c64kernal:
Great thanks @makayla-k!

Issue #1228, opened 2020-05-29
UsdSkel display crash


cameronwhite:
### Description of Issue
This is a crash we've encountered working interactively with UsdSkel in Houdini, but can reproduce standalone in usdview with a stripped down file.

### Steps to Reproduce
1. Extract [usdskel.tar.gz](https://github.com/PixarAnimationStudios/USD/files/4704474/usdskel.tar.gz) and enter the folder
2. Copy over the `HumanFemale` folder from the [UsdSkel sample](http://graphics.pixar.com/usd/downloads.html), which is too large to include as a Github attachment
2. Run `usdview usdskel_1.usda`
3. Ensure that `View` -> `Display Purposes` -> `Guide` is enabled
4. Scrub the timeline a bit.
5. In the interpreter, run
```
l = Sdf.Layer.FindOrOpen("usdskel_2.usda")
usdviewApi.stage.GetRootLayer().TransferContent(l)
```
6. Scrubbing the timeline should crash immediately (example crash log is included in the archive)

### System Information (OS, Hardware)

Ubuntu 18.04, gcc 7.5

### Package Versions

USD 20.05

spiffmon:
Thanks, @cameronwhite - we are IP on pulling in PR #1213

cameronwhite:
Great, thank you!
I think this crash is unrelated to the one fixed in my PR, since it's occuring in the skeleton imaging rather than in BakeSkinning

spiffmon:
Ah - so sorry I conflated the two; I thought I saw "baking" in this report, but now I see I was imagining it.

jtran56:
Filed as internal issue #USD-6104.

Issue #1248, opened 2020-06-17
UsdSkel display crash when changing a primitive type to SkelRoot


cameronwhite:
### Description of Issue

This is a stripped down version of a crash we encountered with UsdSkel in Houdini, and can reproduce in `usdview`.
I think this is a separate issue from the one reported in #1228 - it doesn't require the skeleton prim to be displayed, and has a different stack trace when it crashes

### Steps to Reproduce
1. Extract [usdskel_rebind.tar.gz](https://github.com/PixarAnimationStudios/USD/files/4794559/usdskel_rebind.tar.gz) and enter the folder
2. Copy over the HumanFemale folder from the UsdSkel sample, which is too large to include as a Github attachment. This should be copied twice to `HumanFemale_01` and `HumanFemale_02` - referencing different USD files seems to be required to replicate the sequence of changes that caused the crash in Houdini.
3. Run `usdview usdskel_xform.usda`
4. In the interpreter, run
```
l = Sdf.Layer.FindOrOpen("usdskel_skelroot.usda")
usdviewApi.stage.GetRootLayer().TransferContent(l)
```
5. This usually crashes immediately, but otherwise scrubbing the timeline should do the trick

### System Information (OS, Hardware)

Ubuntu 18.04, gcc 7.5

### Package Versions

USD 20.05 (https://github.com/sideeffects/USD/tree/dev_houdini18.5.197)

jtran56:
Filed as internal issue #USD-6168.

tallytalwar:
Thanks, this was included in the 21.02 release.

PR #1397, opened 2020-11-19
Fix some display crashes after resyncing skeletons.

cameronwhite:
### Description of Change(s)

This fixes an issue where `UsdSkelImagingSkeletonAdapter::_RemovePrim()` failed to remove skeletons that did not have any bindings to skinned prims, which could lead to crashes later on from accessing expired prims.

`UsdSkelImagingSkeletonAdapter::_RemovePrim()` checks `_skelBindingMap` to determine whether the provided path is the skeleton prim. This was inconsistent with `UsdSkelImagingSkeletonAdapter::ProcessPrimResync()`, which checks `_skelDataCache` via `_GetSkelData()`. If there aren't any bindings, the skeleton is in `_skelDataCache` but not `_skelBindingMap`.

### Fixes Issue(s)
Fixes: #1228,  #1248


jtran56:
Filed as internal issue #USD-6489

PR #1398, opened 2020-11-20
Fix links to OpenEXR for Alembic and OpenImageIO plugins

tdelame:
### Description of Change(s)
Fix missing links to OpenEXR for Alembic and OpenImageIO plugins.
For the Alembic plugin, the order of OpenEXR libraries is important when OpenEXR is built as static libraries.

### DIFF
--- a/cmake/modules/FindOpenEXR.cmake
+++ b/cmake/modules/FindOpenEXR.cmake
@@ -62,6 +62,8 @@ foreach(OPENEXR_LIB
     Imath
     IlmImf
     IlmThread
+    IlmImfUtil
+    IexMath
     )
 
     # OpenEXR libraries may be suffixed with the version number, so we search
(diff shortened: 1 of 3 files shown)


meshula:
Please note this is probably insufficient to track OpenEXR 3, which will be part of vfxplatform 2021. Tracking the migration of OpenEXR from 2 to 3 is unfortunately complex. OpenImageIO has in the last week updated their own find algorithm to accommodate the transition away from OpenEXR 2, and I think it makes sense for USD to follow suit in supporting both. Here's the PR from OIIO with gnarly details. https://github.com/OpenImageIO/oiio/pull/2771

jtran56:
Filed as internal issue #USD-6490

tdelame:
@meshula Agreed, this PR focuses only on OpenEXR 2. OpenEXR 3 is definitely something we will have to look at for 2021 ^^.

PR #1399, opened 2020-11-20
Linux compilation

tdelame:
### Description of Change(s)
Fixed the compilation on Linux for different versions of gcc and clang:
- missing `<algorithm>` header and missing `std::` for clang 10.0 using libstdc++
- `std::max(std::initializer_list<T>>)` is not available on gcc 9.3.1 (the one suggested by https://vfxplatform.com/ for 2021)
- `std::less<>` is not available for gcc < 8.0

### DIFF
--- a/pxr/base/tf/token.cpp
+++ b/pxr/base/tf/token.cpp
@@ -39,6 +39,7 @@
 
 #include <tbb/spin_mutex.h>
 
+#include <algorithm>
 #include <string>
 #include <ostream>
 #include <vector>
--- a/pxr/base/trace/eventTreeBuilder.cpp
+++ b/pxr/base/trace/eventTreeBuilder.cpp
@@ -27,7 +27,7 @@
 #include "pxr/pxr.h"
 
 #include "pxr/base/trace/trace.h"
-
+#include <algorithm>
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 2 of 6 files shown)


jtran56:
Filed as internal issue #USD-6491

c64kernal:
Thanks for the PR, @tdelame! The changes look great! There are two that we have a couple of questions about. For the vt/dictionary.h change, we'd love to find a way to not introduce compiler-checks in that header (we try as much as possible to  leave all the platform-specific things in arch).

The other one is the std::max one, could we keep the initializer list somehow if we made the whole thing constexpr for example?

Thanks again!

tdelame:
For `std::max`, I managed to make the version with `std::initialiser_list` pass simply by adding `#include <algorithm>` at the top of the file. I saw that the algorithm header was already added in the dev branch a month ago for this very reason. I will remove the commit. 

For `vt/dictionary.h`, no problem, we won't use gcc < 8.0 for long on our side. I will remove the commit.

nvmkuruc:
Noting that a lot of these `#include <algorithm>`s have already been addressed by other commits.  `glf/udimTexture.cpp` no longer exists.  

`hio/stbImage.cpp` has not been addressed but I wonder if it's still needed given that OpenUSD is building with `gcc 3.9.1` and `clang 13.1.6`.

PR #1408, opened 2020-12-03
Add support for python 3.8 on Windows.

rstelzleni:
With python 3.8 and newer there is a Windows only function that adds a
directory to the dll search path. We need to use that to indicate where
the USD dlls are, instead of using the PATH.

### Description of Change(s)
This works for me in testing on windows with python 3.7 and python 3.8. In both cases I'm able to run usdview and other usd python scripts.

I am not an expert on cmake, so please let me know if there's a better way to do this multi-line string concatenation. I feel like this one looks pretty clean though. The resulting `__init__.py` looks like

```
__all__ = ['Tf', ...etc]

import os, sys
if sys.version_info >= (3, 8, 0):
    dllPath = os.path.split(os.path.realpath(__file__))[0]
    os.add_dll_directory(dllPath)

```

### Fixes Issue(s)
- 1404

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2017,16 +2017,6 @@ if find_executable("python"):
         PrintError("64bit python not found -- please install it and adjust your"
                    "PATH")
         sys.exit(1)
-
-    # Error out on Windows with Python 3.8+. USD currently does not support
-    # these versions due to:
-    # https://docs.python.org/3.8/whatsnew/3.8.html#bpo-36085-whatsnew
-    isPython38 = (sys.version_info.major >= 3 and
-                  sys.version_info.minor >= 8)
-    if Windows() and isPython38:
-        PrintError("Python 3.8+ is not supported on Windows")
-        sys.exit(1)
-
 else:
     PrintError("python not found -- please ensure python is included in your "
                "PATH")
--- a/cmake/macros/Public.cmake
+++ b/cmake/macros/Public.cmake
@@ -379,6 +379,16 @@ function(pxr_setup_python)
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/generated_modules_init.py"
          "__all__ = [${pyModulesStr}]\n")
 
+    # On windows also write out code to add the dll import directory
+    if (WIN32)
+        set(PY_WIN32_DLL_IMPORT_CODE_STRING "\nimport os, sys\n"
+            "if sys.version_info >= (3, 8, 0):\n"
+            "    dllPath = os.path.split(os.path.realpath(__file__))[0]\n"
+            "    os.add_dll_directory(dllPath)\n")
+        file(APPEND "${CMAKE_CURRENT_BINARY_DIR}/generated_modules_init.py"
+             ${PY_WIN32_DLL_IMPORT_CODE_STRING})
+    endif()
+
     install(
         FILES "${CMAKE_CURRENT_BINARY_DIR}/generated_modules_init.py"
         DESTINATION ${installPrefix}


jilliene:
Filed as internal issue #USD-6505

meshula:
For reference, please see the documentation notes here https://docs.python.org/3/library/os.html on os.add_dll_directory and why it's unavoidable on 3.8+.

sunyab:
I gave this a try on a relatively clean Windows 10 machine with Python 3.8.6 and it unfortunately did not work for me. I see the call to os.add_dll_directory in python/pxr/\_\_init\_\_.py as expected but trying to import a module like Tf still gives me:
```
C:\USD>python
Python 3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from pxr import Tf
  File "C:\USD\usd-20.11-py38\lib\python\pxr\Tf\__init__.py", line 85, in <module>
    from . import _tf
ImportError: DLL load failed while importing _tf: The specified module could not be found.
```
I think we're adding the wrong path here. If I understand the [docs](https://docs.python.org/3/library/os.html#os.add_dll_directory) and [porting notes](https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew) correctly, I think we need to add the directories containing the DLLs that the .pyd depends on. For example, for _tf.pyd we'd need to add the directory containing tf.dll as well as any directories containing its dependencies (e.g. boost, tbb.). 

However, we don't have a real great way of enumerating these directories, since in general these DLLs could live anywhere. We could try to extract this information from CMake and use it to generate the calls to add_dll_directory that are needed, but I have a feeling that's going to be really painful.

sunyab:
One possibility is to call os.add_dll_directory with the directories from %PATH%. If I change the code in python/pxr/\_\_init\_\_.py to:
```
import os, sys

# XXX: Only want to do this on Windows
if sys.version_info >= (3, 8, 0):
    for p in os.environ['PATH'].split(';'):
        if os.path.isdir(p):
            os.add_dll_directory(p)
```
then I'm able to import the Python modules without error. This definitely feels like a hack but I think it's equivalent (or close) to the behavior prior to Python 3.8.  Any thoughts?

The other issue I have with this is that it affects global Python behavior: once we import any module from pxr, the DLL search path will be modified permanently so any subsequent imports (including imports of modules outside of pxr) will be affected. Ideally we would scope the DLL search path modifications to just when we import our Python modules. The docs for os.add_dll_directory suggest this is possible.

meshula:
Oof, that's gross. It's going to add EVERYTHING to the python dll load path, including Houdini, every global copy of python, the system folder, and Fortnite. Right now, we ask users to:

```
Add [PATH TO BUILD]\bin to %PATH%
Add [PATH TO BUILD]\lib to %PATH%
Add [PATH TO STAGE]\lib\python to %PYTHONPATH%
```

What if we modify that for windows only, and mandate
```
Add [PATH TO BUILD]\bin to %PATH%
Add [PATH TO BUILD]\lib to %PATH%
Add [PATH TO BUILD]\bin;[PATH TO BUILD]\lib;[PATH TO STAGE]\lib\python to %PYTHONPATH%
```
and then modify your loop to refer to os.environ['PYTHONPATH']? Or maybe least ad hoc, on windows only:
```
Add [PATH TO BUILD]\bin to %PATH%
Add [PATH TO BUILD]\lib to %PATH%
Add [PATH TO STAGE]\lib\python to %PYTHONPATH%
Add [PATH TO BUILD]\bin;[PATH TO BUILD]\lib;[PATH TO STAGE]\lib\python to %USD_PYTHONPATH%
```
and then modify your loop to refer to os.environ['USD_PYTHONPATH']?

sunyab:
That'll only work if you build with build_usd.py, since the build script dictates where the dependencies go. If you're running cmake directly your dependencies could live anywhere. In that case, you'd have to have added those directories to your %PATH% anyway to run USD after you've built it, so you would already have had to deal with ordering your %PATH% appropriately.

rstelzleni:
The Windows error `ImportError: DLL load failed while importing _tf: The specified module could not be found.` could mean that any dependency of _tf wasn't found. If you do the proposed change you originally suggested, does it help or do you still see errors loading _tf? If so this might be more an issue of configuration on the windows test machine. For what it's worth, this did load and run on my test machine, and had been working in the PyPI packages before we removed 3.8 support.

In any case, I agree about the issue of cmake builds potentially not working with this setup. That's why I was originally doing this in the PyPI packages only, because in those we do know exactly what needs to be in the path.

rstelzleni:
I guess I'm saying, I don't think add_dll_directory needs to get anything more added than what we used to do with PATH. Previously, we didn't add every python module directory to PATH.

[Edit, I misunderstood what you were saying]

If the boost, tbb, or whatever else that we built against is not available, yeah, we'd need to add it with add_dll_directory. You can check what dependencies you're missing with a Windows tool called Dependencies https://github.com/lucasg/Dependencies

sunyab:
If I make the change I proposed, then importing tf (and the other modules) does work correctly.

I am a little surprised your change worked on your machine. It looks like your change calls add_dll_directory with <inst>/lib/python/pxr, but there aren't any DLLs in that directory so I don't think that would've done anything. Do you run into the errors on your test machine without this PR?

rstelzleni:
We are currently having a side discussion on this, but I wanted to update here to say my PR is definitely broken in two ways. I messed up the path while I was cleaning up before submitting, and also, I need to add a second path the dll search path. However, this still only works with build_usd.py. We're discussing other solutions at the moment, so I'm not updating the PR until we're sure about the right direction.

rstelzleni:
We've decided to try out some different approaches, so I'm going to close this PR and post a new one soon.

PR #1416, opened 2020-12-15
[test] if a testusdview filepath is invalid, print it

pmolodo:
### Description of Change(s)

This was handy when debugging a test failure I had in testUsdviewTimeSamples
Also fixed some minor formatting issues (ie, sys.stderr.write will not append a trailing newline automatically)

### DIFF
--- a/pxr/usdImaging/bin/testusdview/testusdview.py
+++ b/pxr/usdImaging/bin/testusdview/testusdview.py
@@ -90,16 +90,19 @@ class TestUsdView(Usdviewq.Launcher):
     # launching Usdview and doing the initial load pass
     def _ValidateTestFile(self, filePath):
         if not os.path.exists(filePath):
-            sys.stderr.write('Invalid file supplied, does not exist')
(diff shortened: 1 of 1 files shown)


spitzak:
It woudl be nice if usdview was fixed to display a shorter and more informative error message when the usd file does not exist.

pmolodo:
Hi - not sure if you're referring to the change introduced by this commit, or behavior of usdview that you've noticed... but if it's the latter, note that this PR is only altering the behavior of the test command, which is only used when running the USD test suite (ie, via ctest)... and the specific error messages I modified will probably only happen if you're editing tests, and make a mistake.

Having said that... I do seem to recall that some of the error messages for usdview were sometimes a bit vague, so I'd definitely be all for making them more helpful, too!

jtran56:
Filed as internal issue #USD-6521

tallytalwar:
Thanks, this has been merged with commit: 75fb30faaf6ced16ec9bf607630186b241dafcce

Issue #1417, opened 2020-12-18
Crash updating a point instancer


marktucker:
### Description of Issue
When viewing a point instancer through hydra, updating the point instancer to both deactivate a prototype primitive and set a constant primvar on an ancestor of the prototype in the same update cycle will cause a crash. This happens in the hydra code for determining the dirty prims/bits to send to the render delegate (so is not specific to any particular render delegate).

### Steps to Reproduce
1. Use usdview to open ptinstance1.usda from the following zip file:
[ptinstance.zip](https://github.com/PixarAnimationStudios/USD/files/5713671/ptinstance.zip)

2. In the interpreter window, run:
l = Sdf.Layer.FindOrOpen("ptinstance2.usda")
usdviewApi.stage.GetRootLayer().TransferContent(l)

3. usdview will crash.

### Package Versions
tested with 20.08, and current dev branch

jtran56:
Filed as internal issue #USD-6522

sunyab:
The fix for this issue was merged in commit 85e9400b3853 and released in 21.05. Closing, thanks!

PR #1418, opened 2020-12-18
Attempt to fix a crash on Hydra update when a prototype primitive is removed

marktucker:
Attempt to fix a crash on Hydra update when a prototype primitive is removed
(deleted or deactivated) in the same update cycle where an inheritable primvar
is added or changed to an ancestor of the prototype that used to exist.

### Description of Change(s)
In UsdImagingPointInstancerAdapter::_GetProtoUsdPrim, test that prim is valid before calling prim.IsPrototype().
In UsdImagingPointInstancerAdapter::ProcessPropertyChange, if the prototype UsdPrim is invalid, set the dirty bits to "clean".

### Fixes Issue(s)
- #1417


jtran56:
Filed as internal issue #USD-6523

sunyab:
This PR was merged in commit 85e9400b3853 and released in 21.05. Closing, thanks!

PR #1420, opened 2020-12-22
Added support for python 3.8 on windows by explicitely loading dlls

### LINKED ISSUE #1404: Add support for Python 3.8 on Windows
### Description of Issue

On Windows in Python 3.8, setting the path is no longer the right way to get USD's dlls to load in Python. There is a new api function "os.add_dll_directory(dllPath)" that we need to call instead of updating the path.

rdeioris:
Hi everyone, this patch adds support for python >= 3.8 on windows systems.

The fix works by explicitly adding the dll paths at runtime (by importing an ad-hoc module in the pxr/__init__.py module)

It should fix #1404

### DIFF
--- a/extras/windows_check.py
+++ b/extras/windows_check.py
@@ -1,0 +1,19 @@
+import os
+import platform
+import sys
+
+if platform.system() == 'Windows':
+    """
+    Since python 3.8 we need to explicitely set
+    the dll search path on windows
+    """
+    major = sys.version_info[0]
+    minor = sys.version_info[1]
+    if major > 3 or (major == 3 and minor >= 8):
+        base_dir = os.path.join(os.path.dirname(__file__), '..', '..', '..')
+        bin_dir = os.path.join(base_dir, 'bin')
+        lib_dir = os.path.join(base_dir, 'lib')
+
+        os.add_dll_directory(bin_dir)
+        os.add_dll_directory(lib_dir)
+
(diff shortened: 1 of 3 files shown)


rdeioris:
Note: CI failing on mac during PySide2 installation:

https://dev.azure.com/PixarAnimationStudios/USD/_build/results?buildId=174&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=61b979ba-27c9-588c-740e-a675d85b471e

extras/windows_check.py
```diff
@@ -0,0 +1,19 @@
+import os
+import platform
+import sys
+
+if platform.system() == 'Windows':
+    """
+    Since python 3.8 we need to explicitely set
```
meshula: "explicitly"

meshula (commented):
This is a nice, low touch, way to resolve this problem!

jtran56:
Filed as internal issue #USD-6530

sunyab:
Hi @rdeioris, thanks for this PR! We think there are some issues with this approach though. In particular, this change permanently modifies Python's DLL search path once any USD module is imported. Also, I think this change works well for USD builds done via build_usd.py since USD and all of its dependencies are in a well-known location, but may not work for "manual" builds of USD where the dependencies may live in other locations.

We're actually in progress right now on this issue and hope to have something shortly. So I'm going to close this PR out, but thanks again!

PR #1436, opened 2021-01-27
[usdImaging] Allow prototypes from overs, per https://graphics.pixar.com/usd/docs/api/class_usd_geom_point_instancer.html#UsdGeomPointInstancer_protoProcessing

asluk:
### Description of Change(s)

Allows instancer prototypes to come from within overs, as described in https://graphics.pixar.com/usd/docs/api/class_usd_geom_point_instancer.html#UsdGeomPointInstancer_protoProcessing and https://groups.google.com/g/usd-interest/c/RJnYoYWbgo8/m/Rh6UV2rlCwAJ . Tested with the attached City_set.

[city_protos_over.zip](https://github.com/PixarAnimationStudios/USD/files/5878839/city_protos_over.zip)

### DIFF
--- a/pxr/usdImaging/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/usdImaging/delegate.cpp
@@ -289,6 +289,14 @@ UsdImagingDelegate::_GetDisplayPredicate() const
             UsdPrimDefaultPredicate;
 }
 
+Usd_PrimFlagsConjunction
+UsdImagingDelegate::_GetDisplayPredicateForPrototypes() const
+{
+    return _displayUnloadedPrimsWithBounds ?
+        UsdPrimIsActive && UsdPrimHasDefiningSpecifier && !UsdPrimIsAbstract :
+        UsdPrimIsActive && UsdPrimHasDefiningSpecifier && UsdPrimIsLoaded && !UsdPrimIsAbstract;
+}
+
 // -------------------------------------------------------------------------- //
 // Parallel Dispatch
 // -------------------------------------------------------------------------- //
--- a/pxr/usdImaging/usdImaging/primAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/primAdapter.cpp
@@ -1055,6 +1055,12 @@ UsdImagingPrimAdapter::_GetDisplayPredicate() const
     return _delegate->_GetDisplayPredicate();
 }
 
+Usd_PrimFlagsConjunction
+UsdImagingPrimAdapter::_GetDisplayPredicateForPrototypes() const
+{
+    return _delegate->_GetDisplayPredicateForPrototypes();
+}
+
 size_t
 UsdImagingPrimAdapter::SampleTransform(
     UsdPrim const& prim, 
--- a/pxr/usdImaging/bin/testusdview/CMakeLists.txt
+++ b/pxr/usdImaging/bin/testusdview/CMakeLists.txt
@@ -197,6 +197,16 @@ if (NOT PXR_HEADLESS_TEST_MODE)
     )
 
     pxr_install_test_dir(
+        SRC testenv/testUsdviewInstancingPrototypes
+        DEST testUsdviewPointInstancingPrototypes
+    )
+
+    pxr_install_test_dir(
+        SRC testenv/testUsdviewInstancingPrototypes
+        DEST testUsdviewNativeInstancingPrototypes
+    )
+
+    pxr_install_test_dir(
         SRC testenv/testUsdviewAnimatedBounds
         DEST testUsdviewAnimatedBounds
     )
@@ -431,6 +441,24 @@ if (NOT PXR_HEADLESS_TEST_MODE)
     pxr_register_test(testUsdviewInstancingEdits
         PYTHON
         COMMAND "${CMAKE_INSTALL_PREFIX}/bin/testusdview --testScript testUsdviewInstancingEdits.py test.usda"
+        EXPECTED_RETURN_CODE 0
+    )
+
+    pxr_register_test(testUsdviewPointInstancingPrototypes
+        PYTHON
+        COMMAND "${CMAKE_INSTALL_PREFIX}/bin/testusdview --testScript testUsdviewInstancingPrototypes.py ni_pi.usda"
+        ENV TF_DEBUG=USDIMAGING_INSTANCER
+        STDOUT_REDIRECT point_instancing_proto_test_out
+        DIFF_COMPARE point_instancing_proto_test_out 
+        EXPECTED_RETURN_CODE 0
+    )
+
+    pxr_register_test(testUsdviewNativeInstancingPrototypes
+        PYTHON
+        COMMAND "${CMAKE_INSTALL_PREFIX}/bin/testusdview --testScript testUsdviewInstancingPrototypes.py internal_ni.usda"
+        ENV TF_DEBUG=USDIMAGING_INSTANCER
+        STDOUT_REDIRECT native_instancing_proto_test_out
+        DIFF_COMPARE native_instancing_proto_test_out 
         EXPECTED_RETURN_CODE 0
     )
 
--- a/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInstancingPrototypes/baseline/native_instancing_proto_test_out
+++ b/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInstancingPrototypes/baseline/native_instancing_proto_test_out
@@ -1,0 +1,8 @@
+[Populate NI] prim=/World/Xform_01
+[Add Instance NI] </World/Xform_01>  /World/Xform_01.proto_Sphere_id0 (Sphere), adapter = UsdImagingSphereAdapter
+[Instancer Inserted] /World/Xform_01, adapter = UsdImagingInstanceAdapter
+[Add Instance NI] </World/Xform_01>  /World/Xform_01
+[Populate NI] prim=/World/Xform_02
+[Add Instance NI] </World/Xform_01>  /World/Xform_02
+[Populate NI] prim=/World/Xform_03
+[Add Instance NI] </World/Xform_01>  /World/Xform_03
--- a/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInstancingPrototypes/baseline/point_instancing_proto_test_out
+++ b/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInstancingPrototypes/baseline/point_instancing_proto_test_out
@@ -1,0 +1,10 @@
+[Populate NI] prim=/World/sphereInstancer
+[Add PI] /__Prototype_1/Instancer, parentInstancerCachePath </World/sphereInstancer>
+[Instancer Inserted] /__Prototype_1/Instancer, adapter = UsdImagingInstanceAdapter
+[Add Instance PI] </__Prototype_1/Instancer>  /__Prototype_1/Instancer.proto0_ValidPrototype_id0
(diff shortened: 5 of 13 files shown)


pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -282,11 +282,17 @@ UsdImagingDelegate::_GetHdPrimInfo(const SdfPath &cachePath)
 }
 
 Usd_PrimFlagsConjunction 
-UsdImagingDelegate::_GetDisplayPredicate() const
+UsdImagingDelegate::_GetDisplayPredicate(bool forPrototypes) const
 {
-    return _displayUnloadedPrimsWithBounds ?
-            UsdPrimIsActive && UsdPrimIsDefined && !UsdPrimIsAbstract :
-            UsdPrimDefaultPredicate;
+    Usd_PrimFlagsConjunction retval = UsdPrimIsActive && !UsdPrimIsAbstract;
```
spitzak: This is the same as
```
          UsdPrimIsActive && !UsdPrimIsAbstract &&
          (forPrototypes || UsdPrimIsDefined) &&
          (_displayUnloadedPrimsWithBounds || UsdPrimIsLoaded);
```
Might make it easier to figure out what the patch does.

asluk: Thanks Bill-- to me, it's more clear as is, but I'm happy to change it per your suggestion if there's a consensus to do so. Thanks for looking!

jilliene:
Filed as internal issue #USD-6552

pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -282,11 +282,17 @@ UsdImagingDelegate::_GetHdPrimInfo(const SdfPath &cachePath)
 }
 
 Usd_PrimFlagsConjunction 
-UsdImagingDelegate::_GetDisplayPredicate() const
+UsdImagingDelegate::_GetDisplayPredicate(bool forPrototypes) const
```
spiffmon: Hi @asluk !  In most circumstances, we'd suggest an enum rather than a bool param for clarity, but since this fn is small and simple, we think even greater clarity can be achieved by simply adding a `_GetDisplayPredicateForPrototypes()` that implements the new computation.

pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -282,11 +282,17 @@ UsdImagingDelegate::_GetHdPrimInfo(const SdfPath &cachePath)
...
 {
-    return _displayUnloadedPrimsWithBounds ?
-            UsdPrimIsActive && UsdPrimIsDefined && !UsdPrimIsAbstract :
-            UsdPrimDefaultPredicate;
+    Usd_PrimFlagsConjunction retval = UsdPrimIsActive && !UsdPrimIsAbstract;
+    if (!forPrototypes) {
+        retval = retval && UsdPrimIsDefined;
```
spiffmon: This is actually incorrect. Rather than just leaving out `UsdPrimIsDefined `– which will wind up processing even "overs" underneath the top-most "over" – we want to substitute `UsdPrimHasDefiningSpecifier `for `UsdPrimIsDefined `, which essentially removes the "all ancestor prims must also be defined" clause from the traversal predicate, while still excluding _additional_ overs from consideration.

pxr/usdImaging/usdImaging/primAdapter.cpp
```diff
@@ -1050,9 +1050,9 @@ UsdImagingPrimAdapter::_GetCurrentTimeSamplingInterval()
 }
 
 Usd_PrimFlagsConjunction
-UsdImagingPrimAdapter::_GetDisplayPredicate() const
+UsdImagingPrimAdapter::_GetDisplayPredicate(bool forPrototypes) const
```
spiffmon: And remove the bool here, too?

spiffmon (changes requested):
The additional ask we have is if you could turn your nice test file into a testusdview test, also enhancing it to cover the requested changes (I.e. add some defined geometry under an over inside of one of the prototypes and ensure it doesn't get processed).  Super-extra bonus points if you have a nested PI that exercises the new logic, but won't hold you to that!

Thanks again, @asluk !

pxr/usdImaging/usdImaging/delegate.cpp
asluk: done

pxr/usdImaging/usdImaging/primAdapter.cpp
asluk: done

pxr/usdImaging/usdImaging/delegate.cpp
asluk: [nested_ptinst.zip](https://github.com/PixarAnimationStudios/USD/files/5944002/nested_ptinst.zip)

Thanks Spiff-- here's a zip that has a point instancer with a prototype in an over, and a def underneath, in sphereInstancer.usda -- it doesn't draw as expected unless I remove the UsdPrimHasDefiningSpecifier (or 
UsdPrimIsDefined) in the predicate. Am I misunderstanding something?

There's also an ni_pi.usda in the zip that puts the point instancer behind scenegraph instancing. No predicate as yet seems to get that drawing as expected. I'll continue investigating, unless you already know what's going on.

I'll also study testusdview a bit and see about getting these smaller test cases into it.

Thanks again!

spiffmon: Hmm.. the reason I thought this might not be working doesn't hold up.  There's alot going on right now, but I'll try to take a deeper look and get back to you.  First thing to check is whether, outside of imaging, the prototype root prims (which presumably are children of an undefined "over") pass the predicate.  If they do not, then maybe we have a core bug.

asluk: Hi Spiff, in the sphereInstancer.usda within the zip archive I posted above, I ran the following in usdview.

I think that is correct from a core standpoint, so I don't think there's a way to get the prototype detection logic to descend into overs without removing Usd.PrimHasDefiningSpecifier . But I think I understand your point that you want the prim range to descend into overs, but do not honor prototypes that themselves are pure overs. I'd have to think more about the right logic for that, if that's what the standard should be.

(I haven't had time to dig more into the nested instancing case yet in the zip archive).

Thanks!

(Dummy is an over, Dummy/sphere is a def)

>>> [x for x in Usd.PrimRange( usdviewApi.stage.GetPrimAtPath('/World/Instancer'), Usd.PrimIsActive & Usd.PrimHasDefiningSpecifier & ~Usd.PrimIsAbstract & Usd.PrimIsLoaded)]
[Usd.Prim(</World/Instancer>)]
>>> 
>>> [x for x in Usd.PrimRange( usdviewApi.stage.GetPrimAtPath('/World/Instancer'), Usd.PrimIsActive & ~Usd.PrimIsAbstract & Usd.PrimIsLoaded)]
[Usd.Prim(</World/Instancer>), Usd.Prim(</World/Instancer/Dummy>), Usd.Prim(</World/Instancer/Dummy/sphere>)]

spiffmon: What we were thinking, and what I thought the example showed, is that the parent of the prototypes (itself a child of the PI) would be a pure over. In our experience, having the prototypes themselves be overs would not even be practical to achieve, since they are references to defined models.

asluk: That's correct-- in my attached scene, Dummy is a pure over beneath the PI, and Sphere is a def for the prototype under Dummy. But with UsdPrimHasDefiningSpecifier, traversals won't reach Dummy from the PI.

spiffmon: But it shouldn't need to traverse from the PI, by my read of the code: https://github.com/PixarAnimationStudios/USD/blob/67af68bac3469b328e7f9dff9e0dc6e198ea7a5f/pxr/usdImaging/usdImaging/pointInstancerAdapter.cpp#L199-L233 . It explicitly starts from each prototype root, gleaned via the _prototypes_ relationship.  I'm not super familiar with this code, though.

asluk: Good point, Spiff. I reran the snippet above on the prototype root, which is a pure over with Sphere def'd under it.

>>> [x for x in Usd.PrimRange( usdviewApi.stage.GetPrimAtPath('/World/Instancer/Dummy'), Usd.PrimIsLoaded & ~Usd.PrimIsAbstract & Usd.PrimIsActive)]
[Usd.Prim(</World/Instancer/Dummy>), Usd.Prim(</World/Instancer/Dummy/sphere>)]

>>> [x for x in Usd.PrimRange( usdviewApi.stage.GetPrimAtPath('/World/Instancer/Dummy'), Usd.PrimIsLoaded & ~Usd.PrimIsAbstract & Usd.PrimIsActive & Usd.PrimHasDefiningSpecifier)]
[]

I think the predicate gets applied right away, which makes sense, so it won't descend below the pure over of Dummy.

I kept UsdPrimHasDefiningSpecifier out of the predicate, but now have it as a condition alongside checking for the adapter, such that pure overs are skipped, but defs that have ancestral overs are honored.

I'm still debugging the nested instancing case.

Thanks again!

spiffmon: Aaron, I thought we’d agreed above yhat the proto roots themselves are not meant to be pure overs, but their parents can be... a “protos” Xform or scope, if you will. Can you keep to that design? Then we can have nice simple code.

asluk: Sorry Spiff--- I've been scatterbrained. Fixed. Will continue investigating nested instances and regression testing.

spiffmon:
Hey @asluk !  Both your last force-push and the comment that you're still investigating happened 12 days ago.  We were wondering if you're still working on this?  Thanks!

asluk:
> Hey @asluk ! Both your last force-push and the comment that you're still investigating happened 12 days ago. We were wondering if you're still working on this? Thanks!

Hi Spiff! Yes, I'm still working on it, or rather it's still on my plate but backburnered as I investigate some other more pressing internal issues. I'll update here when I make more progress. Thanks!

asluk:
Not quite ready for re-review yet; working out some kinks with the unit test.

asluk:
Hi @spiffmon , I think this is ready for re-review now. I figured out what was happening with native/nested instances, and added a comment to instanceAdapter.cpp with those findings.

The unit test has a point instancer inside of two native instances, and the point instancer has one prototype under an over, and one prototype that is itself an over, and thus ignored.

Apologies for the delay!

asluk:
Added an additional test case to cover native instancing with overs as direct prototypes rather than in the nested point instancer prototype.

spiffmon:
Thanks, @asluk !  The code looks good, but when @tallytalwar ran our (unfortunately) internal tests with it, there were several failures, including geometry going missing in a test involving nested point and native instancing... which blows our minds a little because your change _should_ be strictly more permissive!  We'll need to make some time to dig into them - sorry for the delay!

spiffmon:
Update: @tcauchois got to the bottom of it!  It's a case where a PI prototype is nested inside a native instance, we never traverse the native instance to find it because the *root* primof NI's has (early) all of its fields stripped out, including its specifier.  We must be doing something special in the implementation of IsDefined that causes the root and all of its children to answer as if the root is defined, but HasDefiningSpecifier has no such special logic for prototype root prims.  We have an imaging-side workaround for now, and we'll followup to hopefully address this at the Usd-level instead.
I guess we had good reason to think this might not be straightforward ;-) Thanks for persevering, @asluk !

asluk:
Awesome-- much appreciated, @tcauchois and @spiffmon !

Issue #1440, opened 2021-01-31
USD can't write layers when executable launched from beneath a Windows junction


marktucker:
### Description of Issue
Related to #1377, but with a somewhat different cause. When a USD executable is launched from within "D:/foo", where "D:/foo" is a symlink or junction to "F:/foo", and then tries to write to "D:/foo/bar.usda", it fails with a permission error (which isn't really a permission error).

### Steps to Reproduce
1. Create a junction on Windows from D:/foo to F:/foo.
2. In a command prompt, navigate to D:/foo, and launch python.
3. Run the following:
from pxr import Sdf
l = Sdf.Layer.CreateAnonymous()
l.Export("d:/foo/bar.usda")
4. The last line fails with `'Insufficient permissions to write to destination directory 'f:\foo\foo''`

### System Information (OS, Hardware)
Windows

### Package Versions
USD 20.08 and 21.02

jilliene:
Filed as internal issue #USD-6556

sunyab:
Closing old issue, this was addressed by #1441.

PR #1441, opened 2021-01-31
Fix for Windows _ExpandSymlinks function

marktucker:
### Description of Change(s)

Fix for Windows _ExpandSymlinks function which would try to look for symlinks defining "X:", which looks at the "current directory" on that drive, not the root of that drive. So if the USD-using process was started in "X:/foo", where "X:/foo" is a junction to "Y:/foo", "X:" would expand to "Y:/foo".

So this change appends a "\" before trying to resolve symlinks on a path that ends with a ":", causing the expansion to be attempted on the root directory of the drive as intended, not on whatever the current directory on that drive happens to be.

### Fixes Issue(s)
- #1440

### DIFF
--- a/pxr/base/tf/pathUtils.cpp
+++ b/pxr/base/tf/pathUtils.cpp
@@ -66,7 +66,13 @@ _ExpandSymlinks(const std::string& path)
     // and the remaining part of the path.
     std::string::size_type i = path.find_first_of("/\\");
     while (i != std::string::npos) {
-        const std::string prefix = path.substr(0, i);
+        std::string prefix = path.substr(0, i);
+        // If the prefix is "X:", this will access the "current" directory on
+        // drive X, when what we really want is the root of drive X, so append
+        // a backslash.
+        if (prefix.at(i-1) == ':') {
+            prefix.push_back('\\');
+        }
         if (TfIsLink(prefix)) {
             // Expand the link and repeat with the new path.
             return _ExpandSymlinks(TfReadLink(prefix) + path.substr(i));


pxr/base/tf/pathUtils.cpp
```diff
@@ -66,7 +66,13 @@ _ExpandSymlinks(const std::string& path)
     // and the remaining part of the path.
     std::string::size_type i = path.find_first_of("/\\");
     while (i != std::string::npos) {
-        const std::string prefix = path.substr(0, i);
+        std::string prefix = path.substr(0, i);
+        // If the prefix is "X:", this will access the "current" directory on
+        // drive X, when what we really want is the root of drive X, so append
+        // a backslash.
+        if (prefix.at(i-1) == ':') {
+            prefix.push_back('\\');
```
spitzak, line 74: I would not use backslash unnecessarily. Forward slash works.

spitzak:
Not sure I understand why you want to ignore the Windows standard meaning of filenames. It seems like the user who really wants the root of X: can put "X:/" at the start of the filename, and may very well want "X:foo" to mean exactly what it does when they type it in a shell.

pxr/base/tf/pathUtils.cpp
marktucker: I thought about that a bit too. But this is pure Windows code, and the returned symlinks all have backslashes in them... In the end I decided it didn't matter. But I'm happy either way.

marktucker:
The place I changed the code only runs after splitting the path on slashes. It is a way to protect against treating "X:/foo" as if it were "X:foo". If you actually pass in "X:foo", this modified code will not run.

spitzak:
ok

jilliene:
Filed as internal issue #USD-6557

marktucker:
We've discovered an issue with this change. The `if (prefix.at(i-1) == ':') {` line should be `if (i > 0 && prefix.at(i-1) == ':') {`, just in case this function is passed a path starting with a slash. This can happen if you ask to save a layer to "/tmp/foo.usda" from a process started on drive C where "C:/tmp" is a valid directory. Should I submit this change as a separate pull request?

c64kernal:
Yes please, @marktucker a separate PR would be great, is it possible to also catch this in a test? Would be great to update the test if reasonable!

PR #1445, opened 2021-02-08
Usd 6554 -- Python3 shebang

bdeda:
### Description of Change(s)
The shebang.py uses the PYTHON_EXECUTABLE as the shebang, and uses "python" as the commands in the windows cmd files. The changes in theis PR make the PXR_PYTHON_SHEBANG settable via cmake, if the caller wants to customize that, and uses "python3" in the windows cmd files for python3 builds.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1438

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,11 +24,13 @@ include(Packages)
 
 # This has to be defined after Packages is included, because it relies on the
 # discovered path to the python executable.
-set(PXR_PYTHON_SHEBANG "${PYTHON_EXECUTABLE}" 
-    CACHE 
-    STRING
-    "Replacement path for Python #! line."
-)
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-6561

sunyab:
Hi @bdeda,

- We need a signed CLA from contributors before we can merge their pull requests in. Could you have a look at the instructions at https://graphics.pixar.com/usd/docs/Contributing-to-USD.html and submit one?

- We're on board with allowing customization of the Python shebang as you have it. However, defaulting to python3 in the .cmd files appears to break on standard Python 3.x installs on Windows which use python.exe and not python3.exe. Would it work if we just used the shebang string passed in to the script instead of hard-coding python or python3?

Thanks!

PR #1453, opened 2021-02-16
Linux arm64 support.

tsahee:
### Description of Change(s)
This change adds support for arm64/linux.
Currently, arm64 is only supported for OSX.

### Fixes Issue(s)
Fixed compilation errors caused trying to compile intel-architecture assembly on arm64.

### DIFF
--- a/pxr/base/arch/stackTrace.cpp
+++ b/pxr/base/arch/stackTrace.cpp
@@ -583,7 +583,6 @@ nonLockingLinux__execve (const char *file,
                          char *const argv[],
                          char *const envp[])
 {
-#if defined(ARCH_BITS_64)
     /*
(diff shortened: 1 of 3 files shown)


jilliene:
Filed as internal issue #USD-6573

AGSaidi:
I'd love to see this merged.

spiffmon:
Considering this PR is on deck for us, pending wrapping our enhanced UTF-8 support, which had gotten bumped by internal priorities.  Thanks!

gitamohr:
Thanks for this!  I tweaked it slightly to use the ARCH_CPU_ARM define we already have to be consistent with the other conditional code.  I will check this in once it passes our internal test suite and then it will show up on github in the next push.

PR #1454, opened 2021-02-16
[build_scripts] fix find_executable for windows .bat (ie, pyside2-uic.bat)

pmolodo:
### Description of Change(s)

There was a known issue with `distutils.spawn.find_executable` - that it fails to find `.bat` files (it only looks for `.exe`):
https://bugs.python.org/issue2200

I noticed this problem because the executable for pyside2-uic that got installed on my system is `pyside2-uic.bat` - presumably this is the same for many other windows users.

Unfortunately, since `distutils` is now deprecated, this will never be fixed, so just made a fixed version in `build.py` and renamed it `FindExecutable`.

For comparison, original code for `find_executable`:
https://github.com/python/cpython/blob/master/Lib/distutils/spawn.py#L91

### Fixes Issue(s)
- Building with pyside2 on windows (when pyside2-uic.bat is created)

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -23,8 +23,6 @@
 #
 from __future__ import print_function
 
-from distutils.spawn import find_executable
-
 import argparse
 import codecs
 import contextlib
@@ -48,8 +46,13 @@ import zipfile
 
 if sys.version_info.major >= 3:
     from urllib.request import urlopen
+    from shutil import which
 else:
     from urllib2 import urlopen
+
+    # Doesn't deal with .bat / .cmd like shutil.which, but only option
+    # available with stock python-2
+    from distutils.spawn import find_executable as which
 
 # Helpers for printing output
 verbosity = 1
@@ -119,7 +122,7 @@ def GetVisualStudioCompilerAndVersion():
     if not Windows():
         return None
 
-    msvcCompiler = find_executable('cl')
+    msvcCompiler = which('cl')
     if msvcCompiler:
         # VisualStudioVersion environment variable should be set by the
         # Visual Studio Command Prompt.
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-6574

c64kernal:
Thanks for this, @pmolodo -- we'd love to not support this in the long term either. What do you think of detecting whether or not we are running Python 3.x and then using shutil.which? That way maybe this PR can be a lot smaller in that case, and the downside is that this would only be fixed on Windows in Python 3.x, but maybe that's okay for your use-case? Bonus points if we can detect the version of pyside installed and warn when the error condition would be hit? Just a thought.

pmolodo:
That could work - I am using python 3 for windows builds, so selfishly, would be fine for me.

As for checking for error condition - I'm not sure I would trust that a .bat is always created for pyside2, and not for (original) pyside.  This certainly wouldn't be the first time that it happens that my computer is special.  We could just always print a warning if we fail to find something on windows + python2.

Still a problem for people who want python-2 on windows, but... one which will go away, eventually...

Only other objection I could think of would be the one mentioned [here](https://bugs.python.org/issue2200#msg131532), which is that this uses `PATHEXT`, which is that the actual function used, `spawnv`, doesn't support everything in `PATHEXT`.  However... when I tested on my machine, all the "extra stuff" came after the standard ones `spawnv` does accept, so... it's probably not really an issue for us.

So... I'd be on board with using `shutil.which`...

c64kernal:
Okay thanks @pmolodo -- maybe best not to overthink this since this is the only report we've gotten of this problem (as far as I know) so if you only need Python 3 on Windows and `shutil.which` is the way of the future, let's switch to that for that platform at least for now until we can characterize further problems better :)

pmolodo:
Pushed the change to use `shutil.which` - sorry to take so long to get to that!

PR #1455, opened 2021-02-17
Support time-sampled computed primvars

cameronwhite:
### Description of Change(s)

This adds support for evaluating time samples of computed primvars (ExtComputation) - the primary use case is motion blur for skinned primitives.

- Added methods to the scene delegate and prim adapter for sampling an `ExtComputation` scene input, similar to `SamplePrimvar()` and related methods
- Implemented `SampleExtComputationInput()` for the skinning computation.
- Added `HdExtComputationUtils::SampleComputedPrimvarValues()` for evaluating time samples of the computed primvars, along with a basic unit test

A couple notes from prior discussions with @rajabala
- We may want to consider adjusting the strategy used when clamping the samples, or make that a parameter. Currently it matches the behaviour of `SamplePrimvar()`
- The hardcoded static capacity for the `HdTimeSampleArray` in `extComputationUtils.h` is a bit unfortunate, but I'm not sure we want to turn `SampleComputedPrimvarValues` into a templated function

### DIFF
--- a/pxr/imaging/hd/CMakeLists.txt
+++ b/pxr/imaging/hd/CMakeLists.txt
@@ -114,3 +114,14 @@ pxr_library(hd
         overview.dox
 )
 
+pxr_build_test(testHdExtComputationUtils
+    LIBRARIES
+        hd
+
+    CPPFILES
+        testenv/testHdExtComputationUtils.cpp
+)
+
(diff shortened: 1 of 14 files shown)


jilliene:
Filed as internal issue #USD-6575

PR #1464, opened 2021-02-25
Remove check for a python instance in your path

rstelzleni:
### Description of Change(s)

The build scripts will use the python interpreter that was used to
invoke them, rather than looking for an executable named python in the
path. Relying on python's name isn't very reliable currently. In Linux
python 3 builds the executable will be named python3, for instance.
Also, in some containerized environments python may be present with a
different name, or not be available in the path.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1414

### Testing
I've done builds with this setup on a few different Mac, Windows and Linux machines. I also verified that the build succeeds in a linux docker container that only has a `python3` and no `python` in the path

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1982,29 +1982,20 @@ if (not find_executable("g++") and
     PrintError("C++ compiler not found -- please install a compiler")
     sys.exit(1)
 
-if find_executable("python"):
-    # Error out if a 64bit version of python interpreter is not found
-    # Note: Ideally we should be checking the python binary found above, but
-    # there is an assumption (for very valid reasons) at other places in the
-    # script that the python process used to run this script will be found.
-    isPython64Bit = (ctypes.sizeof(ctypes.c_voidp) == 8)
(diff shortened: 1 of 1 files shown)


tallytalwar (approved):

spiffmon:
@rstelzleni , this doesn't obviate the need for Pull Request #1445, does it?

rstelzleni:
I think this is a different issue from that one. This change removes a check that seems unnecessary and that can be a nuisance in some build environments. The other one looks like its about setting up shebangs in some environments. I haven't actually ran into the issues reported there.

jilliene:
Filed as internal issue #USD-6587

bdeda:
This is a different issue than [#1445](https://github.com/PixarAnimationStudios/USD/pull/1445). I am not using the python build script, so would not have come across this. 

The shebang issue (1445) is in the cmake part of the builds. In our case, we are explicitly telling cmake the python executable ourselves, where this part of the python build script is trying to get that for the user.

PR #1468, opened 2021-02-26
A collection of fixes for Windows path handling, especially involving symlinks

marktucker:
### Description of Change(s)
This started as an attempt to fix a possible exception when handling paths starting with "/" in the code for #1441 and then write some additional tests. But my test cases kept failing. Turns out there were still issues in handling on non-mountpoint symlinks on Windows. There was also the potential for infinite recursion in Tf_MakeDirsRec when it was passed a malformed path.

Also adding the test cases that were failing without all these fixes.

Relates to #1441, #1378.


jilliene:
Filed as internal issue #USD-6591

PR #1476, opened 2021-03-17
usdview - Don't error or warn in Rollover Prim Info if assetInfo has no identifier entry

ablev-dwa:
### Description of Change(s)

Changes usdview's Rollover Prim Info mode to gracefully handle cases where the "identifier" assetInfo is not authored on a prim. The existing code expects the data to exist, and will flood the terminal with KeyErrors if it's not actually authored, which also causes the rollover to be incomplete.

This change will make it Rollover Prim Info faster and more useful in pipelines that do not author the identifier assetInfo metadata.

### DIFF
--- a/pxr/usdImaging/usdviewq/common.py
+++ b/pxr/usdImaging/usdviewq/common.py
@@ -546,24 +546,23 @@ def GetFileOwner(path):
 # we will change this function to accept a prim rather than a primStack.
 def GetAssetCreationTime(primStack, assetIdentifier):
     """Finds the weakest layer in which assetInfo.identifier is set to
-    'assetIdentifier', and considers that an "asset-defining layer".  We then
-    retrieve the creation time for the asset by stat'ing the layer's
-    real path.
+    'assetIdentifier', and considers that an "asset-defining layer".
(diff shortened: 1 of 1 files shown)


spiffmon:
Hi @ablev-dwa ! Would you consider updating that function's comment about what the (now acceptable) fallback behavior is when there is no identifier-bearing layer to find?

ablev-dwa:
Absolutely, sorry, I should have thought of that. I will update the PR tonight.

ablev-dwa:
Updated the docstring for GetAssetCreationTime()

spiffmon:
Thanks, @ablev-dwa - looks great!  Of course, this function is not Ar 2.0 and non-file-backed-asset friendly at all.  But Ar won't give you concrete time/date information, which is why I wrote the function this way in the first place...

ablev-dwa:
That makes sense. We haven't started our journey to Ar2 yet but it will probably be good to start thinking about it :)

jtran56:
Filed as internal issue #USD-6613

tallytalwar:
This was merged with commit: e7e6067

PR #1480, opened 2021-03-22
Windows: Fix output values reported by PcpPrimIndex::PrintStatistics()

HamedSabri-adsk:
### Description of Change(s)

This PR fixes the values displayed incorrectly when PcpPrimIndex::PrintStatistics() routine is called on Windows. 

According to Linux manual page for printf 

```
      '      For decimal conversion (i, d, u, f, F, g, G) the output is
              to be grouped with thousands' grouping characters if the
              locale information indicates any.  (See setlocale(3).)
              Note that many versions of gcc(1) cannot parse this option
              and will issue a warning.  (SUSv2 did not include %'F, but
              SUSv3 added it.)
```

https://man7.org/linux/man-pages/man3/printf.3.html

There is no such flag for Windows:

https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-160

hence the output values show up wrong on Windows and correct on Linux:

**Windows:**
windows_statistic_printf

**Linux:**
linux_statistic_printf

**After the Fix:**

fix_me

### DIFF
--- a/pxr/usd/pcp/statistics.cpp
+++ b/pxr/usd/pcp/statistics.cpp
@@ -170,11 +170,7 @@ public:
     struct _Helper {
         static std::string FormatNumber(size_t n)
         {
-            #if defined(ARCH_OS_WINDOWS)
-                return TfStringPrintf("%zd", n);
-            #else
-                return TfStringPrintf("%'zd", n);
-            #endif
+            return TfStringPrintf("%zd", n);
         }
 
         static std::string FormatAverage(size_t n, size_t d)
--- a/pxr/usd/pcp/statistics.cpp
+++ b/pxr/usd/pcp/statistics.cpp
@@ -170,7 +170,11 @@ public:
     struct _Helper {
         static std::string FormatNumber(size_t n)
         {
-            return TfStringPrintf("%'zd", n);
+            #if defined(ARCH_OS_WINDOWS)
+                return TfStringPrintf("%zd", n);
+            #else
+                return TfStringPrintf("%'zd", n);
+            #endif
         }
 
         static std::string FormatAverage(size_t n, size_t d)


HamedSabri-adsk:
Hi @spiffmon, here is the fix that I mentioned earlier in our chat on google interest group yesterday. 

Cheers,
H

jtran56:
Filed as internal issue #USD-6617

c64kernal:
Thanks for the fix @HamedSabri-adsk! We generally try very hard to avoid any platform-specific ifdefs outside of arch. Would it be possible to just remove offending "''" and instead do something a little more standard across all platforms but still visually easy to compare (e.g., everything right justified)? Thanks again!

HamedSabri-adsk:
Hi @c64kernal, 

sure, that's a good point.  I removed the offending "'"  all together. As for the text Alignment, I didn't want to make a bigger change than it is necessary since the alignment looks visually nice and readable even with a very large number (e.g 15 digits). Hope this Ok?

Addressed: 76705d8198959d0f3918256487bf65d9945ee1b9

```
  Total nodes:                       922337200000000
  Total culled* nodes:               922337200000000
  By type (total / culled*):         
    root:                            922337200000000 / 922337200000000
    inherit:                         922337200000000 / 922337200000000
      implied inherits:              922337200000000 / 922337200000000
    variant:                         922337200000000 / 922337200000000
    relocate:                        922337200000000 / 922337200000000
    reference:                       922337200000000 / 922337200000000
    payload:                         922337200000000 / 922337200000000
    specialize:                      922337200000000 / 922337200000000
  (*) This does not include culled nodes that were erased from the graph
```

c64kernal:
This seems totally reasonable to me, thanks so much @HamedSabri-adsk! We'll try to get it merged soon.

PR #1486, opened 2021-04-06
base: Use std::atomic instead of tbb.

waywardmonkeys:
This converts the code in `pxr/base` to use `std::atomic` instead of `tbb::atomic`, which is deprecated.

This leaves only a couple of remaining usages of `tbb::atomic` in the codebase.


jilliene:
Filed as internal issue #USD-6640

PR #1491, opened 2021-04-15
[usdImagingGL,usdview] Add support to set the OCIO display, view, and colorspace

marsupial:
As we're testing more assets and images in usdview, there have been some request to be able to actual configure OCIO.
Additionally some internal usage of **usdImagingGL** will benefit from being able to tweak the profile in use.
For usdview:
- This keeps the behavior that selecting *openColorIO* from the *Color Mangement* menu will load a configs default behavior.
- It adds a sub menu where one can set the *display*, *view*, and *colorspace* for more fine-grained control.

### Description of Change(s)
Just pushes through some string to the right place to get picked up in **HdxColorCorrectionTaskParams**
Dynamically builds an *OpenColorIO* menu in usdview based on the current config if **PyOpenColorIO** is available.

Open to suggestions on how to make this better.

### DIFF
--- a/pxr/usdImaging/usdImagingGL/engine.cpp
+++ b/pxr/usdImaging/usdImagingGL/engine.cpp
@@ -274,7 +274,8 @@ UsdImagingGLEngine::RenderBatch(
 
     _PrepareRender(params);
 
-    SetColorCorrectionSettings(params.colorCorrectionMode);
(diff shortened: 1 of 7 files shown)


jilliene:
Filed as internal issue #USD-6656

PR #1496, opened 2021-04-19
Fixing vector iterator error found using Debug build

seando-adsk:
### Description of Change(s)
In the maya-usd repo during the configure stage (when calling usdGenSchema) I encountered the following runtime error:
image

### Fixes Issue(s)
- According to the vector::push_back() doc, _If the new size() is greater than capacity() then all iterators and references (including the past-the-end iterator) are invalidated. Otherwise only the past-the-end iterator is invalidated._ Therefore after the resize and push_back the end iterator is invalid and on the next iteration of the for loop, during the find() it will throw this error.

### DIFF
--- a/pxr/usd/sdf/layer.cpp
+++ b/pxr/usd/sdf/layer.cpp
@@ -3231,10 +3231,14 @@ SdfLayer::_ListFields(SdfSchemaBase const &schema,
             // more than one additional allocation here.
             if (mightAlloc && dataList.size() == dataList.capacity()) {
                 dataList.reserve(dataList.size() + (reqSz - reqIdx));
-                dataListBegin = dataList.begin(), dataListEnd = dataList.end();
                 mightAlloc = false;
+                dataList.push_back(reqName);
+                dataListBegin = dataList.begin();
+                dataListEnd = dataList.end();
             }
-            dataList.push_back(reqName);
+            else {
+                dataList.push_back(reqName);
+            }
         }
     }
     return dataList;


pxr/usd/sdf/layer.cpp
```diff
@@ -3231,10 +3231,14 @@ SdfLayer::_ListFields(SdfSchemaBase const &schema,
...
             if (mightAlloc && dataList.size() == dataList.capacity()) {
                 dataList.reserve(dataList.size() + (reqSz - reqIdx));
-                dataListBegin = dataList.begin(), dataListEnd = dataList.end();
                 mightAlloc = false;
+                dataList.push_back(reqName);
+                dataListBegin = dataList.begin();
+                dataListEnd = dataList.end();
+            }
+            else {
+                dataList.push_back(reqName);
```
seando-adsk, line 3240: I believe this keeps the original optimization intent by only searching thru the original vector after pushing new items as long as we don't have to resize. Once we resize, we must re-get the begin/end iterators as they will have been invalidated.

seando-adsk:
@gitamohr I believe this was something you recently changed. Can you have a look at this fix please?

pxr/usd/sdf/layer.cpp
gitamohr: Thanks!  So as I read it, the reserve() call can invalidate the iterators since it changes the capacity.  That's why I refetch the iterators directly after that call.  But then we've reserved enough capacity to accommodate all the rest of the elements even if the remainder are all unique.  So I believe that all future push_back()s should not invalidate the iterators.  That was the intent, anyway.

jtran56:
Filed as internal issue #USD-6660

pxr/usd/sdf/layer.cpp
seando-adsk: That is what I thought as well, but then upon re-reading the doc of push_back the bolded part caught my eye:
If the new size() is greater than capacity() then all iterators and references (including the past-the-end iterator) are invalidated. **Otherwise only the past-the-end iterator is invalidated.** 

So I believe that is why the original code has a problem. And I now also think that my fix might also not be correct and could still have a potential problem with the cached end iterator. To be 100% safe I would suggest not caching the begin/end:

        auto iter = std::find(dataList.begin(), dataList.end(), reqName);
        if (iter == dataList.end()) {

If there is a concern about performance and always searching to the end() from the find() we could add some iterator math to subtract the number of elements pushed. That way we don't bother searching thru the newly pushed items.

@gitamohr What do you think?

seando-adsk: I also found this stackoverflow post: https://stackoverflow.com/questions/46132706/does-stdvectorreserve-guarantee-that-the-implementation-will-not-invalidate/46132842#46132842 where in one of the replies someone comments, "Yes, push_back invalidates end()"

gitamohr: Ah, got it!  push_back unconditionally invalidates end().  I'll get this fixed up, thanks!

c64kernal:
Thanks for raising this @seando-adsk -- we'll close this out as @gitamohr mentioned we're working on a fix which we'll close out against #1501. Thanks again!

PR #1497, opened 2021-04-20
[usd] Use more specific constants for crate.

marsupial:
### Description of Change(s)
Rename **PAGEXXXX** contants to **CRATE_PAGEXXXX**

### Fixes Issue(s)
The constants used in crate.cpp are named a little to general for some platforms, notably [emscripten](https://github.com/PixarAnimationStudios/USD/issues/1492), where **PAGESIZE** is a pre-processor define.


jtran56:
Filed as internal issue #USD-6661

PR #1498, opened 2021-04-20
[hio] Don't export stbimage functions for linkage with things that may also use stbimage.

marsupial:
### Description of Change(s)
Add pre-processor macros so _stbimage_ functions aren't exported.

### Fixes Issue(s)
When building/linking with other things that use _stbimage_, symbols can be defined multiple times.

### DIFF
--- a/pxr/imaging/hio/stbImage.cpp
+++ b/pxr/imaging/hio/stbImage.cpp
@@ -39,12 +39,15 @@
 #include "pxr/base/tf/stringUtils.h"
 #include "pxr/base/tf/type.h"
 
+#define STB_IMAGE_STATIC
 #define STB_IMAGE_IMPLEMENTATION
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-6662

sunyab:
Hi @marsupial, this change caused one of our internal test builds on Linux in strict mode to fail. This was because gcc was emitting a bunch of "unused function" warnings now that the stb_* functions are marked static. I ended up adding a new pragma to arch/pragmas.h so that we could suppress those warnings in this file and get things working again.

Let me know if you feel we should solve this another way, otherwise this should land shortly. Thanks!

marsupial:
If I understand, you'll keep whats in this MR, but just surround it in a new `#pragma unused` macro ?
Sounds good and useful if so !

PR #1499, opened 2021-04-20
[tf,arch] Add missing includes.

marsupial:
Explicitly include `<ctime>` which doesn't automatically get included when compiling for Android or [emscripten](https://github.com/PixarAnimationStudios/USD/issues/1492).


jtran56:
Filed as internal issue #USD-6663

PR #1509, opened 2021-05-04
Support building of usdDraco plugin with monolithic build on Windows

furby-tm:
Previously when building a monolithic library of USD, the `usdDraco` plugin would result in missing symbols. This is because the python bindings require the methods in `writer.h` to be exported. This change exports those required symbols, and `usdDraco` can now be built with a monolithic build of USD on Windows.


jilliene:
Filed as internal issue #USD-6680

PR #1510, opened 2021-05-04
[arch] Fix demangling of symbols that have keywords as substrings.

asluk:
On Windows, USD schemas with "struct' in their names, such as "DestructionAPI" for fracture simulations, were not getting their TfType aliases registered properly, due to the demangling of their symbols. The relevant logic in Arch demangling was already stripping off a trailing space, so the fix is for the substring search to also include the trailing space for  "class ", "struct ", and "enum ".


jilliene:
Filed as internal issue #USD-6681

PR #1511, opened 2021-05-04
Support for Python 3.8+ on Windows

rstelzleni:
### Description of Change(s)

Previously this had been disabled because of a change to how the python
interpreter searches for DLLs. This change should restore the old
behavior, and give software builders the ability to opt into the new
python behavior where possible. See

https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

We've discussed a few approaches to this and went with this full path searching approach to try to be as close to the pre python 3.8 behavior as possible. You can disable it with an environment variable, which we'll make use of in the python pypi packages, since those are better off not searching the path.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1404

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -603,3 +603,10 @@ platforms to avoid issues with Boost config files (introduced in Boost version
 to use Boost specified config files for their USD build, specify 
 -DBoost_NO_BOOST_CMAKE=OFF when running cmake.
 
+2. Windows and Python 3.8+
+Python 3.8 and later on Windows will no longer search PATH for DLL dependencies.
+Instead, clients can call `os.add_dll_directory(p)` to set paths to search.
+By default on that platform USD will iterate over PATH and add all paths using
+`os.add_dll_directory()` when importing Python modules. Users may override
+this by setting the environment variable `PXR_USD_WINDOWS_DLL_PATH` to a PATH-like
+string. If this is set, USD will use these paths instead.
--- a/pxr/base/tf/__init__.py
+++ b/pxr/base/tf/__init__.py
@@ -25,6 +25,44 @@
 Tf -- Tools Foundation
 """
 
+# Type to help handle DLL import paths on Windows with python interpreters v3.8
+# and newer. These interpreters don't search for DLLs in the path anymore, you
+# have to provide a path explicitly. This re-enables path searching for USD 
+# dependency libraries
+import platform, sys
+if sys.version_info >= (3, 8) and platform.system() == "Windows":
+    import contextlib
+
+    @contextlib.contextmanager
+    def WindowsImportWrapper():
+        import os
+        dirs = []
+        import_paths = os.getenv('PXR_USD_WINDOWS_DLL_PATH')
+        if import_paths is None:
+            import_paths = os.getenv('PATH', '')
+        for path in import_paths.split(os.pathsep):
+            # Calling add_dll_directory raises an exception if paths don't
+            # exist, or if you pass in dot
+            if os.path.exists(path) and path != '.':
+                dirs.append(os.add_dll_directory(path))
+        # This block guarantees we clear the dll directories if an exception
+        # is raised in the with block.
+        try:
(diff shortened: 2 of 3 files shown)


jilliene:
Filed as internal issue #USD-6682

azure-pipelines.yml
```diff
@@ -4,7 +4,7 @@ trigger:
 
 jobs:
 - job: Linux
-  timeoutInMinutes: 90
+  timeoutInMinutes: 150
```
sunyab: I'm not sure why the Azure jobs were taking so long before but they seem to have recovered as of the last push to dev. Can we revert these timeout increases?

azure-pypi-pipeline.yml
```diff
@@ -30,7 +30,7 @@ stages:
         Python37:
           PYTHON_INTERPRETER: /opt/python/cp37-cp37m/bin/python
           PYTHON_TAG: cp37
-    timeoutInMinutes: 90
+    timeoutInMinutes: 150
```
sunyab: See above

pxr/base/tf/__init__.py
```diff
@@ -25,6 +25,44 @@
...
+# dependency libraries
+import platform, sys
+if sys.version_info >= (3, 8) and platform.system() == "Windows":
+    import contextlib
+
+    @contextlib.contextmanager
+    def WindowsImportWrapper():
+        import os
+        dirs = []
+        import_paths = os.getenv('USD_WINDOWS_DLL_DIRECTORY')
```
sunyab: Can we change this to something like PXR_USD_WINDOWS_DLL_PATH? The PXR prefix would be consistent with other env vars used by USD, and the PATH bit makes it more similar to the regular PATH env var.

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -41,7 +41,8 @@
 from pxr.UsdUtils.constantsGroup import ConstantsGroup
 
 # UI Components
-from ._usdviewq import Utils
+with Tf.WindowsImportWrapper():
```
sunyab: I thought it'd be nice to avoid having to use the import wrapper outside of Tf so I added commit d9580dbec27 to make sure we initialize usdviewq with Tf.PreparePythonModule. I tested this out and it seems to work, so I think the changes in usdviewq can be reverted.

BUILDING.md
```diff
@@ -603,3 +603,11 @@ platforms to avoid issues with Boost config files (introduced in Boost version
 to use Boost specified config files for their USD build, specify 
 -DBoost_NO_BOOST_CMAKE=OFF when running cmake.
 
+2. Windows and Python 3.8+
+After python 3.8 on windows, python will no longer search the whole path for
```
sunyab: Some tweaks, lmk what you think:

Python 3.8 and later on Windows will no longer search PATH for DLL dependencies.
Instead, clients can call `os.add_dll_directory(p)` to set paths to search.
By default on that platform USD will iterate over PATH and add all paths using
`os.add_dll_directory()` when importing Python modules. Users may override
this by setting the environment variable `USD_WINDOWS_DLL_PATH` to a PATH-like
string. If this is set, USD will use these paths instead.

sunyab (changes requested):
This looks great to me, just left a few cleanup-ish notes.

BUILDING.md
rstelzleni: I took this wording exactly, but also added PXR_ to the env var name

rstelzleni:
I think I addressed all the review notes. Let me know if I missed anything.

Would you prefer I rebase this so it's just one change, or is it better to keep the commit history? Let me know which you'd prefer, then I'll rebase the two followup changes to pick this up, and to use the new env var name.

sunyab:
I'd lean towards rebasing it to just one change, thanks!

sunyab (approved):

PR #1512, opened 2021-05-04
Enable building PyPI packages for Python 3.8

rstelzleni:
### Description of Change(s)

This PR is a followup to the change to add support for python 3.8 on windows (https://github.com/PixarAnimationStudios/USD/pull/1511). Now that we have that we can publish pypi packages for python 3.8

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1405


jilliene:
Filed as internal issue #USD-6683

PR #1513, opened 2021-05-04
Enable building PyPI packages for Python 3.9

rstelzleni:
### Description of Change(s)

Now that we can support python 3.8+ on our main platforms, we can also enable builds for python 3.9. This is a followup for https://github.com/PixarAnimationStudios/USD/pull/1512 that also adds 3.9.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1405


jilliene:
Filed as internal issue #USD-6684

PR #1519, opened 2021-05-13
Make virtual SdfFileFormat functions protected instead of private

BrianHarris:
https://groups.google.com/g/usd-interest/c/tGCnsGpl8iQ


jtran56:
Filed as internal issue #USD-6696

PR #1527, opened 2021-06-02
Correctly locate the dependent library in Debug mode

aaronmack:
### Description of Change(s)
In debug mode, FindOpenEXR failed to find the correct library

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/testenv/testSdfParsing.py
+++ b/pxr/usd/sdf/testenv/testSdfParsing.py
@@ -33,8 +33,19 @@
 # escaping errors.
 
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


meshula:
This note is to confirm that OpenEXR uses a CMAKE_DEBUG_POSTFIX set to "_d".

WinsingAaron:
I see. Thank you.

meshula:
To clarify what I meant, OpenEXR library names are suffixed with _d, so a debug build of USD should be linking against e.g. OpenEXR_d.lib, not OpenEXR.lib. Sorry if that wasn't clear.

PR #1528, opened 2021-06-03
Screen-space based scaling of curve width

abucior:
Submitting this pull request on behalf of Rakesh Roopavataram.

### About the proposed feature
The proposed change introduces a new feature for screen-space based widening of OpenGL curve rendering in HdStorm. It is in effect during the patch rendering mode of the curve only. The widening is based on simple metrics computed in screen space. The automatic widening kicks in when the computed curve width during the tessellation stage falls below one pixel. 

### Previous discussions and usage
The above change has been developed and tested at our site. The proposed change was also discussed with a few Pixar engineers offline. It was agreed upon that this feature might be general enough to be included in the PR.

### Implementation details
The current code change is structured in a way that is "always on". While this mechanism works for our production needs, we do realize other users may have different workflows. So the proposed code change is only for the core feature. We are open to adding "opt-in" features such as conditional execution of the shader code. 

### Steps to reproduce
We are attaching a simple USD scene file. 
[hairLayerLow_USD.usda.txt](https://github.com/PixarAnimationStudios/USD/files/6588509/hairLayerLow_USD.usda.txt)

Before
1. Open the scene file in USDView.
2. Set complexity to High  or Very High (View->Complexity->(High/VeryHigh)

Introduce changes to basisCurves.glslfx as shared

After
Repeat 1 and 2. 

The curve rendering should look widened and less aliased.

Before:
widening_before

After:
widening_after

Before:
hairWidth_before

After:
hairWidth_after1

### DIFF
--- a/pxr/imaging/hdSt/shaders/basisCurves.glslfx
+++ b/pxr/imaging/hdSt/shaders/basisCurves.glslfx
@@ -509,6 +509,25 @@ void main()
     MAT4 transform = ApplyInstanceTransform(HdGet_transform());
     vec4 scaledDirection = GetWorldToViewMatrix() * transform 
                   * vec4(direction, 0);
+
+    //
+    // Screen space curve width scaling
+    // if the width falls below one pixel we widen it 
+    // optional minCurveWidth() for user driven controls 
+    //
+    MAT4 projMat = GetProjectionMatrix();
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-6713

Issue #1535, opened 2021-06-14
[usdview] Potentially opens in the background under Windows


pmintus:
### Description of Issue
When opening usdview, it may sometimes launch itself behind other windows in Windows. This can happen when usdview is associated with a file extension and launched out of Windows Explorer.

### Steps to Reproduce
1. Associate the .usd extension with usdview
2. Double-click on a usd file in Windows Explorer

I have tested this on 2 machines. It's a 100% repro on one, and 0% on the other. They are both running Windows 10 20H2.

You can see it happen in this YouTube video. From 3:12 to 3:25.
https://youtu.be/3QZCT6y2iXc?t=192

### System Information (OS, Hardware)
Windows 10 20H2

### Package Versions
USD 21.05

pmintus:
I plan on submitting a fix.

jilliene:
Filed as internal issue #USD-6732

sunyab:
This should be addressed by PR #1537 in the 21.08 release. Closing, thanks!

PR #1537, opened 2021-06-14
[usdview] Call ActivateWindow on MainWindow after Show

pmintus:
### Description of Change(s)
Calling qtWidget::activateWindow on mainWindow after it is created to bring it to the foreground. On Windows this will call Win32 SetForegroundWindow.

### Fixes Issue(s)
- #1535


jilliene:
Filed as internal issue #USD-6733

PR #1547, opened 2021-06-24
Add docker file + build, test, and run scripts to USD repo

lucbettaieb:
### Description of Change(s)
I had to dockerize the USD install for a project I am working on and figured it might make for a good upstream commit.

### Fixes Issue(s)
- Doing that "engineer thing" where I'm fixing problems that probably don't exist.
- I had a colleague who was having a lot of trouble with the Linux install so I dockerized it.

### DIFF
--- a/docker/21.05/Dockerfile
+++ b/docker/21.05/Dockerfile
@@ -1,0 +1,48 @@
+FROM ubuntu:focal
+
+LABEL description="ubuntu-focal-usd"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+
+# Install build dependencies 
+RUN apt-get update -qq
+RUN apt-get install -y cmake \
+                       build-essential \
+                       curl \
+                       git \
+                       software-properties-common \
+                       wget \
+                       libgl1-mesa-dev \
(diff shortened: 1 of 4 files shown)


lucbettaieb:
**Demo**
Screencast-2021-06-24-115004

lucbettaieb:
If you don't want to merge this into your main branch, I'm happy to just host a repo with some Dockerfiles for releases since I'm calling a `git clone` in the Dockerfile, anyway.

jilliene:
Filed as internal issue #USD-6774

spiffmon:
> If you don't want to merge this into your main branch, I'm happy to just host a repo with some Dockerfiles for releases since I'm calling a `git clone` in the Dockerfile, anyway.

That would actually be great, @lucbettaieb !  We have no problems with this change, but we don't use Docker at all and are already stretched pretty thin for testing all our current configurations and options.

aloysbaillet:
Hi @lucbettaieb , I did also write a few docker files a few years back, see here: https://github.com/aloysbaillet/docker-usd

sunyab:
As @spiffmon mentioned we think this would be better served living outside our repo since our resources to exercise and maintain this are unfortunately limited. I'm going to close this out, thank you for filing this!

PR #1553, opened 2021-07-07
Add missing quotes in examples of UsdGeom overview

charlesfleche:
### Description of Change(s)

This patch adds missing quotes around prim names in the example USD of the Imageable Purpose section of the UsdGeom overview.


jilliene:
Filed as internal issue #USD-6780

PR #1557, opened 2021-07-12
Resolved unclosed file ResourceWarnings

jmather-sesi:
The original code would emit unclosed file ResourceWarnings when the PYTHONWARNINGS environment variable is active.


jilliene:
Filed as internal issue #USD-6785

Issue #1558, opened 2021-07-13
Resolution of UDIM texture paths inside USDZ archives fails


dalorin:
### Description of Issue
It is not possible to resolve paths to UDIM textures that are located inside a USDZ archive. This is because the utility function _ResolveAssetAttribute located in [materialParamUtils.cpp](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/materialParamUtils.cpp#L156) does not account for the presence of usdz reference syntax e.g. /path/to/file.usdz[0/texture.png] when sanity checking the suffix post-resolution [here](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/materialParamUtils.cpp#L187). As a result, the return SdfAssetPath object does not contain a resolved path and renderers will treat texture paths such as "0/texture.png" as absolute and fail.

### Steps to Reproduce
1. Create a USD file containing a model with attached texture using UDIMs.
2. Run usdzip with -a flag against the USD file to create a usdz archive containing the texture files.
3. Attempt to render the usdz archive with a hydra render delegate. In the case of arnold, the following output is produced:
00:00:00   535MB ERROR   |  [texturesys] could not read resolved image filename = "2/alien_Opacity.1007.tif"

### System Information (OS, Hardware)
CentOS 7, python 2.7.17

I have a patch for this issue that I will raise in a pull request if this issue is confirmed.

jilliene:
Filed as internal issue #USD-6786

sunyab:
Hi @dalorin, I can confirm the issue is exactly as you've said. We have a fix internally pending review and testing but it'll be at least a week or so before we can push that up due to some office outages. I'll add some notes to the PR as well. Thanks!

PR #1560, opened 2021-07-14
Fix resolution of UDIM texture paths inside USDZ archives

unionvfx:
### Description of Change(s)
Adding logic to asset path resolution utility function to allow for the presence of assets internal to USDZ archives i.e. /path/to/asset.usdz[0/texture.jpg]

### Fixes Issue(s)
- Issue #1558 - Resolution of UDIM texture paths inside USDZ archives fails

### DIFF
--- a/pxr/usdImaging/usdImaging/materialParamUtils.cpp
+++ b/pxr/usdImaging/usdImaging/materialParamUtils.cpp
@@ -21,6 +21,8 @@
 // KIND, either express or implied. See the Apache License for the specific
 // language governing permissions and limitations under the Apache License.
 //
+#include <regex>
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-6788

sunyab:
Thanks for filing this PR! We had already been looking into this issue before this was uploaded and have a fix we'll be pushing in the near future -- see comments in #1558. The primary difference with our internal fix is that we use API in ar/packageUtils.h for slicing and dicing .usdz paths instead of a special regex. I'm going to close this PR in favor of that upcoming fix. 

Thanks again!

PR #1568, opened 2021-07-22
Fix a bug which could cause a crash during composition

marktucker:
Fix a bug which could cause a crash during composition or recomposition
after a change to a stage. Sufficiently frequent/repetitive calls to
_EvalRefOrPayloadArcs in primIndex.cpp, where ComputeLayerStack is called
many times for the same layer could result in Pcp_LayerStackRegistry
returning a PcpLayerStack that was either already deleted, or about to be
deleted. Fixing this required moe careful management of the table mapping
identifiers to layer stacks in the registry object.

### Description of Change(s)
I believe there was unsafe code in the layer stack registry. Given enough threads simultaneously being in _EvalRefOrPayloadArcs in pcp/primIndex.cpp, with all those threads calling ComputeLayerStack for the same layer file, there is lots of opportunity for the registry cleanup code in the PcpLayerStack destructor to interact badly with the Pcp_LayerStackRegistry::FindOrCreate method.

I wish I could provide a reproduceable test case for this crash, but I was only able to make it happen within Houdini. But the crash would happen within a call to UsdUtilsFlattenLayerStack, in the SdfChangeBlock destructor when exiting UsdFlattenLayerStack. With this change I can no longer reproduce the crash in Houdini, and all the USD tests still pass (so at least I'm pretty sure I didn't break anything). Really curious to hear thoughts about this change...

### DIFF
--- a/pxr/usd/pcp/layerStackRegistry.cpp
+++ b/pxr/usd/pcp/layerStackRegistry.cpp
@@ -145,23 +145,31 @@ Pcp_LayerStackRegistry::FindOrCreate(const PcpLayerStackIdentifier& identifier,
     }
 
     tbb::queuing_rw_mutex::scoped_lock lock(_data->mutex, /*write=*/false);
+    PcpLayerStackRefPtr refLayerStack;
 
     if (const PcpLayerStackPtr & layerStack = _Find(identifier)) {
-        return layerStack;
-    } else {
+        refLayerStack = TfCreateRefPtrFromProtectedWeakPtr(layerStack);
+    }
+    if (!refLayerStack) {
         lock.release();
 
-        PcpLayerStackRefPtr refLayerStack =
+        PcpLayerStackRefPtr createdLayerStack =
             TfCreateRefPtr(new PcpLayerStack(
                 identifier, _GetFileFormatTarget(), _GetMutedLayers(), 
                 _IsUsd()));
 
-        // Take the lock and see if we get to install the layerstack.
+        // Take the lock and check again for an existing layer stack, or
+        // install the one we just created.
         lock.acquire(_data->mutex);
-        auto iresult =
-            _data->identifierToLayerStack.emplace(identifier, refLayerStack);
-        if (iresult.second) {
-            // If so give it a link back to us so it can remove itself upon
+        if (const PcpLayerStackPtr & layerStack = _Find(identifier)) {
+            refLayerStack = TfCreateRefPtrFromProtectedWeakPtr(layerStack);
+        }
+        if (!refLayerStack) {
+            // No existing entry, or it is being deleted. Add the one we just
+            // create to the map.
+            refLayerStack = createdLayerStack;
+            _data->identifierToLayerStack[identifier] = refLayerStack;
+            // Also give it a link back to us so it can remove itself upon
             // destruction, and install its layers into our structures.
             refLayerStack->_registry = TfCreateWeakPtr(this);
             _SetLayers(get_pointer(refLayerStack));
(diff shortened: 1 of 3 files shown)


marktucker:
FWIW, here is a simple Houdini file that demonstrates the crash:
EDIT: The original file posted here was missing a file asset.usda. New zip archive here:
[stitching_crash_2.zip](https://github.com/PixarAnimationStudios/USD/files/6864566/stitching_crash_2.zip)

jilliene:
Filed as internal issue #USD-6804

lbiasco-dwa:
I believe I'm also seeing this bug pop up. I haven't found a consistent repro with just USD either, but if I do find one I'll try testing this fix against it.

pxr/usd/pcp/layerStackRegistry.cpp
```diff
@@ -224,13 +232,21 @@ Pcp_LayerStackRegistry::GetAllLayerStacks() const
...
         _data->identifierToLayerStack.find(identifier);
-    if (TF_VERIFY(i != _data->identifierToLayerStack.end()) &&
-        TF_VERIFY(i->second.operator->() == layerStack)) {
+    // It's possible that layerStack has already been removed from the
+    // map if a FindOrCreate call intercedes between the moment when the
+    // layer stack's ref count drops to zero and the time the layer stack
+    // destructor is called (which is how we get into this method).
+    if (i != _data->identifierToLayerStack.end() &&
+        i->second.operator->() == layerStack) {
+        _SetLayers(layerStack);
```
sunyab: I think we should call _SetLayers outside of this if block. If we're in here, the layerStack is being destroyed so we'd want to make sure any mappings between it and other layers are cleaned up regardless of whether the identifierToLayerStack map has already been updated with a new entry for the given identifier. Does that seem right to you?

pxr/usd/pcp/layerStackRegistry.h
```diff
@@ -112,8 +112,8 @@ class Pcp_LayerStackRegistry : public TfRefBase, public TfWeakBase {
     PcpLayerStackPtr _Find(const PcpLayerStackIdentifier&) const;
 
     // Remove the layer stack with the given identifier from the registry.
-    void _Remove(const PcpLayerStackIdentifier&,
-                 const PcpLayerStack *);
+    void _SetLayersAndRemove(const PcpLayerStackIdentifier&,
+                             const PcpLayerStack *);
```
sunyab, line 116: Can we update the comment at line 135 to accommodate the renamed function?

sunyab (changes requested):
Nice find! I added a few notes, let me know what you think.

pxr/usd/pcp/layerStackRegistry.h
marktucker: Thanks for catching this. Done.

pxr/usd/pcp/layerStackRegistry.cpp
marktucker: Yes, I think you're right about this. My misunderstanding was that I thought the other mappings from the layer stack to layers and the layer stack to muted layers were also based on the layer stack identifier. But reading the code I see the other mappings are based on the layer stack pointer, so I agree they should always be cleaned up. Thanks!

sunyab (approved):

sunyab:
Thanks @marktucker! Changes look good to me, I've merged them internally and they should land in dev soon.

lbiasco-dwa:
Just to follow up with some more affirmation, I found a couple fairly consistent repro, the best being rapidly running LoadAndUnload on some heavy assets at my studio (failing on anywhere between 1 to 20 load/unloads), and this change completely fixes the problem! It hasn't crashed once, even on 50 load/unloads. Sadly the repro requires production assets that I can't share so I can't pass along a test, but I at least wanted to confirm the stability of the fix in case there was any uncertainty about the fix working with more cases.

marktucker:
Thanks @lbiasco-dwa! It's great to have additional evidence that the bug is not Houdini-specific, and that this fix works in more scenarios than the one I was testing with.

PR #1575, opened 2021-07-28
Fix OpenEXR libraries installation

hshakula:
### Description of Change(s)
Add `OPENEXR_PACKAGE_PREFIX` option to OpenEXR build script. This is required to install the `IlmImf` library into the correct directory. Without `OPENEXR_PACKAGE_PREFIX` option this [line](https://github.com/AcademySoftwareFoundation/openexr/blob/v2.3.0/OpenEXR/IlmImf/CMakeLists.txt#L6) sets `RUNTIME_DIR` to `/bin` which leads the `IlmImf` library to be [installed](https://github.com/AcademySoftwareFoundation/openexr/blob/v2.3.0/OpenEXR/IlmImf/CMakeLists.txt#L259) to the root of the filesystem.

### Fixes Issue(s)
- `IlmImf` library is not installing to `instDir/bin`


jilliene:
Filed as internal issue #USD-6810

Issue #1577, opened 2021-07-29
USDZ archive lookup is O(n^2)


marsupial:
### Description of Issue
USDZ is linearly traversing the archive for every lookup, which creates a huge bottleneck and is very noticeable once you pass ~10000 entries.

This was noticed by someone who is trying to use USD and Draco compression, which means every mesh will be referencing an external file.  While not the most common use-case, the issue will also be at play even in a **flat** USD scene that makes heavy use of unique textures, or for a heavily referenced scene that just wants to be bundled into one file.

### Steps to Reproduce
1. Generate a usdz file with around 25000 external references, and then Open the stage.  Even when those references are all nothing more than a single plane, load time will be above 15 seconds, while the original **usdc** takes under 3 seconds.

jilliene:
Filed as internal issue #USD-6816

PR #1578, opened 2021-07-29
Cache usdz archive traversal for subsequent lookup/queries on external references.

marsupial:
### Description of Change(s)
While linearly traversing the **usdz** archive, cache the path and header information so that subsequent lookups can be accelerated.

### Fixes Issue(s)
#1577, #1579

This brings down load time of a **usdz** file that was taking ~3 minutes to ~16 seconds, and can often wind up performing better than parsing the original **usdc** file that generated the **usdz** (likely because IO is avoided on the references as the files are _in memory_ from the archive).

The test case is an example that takes more than 15 seconds to load currently, but with caching goes down to 2 seconds.  It has 25000 entries, chosen to demonstrate the time difference in a way that can be tested against regression.  The amount of entries is also good to  test against any race conditions filling in the cache (though per the comments in code, a relatively simplistic filling is used to as it performs well enough and keeps the code simple)

### DIFF
--- a/pxr/usd/ar/packageUtils.cpp
+++ b/pxr/usd/ar/packageUtils.cpp
@@ -85,11 +85,11 @@ _FindMatchingOpeningDelimiter(
 {
     size_t numOpenNeeded = 1;
     std::string::const_reverse_iterator revIt(closingDelimIt);
-    for (; revIt != path.rend() && numOpenNeeded != 0; ++revIt) {
+    for (const auto rend = path.rend(); revIt != rend && numOpenNeeded != 0; ++revIt) {
         if (*revIt == '[' || *revIt == ']') {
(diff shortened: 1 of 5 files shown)


jilliene:
Filed as internal issue #USD-6817

sunyab:
Just wanted to mention I'm looking at merging this PR in now. Thanks for your patience!

sunyab:
Thanks for your patience. I ended up reworking some of these changes and separating the fix for #1577 and #1579 into different commits. I also pulled the performance test into an internal test suite instead of committing them, as I'm hesitant to add a semi-large-ish file to the repo. We could certainly revisit that if necessary.

I'm going to close this PR out in favor of those separate commits, which will land with the next dev push. Thanks again!

Issue #1443, opened 2021-02-04
Unicode characters in filenames are not preserved


rstelzleni:
### Description of Issue

Recently we've run into some issues with USD files that have non-ascii characters in their filenames. We initially found that a USD file with an umlaut in the filename opened fine on Linux but failed to open on Windows. I noticed that ArchOpenFile calls fopen, which might explain that difference. On Linux fopen can accept a utf8 filename, but the Windows version doesn't.

Based on that I wasn't sure if USD is expected to handle utf8 filenames or not. I did some testing, and it seems like you can use utf8 filenames in sublayer lists and in references, and things seem to work. However, if I create such a file by hand, open it and save it out I find that any special characters get stripped. For example, this:

```
#usda 1.0

def "Sphere" (
    references = @Grüße/Löwenbräu.usd@</Sphere>
)
{
}
```

to this

```
#usda 1.0

def "Sphere" (
    references = @Gre/Lwenbru.usd@</Sphere>
)
{
}
```

I didn't test this with string valued attributes or asset paths in USD files.

I posted this to the mailing list, and Alex Mohr suggested filing an issue for tracking this case.

### Steps to Reproduce
1. Create a usd file and name it with some utf8 characters
2. Attempt to open on windows, open will fail
3. On linux, it will open. Create another usd file that references it
4. Check that you can open that usd file and the reference works
5. Save that usd file, the non ascii characters are stripped from the reference

### System Information (OS, Hardware)

Tested on Windows 10 and Ubuntu 20.04

### Package Versions

USD 21.02

jilliene:
Filed as internal issue #USD-6560

rstelzleni:
Hello! When I originally filed this issue it sounded like Pixar was already working on utf8 support. I just wanted to check in to see if there is a plan for that. Would we expect to see it in 21.04, for instance?

Thanks!

spiffmon:
Hi @rstelzleni ! Yes, we are working on it, and had hoped to have it ready for 21.05; however, some pretty intense production priorities have intervened, which pushes this out to 21.08.

rstelzleni:
Sounds good, thanks Spiff. If there are changes that we could cherry pick and test out sooner we'd be happy to get some testing done. I didn't see anything in the current public repo. We could also help with implementation under Pixar's direction if that helps, just let us know!

DDoS:
[...]

PR #1580, opened 2021-07-30
Explicit unicode support for windows

aloysbaillet:
### Description of Change(s)
This change adds explicit windows support for UTF-8 characters in file paths.
As discussed in #1443 these changes are not necessary if the calling executable is patched to use the UTF-8 code page on recent versions of Windows (see https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page for details).
Unfortunately these changes are necessary for applications that need to be used without changing the OS default settings.

### Fixes Issue(s)
- #1443

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -74,13 +74,17 @@ static inline HANDLE _FileToWinHANDLE(FILE *file)
 
 FILE* ArchOpenFile(char const* fileName, char const* mode)
 {
+#if defined(ARCH_OS_WINDOWS)
+    return _wfopen(ArchWindowsUtf8ToUtf16(fileName).c_str(), ArchWindowsUtf8ToUtf16(mode).c_str());
+#else
     return fopen(fileName, mode);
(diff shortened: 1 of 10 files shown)


aloysbaillet:
I just updated the PR by replacing the `wstring_convert` which caused big slowdowns on our side, with the Windows API `MultiByteToWideChar` and `WideCharToMultiByte` which are much faster and do not seem to impact performance at all.
BTW the slowdown was found in a simple usdz file containing 25k references to a tiny geometry file, which probably is the worst case scenario for this change (loading that file went from 10s to 45s on my machine without the last change!).

jilliene:
Filed as internal issue #USD-6819

gitamohr:
Hey Aloys -- I'm getting back to this one now.  Was there more you wanted to do here or do you feel this is ready to go?

aloysbaillet:
Hi Alex, I believe this PR is good to go, it has been used for some time now and no regression was reported until now...
That said I'd love to get your feedback on if there's a nicer way to do this!

PR #1606, opened 2021-08-30
[usdPhysics] Fix python bindings for metrics and kilogramsPerUnit metadata registration.

asluk:
[usdPhysics] Fix python bindings for metrics and kilogramsPerUnit metadata registration.

(Also register testUsdGeomMetrics with tests)


jilliene:
Filed as internal issue #USD-6870

Issue #1607, opened 2021-08-30
UsdImaging slow to update from transforming many instanceable prims


marktucker:
### Description of Issue
The UsdImagingDelegate::ApplyPendingUpdates method has very bad performance when applying a large number of edits to instanceable references (with a shared prototype). Each modified instance causes the prototype's TrackVariability method to be called, which iterates over all instances, resulting in O(n^2) behavior if you modify all n instances.

### Steps to Reproduce
1. Extract the attached file and run "usdview fans_1.usda"
[updateperformancelarge.zip](https://github.com/PixarAnimationStudios/USD/files/7078148/updateperformancelarge.zip)
2. In the interpreter window, do:
l = Sdf.Layer.FindOrOpen('fans_2.usda')
usdviewApi.stage.GetRootLayer().TransferContent(l)
3. Everything updates as desired, but it takes about 5 seconds (on my machine anyway). An equivalent animated layer (usdstitch fans_1 and fans_2 together) takes a tiny fraction of a second to transition from frame 1 to frame 2.

### Package Versions
USD 21.08

jilliene:
Filed as internal issue #USD-6871

PR #1608, opened 2021-08-30
Improve UsdImagingDelegate performance when editing many instances of the same prototype

marktucker:
Address a performance issue when applying a large number of edits to
instanceable references of the same prototype. The TrackVariability call
is O(n) on the number of instances, and it was being called for each
modified instance, resulting in O(n^2) cost to such an edit. This change
globally keeps track of all paths for which we call TrackVariability as
part of an update to avoid calling it more than once for any particular
primitive.

### Fixes Issue(s)
- #1607


jilliene:
Filed as internal issue #USD-6872

PR #1610, opened 2021-09-01
Make mtlx functionality stack relocatable

hshakula:
### Description of Change(s)
Instead of using compile-time constant paths to MaterialX stdlib, evaluate those paths in runtime.

This allows us to move the USD installation directory without the need to set many different additional environment settings that are not that straightforward to use (see https://github.com/PixarAnimationStudios/USD/issues/1586)

This [commit](https://github.com/PixarAnimationStudios/USD/commit/225138be467814849882688e25adcc1f0074d1a6) allegedly fixes this issue. But it still uses compile-time constant...

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/USD/issues/1586

### DIFF
--- a/pxr/imaging/hdSt/CMakeLists.txt
+++ b/pxr/imaging/hdSt/CMakeLists.txt
@@ -21,9 +21,6 @@ set(optionalIncludeDirs "")
 set(optionalPublicClasses "")
 set(optionalPrivateClasses "")
 if (${PXR_ENABLE_MATERIALX_SUPPORT})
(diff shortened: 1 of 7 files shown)


hshakula:
Probably, there is a simpler way to query the USD root directory than in my implementation (`UsdGetRootDir`). Let me know and I will fix it.

jilliene:
Filed as internal issue #USD-6888

klucknav:
Thank you for raising this issue, we have a fix that should be in dev soon.

PR #1623, opened 2021-09-16
'cameraFacing' and 'pixelScale' attributes for HdMesh and HdBasisCurves in hdStorm

vlasovi:
### Description of Change(s)

In order for Autodesk to use HdStorm as a viewport solution in its products, it has to support secondary graphics functionality.

This submission implements 'cameraFacing' and 'pixelScale' attributes for HdMesh and HdBasisCurves in hdStorm.

- A new attribute 'primvars:cameraFacing' was introduced in HdStorm that forces the rendered geometry to be always turned towards the camera.

Video: https://user-images.githubusercontent.com/71097162/133627981-f0cd9208-423a-4f74-bb2d-6c0ca9082b4a.mp4 

- A new attribute 'primvars:pixelScale' was introduced in HdStorm that forces the rendered geometry to have always the same size on the screen, which is independent on its position in 3d space.

Video: https://user-images.githubusercontent.com/71097162/133628187-6b1e40dc-57e3-4784-9da8-b869bbc657f3.mp4

### DIFF
--- a/pxr/imaging/hd/tokens.h
+++ b/pxr/imaging/hd/tokens.h
@@ -40,6 +40,7 @@ PXR_NAMESPACE_OPEN_SCOPE
     (bezier)                                    \
     (bSpline)                                   \
     (camera)                                    \
+    (cameraFacing)                              \
     (catmullRom)                                \
     (collection)                                \
     (computeShader)                             \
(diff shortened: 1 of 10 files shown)


jilliene:
Filed as internal issue #USD-6901

erikaharrison-adsk:
We'd like to close this PR and resubmit an updated implementation. cc @tallytalwar as you seem to track these.

Issue #1449, opened 2021-02-11
Material bindings are not honored on scenegraph instances inside of point instances


asluk:
### Description of Issue
Materials bound to point instancer prototypes are not honored by scenegraph instances within those protoypes.

### Steps to Reproduce

[grass_scatter0.zip](https://github.com/PixarAnimationStudios/USD/files/5966452/grass_scatter0.zip)

1. Unzip the attached archive.
2. Open grass_scatter0/layout/grass_scatter_layout.usda in usdview
3. The spheres all draw with no materials.
4. Select /grass_plain/tree_instancer/Prototypes/tree_1/tree_1_geo/tree_1_leaves/obj_0 in the Browser
5. in the interpreter, usdviewApi.prim.SetInstanceable(False)
6. The spheres will turn blue, now honoring the bound UsdPreviewSurface

### System Information (OS, Hardware)

Windows
### Package Versions

current dev branch

### Build Flags

python build_scripts\build_usd.py _install

jilliene:
Filed as internal issue #USD-6570

PR #1627, opened 2021-09-19
Fix display of materials on instanceable prototype on point instancers.

### LINKED ISSUE #1626: Materials not showing up on point instancer prototypes marked as instanceable
### Description of Issue
When a prototype of a point instancer is an instancebale reference, that prototype doesn't show its bound materials.
A variety of minor tweaks can be made to make the materials show up:
* bind the material to the instanceable prototype prim instead of to the prims inside
* Add a geometry subset to the mesh inside the prototype
* Turn off the instanceable flag on the prototype
This appears to be aregression since 20.08, but I've only tested with 21.05 and 21.08, and both fail, so I believe this broke sometime between 20.08 and 21.05. My suspicion is that it has to do with this change: https://github.com/PixarAnimationStudios/USD/commit/d9fa8e07b78592b43401fe38d1c13f35374a963d

### Steps to Reproduce
1. Open cuberef.usda and cubeinst.usda in usdview from the attached zip file:
[instancematerials.zip](https://github.com/PixarAnimationStudios/USD/files/7192864/instancematerials.zip)
2. Notice that cueref shows a blue cube (from the bound USD preview surface), but cubeinst shows a grey cube (no material).

Tested on Windows, almost certainly affects all.

### Package Versions
Broken in 21.05 and 21.08, issue does not exist in 20.08

marktucker:
### Description of Change(s)

The fix for #1449 looks like it broke material bindings in the case where
a point instancer prototype was marked as instanceable, but the material
was bound inside the prototype (on an instance proxy). The bound material
was being expressed with the instance proxy path instead of the prototype
path, which is how the material is actually identified in hydra. So now
the fix for #1449 is only applied to the instanceable prim, not the
contained instance proxy prims.

In my limited testing, this change seems to fix #1626 without breaking #1449.
But I don't understand this stuff deeply enough to really be sure this is right.

### Fixes Issue(s)
- #1626

### DIFF
--- a/pxr/usdImaging/usdImaging/pointInstancerAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/pointInstancerAdapter.cpp
@@ -387,7 +387,10 @@ UsdImagingPointInstancerAdapter::_PopulatePrototype(
                     populatePrim = _GetPrim(instancerChain.at(1));
                 }
 
-                SdfPath const& materialId = GetMaterialUsdPath(instanceProxyPrim);
+                SdfPath const& materialId =
+                    populatePrim.IsInstance()
+                        ? GetMaterialUsdPath(instanceProxyPrim)
+                        : GetMaterialUsdPath(populatePrim);
                 TfToken const& drawMode = GetModelDrawMode(instanceProxyPrim);
                 TfToken const& inheritablePurpose = 
                     GetInheritablePurpose(instanceProxyPrim);


jilliene:
Filed as internal issue #USD-6908

tcauchois:
Hey Mark,

Do you have a failing example scene we can take a look at?  That would help me think through the consequences of this change.

Thanks,
Tom

marktucker:
Yep, there is a file attached to #1626.

tcauchois:
Thanks as always for the great test cases :).  I have an alternative fix that should be going into the pixar tree by EOD.

tallytalwar:
@tcauchois 's fix e0da0a7f55c26fb4acd7717c431e7ab3755ca710, has been pushed to dev branch. Closing this PR in favor of that change. Thanks

PR #1628, opened 2021-09-20
Bring MaterialX stdlib search paths in HdStorm on par with the ones used in MtlxPlugin

vlasovi:
### Description of Change(s)

HdStorm now uses environment variable "PXR_USDMTLX_STDLIB_SEARCH_PATHS" to set additional MaterialX stdlib search paths, which fully mimics the behaviour of MtlxPlugin.

### Fixes Issue(s)

Partially addresses https://github.com/PixarAnimationStudios/USD/issues/1586 (MaterialX search paths are unstable not relocatable and code gen breaks when relocating libraries)

### DIFF
--- a/pxr/imaging/hdSt/materialXFilter.cpp
+++ b/pxr/imaging/hdSt/materialXFilter.cpp
@@ -28,11 +28,14 @@
 #include "pxr/usd/sdr/registry.h"
 #include "pxr/imaging/hio/glslfx.h"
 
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-6909

klucknav:
Thank you for the change. Unfortunately we cannot directly merge your change as is, there are a handful of changes we would like to make. But we have a fix for the underlying issue based on your change that should be in dev soon. Thanks again.

PR #1633, opened 2021-09-29
support for MaterialX-1.38.1

pmolodo:
### Description of Change(s)
Allows building against MaterialX-1.38.1 or 1.38.0

1.38.1 changes the return type of `getShaderNodes` from an `unordered_set` to a `vector`.

I use `auto` to handle this - I know `auto` isn't generally encouraged at Pixar, but since MaterialX apparently provides [no way to tell the version at compile time](https://github.com/materialx/MaterialX/issues/704), `#define` isn't an option (and I didn't want to use run-time branching, which would result in a lot of duplicated code!)

### Fixes Issue(s)
- Building against MaterialX-1.38.1


jilliene:
Filed as internal issue #USD-6929

PR #1634, opened 2021-09-30
MaterialX: fix for GLSL shader generation if not using bindless textures

pmolodo:
### Description of Change(s)

Currently, if you try to use usdview to view an object with a materialX shader applied, and bindless textures are not enabled, then you will get an error:

```
error C7554: OpenGL requires sampler variables to be explicitly declared as uniform
```

This can be reproduced on the dev branch by building a stock USD, then setting `GLF_ENABLE_BINDLESS_SHADOW_TEXTURES=0` in your environment before launching usdview on any file displaying materialx shaders.

### Fixes Issue(s)
- Ability to view MaterialX in OpenGL if not using bindless textures.

### DIFF
--- a/pxr/imaging/hdSt/materialXShaderGen.cpp
+++ b/pxr/imaging/hdSt/materialXShaderGen.cpp
@@ -21,6 +21,7 @@
 // KIND, either express or implied. See the Apache License for the specific
 // language governing permissions and limitations under the Apache License.
 //
+#include "pxr/imaging/glf/contextCaps.h"
 #include "pxr/imaging/hdSt/materialXShaderGen.h"
 #include "pxr/imaging/hdSt/materialXFilter.h"
 #include "pxr/base/tf/stringUtils.h"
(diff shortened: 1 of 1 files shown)


pmolodo:
This is a bit messy, as it adds more special case branching. Wasn't sure how to do a cleaner version that didn't require much more intrusive changes, though...

pmolodo:
Attaching a simple scene that can be used to show the issue - thanks to Jerry Gamache and Spiff (https://groups.google.com/g/usd-interest/c/8Wy8zi1SN-o/m/YS4P1TFPCAAJ)

Just unzip and do:

Windows:
```
set GLF_ENABLE_BINDLESS_SHADOW_TEXTURES=0
usdview jerry_gamache_mtlx_example.usda
```
 
Linux/MacOS:
```
export GLF_ENABLE_BINDLESS_SHADOW_TEXTURES=0
usdview jerry_gamache_mtlx_example.usda
```

[jerry_gamache_mtlx_example.zip](https://github.com/PixarAnimationStudios/USD/files/7255937/jerry_gamache_mtlx_example.zip)

jilliene:
Filed as internal issue #USD-6939

klucknav:
Thank you for the change. Unfortunately, we cannot directly merge your change as there are a handful of changes we would like to make. But we have a fix for the underlying issue based on your change that should be in dev soon. Thanks again.

pmolodo:
Great! Was by no means enamored of my particular solution - as long as the underlying issue gets fixed!

PR #1635, opened 2021-09-30
removes redundant "vals" namespace from the primvars namespace when r…

BSalem:
removes redundant "vals" namespace from the primvars namespace when reading array primvars from alembic


spiffmon:
Thanks, @BSalem !  If there is no existing test case that changes as a result, would it be possible for you to add one?

BSalem:
would it be enough to add to the PR an alembic archive with two primvars (two color sets, one called vals - just in case! And the other called colorSet), that then read in usdview showing properly named primvars and properly returned values? (actually that's only how I tested it at my end).
Otherwise, guide me to what should the test look like in this case please.

spiffmon:
Sure, @BSalem !  Have a look at [the existing usdAbc tests](https://github.com/PixarAnimationStudios/USD/tree/release/pxr/usd/plugin/usdAbc/testenv), the majority of which are baseline-comparison tests.  So you just take your simple abc file, run it through usdcat to get a usda equivalent (good to verify that it produces the bad output without your fix), and add a py file that effectively reruns usdcat.  I'd maybe look at testUsdAbcIndexedProperties, including what gets added to the CMakeLists.txt to perform the baseline diff.

Thanks!

BSalem:
Hi @spiffmon, I implemented a similar test, please let me know if this is OK

pxr/usd/plugin/usdAbc/CMakeLists.txt
```diff
@@ -205,6 +211,15 @@ pxr_register_test(testUsdAbcIndexedProperties
         USD_ABC_TESTSUFFIX=def
 )
 
+pxr_register_test(testUsdAbcIndexedGeomArb
+        PYTHON
+        COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testUsdAbcIndexedGeomArb"
+        DIFF_COMPARE bad_primvars_namespace_before_fix.usda
```
spiffmon: Shouldn't we be comparing to the "good" , after-fix baseline, here?  And I don't think we need the "bad" version included in the PR - I only mentioned it so that you could have an extra validation that your fix works by running the test against the *good* baseline, *without your fix*, to see that the test fails, a la test-driven development.

Thanks, @BSalem !

BSalem: to be honest, I didn't ever work with tests modules, so I'm not familiar with their dependencies and how it work... I blindly followed the example you shared.
I just pushed an updated one though.

jilliene:
Filed as internal issue #USD-6940

PR #1641, opened 2021-10-04
Fix MaterialX Nodedef Discovery

kxl-adsk:
### Description of Change(s)
Updated the discovery process to request all NodeDefs directly from the library. This fixes PixarAnimationStudios/USD#1629 where implementation were not able to provide their NodeDef, and supports a future where multiple NodeDef could be implemented with the same NodeGraph.

Also fixed issue where the versionless name discovered for ND_standard_surface_surfaceshader was wrong. In this case, the latest version has the versionless name, while the predecessor is version tagged, so the name mapping algorithm was updated to choose the shortest name in the hierarchy.
Updated the parser to also fetch the NodeDef from the library.

Use the correct API to get the complete port interface from the inheritance hierarchy.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1629


jilliene:
Filed as internal issue #USD-6944

Issue #1602, opened 2021-08-26
usd-core failing on Windows + conda env + py3.9


Caenorst:
### Description of Issue
Hi, I tested on two different machines to install and import `usd-core` pypi package and it's failing with the following logs:

```
test_usd.py:21: in <module>
    from pxr import Usd, UsdGeom
C:\Users\myusername\Anaconda3\envs\kaolin_release\lib\site-packages\usd_core-21.8-py3.9-win-amd64.egg\pxr\Usd\__init__.py:24: in <module>
    from pxr import Tf
C:\Users\myusername\Anaconda3\envs\kaolin_release\lib\site-packages\usd_core-21.8-py3.9-win-amd64.egg\pxr\Tf\__init__.py:163: in <module>
    PreparePythonModule()
C:\Users\myusername\Anaconda3\envs\kaolin_release\lib\site-packages\usd_core-21.8-py3.9-win-amd64.egg\pxr\Tf\__init__.py:88: in PreparePythonModule
    module = importlib.import_module(
C:\Users\myusername\Anaconda3\envs\kaolin_release\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ImportError: DLL load failed while importing _tf: The specified module could not be found.
```

#### Remarks
On the same machine there is no problem with `virtualenv`, neither with python version 3.7.

### Steps to Reproduce
1. Install conda
2. Make a conda environment with python 3.9 (I believe it's also failing on 3.8)

```
conda create -n myenv python=3.9
conda activate myenv
```

3. pip install usd-core
4. python -c "from pxr import Usd, UsdGeom"

### System Information (OS, Hardware)
Windows 10

### Package Versions
usd-core == 21.8
conda, python == 3.9

rstelzleni:
Starting with python 3.8 on Windows the logic for finding dlls in the path changed. We had written code to handle this for the python packages, but it sounds like conda is exposing some flaw in that solution. I'm able to reproduce this in Windows conda, with the python 3.8 and 3.9 packages.

rstelzleni:
I haven't been able to completely get to the bottom of this yet, but I can report that setting the environment variable `CONDA_DLL_SEARCH_MODIFICATION_ENABLE` to 1 makes the package work for me. Just randomly googling that env var, it seems like some packages work when this is set but others may not. So, your mileage may vary, but this could be a way to get this working while we look for a permanent fix.

[...]

PR #1642, opened 2021-10-04
fix for importing pxr.Tf on windows anaconda 3.8+ interpreters

pmolodo:
cpython changed the way the dlls are loaded on windows for python-3.8+.  USD implemented a workaround for this change, installed into `pxr.__init__` (and `Tf.__init__`).  However, anaconda python interpreters were patched to behave more like pre-3.8:

   https://github.com/conda-forge/python-feedstock/blob/8195ba1178041b7461238e8c5680eee62f5ea9d0/recipe/patches/0020-Add-CondaEcosystemModifyDllSearchPath.patch#L37

Note: I'm not crazy about adding code that does a filesystem operation (`os.path.exists`) when the `pxr` module is imported.  However, it [seems to be the most robust way to check for conda](https://stackoverflow.com/a/47610844/920545).  The best alternative I could come up with was to search through the `CDLL.__init__` code object varnames:

```python
import ctypes
is_conda = 'LOAD_WITH_ALTERED_SEARCH_PATH' in ctypes.CDLL.__init__.__code__.co_varnames
```

...but for some reason that feels more fragile to me.  I can switch to use that approach if you'd prefer, though.

### Description of Change(s)

We now check for anaconda interpreters before using the 3.8+ only fix.

### Fixes Issue(s)
- #1602

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -603,10 +603,15 @@ platforms to avoid issues with Boost config files (introduced in Boost version
 to use Boost specified config files for their USD build, specify 
 -DBoost_NO_BOOST_CMAKE=OFF when running cmake.
 
-2. Windows and Python 3.8+
+2. Windows and Python 3.8+ (non-Anaconda)
 Python 3.8 and later on Windows will no longer search PATH for DLL dependencies.
 Instead, clients can call `os.add_dll_directory(p)` to set paths to search.
 By default on that platform USD will iterate over PATH and add all paths using
 `os.add_dll_directory()` when importing Python modules. Users may override
 this by setting the environment variable `PXR_USD_WINDOWS_DLL_PATH` to a PATH-like
 string. If this is set, USD will use these paths instead.
+
+Note that the above does not apply to Anaconda python 3.8+ interpreters, as they
+are modified to behave like pre-3.8 python interpreters, and so continue to use
+the PATH for DLL dependencies.  When running under Anaconda users should
+configure their system the same way they did for pre-python 3.8.
--- a/build_scripts/pypi/package_files/setup.py
+++ b/build_scripts/pypi/package_files/setup.py
@@ -74,8 +74,15 @@ import os, sys
 dllPath = os.path.split(os.path.realpath(__file__))[0]
 if sys.version_info >= (3, 8, 0):
     os.environ['PXR_USD_WINDOWS_DLL_PATH'] = dllPath
-else:
-    os.environ['PATH'] = dllPath + os.pathsep + os.environ['PATH']
+# Note that we ALWAYS modify the PATH, even for python-3.8+. This is because:
+#    - Anaconda python interpreters are modified to use the old, pre-3.8, PATH-
+#      based method of loading dlls
+#    - extra calls to os.add_dll_directory won't hurt these anaconda
+#      interpreters
+#    - similarly, adding the extra PATH entry shouldn't hurt standard python
+#      interpreters
+#    - there's no canonical/bulletproof way to check for an anaconda interpreter
+os.environ['PATH'] = dllPath + os.pathsep + os.environ['PATH']
 ''')
 
 # Get the readme text


meshula:
This will accommodate conda, if conda is in the path, but overridden by another python with higher path-precedence. I was debugging an issue around that exact scenario today... What about adding a --conda option to build_usd.py itself to remove any heuristics, and relying on the end user to supply the information instead?

sunyab:
A `--conda` option won't work for PyPI packages since (AFAIK) there's no difference between the package downloaded by pip in a conda environment vs. a standard environment.

meshula:
I should have read the linked issue :) My mind went to running the build_usd.py script under conda. FWIW the test build that I did while not reading #1602 succeeded under conda/3.9 LOL

pmolodo:
> This will accommodate conda, if conda is in the path, but overridden by another python with higher path-precedence.

If we're worried about mixed conda / non-conda scenarios `PYTHONPATH` scenarios - then maybe we should use the `ctypes.CDLL` check I mentioned above?  Since that's more directly related to the change we're interested in...

> FWIW the test build that I did while not reading #1602 succeeded under conda/3.9 LOL

 Meaning a test build using the stock USD? Or a test build using this PR?

meshula:
Yes, a test build using stock USD, under conda, from top of dev branch without a patch.

The reason I am concerned about the mixed conda environment is due to the fact that on mac the system has always got a python 2.7 lurking in the path. Normally it's completely masked by conda in the path of course, but a shell script I encountered today referenced /usr/bin/python by full path, and things got weird.

This sort of scenario has to be much less problematic on Windows, since there's no one magic place to rely on for a canonical python installation.

rstelzleni:
I've seen other projects use `'conda' in sys.version` to detect when they're running in a conda python interpreter. The stackoverflow thread linked above says that might be fragile because they could rename their version string, but it seems like they could just as easily change their build setup to not create a folder named conda-meta.

Side note, it might be good to add a test to the pypi azure pipeline that uses a conda environment instead of a default installed python.

BUILDING.md
```diff
@@ -603,10 +603,12 @@ platforms to avoid issues with Boost config files (introduced in Boost version
...
+2. Windows and Python 3.8+ (non-Anaconda)
 Python 3.8 and later on Windows will no longer search PATH for DLL dependencies.
 Instead, clients can call `os.add_dll_directory(p)` to set paths to search.
 By default on that platform USD will iterate over PATH and add all paths using
 `os.add_dll_directory()` when importing Python modules. Users may override
 this by setting the environment variable `PXR_USD_WINDOWS_DLL_PATH` to a PATH-like
-string. If this is set, USD will use these paths instead.
+string. If this is set, USD will use these paths instead.  (Note that Anaconda
+python interpreters are modified to behave like pre-3.8 python interpreters,
+and so they continue to use the PATH.)
```
rstelzleni: I don't think this comment is accurate, with this change it will be true for the pypi packages but not for builds in general. If we want to make this something that affects all builds, we could move the new logic into `Tf/__init__.py`.

pmolodo:
> I've seen other projects use `'conda' in sys.version` to detect when they're running in a conda python interpreter. The stackoverflow thread linked above says that might be fragile because they could rename their version string, but it seems like they could just as easily change their build setup to not create a folder named conda-meta.
> 

The problem is that the `'conda' in sys.version` check is already broken - in my conda 3.9.7, this is my sys.version:

```
'3.9.7 (default, Sep 16 2021, 16:59:28) [MSC v.1916 64 bit (AMD64)]'
```

BUILDING.md
rstelzleni: Talked to @pmolodo out of band, and my note about changing it in Tf is not right. Basically anything we do there won't have any effect under conda unless CONDA_DLL_SEARCH_MODIFICATION_ENABLE is set. I think what he was trying to say here is that conda just doesn't do the new thing so users need to configure the old way, but I misunderstood.

rstelzleni:
> '3.9.7 (default, Sep 16 2021, 16:59:28) [MSC v.1916 64 bit (AMD64)]'

ha, well ok then. Already broken

BUILDING.md
pmolodo: I pushed a new version that hopefully clarifies the wording.  Thanks for pointing this out as potentially confusing!

pmolodo:
So, have another potential way of checking for conda:

```python
import os
import sys
CONDA_ENV_VAR = "CONDA_DLL_SEARCH_MODIFICATION_ENABLE"

def is_windows_conda():
    try:
        import _winapi
        lib_path = _winapi.GetModuleFileName(sys.dllhandle)
        with open(lib_path, "rb") as f:
            lib_bytes = f.read()
    except Exception:
        return False
    # windows uses utf-16 encoding in it's binaries, with leading byte-order
    # mark stripped off
    env_var_bytes = CONDA_ENV_VAR.encode("utf16")[2:]
    return env_var_bytes in lib_bytes

def using_py38_dll_loading():
    return (sys.version_info >= (3, 8, 0) and (not is_windows_conda()
                                               or CONDA_ENV_VAR in os.environ))

print(using_py38_dll_loading())
```

This checks, ie, `python39.dll` for the binary string of `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`, which should obviously only be there for conda-modified interpreters.  The advantage of this is that is a value that is unlikely to change, as doing so would break code for those relying on it, and we're checking for something directly tied to the feature we care about.  It should also be fairly accurate, even in "mixed path" scenarios (such as when sys.executable and the underlying python library are coming from different places, which I've seen!)  Downside is that we have to rely on details like how windows encodes strings into dlls.  Also, we're relying on being able to read the dll from the filesystem - so, for instance, if the entire python install is zipped, this won't work.  But that would also cause problems for the `conda-meta` directory check.

pmolodo:
I modified the PR to use the above method of checking for conda, since I think it's the best option thus far.  Let me know if you disagree - I don't feel too strongly about this.

pmolodo:
On further thought - the simplest solution might be to not even bother checking for conda, and just ALWAYS modify `PATH`... and then additionally set `PXR_USD_WINDOWS_DLL_PATH` if we're `>= 3.8`.  The advantage here is this should "just work" regardless of which python interpreter we're using - since extra calls to `os.add_dll_directory` shouldn't do anything to the conda interpreter, and the extra add to the `PATH` won't hurt non-conda 3.8 (I mean, if it didn't do any harm for <3.7, it shouldn't hurt now.  You could even make the case that by continuing to add to the path, we're keeping more consistent behavior, if for some reason someone was relying on some other binary they've added to that folder.)

pmolodo:
Ok - did another push where I just went with the always-modify-the PATH approach. I think it's the safest, and it's definitely the simplest solution.

sunyab:
Filed as internal issue #USD-6948

pmolodo:
Hi - now that 21.11 is out the door, just wanted to ping here to see if we could get this into the next release. Thanks!

spiffmon:
We are trying to get a conda setup so that we can test, @pmolodo !

PR #1647, opened 2021-10-07
Enable selection highlighting for MaterialX shaders in HdStorm

vlasovi:
### Description of Change(s)

Emit shader code that enables selection highlightin during MaterialX shader generation.

### Fixes Issue(s)

- The submission fixes the bug in HdStorm where there was no selection highlighting for geomteric primitives with MaterialX materials applied.


jilliene:
Filed as internal issue #USD-6955

PR #1649, opened 2021-10-11
Only find program_options when building tools

charlesfleche:
### Description of Change(s)

Boost program_options was always checked, regardless of PXR_BUILD_USD_TOOLS's status. That would force program_options to be available, even if not actually used.

This patch ensures program_options is looked for only when the USD tools are being built.

### Fixes Issue(s)
-


jilliene:
Filed as internal issue #USD-6957

PR #1653, opened 2021-10-12
Ensure light collections are dirtied on creation

aloysbaillet:
### Description of Change(s)

We found that when undoing the deletion of a light with an active collection, the re-creation of that light was not dirtying the content of the collection.

It is unfortunately quite tricky to provide a test for this use case, but we have one internally in Omniverse...
Our test goes like this:
* create 2 cubes
* create a light
* set the light's lightLink collection to refer to cube2
* check in viewport with a light-collection-enabled renderer that things are OK
* delete the light
* undo that deletion
* check in viewport with a light-collection-enabled renderer that things are OK again (it wasn't without this fix)

### DIFF
--- a/pxr/usdImaging/usdImaging/collectionCache.cpp
+++ b/pxr/usdImaging/usdImaging/collectionCache.cpp
@@ -38,10 +38,29 @@ _IsQueryTrivial(UsdCollectionAPI::MembershipQuery const& query)
         ruleMap.begin()->second == UsdTokens->expandPrims;
 }
 
+void
+UsdImaging_CollectionCache::_MarkCollectionContentDirty(UsdStageWeakPtr const& stage, 
+    UsdCollectionAPI::MembershipQuery const& query)
+{
+    if(query.HasExcludes())
+    {
+        // If there are any exlusion we have to consider all previously excluded paths dirty
+        // As there is no API to query excluded paths we mark all prims in the Stage dirty
+        _allPathsDirty = true;
+    }
+    else
+    {
+        SdfPathSet linkedPaths = UsdComputeIncludedPathsFromCollection(query, stage);
+        std::merge(_dirtyPaths.begin(), _dirtyPaths.end(), linkedPaths.begin(), linkedPaths.end(),
+            std::inserter(_dirtyPaths, _dirtyPaths.begin()));
+    }
+}
+
 TfToken
 UsdImaging_CollectionCache::UpdateCollection(UsdCollectionAPI const& c)
 {
-    RemoveCollection(c);
+    const UsdStageWeakPtr& stage = c.GetPrim().GetStage();
+    RemoveCollection(stage, c.GetCollectionPath());
 
     std::lock_guard<std::mutex> lock(_mutex);
 
@@ -78,15 +97,16 @@ UsdImaging_CollectionCache::UpdateCollection(UsdCollectionAPI const& c)
     _pathsForQuery[query].insert(path);
     _idForPath[path] = id;
 
+    _MarkCollectionContentDirty(stage, query);
+    
     return id;
 }
 
 void
-UsdImaging_CollectionCache::RemoveCollection(UsdCollectionAPI const& c)
+UsdImaging_CollectionCache::RemoveCollection(UsdStageWeakPtr const& stage, SdfPath const& path)
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    SdfPath path = c.GetCollectionPath();
     auto const& pathEntry = _idForPath.find(path);
     if (pathEntry == _idForPath.end()) {
         // No pathEntry -- bail.  This can happen if the collection was
@@ -102,6 +122,9 @@ UsdImaging_CollectionCache::RemoveCollection(UsdCollectionAPI const& c)
         return;
     }
     UsdCollectionAPI::MembershipQuery const& queryRef = queryEntry->second;
+    _idForQuery.erase(queryRef);
+
+    _MarkCollectionContentDirty(stage, queryRef);
 
     auto const& pathsForQueryEntry = _pathsForQuery.find(queryRef);
     pathsForQueryEntry->second.erase(path);
@@ -113,12 +136,30 @@ UsdImaging_CollectionCache::RemoveCollection(UsdCollectionAPI const& c)
     // This also reaps the associated identifier.
     if (pathsForQueryEntry->second.empty()) {
         _pathsForQuery.erase(pathsForQueryEntry);
-        _idForQuery.erase(queryRef);
         _queryForId.erase(queryEntry);
         TF_DEBUG(USDIMAGING_COLLECTIONS)
             .Msg("UsdImaging_CollectionCache: Dropped id '%s'", id.GetText());
     }
 };
+
+SdfPathSet const&
+UsdImaging_CollectionCache::GetDirtyPaths() const
+{
+    return _dirtyPaths;
+}
+
+bool
+UsdImaging_CollectionCache::AreAllPathsDirty() const
+{
+    return _allPathsDirty;
+}
+
+void
+UsdImaging_CollectionCache::ClearDirtyPaths()
+{
+    _dirtyPaths.clear();
+    _allPathsDirty = false;
+}
 
 TfToken
 UsdImaging_CollectionCache::GetIdForCollection(UsdCollectionAPI const& c)
--- a/pxr/usdImaging/usdImaging/collectionCache.cpp
+++ b/pxr/usdImaging/usdImaging/collectionCache.cpp
@@ -42,16 +42,9 @@ void
 UsdImaging_CollectionCache::_MarkCollectionContentDirty(UsdStageWeakPtr const& stage, 
     UsdCollectionAPI::MembershipQuery const& query)
 {
-    if(query.HasExcludes())
-    {
-        // If there are any exlusion we have to consider all previously excluded paths dirty
-        // As there is no API to query excluded paths we mark all prims in the Stage dirty
-        _allPathsDirty = true;
-    } else {
-        SdfPathSet linkedPaths = UsdComputeIncludedPathsFromCollection(query, stage);
-        std::merge(_dirtyPaths.begin(), _dirtyPaths.end(), linkedPaths.begin(), linkedPaths.end(),
-            std::inserter(_dirtyPaths, _dirtyPaths.begin()));
-    }
+    SdfPathSet linkedPaths = UsdComputeIncludedPathsFromCollection(query, stage);
+    std::merge(_dirtyPaths.begin(), _dirtyPaths.end(), linkedPaths.begin(), linkedPaths.end(),
+        std::inserter(_dirtyPaths, _dirtyPaths.begin()));
 }
 
 bool
@@ -97,16 +90,18 @@ UsdImaging_CollectionCache::UpdateCollection(UsdCollectionAPI const& c)
     _idForPath[path] = id;
 
     _MarkCollectionContentDirty(stage, query);
-    
+    // Also add the light in the dirty set for it to be marked as collection dirty
+    _dirtyPaths.insert(c.GetPrim().GetPath());
+
     return changed;
 }
 
 size_t
-UsdImaging_CollectionCache::RemoveCollection(UsdStageWeakPtr const& stage, SdfPath const& path)
+UsdImaging_CollectionCache::RemoveCollection(UsdStageWeakPtr const& stage, SdfPath const& collectionPath)
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    auto const& pathEntry = _idForPath.find(path);
+    auto const& pathEntry = _idForPath.find(collectionPath);
     if (pathEntry == _idForPath.end()) {
         // No pathEntry -- bail.  This can happen if the collection was
         // trivial; see _IsQueryTrivial().
@@ -125,12 +120,13 @@ UsdImaging_CollectionCache::RemoveCollection(UsdStageWeakPtr const& stage, SdfPa
     _idForQuery.erase(queryRef);
 
     _MarkCollectionContentDirty(stage, queryRef);
+    _dirtyPaths.insert(collectionPath.GetPrimPath());
 
     auto const& pathsForQueryEntry = _pathsForQuery.find(queryRef);
-    pathsForQueryEntry->second.erase(path);
+    pathsForQueryEntry->second.erase(collectionPath);
     TF_DEBUG(USDIMAGING_COLLECTIONS)
         .Msg("UsdImaging_CollectionCache: Id '%s' disused <%s>\n",
-             id.GetText(), path.GetText());
+             id.GetText(), collectionPath.GetText());
 
     // Reap _pathsForQuery entries when the last path is removed.
     // This also reaps the associated identifier.
@@ -138,7 +134,7 @@ UsdImaging_CollectionCache::RemoveCollection(UsdStageWeakPtr const& stage, SdfPa
         _pathsForQuery.erase(pathsForQueryEntry);
         _queryForId.erase(queryEntry);
         TF_DEBUG(USDIMAGING_COLLECTIONS)
-            .Msg("UsdImaging_CollectionCache: Dropped id '%s'", id.GetText());
+            .Msg("UsdImaging_CollectionCache: Dropped id '%s'\n", id.GetText());
     }
     return hash;
 };
@@ -149,17 +145,10 @@ UsdImaging_CollectionCache::GetDirtyPaths() const
     return _dirtyPaths;
 }
 
-bool
-UsdImaging_CollectionCache::AreAllPathsDirty() const
-{
-    return _allPathsDirty;
-}
-
 void
 UsdImaging_CollectionCache::ClearDirtyPaths()
 {
     _dirtyPaths.clear();
-    _allPathsDirty = false;
 }
 
 TfToken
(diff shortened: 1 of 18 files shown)


spitzak:
We ran into this as well, but it was pretty simple to have the renderDelegate update things due to the creation of the light and the calling of Sync() on it, so I am not absolutely certain this dirtying is needed

aloysbaillet:
Thanks Bill, I'm still discovering the Hydra world... but in our render delegate the Light::Sync was indeed called, but each linked geometry rprim wasn't and we need them called to re-register them with the light link. I'm curious how you made it work?

spitzak:
This is lame but in Light::Sync() it checks if the categories changed and
if so it calls
sceneDelegate->GetRenderIndex().GetChangeTracker().MarkAllRprimsDirty(pxr::HdChangeTracker::DirtyCategories);
which calls Sync() on all the Geometry.

aloysbaillet:
Fair enough :)
I wanted to avoid marking all prims dirty though, the proposed change only marks the collection members dirty...

jilliene:
Filed as internal issue #USD-6961

tcauchois:
Hey Aloys,

Like I mentioned on your other PR, I'm a bit hesitant to have special case code in delegate.cpp since it becomes hard to maintain.  There's also a few bits you're missing.  This will be a bit involved to fix but I'll try to lay out the pieces and answer any questions.

First up, we have an object UsdImaging_CollectionCache that actually answers all of the queries, both GetLightParamValue(lightLink) and GetCategories(rprim).  Here's the class: https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/collectionCache.h ... and here's where we populate it: https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/lightAdapter.cpp#L131 ...  In order to keep the collection cache up to date with any changes, I'd recommend updating lightAdapter.cpp and lightFilterAdapter.cpp to call UpdateCollection in TrackVariability (which is already there); call UpdateCollection in ProcessPropertyChange if the collection changes (and return HdLight::DirtyParams); and call RemoveCollection in _RemovePrim.  This way the results of any calls into the collection cache will always be up-to-date.

The next piece you need is hydra dirty bit tracking.  Ideally, you could do this when you call UpdateCollection; the collection cache removes a bunch of gprims from the collection list, and adds a bunch of other ones, and you could just produce a list of paths to mark with DirtyCategories.  However, TrackVariability (for example) is called multi-threaded and can't call the hydra change-tracker, which isn't threadsafe.  And neither TrackVariability nor ProcessPropertyChange have a UsdImagingIndexProxy, which is the class needed to call MarkDirty.

The way out of this that I see is for the collection cache to keep a dirty gprim list.  Every time you call RemoveCollection or UpdateCollection, add hydra paths to an SdfPathSet (or Vector) on the collection cache itself.  At the end of UsdImagingDelegate::ApplyPendingUpdates(), which is our function that applies USD edits to imaging, loop through this set calling MarkDirty(DirtyCategories); I'm imagining putting that loop here https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/delegate.cpp#L1040 ... right after _ExecuteWorkForVariabilityUpdate.  Or the other option, of course, is Bill's approach of marking all rprims dirty with DirtyCategories whenever the collections change.  That's faster if your light collections tend to be big.  (Note that the trivial collection of "/" avoids all of these codepaths entirely).

That should do a much better job of handling the category change-tracking.

Hopefully this makes sense!  Any work to fix this up is super appreciated, and I'm happy to answer whatever questions you have about the code.

Thanks,
Tom

aloysbaillet:
Thanks Tom, I'll look into this!

pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -1348,10 +1348,35 @@ UsdImagingDelegate::_ResyncUsdPrim(SdfPath const& usdPath,
...
+                        TF_VERIFY(primInfo->adapter, "%s", affectedCachePath.GetText())) {
+                        UsdImagingPrimAdapterSharedPtr& adapter = primInfo->adapter;
+                        adapter->MarkDirty(primInfo->usdPrim, affectedCachePath,
+                            HdChangeTracker::DirtyCategories, proxy);
+                    }
+                }
+            };
+            UsdLuxLightAPI light(usdPrim);
+            UsdCollectionAPI lightLink{ light.GetLightLinkCollectionAPI() };
+            _MarkLinkPathsDirty(lightLink.ComputeIncludedPaths(lightLink.ComputeMembershipQuery(), _stage));
```
tcauchois: I think the "Compute" here won't go through the collection cache, right?  Is there a way to do that?

pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -1082,6 +1082,39 @@ UsdImagingDelegate::ApplyPendingUpdates()
...
     _ExecuteWorkForVariabilityUpdate(&worker);
+
+    // Mark all dirty collection prims
+    if (_collectionCache.AreAllPathsDirty())
+    {
+        TRACE_SCOPE("Mark all dirty collection members");
+        TF_FOR_ALL(it, _hdPrimInfoMap) {
+            UsdImagingPrimAdapterSharedPtr& adapter = it->second.adapter;
+            TF_DEBUG(USDIMAGING_CHANGES).Msg("[Update]: invalidate collection member prim (all paths dirty) %s\n", it->first.GetText());
+            adapter->MarkDirty(it->second.usdPrim, it->first,
```
tcauchois: We should only do this for gprims.  The standard way to do this is to add a "MarkCollectionsDirty" on prim adapter that does nothing, and an override on gprim adapter that calls MarkDirty(DirtyCategories).

"DirtyCollections" doesn't mean anything on other types of prim and will invalidate a random attribute.

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -177,6 +216,9 @@ UsdImagingLightAdapter::ProcessPropertyChange(UsdPrim const& prim,
     if (UsdGeomXformable::IsTransformationAffectedByAttrNamed(propertyName)) {
         return HdLight::DirtyBits::DirtyTransform;
     }
+
+    _UpdateCollectionsChanged(prim, cachePath);
```
tcauchois: We'd want to guard this in "if (propertyName == lightLinkAttrToken)" or something, not sure what the right token is offhand.

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -127,11 +163,14 @@ UsdImagingLightAdapter::TrackVariability(UsdPrim const& prim,
...
-        UsdImaging_CollectionCache &collectionCache = _GetCollectionCache();
-        collectionCache.UpdateCollection(light.GetLightLinkCollectionAPI());
-        collectionCache.UpdateCollection(light.GetShadowLinkCollectionAPI());
-        // TODO: When collections change we need to invalidate affected
-        // prims with the DirtyCollections flag.
+        if (_UpdateCollectionsChanged(prim, cachePath))
+        {
+            *timeVaryingBits |= HdLight::DirtyBits::DirtyCollection;
+        }
+        else
```
tcauchois: Can skip the else here.

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -80,6 +80,42 @@ UsdImagingLightAdapter::_RemovePrim(SdfPath const& cachePath,
...
+    UsdImaging_CollectionCache &collectionCache = _GetCollectionCache();
+    auto getCollectionHash = [&collectionCache] (const UsdCollectionAPI& api) -> size_t {
+        const TfToken id = collectionCache.UpdateCollection(api);
+        const UsdImaging_CollectionCache::Query* query = nullptr;
+        collectionCache.GetMembershipQuery(id, &query);
+        return query != nullptr ? query->GetHash() : 0;
+    };
+    UsdLuxLight light(prim);
+    const size_t newLightCollectionHash = getCollectionHash(light.GetLightLinkCollectionAPI());
+    const size_t newShadowCollectionHash = getCollectionHash(light.GetShadowLinkCollectionAPI());
```
tcauchois: We might have gone over this already, but why are we hashing the collections?  Can we not have "UpdateCollection" tell us whether the membership has changed?

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -80,6 +80,42 @@ UsdImagingLightAdapter::_RemovePrim(SdfPath const& cachePath,
                                     UsdImagingIndexProxy* index)
 {
     index->RemoveSprim(HdPrimTypeTokens->light, cachePath);
+    UsdImaging_CollectionCache &collectionCache = _GetCollectionCache();
+    SdfPath lightLinkPath = cachePath.AppendProperty(UsdImagingTokens->collectionLightLink);
+    collectionCache.RemoveCollection(GetDelegate()->GetStage(), lightLinkPath);
+    SdfPath shadowLinkPath = cachePath.AppendProperty(UsdImagingTokens->collectionShadowLink);
+    collectionCache.RemoveCollection(GetDelegate()->GetStage(), shadowLinkPath);
```
tcauchois: Might want to remove _collectionHashes[cachePath] here...

pxr/usdImaging/usdImaging/collectionCache.cpp
```diff
@@ -78,15 +97,16 @@ UsdImaging_CollectionCache::UpdateCollection(UsdCollectionAPI const& c)
     _pathsForQuery[query].insert(path);
     _idForPath[path] = id;
 
+    _MarkCollectionContentDirty(stage, query);
+
```
tcauchois: I feel like there's some missing bits here...  If _queryForId[id] before you update it != query, also _MarkContentDirty(stage, _queryForId[id])?

pxr/usdImaging/usdImaging/delegate.h
```diff
@@ -520,6 +520,12 @@ class UsdImagingDelegate : public HdSceneDelegate, public TfWeakBase {
     USDIMAGING_API
     bool IsInInvisedPaths(const SdfPath &usdPath) const;
 
+    USDIMAGING_API
```
tcauchois: I'd add this as protected API on primAdapter.cpp instead, if that's ok.

pxr/usdImaging/usdImaging/lightAdapter.cpp
tcauchois: Actually, can just call _UpdateCollectionsChanged.  I don't think collections are allowed to be time-varying?  And if you did want to check that, instead of seeing if the collection has new contents, you'd want to ask whether the attributes are time varying.

pxr/usdImaging/usdImaging/collectionCache.cpp
```diff
@@ -38,10 +38,29 @@ _IsQueryTrivial(UsdCollectionAPI::MembershipQuery const& query)
...
 
+void
+UsdImaging_CollectionCache::_MarkCollectionContentDirty(UsdStageWeakPtr const& stage, 
+    UsdCollectionAPI::MembershipQuery const& query)
+{
+    if(query.HasExcludes())
+    {
+        // If there are any exlusion we have to consider all previously excluded paths dirty
+        // As there is no API to query excluded paths we mark all prims in the Stage dirty
+        _allPathsDirty = true;
```
tcauchois: Even if the query has exclude paths, I'd still expect UsdComputeIncludedPathsFromCollection to come up with the right list.  That should return the set of paths that pass both the include & exclude rules, right?

pxr/usdImaging/usdImaging/lightAdapter.cpp
aloysbaillet: Indeed! Would this be OK? Some string parsing as the attributes can be numerous and are all under the collection:lightLink/shadowLink prefix:
```
    if (TfStringStartsWith(propertyName.GetString(), UsdImagingTokens->collectionShadowLink.GetString()) || 
        TfStringStartsWith(propertyName.GetString(), UsdImagingTokens->collectionLightLink.GetString()))
    {
        _UpdateCollectionsChanged(prim, cachePath);
    }
```

tcauchois: Yeah, that would be great!

pxr/usdImaging/usdImaging/collectionCache.cpp
aloysbaillet: Indeed, but prims that were previously excluded from the collection would not be returned in the include set, and these also need to be marked dirty...

pxr/usdImaging/usdImaging/collectionCache.cpp
aloysbaillet: The `RemoveCollection` call at the beginning of `UpdateCollection` will mark the previous collection content dirty, wouldn't this cover it?

pxr/usdImaging/usdImaging/lightAdapter.cpp
aloysbaillet: Indeed! Would this signature be OK?
```
TfToken
UsdImaging_CollectionCache::UpdateCollection(UsdCollectionAPI const& c, bool* changed)
```

aloysbaillet: Actually I'll just return a bool instead of the TfToken as the return value isn't needed anymore...

pxr/usdImaging/usdImaging/lightAdapter.cpp
aloysbaillet: Good point, but not needed anymore as I'm not storing the hashes any longer!

pxr/usdImaging/usdImaging/delegate.cpp
aloysbaillet: This code is now deleted

pxr/usdImaging/usdImaging/lightAdapter.cpp
aloysbaillet: I removed the call altogether as the attributes cannot be animated... I'm guessing that if timesamples are added on the includeRoot bool that could cause some issues though?

pxr/usdImaging/usdImaging/collectionCache.cpp
aloysbaillet: Actually as the RemoveCollection call will catch all paths from the outgoing collection, the UpdateCollection should be able to safely ignore the excluded paths... Will test that further!

aloysbaillet:
I've cleaned and simplified the code further, also fixed some formatting inconsistencies.
The `UsdImagingLightAdapter::ProcessPropertyChange` method now returns `HdLight::DirtyBits::DirtyCollection` when it detects a collection change (using hashes for comparison, but without having to store them).

I'm still having one refresh issue that I'm chasing but I'm confident it is a problem in our render delegate (I had to change it quite a lot to handle `HdLight::DirtyCollection` properly).
I'm quite confident the prman render delegate will need updating to take advantage of the now-working DirtyCollection flag, but I'm not sure if I'll be able to do that (I think I would need a license first ;) ).

aloysbaillet:
I've fixed a compilation error (`UsdLuxLight`->`UsdLuxLightAPI` as I was testing on an older USD version sorry!) and @sirpalee kindly tested these changes with success using the Arnold render delegate.

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -71,15 +71,50 @@ UsdImagingLightAdapter::Populate(UsdPrim const& prim,
...
+    collectionCache.RemoveCollection(_GetStage(), shadowLinkPath);
+}
+
+void
+UsdImagingLightAdapter::_RegisterLightCollections(UsdPrim const& prim) {
+    _UpdateCollectionsChanged(prim);
+}
+
+bool
+UsdImagingLightAdapter::_UpdateCollectionsChanged(UsdPrim const& prim) const {
```
tcauchois: Probably just a merge issue, but _UpdateCollectionsChanged shows up twice in this file (seemingly with the same implementation).

tcauchois: Probably just a merge issue, but _UpdateCollectionsChanged shows up twice in this file (seemingly with the same implementation).

aloysbaillet: oh wow, sorry about this!

pxr/usdImaging/usdImaging/lightAdapter.cpp
```diff
@@ -71,15 +71,50 @@ UsdImagingLightAdapter::Populate(UsdPrim const& prim,
 {
     index->InsertSprim(HdPrimTypeTokens->light, prim.GetPath(), prim);
     HD_PERF_COUNTER_INCR(UsdImagingTokens->usdPopulatedPrimCount);
+    _RegisterLightCollections(prim);
 
     return prim.GetPath();
 }
 
 void
-UsdImagingLightAdapter::_RemovePrim(SdfPath const& cachePath,
```
tcauchois, line 79: Did you mean to remove this?  It's still marked overridden in the header, so it's causing build issues.

tcauchois, line 79: Did you mean to remove this?  It's still marked overridden in the header, so it's causing build issues.

aloysbaillet: Not really sorry about this, the USD version I've been testing with didn't have that method implemented!

Issue #1658, opened 2021-10-17
HdMtlx does not work with Houdini's mtlx UsdShade graphs


hshakula:
### Description of Issue
Using `HdMtlxCreateMtlxDocumentFromHdNetwork` on `HdNetwork` when `UsdShade` graph is constructed from Houdini's Mtlx nodes leaks `MaterialX::Exception` due to elements with duplicate names.

### Steps to Reproduce
1. Set up material with 2 or more Mtlx nodes in Houdini
2. Pass resulting `HdNetwork` to `MaterialX::Document` using `HdMtlxCreateMtlxDocumentFromHdNetwork`

spiffmon:
Hi @hshakula , it would be helpful if you could attach a usda file that illustrates the problem, which would help identify whether it is actually a Solaris issue or an Hd issue.

Thanks!

hshakula:
[mtlxTest.zip](https://github.com/PixarAnimationStudios/USD/files/7362857/mtlxTest.zip)

My proposed solution is #1659

jilliene:
Filed as internal issue #USD-6972

hshakula:
Fixed in #1659

PR #1659, opened 2021-10-18
[HdMtlx] Fix duplicate name error when processing HdNetwork from Houdini

hshakula:
### Description of Change(s)
Resolve duplicate names by appending a suffix to nodegraph that encapsulates HdNetwork nodes.
The reason why there are duplicates is that [this name](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/imaging/hdMtlx/hdMtlx.cpp#L280) and [this one](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/imaging/hdMtlx/hdMtlx.cpp#L335) are the same. It leads to attempting to create nodegraph with the same name as the already created material node.

### Fixes Issue(s)
#1658

### DIFF
--- a/pxr/imaging/hdMtlx/hdMtlx.cpp
+++ b/pxr/imaging/hdMtlx/hdMtlx.cpp
@@ -277,7 +277,7 @@ _GatherUpstreamNodes(
 
     // Initilize the mxNodeGraph if needed
     if (!(*mxNodeGraph)) {
-        const std::string & nodeGraphName  = hdNodePath.GetParentPath().GetName();
+        std::string nodeGraphName = hdNodePath.GetParentPath().GetName() + "_Graph";
         *mxNodeGraph = mxDoc->addNodeGraph(nodeGraphName);
     }
     


jilliene:
Filed as internal issue #USD-6973

pxr/imaging/hdMtlx/hdMtlx.cpp
```diff
@@ -277,7 +277,7 @@ _GatherUpstreamNodes(
 
     // Initilize the mxNodeGraph if needed
     if (!(*mxNodeGraph)) {
-        const std::string & nodeGraphName  = hdNodePath.GetParentPath().GetName();
+        std::string nodeGraphName = hdNodePath.GetParentPath().GetName() + "_Graph";
```
bernardkwok, line 280: As a suggestion here, there is a `makeUniqueChildName()` method which can be called on `mxDoc`, and should be more robust than appending "_Graph" -- in case this is also a duplicate.

hshakula: Seems like you mean `createValidChildName`. Yeah we can use it here

PR #1664, opened 2021-10-21
HdStorm: Move MaterialX parameter values out of the shader and into parameter buffer

vlasovi:
### Description of Change(s)

- MaterialX input parameters are no longer hardcoded in the shader but set from the parameter buffer
- Support float and integer types with sizes between 1 and 4. Make sure that all other types are still hardcoded, so that there are no gaps

### Fixes Issue(s)

- This change is the first step (out of two) to make MaterialX parameter updates interactive in HdStorm.

- The second step will be in fixing this issue in MaterialX code: https://github.com/materialx/MaterialX/issues/710 (Shader code generated by MaterialX is not deterministic)

- With the two steps in, shader source code caching in HdStorm will work as expected, HdStorm will be no longer recompiling shaders on MaterialX parameter changes, thus allowing for their interactive updates/editing


jilliene:
Filed as internal issue #USD-6980

PR #1670, opened 2021-10-28
Fix missing materials after enabling parent prim

### LINKED ISSUE #1669: Hydra does not apply assigned material when missing at gprim populate time
### Description of Issue
If a geometry has a material assignment which does not exist at load time, loading the materials after the stage has been rendered will not detect the new material and will not apply it.
This can happen with deactivating the parent prim of a Material being used by a mesh, then reactivating that parent prim.

The reason seems to be that material delegates only get created when a gprim is populated: https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/gprimAdapter.cpp#L157
And if the Material prim does not exist then the dependency info is never generated for the missing material.

There could be quite a few ways to fix this issue, and I'm considering which way would be best. In order to maintain current performance I'm leaning towards adding a new `UsdImagingDelegate::_missingDependencyInfo` that would gather all missing material paths, and would be checked against when `_ResyncUsdPrim` is called.
Thoughts?

### Steps to Reproduce
1. Download the test file [simpleMatBind.usda.txt](https://github.com/PixarAnimationStudios/USD/files/7422944/simpleMatBind.usda.txt) (kindly prepared by @asluk !) and rename to simpleMatBind.usda
2. Open the file in usdview
3. Select the `/World/Looks` prim and right-click  menu : Deactivate
4. Blue cube becomes white: expected
5. With `/World/Looks` still selected, right click and select Activate
6. Cube stays white: unexpected, the cube should be back to blue as the material now exists in the stage

N.B. that the same thing happens if you save the stage with the `/World/Looks` prim unloaded, loading the prim does not make the re-attach the blue material to the cube.

N.B. that if the `/World/Looks/PreviewSurface` itself is unloaded the material reappears when the Material prim is loaded, the reason seems that the prim path still appears in the `_dependencyInfo` delegate cache.

Windows and Linux

### Package Versions
USD 20.08 to USD 21.11

### Build Flags
`--imaging`

aloysbaillet:
### Description of Change(s)
This is a potential fix for #1669 with minimal performance overhead as far as I've been able to test.
I would love some feedback on potentially better ways to achieve the right behavior for Hydra when materials appear on the stage after the geometry that needs it!

### Fixes Issue(s)
- Fixes #1669

### DIFF
--- a/pxr/usdImaging/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/usdImaging/delegate.cpp
@@ -613,6 +613,8 @@ UsdImagingDelegate::_Populate(UsdImagingIndexProxy* proxy)
 
     std::vector<UsdPrim> primsToPopulateWithoutInstancing;
 
+    bool anyPrimRemoved = false;
+
     for (SdfPath const& usdPath: usdPathsToRepopulate) {
 
         // _Populate should never be called on prototype prims or prims in
@@ -651,6 +653,42 @@ UsdImagingDelegate::_Populate(UsdImagingIndexProxy* proxy)
                             "due to prim type <%s>\n",
                             iter->GetPath().GetText(),
                             iter->GetTypeName().GetText());
+
+                // Use this iteration of the subtree to check for any newly added material
+                // that should be reconnected to a gprim
+                if (iter->IsA<UsdShadeMaterial>()) {
+                    TRACE_SCOPE("Material re-assigment detection")
+                    // Check if any existing gprim has this material assigned
+                    for (const auto& dep : _dependencyInfo) {
+                        if (dep.first != dep.second) {
+                            // Only consider self dependencies as they can
+                            // indicate a missing material assignment
+                            continue;
+                        }
+                        _HdPrimInfo *primInfo = _GetHdPrimInfo(dep.second);
+                        if (primInfo == nullptr || primInfo->adapter == nullptr) {
+                            continue;
+                        }
+                        UsdPrim& prim = primInfo->usdPrim;
+                        if (!prim.IsA<UsdGeomImageable>()) {
+                            continue;
+                        }
+                        if (_materialBindingCache.GetValue(prim) == iter->GetPath()) {
+                            // found a prim that uses this material, need to repopulate the prim
+                            // first remove it
+                            proxy->RemoveRprim(dep.second);
+                            anyPrimRemoved = true;
+
+                            // then re-add it
+                            if (UsdImagingPrimAdapterSharedPtr adapter = _AdapterLookup(prim)) {
+                                TF_DEBUG(USDIMAGING_CHANGES).Msg("[Repopulate] re-resolving material bindings on %s\n",
+                                    prim.GetPath().GetText());
+                                leafPaths.push_back(std::make_pair(prim, adapter));
+                            }
+                        }
+                    }
+                }
+
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #USD-6992

tcauchois:
Hey Aloys,

This approach worries me; having special case code in delegate.cpp (instead of the adapters) tends to be a maintenace burden and I'd expect the loop over _dependencyInfo to get expensive in a big scene.

I'm not sure, but I *think* you can fix this case in gprimAdapter.cpp.  In _AddRprim, we look up the material binding, then look up the material prim, and if the material prim exists we populate it and add a dependency from the material prim to the hydra gprim.  https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usdImaging/usdImaging/gprimAdapter.cpp#L163 ...  If you pull the index->AddDependency() call outside of the if statement, it should add the dependency even if the material prim is missing/null (since the dependency map is just path-based).  In that case, when you load/unload your looks you'd resync the material, which would trigger a resync on any gprims bound to the material, which would correctly repopulate (or remove) the material.

Can you try that and let me know if it works?

Thanks,
Tom

aloysbaillet:
Thanks Tom!
I agree with you. And this is the approach I actually tried initially, unfortunately the `materialUsdPath` was empty as the material resolver checks if the material prim exists (which kinda makes sense!), and modifying the material resolver to allow resolving to non-existing prims seemed a bit scary to me...
That said it could be done as a last try: first resolve the material normally, if it comes back empty try again with an added flag to allow non-existing materials to be resolved. What do you think?

tcauchois:
Can you point me to where we check for prim existence?  How we'd edit it depends on whether it's usd or usdImaging code.  But yeah, I think adding a way to get a missing material binding is worthwhile, via a flag or another function signature or something.

aloysbaillet:
This is coming from [UsdImagingPrimAdapter::GetMaterialUsdPath](https://github.com/PixarAnimationStudios/USD/blob/7a5f8c4311fed3ef2271d5e4b51025fb0f513730/pxr/usdImaging/usdImaging/primAdapter.cpp#L1277) which in turn uses `UsdShadeMaterialBindingAPI(prim).ComputeBoundMaterial()` which lives in UsdShade...

aloysbaillet:
I believe this is the line that filters out invalid material prims: https://github.com/PixarAnimationStudios/USD/blob/7a5f8c4311fed3ef2271d5e4b51025fb0f513730/pxr/usd/usdShade/materialBindingAPI.cpp#L745

tcauchois:
That's an interesting bit of code, for sure!  I think my preference is building out what dependency maps we can, to keep things moving fast, so I'm talking to some folks about that bit of MaterialBindingAPI code and I'll let you know what options we come up with.

tcauchois:
Just to follow up, I've been looking into exporting dependency information from the material binding cache but it's a bit of a rabbit hole; I'll let you know how it goes.  In the meantime, the USD edit APIs (including the python in the usdview interpreter) are pretty low level but it looks like in all of our authoring applications relationships aren't allowed to point at nonexistent prims; that's considered badly conditioned scene description, because it can lead to thorny change-tracking and scene analysis issues.  We can't enforce that in usdImaging, but maybe you can at the app/asset level?

tallytalwar:
I believe UsdShade and UsdImaging logic were updated wrt this PR here: 161873983c86c24acf7953430a6277713cc0a202

Going to close it, @aloysbaillet please reopen or file a new issue if thats not the case.

Thanks

PR #1672, opened 2021-11-03
collection of fixes for building without pre-compiled headers

pmolodo:
### Description of Change(s)

ie, if -DPXR_ENABLE_PRECOMPILED_HEADERS=OFF

wanted to build without precompiled headers to help with debugging a
strange build issue (ie, create more transparent + atomic build steps)

also, turning off PCH usage allows stashed.io (a
caching compiler wrapper for MSVC) to operate when using PDB
(see: https://github.com/playscale/stashed.io/wiki/PDB-and-PCH-generation)

most fixes are pretty self-explanatory - adding some missing headers to
individual source / header files

in stackTrace.cpp, the include of Winsock2.h must happen before windows.h,
which is brought in via fileSystem.h

### DIFF
--- a/pxr/base/arch/stackTrace.cpp
+++ b/pxr/base/arch/stackTrace.cpp
@@ -32,6 +32,11 @@
 #include "pxr/base/arch/error.h"
 #include "pxr/base/arch/errno.h"
 #include "pxr/base/arch/export.h"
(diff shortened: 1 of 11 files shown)


jilliene:
Filed as internal issue #USD-7002

Issue #1675, opened 2021-11-07
Cannot compile with -Werror=format-security


musicinmybrain:
### Description of Issue

USD 21.11 cannot compile with `-Werror=format-security`.

### Steps to Reproduce
1. Build USD with `-Werror=format-security`.

### System Information (OS, Hardware)

Fedora Linux Rawhide (development version), x86_64.

### Package Versions

21.11

### Build Flags

`-Werror=format-security`

----

See the downstream issue for upgrading to 21.11, beginning with [Comment 6](https://bugzilla.redhat.com/show_bug.cgi?id=2019440#c5).

The error is:

```
/builddir/build/BUILD/USD-21.11/pxr/imaging/hd/dirtyList.cpp: In member function 'void pxrInternal_v0_21__pxrReserved__::HdDirtyList::UpdateRenderTagsAndReprSelectors(const TfTokenVector&, const HdReprSelectorVector&)':
/builddir/build/BUILD/USD-21.11/pxr/imaging/hd/dirtyList.cpp:193:38: error: format not a string literal and no format arguments [-Werror=format-security]
  193 |                 TfDebug::Helper().Msg(ss.str().c_str());
      |                 ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

The lines around the error are:

```cpp
            // Reset tracked repr set.
            // XXX An alternative is to grow the tracked repr set similar to 
            //     render tags (above). This will require the render index to
            //     sync the tracked reprs rather than ones requested by the
            //     tasks.
            if (TfDebug::IsEnabled(HD_DIRTY_LIST)) {
                std::stringstream ss;
                ss << "Resetting tracked reprs in dirty list from "
                   << _trackedReprs << " to " << reprs << "\n";
                TfDebug::Helper().Msg(ss.str().c_str());
            }
```

For the definition of `class TfDebug`, see `pxr/base/tf/debug.h`. We find that `TfDebug::Helper` is a `struct` defined as follows:

```cpp
    struct Helper {
        static TF_API void Msg(const std::string& msg);
        static TF_API void Msg(const char* msg, ...) ARCH_PRINTF_FUNCTION(1,2);
    };
```

[...]

PR #1676, opened 2021-11-07
Fix compiling with -Werror=format-security

musicinmybrain:
### Description of Change(s)

Use the `std::string` copied from a stringstream’s value, rather than the `char const *` obtained from that `std::string`’s `c_str()` method, so that the correct overload of `TfDebug::Helper::Msg` is selected, and the message built in the stringstream is not accidentally treated as a `printf`-style format string.

### Fixes Issue(s)
- #1675

### DIFF
--- a/pxr/imaging/hd/dirtyList.cpp
+++ b/pxr/imaging/hd/dirtyList.cpp
@@ -191,7 +191,7 @@ HdDirtyList::UpdateRenderTagsAndReprSelectors(
                 std::stringstream ss;
                 ss << "Resetting tracked reprs in dirty list from "
                    << _trackedReprs << " to " << reprs << "\n";
-                TfDebug::Helper().Msg(ss.str().c_str());
+                TfDebug::Helper().Msg(ss.str());
             }
             _trackedReprs = reprs;
             trackedReprsChanged = true;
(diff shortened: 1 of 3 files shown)


sunyab:
Hi @musicinmybrain, thank you for this PR! We ask that contributors sign a CLA before merging their PRs. Could you follow the instructions at https://graphics.pixar.com/usd/release/contributing_to_usd.html?

Once we receive your signed CLA we can pull this in. Thanks!

jilliene:
Filed as internal issue #USD-7016

musicinmybrain:
> Once we receive your signed CLA we can pull this in. Thanks!

I sent it to usd-cla@pixar.com on 2021-11-10, but I haven’t yet received an acknowledgement. How long does that usually take?

musicinmybrain:
> > Once we receive your signed CLA we can pull this in. Thanks!
> 
> I sent it to [usd-cla@pixar.com](mailto:usd-cla@pixar.com) on 2021-11-10, but I haven’t yet received an acknowledgement. How long does that usually take?

Never mind! I just got the CLA acknowledgement, so you should be able to pull this in at your leisure.

PR #1677, opened 2021-11-08
Physics rigid body mass computation and tests

AlesBorovicka:
### Description of Change(s)
Exposing UsdPhysicsCollisionAPI::GetCollisionPhysicsMaterialPrim that gathers material prim for the collision and added tests for it.
Exposing UsdPhysicsRigidBodyAPI::ComputeMassProperties that can compute mass properties for the rigid body and added tests for it.
Added massProperties.h that include basic mass computer.


jilliene:
Filed as internal issue #USD-7017

pxr/usd/usdPhysics/massProperties.h
```diff
@@ -0,0 +1,251 @@
...
+
+#include "pxr/base/gf/vec3f.h"
+#include "pxr/base/gf/quatf.h"
+#include "pxr/base/gf/rotation.h"
+#include "pxr/base/gf/matrix3f.h"
+#include "pxr/base/gf/matrix4f.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+GfQuatf IndexedRotation(uint32_t axis, float s, float c)
```
spiffmon: @AlesBorovicka , since this is a public header, our coding conventions require that every public function, type, and class, be prefixed with the ProperCase module name, so `UsdPhysicsIndexedRotation()` etc.  We do not allow module namespaces, _but_ do allow a nesting **class** of static methods, if you prefer.  So `UsdPhysicsMass::IndexedRotation()` etc, if that seems more appealing.  Thanks!

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+
+TF_DEFINE_PRIVATE_TOKENS(
+    _physicsPurposeTokens,
+    ((materialPurposePhysics, "physics"))
+);
+
+
+const float COMPARE_TOLERANCE = 1e-05f;
+
+struct MassApiData
```
spiffmon: Conventions: non-public types prefixed with "_", e.g. `_MassApiData`

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+    float       mass = -1.0f;
+    float       density = -1.0f;
+    bool        hasInertia = false;
+    GfVec3f     diagonalInertia = { 1.0f, 1.0f, 1.0f };
+    bool        hasPa = false;
+    GfQuatf     principalAxes;
+};
+
+// gather all the mass information for the gien prim, based on UsdPhysicsMassAPI
+MassApiData ParseMassApi(const UsdPrim& usdPrim)
```
spiffmon: Conventions: non-public functions prefixed with "_", e.g. `_ParseMassApi`

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+{
+    bool comSet = false;
+    if (usdPrim.HasAPI<UsdPhysicsMassAPI>())
+    {
+        // com
+        const UsdPhysicsMassAPI massAPI(usdPrim);
+        const UsdAttribute comAttribute = massAPI.GetCenterOfMassAttr();
+
+        // no sentinel value we need to check for authored value        
+        if (comAttribute.HasAuthoredValue())
```
spiffmon: We must have overlooked these properties in our epic discussion about sentinels last year.  To recap, predicating computation results on "isAuthored" state is fragile, because simple (common, and unwitting) "breakdown" scene mutations can then change the outcome of a computation.  Can we give _centerOfMass_ and any other needful schema properties workable sentinel fallbacks, and then compare to sentinel here rather than HasAuthoredValue?

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+            result.hasPa = true;
+            principalAxes.Get(&result.principalAxes);
+        }
+
+    }
+    return result;
+}
+
+// custom get center of mass, using also transformation to apply scaling
+bool GetCoM(const UsdPrim& usdPrim, GfVec3f& com, UsdGeomXformCache& xfCache)
```
spiffmon: Another coding convention: mutable objects passed by pointer rather than reference so that it's easy to tell at calling sites what's boing to be mutated.  Of all the conventions I've called out, this is the one that there's some internal contention about, but it **is** still enforced.

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+    // setup collision transformation
+    transform.SetTranslate(massInfo.localPos);
+    transform.SetRotateOnly(GfQuatd(massInfo.localRot));
+
+    // return final collision mass properties
+    return MassProperties(shapeMassInfo.mass, inertia, massInfo.centerOfMass);
+}
+
+// compute mass properties for given rigid body
+float UsdPhysicsRigidBodyAPI::ComputeMassProperties(GfVec3f& _diagonalInertia, GfVec3f& _com, GfQuatf& _principalAxes, const MassInformationFn& massInfoFn) const
```
spiffmon: 80 columns, please?

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+
+        // traverse all collisions below this body and get their collision information
+        // first gather all collisions
+        UsdPrimRange range(usdPrim);
+        for (auto it = range.begin(); it != range.end(); ++it)
+        {
+            const UsdPrim collisionPrim = *it;
+            if (collisionPrim && collisionPrim.HasAPI<UsdPhysicsCollisionAPI>())
+            {
+                collisionPrims.push_back(collisionPrim);
```
spiffmon: If you remove the `const` above, which isn't buying you much, then you can `std::move(collisionPrim)` here, which _does_ buy you several refcount increments.

pxr/usd/usdPhysics/rigidBodyAPI.cpp
```diff
@@ -250,3 +250,402 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+                {
+                    const UsdShadeMaterial& generalMaterial = generalMaterials[i];
+                    if (generalMaterial && generalMaterial.GetPrim().HasAPI<UsdPhysicsMaterialAPI>())
+                    {
+                        physicsMaterials[i] = generalMaterial;
+                    }
+                }
+            }
+
+        }
```
spiffmon: Ok, weird.  I added a comment in what must have been a forked branch version of this code that you don't need this second pass, here.

spiffmon (changes requested):
I haven't checked out the test, but left a few things in the code itself to follow up on - thanks, @AlesBorovicka !

PR #1680, opened 2021-11-10
Add shadowing support to MaterialX shaders in HdStorm

vlasovi:
### Description of Change(s)

Insert HdStorm's shadowing code into MaterialX shader during the code gen.

### Fixes Issue(s)

Fixes the bug that there are no shadows in HdStorm on geometry with MaterialX material applied


jilliene:
Filed as internal issue #USD-7019

klucknav:
Thank you for the change. Unfortunately there are a handful of changes we would like to make, so we cannot directly merge it. But we have a fix for the underlying issue, based on your change, that should be in dev soon. Thanks again.

PR #1690, opened 2021-11-23
Maya/krickw/check if the next prim is not in the subtree before starting the wider search

williamkrick:
### Description of Change(s)
Do not do the O(N) search for the first path that does not have subtree as a prefix if subtree has not children.

Please see https://github.com/PixarAnimationStudios/USD/issues/1689 for a detailed description of the problem.
Internal issue MAYA-114875.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1689


williamkrick:
I messed up the base commit for this PR. I will close this one and open another.

PR #1691, opened 2021-11-23
Do not search for the end of the list of children if there are no children

williamkrick:
### Description of Change(s)
Do not do the O(N) search for the first path that does not have subtree as a prefix if subtree has no children.

Please see https://github.com/PixarAnimationStudios/USD/issues/1689 for a detailed description of the problem.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1689

### DIFF
--- a/pxr/usdImaging/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/usdImaging/delegate.cpp
@@ -924,12 +924,19 @@ UsdImagingDelegate::_GatherDependencies(SdfPath const& subtree,
         return;
     }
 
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-7037

sunyab:
This PR has been merged internally and will be available in the dev branch with the next push. 

I wanted to mention that this change was modified to fix a bug where the `end` iterator was invalid and dereferenced, leading to crashes in certain cases.

williamkrick:
Awesome, thank you so much @sunyab!

Issue #1542, opened 2021-06-22
Boost install fails when building


MATTJ0NES:
Hi, I'm running into an error on building at the installing Boost step. I've seen similar previous issues but all with much earlier versions of things and not seen a solution that applies here. I've added the relevant info below. Any help much appreciated, thanks.

### Description of Issue
Error when building, due to a failure on the Boost install:
_ERROR: Failed to run 'b2 --prefix="C:\USD" --build-dir="C:\USD\build" -j12 address-model=64 link=shared runtime-link=shared threading=multi variant=release --with-atomic --with-program_options --with-regex --with-python --user-config=python-config.jam toolset=msvc-14.1 install'_

### Steps to Reproduce
With all dependecies already installed, I'm running this command:
python USD-release/build_scripts/build_usd.py "C:\USD"

### System Information (OS, Hardware)
Windows 10, 64 bit on Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz   2.59 GHz

### Package Versions
Python 3.6
Latest USDpackage as of 21 Jun 2021 (v21.05)

### Build Flags
None. See command above

Log file text:

2021-06-22 14:29
b2 --prefix="C:\USD" --build-dir="C:\USD\build" -j12 address-model=64 link=shared runtime-link=shared threading=multi variant=release --with-atomic --with-program_options --with-regex --with-python --user-config=python-config.jam toolset=msvc-14.1 install
Performing configuration checks

    - default address-model    : 32-bit (cached)
    - default architecture     : x86 (cached)
    - has_icu builds           : no  (cached)

Component configuration:

    - atomic                   : building
    - chrono                   : not building
    - container                : not building
    - context                  : not building
    - contract                 : not building
    - coroutine                : not building
    - date_time                : not building
    - exception                : not building
    - fiber                    : not building
    - filesystem               : not building
    - graph                    : not building
    - graph_parallel           : not building
    - headers                  : not building
    - iostreams                : not building
    - locale                   : not building
    - log                      : not building
    - math                     : not building
    - mpi                      : not building
    - program_options          : building
    - python                   : building
    - random                   : not building
    - regex                    : building
    - serialization            : not building
    - stacktrace               : not building
    - system                   : not building
    - test                     : not building
    - thread                   : not building
    - timer                    : not building
    - type_erasure             : not building
    - wave                     : not building

...patience...
...patience...
...found 31927 targets...
...updating 6 targets...
[...]

PR #1692, opened 2021-11-25
better detection of python lib / include dirs in build_usd.py

pmolodo:
### Description of Change(s)

This addresses two issues:

- Building using a venv-installed python
  - The venv-installed pythons would try to pull the libraries from
    the venv-directory, when they only reside at the "original"\
    installed python location

- Building with a python that was moved after building on Linux
  - We experienced issues where the `INCLUDEPY` and `LIBDIR` sysconfig
    vars were baked at build time.  Thus these would no longer be
    reliable if running a python executable that was moved to a
    different location after building.

Note that on python-2.7, `installed_base` doesn't exist, so we fall back
to `base`.  Also we fall back to `INCLUDEPY` and `LIBDIR` if the
`installed_base` / `base` method doesn't work, so this should be fairly
backward compatible.

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -226,20 +226,38 @@ def GetPythonInfo(context):
         pythonLibPath = os.path.join(pythonBaseDir, "lib",
                                      _GetPythonLibraryFilename(context))
     else:
-        pythonIncludeDir = sysconfig.get_config_var("INCLUDEPY")
-        if Windows():
+        pythonIncludeDir = sysconfig.get_path("include")
+        if not pythonIncludeDir or not os.path.isdir(pythonIncludeDir):
+            # as a backup, and for legacy reasons - not preferred because
+            # it may be baked at build time
+            pythonIncludeDir = sysconfig.get_config_var("INCLUDEPY")
+
+        # if in a venv, installed_base will be the "original" python,
+        # which is where the libs are ("base" will be the venv dir)
+        pythonBaseDir = sysconfig.get_config_var("installed_base")
+        if not pythonBaseDir or not os.path.isdir(pythonBaseDir):
+            # for python-2.7
             pythonBaseDir = sysconfig.get_config_var("base")
+
+        if Windows():
             pythonLibPath = os.path.join(pythonBaseDir, "libs",
                                          _GetPythonLibraryFilename(context))
         elif Linux():
-            pythonLibDir = sysconfig.get_config_var("LIBDIR")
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-7040

pmolodo:
Any word on this PR? I understand that it might be hard to test, but the various `isdir` checks should ensure it falls back to old behavior if needed.

build_scripts/build_usd.py
```diff
@@ -226,24 +226,36 @@ def _GetPythonLibraryFilename(context):
...
+
+        if Windows():
+            pythonLibDir = os.path.join(pythonBaseDir, "libs")
         elif Linux():
-            pythonLibDir = sysconfig.get_config_var("LIBDIR")
+            pythonLibDir = os.path.join(pythonBaseDir, "lib")
+            if not os.path.isdir(pythonLibDir):
+                # as a backup, and for legacy reasons - not preferred because
+                # it may be baked at build time
+                pythonLibDir = sysconfig.get_config_var("LIBDIR")
```
sunyab: This doesn't give me the right results on my RHEL7 workstation running Python 3.6.8. On this system, the Python library is located at `/usr/lib64/libpython3.6m.so`. That's what I get with the old code:

```
sysconfig.get_config_var("LIBDIR") => '/usr/lib64'
pythonLibDir = '/usr/lib64'
pythonLibPath = '/usr/lib64/libpython3.6m.so'
```
but with this change I'd get a different path that doesn't exist:

```
sysconfig.get_config_var("installed_base") => '/usr'
sysconfig.get_config_var("base") => '/usr'
pythonLibDir = '/usr/lib'
pythonLibPath = '/usr/lib/libpython3.6m.so'
```

I wonder if, instead of checking if the *directory* generated via the new method exists, we should check whether the *library* exists instead. If it doesn't, then we fall back to the old method.

sunyab (changes requested):
I tested this out in a few places -- on Windows this seems to work great and fixes venv-related build issues reported elsewhere (#1542, for example). However on my Linux workstation it doesn't give me the right results. I left a comment below with further details.

build_scripts/build_usd.py
pmolodo: @sunyab - Sorry for the long response time.  Tried doing as you suggested, and I now check if the library exists, and go onto another method if it doesn't.  Also made it try the old way first, for better backward compatibility.

sunyab (approved):

PR #1695, opened 2021-11-25
BakeSkinning: add support for skinning normals with faceVarying interpolation

cameronwhite:
### Description of Change(s)
Previously, only normals with vertex or varying interpolation were deformed. This was something we noticed while experimenting with using UsdSkel to deform fractured meshes from rigid body simulations, where faceted normals are quite common.

To reproduce, load the [attached file](https://github.com/PixarAnimationStudios/USD/files/7605237/facevarying_normals.zip) with `usdview facevarying_normals.usda --camera /cameras/camera1`, and then run `UsdSkel.BakeSkinning(usdviewApi.stage.Traverse())` in the interpreter. The undeformed normals can be seen as the cube spins around.

A few notes:

- I debated between adding a separate `UsdSkelSkinFaceVaryingNormalsLBS()` method rather than more overloads of `UsdSkelSkinNormalsLBS()` - feedback on this is welcome!
- faceVarying normals also are not deformed in hydra, but the issue is less noticeable depending on the renderer (the normals are suppressed, leaving them to be auto-computed post skinning). I'd like to look at this for a future PR, but wanted to keep this one at a reasonable scope
- faceVarying normals are also not handled for blendshapes. Doing this would require extensions to the schema, which assumes a 1:1 mapping between normals and points.


jilliene:
Filed as internal issue #USD-7043

ucbearkat:
@cameronwhite Thanks for the change! To answer your first question - using a separate UsdSkelSkinFaceVaryingNormalsLBS() method instead of adding yet another override is probably ok in this case since this is about the caller facing api and internally they all end up calling _SkinNormalsLBS anyway. 
Your caveats about the scope of the change are also noted. 
I've staged this internally for review and will get back to you with any comments that come up.

pxr/usd/usdSkel/utils.cpp
```diff
@@ -1236,7 +1236,34 @@ struct _InterleavedInfluencesFn {
...
+/// Functor which uses the faceVertexIndices attribute to find the
+/// corresponding point index. Use for deforming faceVarying normals.
+struct _FaceVaryingPointIndexFn
+{
+    TfSpan<const int> faceVertexIndices;
+    const int numPoints;
+
+    size_t GetPointIndex(size_t index) const
+    {
+        int pointIndex = faceVertexIndices[index];
```
ucbearkat: Can this be changed to const int?

pxr/usd/usdSkel/utils.h
```diff
@@ -595,6 +596,18 @@ UsdSkelSkinNormalsLBS(const GfMatrix3f& geomBindTransform,
...
+bool
+UsdSkelSkinFaceVaryingNormalsLBS(const GfMatrix3d& geomBindTransform,
+                                 TfSpan<const GfMatrix3d> jointXforms,
+                                 TfSpan<const int> jointIndices,
+                                 TfSpan<const float> jointWeights,
+                                 int numInfluencesPerPoint,
+                                 TfSpan<const int> faceVertexIndices,
+                                 TfSpan<GfVec3f> normals,
+                                 bool inSerial=false);
+
```
ucbearkat, line 610: For completeness and consistency with the other UsdSkelSkin* utility methods here, would it be possible to also provide an overload that takes GfMatrix3f transforms?

ucbearkat (commented):
@cameronwhite Sorry it took a while to get back to you on this. 
This change is pretty great overall. Noted a couple comments that came up.
Thanks!

cameronwhite:
Thanks! I've pushed a commit with the noted changes

PR #1696, opened 2021-11-26
Address a compilation error with Clang

christophercrouzet:
### Description of Change(s)

Address a couple of compilation errors from Clang related to default constructors:

```
usd/pxr/imaging/hd/command.cpp:30:22: error: defaulting this default constructor would delete it after its first declaration
HdCommandDescriptor::HdCommandDescriptor() = default;
                     ^
usd/pxr/imaging/hd/command.h:104:35: note: default constructor of 'HdCommandDescriptor' is implicitly deleted because field 'commandArgs' of const-qualified type 'const pxrInternal_v0_21__pxrReserved__::HdCommandArgDescriptors' (aka 'const vector<pxrInternal_v0_21__pxrReserved__::HdCommandArgDescriptor>') would not be initialized
    const HdCommandArgDescriptors commandArgs;
```

Tested with Clang 14.

### DIFF
--- a/pxr/imaging/hd/command.cpp
+++ b/pxr/imaging/hd/command.cpp
@@ -25,8 +25,4 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-HdCommandArgDescriptor::HdCommandArgDescriptor() = default;
-
-HdCommandDescriptor::HdCommandDescriptor() = default;
-
 PXR_NAMESPACE_CLOSE_SCOPE
--- a/pxr/imaging/hd/command.h
+++ b/pxr/imaging/hd/command.h
@@ -44,7 +44,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 ///
 struct HdCommandArgDescriptor
 {
-    HdCommandArgDescriptor();
+    HdCommandArgDescriptor() noexcept = default;
 
     HdCommandArgDescriptor(const TfToken &argName,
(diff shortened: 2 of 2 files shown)


christophercrouzet:
Interestingly, the build compiled fine on my Ubuntu 20.04 + Clang 14 environment however it failed when using the runner's macOS + Clang 10 combo:

```
/Users/runner/work/1/s/pxr/imaging/hd/command.h:47:5: error: exception specification of explicitly defaulted default constructor does not match the calculated one
    HdCommandArgDescriptor() noexcept = default;
    ^
```

As a result, I removed the `noexcept` specifier from that one definition and it's now going through.

I assume that the error comes from the class `HdCommandArgDescriptor` having a member of type `VtValue` that has a default constructor not defined with the `noexcept` specifier?

jilliene:
Filed as internal issue #USD-7044

sunyab:
Hello, just wanted to mention that we have an alternate fix for this which will be part of the next push to the dev branch. I wasn't able to set up a clang USD build on Linux to test this specifically, but I did set up a similar test case independent of USD and confirmed that was fixed. I'll leave this PR open for now but close it when the other fix gets pushed to dev. Thanks!

christophercrouzet:
Awesome, thank you @sunyab!

sunyab:
Closing this PR out in favor of b467dd7, thanks!

PR #1715, opened 2021-12-14
Minor fixes to build with gcc-11.

anderev:
### Description of Change(s)
Minor updates so the code will compile with gcc-11.

### Fixes Issue(s)
-size_t moved to std namespace
-numeric_limits defined in limits header


anderev:
Will reopen against dev branch

PR #1717, opened 2021-12-15
Minor fixes to build with gcc-11.

anderev:
### Description of Change(s)
Minor updates so the code will compile with gcc-11.
### Fixes Issue(s)
- size_t moves to std namespace
- numeric_limits is defined in limits header
### Does not fix
- external plugins, e.g. MaterialX, but similar updates can be applied where necessary.


jilliene:
Filed as internal issue #USD-7080

sunyab:
Hi @anderev, I was unable to confirm whether you've submitted or are covered by a CLA. Could you let me know if that's the case? If not, we'll need a signed CLA before we can merge this PR, the instructions are here: https://graphics.pixar.com/usd/release/contributing_to_usd.html

Thank you!

PR #1727, opened 2021-12-20
Restore Dependency to OpenEXR's Half Lirbary

christophercrouzet:
### Description of Change(s)

Revert commit 52bfa939cf83980efcb94e61be0bd446f8c2f674.

### Fixes Issue(s)
- an oversight on my end that causes a compilation error, as described by @asluk here: https://github.com/PixarAnimationStudios/USD/commit/52bfa939cf83980efcb94e61be0bd446f8c2f674#r62233360.

### DIFF
--- a/pxr/imaging/hioOpenVDB/CMakeLists.txt
+++ b/pxr/imaging/hioOpenVDB/CMakeLists.txt
@@ -14,6 +14,7 @@ pxr_library(hioOpenVDB
         hio
         tf
         usd
+        ${OPENEXR_Half_LIBRARY}
         ${OPENVDB_LIBRARY}
 
     INCLUDE_DIRS


spiffmon:
Thank you for the quick response, @christophercrouzet , but is this the only remedy?  It's extremely unfortunate to introduce a direct dependency in USD on exr/half, because:
a) hioOpenVdb does **not** itself directly depend on it
b) half is being moved out of exr in exr 3, and thus those dependencies are problematic for anyone trying to upgrade to exr 3, which is why we've been very happy that USD is currently free of _any_ direct dependencies on exr or ilmbase/exr half.

It seems problematic that if OpenVdb depends on half, it does not provide the dependency itself - is this possibly a problem with @asluk 's Vdb build?  Is this only a Windows problem?

Wondering if @meshula has any insights?

meshula:
OpenEXR 2.x did not always have a fully specified config file, therefore various projects have been relying on a findOpenEXR cmake script that attempted to set things up. These scripts have never been consistently implemented or distributed.

OpenVdb itself does set Imath or IlmBase as a `PUBLIC` dependency for `target_link_libraries` for openvdb_shared, and openvdb_static, which are aliased to the library `openvdb`. Since the dependency is NOT propagating, that means that OpenVdb's openEXR findpackage script is not specifying the libraries in a way that allows the propagation.

I tried to work out what that script is doing, it's .... complex. Short answer, I don't think it's output is predictable without access to your build environment.

There has been an option to use a built-in version of Half for some time, at one point that was under a `USE_EXR` cmake option, default ON, and later, that option was renamed to `USE_IMATH_HALF`, default OFF. 

I recommend modifying your build set up to build OpenVdb with `USE_EXR` OFF, or `USE_IMATH_HALF` OFF since end of Feb 2021.

The latest version of OpenVdb are compatible with OpenEXR 3, and under an OpenEXR 3 build scenario, these issues go away.

In the case of a build under USD's `build_scripts/build_usd.py`, I recommend that we remove the specification of `-DILMBASE_ROOT`, and instead set `-DUSE_EXR=OFF`, which is available in the v6.2.1 specified in USD's script. 

Edit:

to clarify ~ if you need an OpenVdb build that exposes OpenEXR's Half as a native OpenVdb type, then we should discuss solutions further, as I don't think we should be by default reintroducing that dependency.

to meet @spiffmon's note of not requiring OpenEXR as a USD dependency, we should be specifying USE_EXR OFF, rather than reintroducing the linkage. I am very doubtful that people building USD with our build script would be affected by such a change, and people building USD and OpenVdb into specialized environments, such as a vendor might into their DCC, they won't be using our python build scripts in the first place.

asluk:
I'll check tomorrow or next week to see if Linux builds are working without the new PR. Thanks all!

christophercrouzet:
Hi @spiffmon, thank you for your insights!

I do agree that this pull request is not ideal. I was also under the assumption that this dependency would have been provided through OpenVDB, which is why I took the liberty to remove it in the previous commit since I couldn't spot any code in that hioOpenVDB module that directly used Half. Also my own tests + the CI seemed to confirm that it was fine doing so.

However, I am not aware of the specifics and realized that maybe I missed something, and I'm not sure neither how Pixar handles such situations that are breaking for some customers of USD, which is why I though that opening a new pull request to revert that commit for now seemed like a fair option to consider, even if only to start a discussion.

I'll leave it to you to pick the best course of action, thanks again!

spiffmon:
Thank you all!  @asluk , we won't be building a 22.02 RC until we're back in January, so we have time to hear the results of your testing.  I'm inclined to try @meshula 's last suggestion, but we'll see what the mighty @sunyab has to say in the New Year.

asluk:
I like @meshula's suggestion as well. I tried it out, and also removed OPENEXR as a requirement for OpenVDB in build_usd.py, but the OpenVDB build still looks for IlmBase. I'm not sure why yet, as the cmake files for OpenVDB 6.1.0 found by build_usd.py in dev seem to put IlmBase requirements behind if statements for EXR. Will look into it more tomorrow. Thanks all!

[CMakeCache.txt](https://github.com/PixarAnimationStudios/USD/files/7786308/CMakeCache.txt)
[log.txt](https://github.com/PixarAnimationStudios/USD/files/7786309/log.txt)
[build_usd.zip](https://github.com/PixarAnimationStudios/USD/files/7786312/build_usd.zip)

asluk:
Also verified that only Windows seems to pick up the hioopenvdb dependency on libHalf; Linux builds are linking successfully.

asluk:
I got a build working with openvdb support on Windows that doesn't depend on openexr, and posted the PR https://github.com/PixarAnimationStudios/USD/pull/1728 for further discussion.

In particular, openvdb seems to always require libHalf even when the build flags are set to disable it, until this change in v8.1.0 - https://github.com/AcademySoftwareFoundation/openvdb/pull/927/files#diff-6fcdb7772bbfb37017a5b670d3d3a51342f3242a0e0bc8fb46d2def275867c83L451 

I'm not sure if we want to jump that far ahead in openvdb versioning, but hope that these findings are useful to carry the conversation forward.

Thanks for all the help, and Happy New Year!

spiffmon:
Thanks again for jumping on this, @christophercrouzet !  I'm going to close this for now in hopes that PR #1728 or something like it can keep us on the exr-free path!

christophercrouzet:
No, thanks to all of you!

meshula:
Sorry to necro a closed thread ~ to help the OpenVdb team understand the issue, they would obviously prefer the problem be solved by moving to at least version 8.1. Are there any notes we can pass to them on why updating to a later OpenVdb wasn't on the table as a potential fix?

PR #1728, opened 2021-12-29
Support OpenVDB without depending on OpenEXR

asluk:
Posting for further discussion... this seems to be the minimal change contents to support OpenVDB without depending on IlmBase for libHalf via OpenEXR.

See https://github.com/PixarAnimationStudios/USD/pull/1727 and https://github.com/AcademySoftwareFoundation/openvdb/pull/927

In particular, it appears that OpenVDB v8.1.0 is the first release that doesn't require IlmBase, even if flags are set to disable the OpenEXR dependency, per the change contents in https://github.com/AcademySoftwareFoundation/openvdb/pull/927/files#diff-6fcdb7772bbfb37017a5b670d3d3a51342f3242a0e0bc8fb46d2def275867c83L451

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1068,7 +1068,7 @@ BLOSC = Dependency("Blosc", InstallBLOSC, "include/blosc.h")
 # not require additional dependencies such as GLFW. Note that version
 # 6.1.0 does require CMake 3.3 though.
 
-OPENVDB_URL = "https://github.com/AcademySoftwareFoundation/openvdb/archive/v6.1.0.zip"
+OPENVDB_URL = "https://github.com/AcademySoftwareFoundation/openvdb/archive/v8.1.0.zip"
 
 def InstallOpenVDB(context, force, buildArgs):
     with CurrentWorkingDirectory(DownloadURL(OPENVDB_URL, context, force)):
@@ -1087,9 +1087,10 @@ def InstallOpenVDB(context, force, buildArgs):
                          .format(instDir=context.instDir))
         extraArgs.append('-DTBB_ROOT="{instDir}"'
                          .format(instDir=context.instDir))
-        # OpenVDB needs Half type from IlmBase
-        extraArgs.append('-DILMBASE_ROOT="{instDir}"'
-                         .format(instDir=context.instDir))
+        # Use built-in libHalf instead of depending on IlmBase via OpenEXR
+        # https://github.com/PixarAnimationStudios/USD/pull/1727#issuecomment-998263059
+        # https://github.com/AcademySoftwareFoundation/openvdb/pull/927
+        extraArgs.append('-DUSE_IMATH_HALF=OFF')
 
         # Add on any user-specified extra arguments.
         extraArgs += buildArgs
@@ -2036,7 +2037,7 @@ if context.buildImaging:
     requiredDependencies += [OPENSUBDIV]
 
     if context.enableOpenVDB:
-        requiredDependencies += [BLOSC, BOOST, OPENEXR, OPENVDB, TBB]
+        requiredDependencies += [BLOSC, BOOST, OPENVDB, TBB]
     
     if context.buildOIIO:
         requiredDependencies += [BOOST, JPEG, TIFF, PNG, OPENEXR, OPENIMAGEIO]
--- a/cmake/defaults/Packages.cmake
+++ b/cmake/defaults/Packages.cmake
@@ -253,7 +253,6 @@ if (PXR_BUILD_IMAGING)
     endif()
     # --OpenVDB
     if (PXR_ENABLE_OPENVDB_SUPPORT)
-        find_package(OpenEXR REQUIRED)
         find_package(OpenVDB REQUIRED)
         add_definitions(-DPXR_OPENVDB_SUPPORT_ENABLED)
     endif()
--- a/pxr/imaging/hioOpenVDB/vdbTextureData.cpp
+++ b/pxr/imaging/hioOpenVDB/vdbTextureData.cpp
@@ -22,6 +22,12 @@
 // language governing permissions and limitations under the Apache License.
 //
 
+#include "pxr/base/arch/defines.h"
+#ifdef ARCH_OS_WINDOWS
+    // Otherwise M_PI_2 is undefined from openvdb's Mat.h
+    #define _USE_MATH_DEFINES
+#endif
+
 #include "pxr/imaging/hioOpenVDB/vdbTextureData.h"
 #include "pxr/imaging/hioOpenVDB/debugCodes.h"
 #include "pxr/imaging/hioOpenVDB/utils.h"


meshula:
ASWF projects are supporting vfxplatform current year, plus three, which means there's some expectation of support for openvdb 6, 7, 8, and 9. Thanks to your detective work, we know that 8.1 and greater properly elide the IlmBase dependency when requested.

The vfxplatform also says that the specified versions are minimums, not maximums.

In order for both the ASWF guideline and vfxplatform guidelines to be true, 

|      | EXR2 | EXR3 |
| ---- | ---- | ---- |
| vdb6 | Y    | Yx   |
| vdb7 | Y    | Yx   |
| vdb8 | Y    | Y    |
| vdb9 | N    | Y    |

has to also be true.

The most pedantically correct thing to do, and the most end-user-developer-friendly thing, would be to ask the OpenVdb team to do maintenance releases of 6 and 7 that correct the issue with IlmBase.

A workaround could be to add a bunch of logic to build_usd.py that checks for the Yx combinations, and rejects an attempt to build an impossible combo.

Since the local copy of half.h was already available in vdb6, IIRC, it seems like maintenance point releases might not be too much to request.

jilliene:
Filed as internal issue #USD-7097

sunyab:
Thanks everyone for digging into this! My main concern with this change and the other approaches mentioned in #1727 is that they depend on OpenVDB being built in a specific way. That would be fine for builds via build_usd.py where we control the entire ecosystem of dependencies. But it doesn't seem good to impose that requirement on users who have built out the dependencies separately.

I wonder if an alternative fix would be to use the FindOpenVDB.cmake module that ships with OpenVDB [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/v6.1.0/cmake/FindOpenVDB.cmake) instead of the homegrown FindOpenVDB.cmake we currently use [here](https://github.com/PixarAnimationStudios/USD/blob/release/cmake/modules/FindOpenVDB.cmake). The OpenVDB version appears to have more extensive handling of dependencies when setting up import targets. My hope is that using those import targets would just do the right thing regardless of how OpenVDB itself was built.

I'll see if I can play around with this a bit, but if anyone has a chance to try this out that'd be great too. I'm still shaking off some of the holiday cobwebs so apologies if I'm missing anything.

meshula:
Unless I'm misreading, there's a bug in OpenVdb's findpackage script. Although they have checks in some places in the script, they don't here:

https://github.com/AcademySoftwareFoundation/openvdb/blob/ef214a4393c5374952329bdc8675cba21700ef63/cmake/FindOpenVDB.cmake#L317

To your point though, it might shift the responsibility for fulfilling that prerequisite to the OpenVdb package? But it doesn't seem like it would enable supporting OpenVDB without depending on OpenEXR.

sunyab:
Hm, not quite sure I see the bug. The FindOpenVDB.cmake file I linked to was from OpenVDB 6.1 and not latest, and it seems like in that version the Half dependency is consistently treated as required and not gated by the `USE_EXR` flag.

If we use the imported target set up by OpenVDB, then I think it should hide the OpenEXR dependency from USD? The target should have the appropriate dependencies set up based on how the user built OpenVDB (see [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/ef214a4393c5374952329bdc8675cba21700ef63/cmake/FindOpenVDB.cmake#L472)). We'd still need at least part of this PR though.

meshula:
Yes, that's the file I linked, this is the bit of code that is unguarded as far as I can tell:

```
# Add standard dependencies

find_package(IlmBase REQUIRED COMPONENTS Half)
find_package(TBB REQUIRED COMPONENTS tbb)
find_package(ZLIB REQUIRED)
find_package(Boost REQUIRED COMPONENTS iostreams system)
```

50 lines later, and every other place necessary in the file there are guards like this

```
if(OpenVDB_USES_EXR)
  find_package(IlmBase REQUIRED)
  find_package(OpenEXR REQUIRED)
endif()
```

according to the docs, https://cmake.org/cmake/help/latest/command/find_package.html, that first one should cause a fatal error if Half can't be found in the IlmBase package, which must also be found. That first chunk may be guarded but I can't spot it... Everywhere else in the file, EXR references are obviously guarded.

sunyab:
Hm, I guess the way I understand it, guards aren't needed for that first block since half is a requirement in OpenVDB v6.1 -- see [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/ef214a4393c5374952329bdc8675cba21700ef63/openvdb/doc/dependencies.txt#L40). The USE_EXR flag isn't intended to turn off the half requirement, that's not available until v8.1 as @asluk found.

I was able to hack together a change that uses OpenVDB's CMake module but that got ugly real fast. In the interim, I'm going to check in a change that restores the explicit half dependency in hioOpenVDB to get things working again. We'll still need to figure out what to do to better support OpenVDB versions that don't require half from OpenEXR (i.e., v8.1+).

FlorianZ:
@asluk, where do you stand on this right now?

meshula:
(side note as reference ~ I managed to get OpenTimelineIO compatible with OpenEXR 2 and 3 simultaneously in this commit https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/ebd15cd618b98ce432c4d5efdd95368422cdb256 ~ perhaps there's a useful trick to be found there, mostly in the code that assigns the discovered Imath to a variable.)

asluk:
Apologies, @FlorianZ -- I lost track of next steps for this. @meshula , shall I try to take an approach similar to https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/ebd15cd618b98ce432c4d5efdd95368422cdb256 before reaching out to the OpenVDB maintainers as you suggested in https://github.com/PixarAnimationStudios/USD/pull/1728#issuecomment-1002768918 ? Thanks!

meshula:
I think it is worth trying the OTIO_IMATH_TARGETS approach, as it only requires local changes and accommodates the existing build environment. That pattern is adapted from OpenImageIO btw, that's what Larry invented to solve the problem there ~ doing the same for USD would be consistent.

asluk:
> I think it is worth trying the OTIO_IMATH_TARGETS approach, as it only requires local changes and accommodates the existing build environment. That pattern is adapted from OpenImageIO btw, that's what Larry invented to solve the problem there ~ doing the same for USD would be consistent.

Thanks @meshula -- it looks like @dgovil has something similar to https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/ebd15cd618b98ce432c4d5efdd95368422cdb256 in https://github.com/PixarAnimationStudios/USD/pull/1829 ? But since the imath discovery happens at from cmake, we wouldn't be able to remove OpenEXR as a dependency for the OpenVDB plugin on the fly in build_usd.py, if I understand correctly?

meshula:
@asluk the short answer is that FOO_IMATH_TARGETS approach could work from build_usd.py; the variables would need to be configured and managed by build_usd.py and supplied to cmake's individual calls via `-DFOO_IMATH_TARGETS`, which would leverage Drhuv's proposed mechanism.

PR #1729, opened 2022-01-04
Make no-python build ABI compatible with python build

foisya-adsk:
### Description of Change(s)
This change makes sure clients compiling against builds with Python
support disabled can still run their application with a runtime USD
with Python enabled.

This is mainly done by wrapping Python binding types in wrappers which
have the same size and alignment in no-python builds.

This is mainly done in existing pyObjWrapper.h and new
pyBoostObjWrapper.h.   For example, the applied pattern looks like:

```
#ifdef PXR_PYTHON_SUPPORT_ENABLED

PXR_NAMESPACE_OPEN_SCOPE

// We define the empty stub for ABI compatibility even if Python support is 
// enabled so we can make sure size and alignment is the same.
class TfPyObjWrapperStub
{
public:
    static constexpr std::size_t Size = 16;
    static constexpr std::size_t Align = 8;

private:
    std::aligned_storage<Size, Align>::type _stub;
};

/// \class TfPyObjWrapper
///
/// Boost Python object wrapper.
/// …
#ifdef PXR_PYTHON_SUPPORT_ENABLED
class TfPyObjWrapper
    : public boost::python::api::object_operators<TfPyObjWrapper>
{
    typedef boost::python::object object;

public:

    TF_API TfPyObjWrapper();

    TF_API TfPyObjWrapper(object obj);

    object const &Get() const {
        return *_objectPtr;
    }

    TF_API PyObject *ptr() const;

    friend inline size_t hash_value(TfPyObjWrapper const &o) {
        return (size_t) o.ptr();
    }

    TF_API bool operator==(TfPyObjWrapper const &other) const;

    TF_API bool operator!=(TfPyObjWrapper const &other) const;

private:

    // Befriend object_operators to allow it access to implicit conversion to
    // boost::python::object.
    friend class boost::python::api::object_operators<TfPyObjWrapper>;
    operator object const &() const {
        return Get();
    }

    // Store a shared_ptr to a python object.
    std::shared_ptr<object> _objectPtr;
};

static_assert(sizeof(TfPyObjWrapper) == sizeof(TfPyObjWrapperStub),
              "ABI break: Incompatible class sizes.");
static_assert(alignof(TfPyObjWrapper) == alignof(TfPyObjWrapperStub),
              "ABI break: Incompatible class alignments.");

#else // PXR_PYTHON_SUPPORT_ENABLED

class TfPyObjWrapper : TfPyObjWrapperStub
{
};

#endif // PXR_PYTHON_SUPPORT_ENABLED

```

### Fixes Issue(s)
- External issue https://github.com/PixarAnimationStudios/USD/issues/1716

Co-authored-by: Martin Bisson <martin.bisson@autodesk.com>

### DIFF
--- a/pxr/base/tf/anyWeakPtr.h
+++ b/pxr/base/tf/anyWeakPtr.h
@@ -36,9 +36,9 @@
 
 #ifdef PXR_PYTHON_SUPPORT_ENABLED
 #include "pxr/base/tf/pyUtils.h"
-#include <boost/python/object.hpp>
 #endif // PXR_PYTHON_SUPPORT_ENABLED
 
+#include "pxr/base/tf/pyObjWrapper.h"
 #include <boost/operators.hpp>
 
 #include <cstddef>
@@ -157,9 +157,7 @@ public:
         virtual TfWeakBase const *GetWeakBase() const = 0;
         virtual operator bool() const = 0;
         virtual bool _IsConst() const = 0;
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
-        virtual boost::python::api::object GetPythonObject() const = 0;
-#endif // PXR_PYTHON_SUPPORT_ENABLED
+        virtual TfPyObjWrapper GetPythonObject() const = 0;
         virtual const std::type_info & GetTypeInfo() const = 0;
         virtual TfType const& GetType() const = 0;
         virtual const void* _GetMostDerivedPtr() const = 0;
@@ -174,9 +172,7 @@ public:
         TF_API virtual TfWeakBase const *GetWeakBase() const;
         TF_API virtual operator bool() const;
         TF_API virtual bool _IsConst() const;
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
-        TF_API virtual boost::python::api::object GetPythonObject() const;
-#endif // PXR_PYTHON_SUPPORT_ENABLED
+        TF_API virtual TfPyObjWrapper GetPythonObject() const;
         TF_API virtual const std::type_info & GetTypeInfo() const;
         TF_API virtual TfType const& GetType() const;
         TF_API virtual const void* _GetMostDerivedPtr() const;
@@ -195,9 +191,7 @@ public:
         virtual TfWeakBase const *GetWeakBase() const;
         virtual operator bool() const;
         virtual bool _IsConst() const;
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
-        virtual boost::python::api::object GetPythonObject() const;
-#endif // PXR_PYTHON_SUPPORT_ENABLED
+        virtual TfPyObjWrapper GetPythonObject() const;
         virtual const std::type_info & GetTypeInfo() const;
         virtual TfType const& GetType() const;
         virtual const void* _GetMostDerivedPtr() const;
@@ -263,15 +257,16 @@ TfAnyWeakPtr::_PointerHolder<Ptr>::operator bool() const
     return bool(_ptr);
 }
 
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
 template <class Ptr>
-boost::python::api::object
+TfPyObjWrapper
 TfAnyWeakPtr::_PointerHolder<Ptr>::GetPythonObject() const
 {
+#ifdef PXR_PYTHON_SUPPORT_ENABLED
     return TfPyObject(_ptr);
-}
+#else
+    return {};
 #endif // PXR_PYTHON_SUPPORT_ENABLED
-
+}
 template <class Ptr>
 const std::type_info &
 TfAnyWeakPtr::_PointerHolder<Ptr>::GetTypeInfo() const
--- a/pxr/base/plug/info.h
+++ b/pxr/base/plug/info.h
@@ -43,9 +43,7 @@ public:
     enum Type {
         UnknownType,
         LibraryType,
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
         PythonType,
-#endif // PXR_PYTHON_SUPPORT_ENABLED
         ResourceType
     };
 
--- a/pxr/base/plug/plugin.h
+++ b/pxr/base/plug/plugin.h
@@ -117,9 +117,7 @@ public:
 private:
     enum _Type {
         LibraryType, 
-#ifdef PXR_PYTHON_SUPPORT_ENABLED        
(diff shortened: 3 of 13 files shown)


jilliene:
Filed as internal issue #USD-7107

spiffmon:
Hi @foisya-adsk !  We want to integrate this, but we do not have you listed in Autodesk's CLA.  Can you please ask Gordon to add you?  Thank you!!

pxr/base/tf/anyWeakPtr.h
```diff
@@ -174,9 +168,7 @@ class TfAnyWeakPtr : boost::totally_ordered<TfAnyWeakPtr>
         TF_API virtual TfWeakBase const *GetWeakBase() const;
         TF_API virtual operator bool() const;
         TF_API virtual bool _IsConst() const;
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
-        TF_API virtual boost::python::api::object GetPythonObject() const;
-#endif // PXR_PYTHON_SUPPORT_ENABLED
+        TF_API virtual TfPyBoostObjWrapper GetPythonObject() const;
```
sunyab: Is there any particular reason why a separate TfPyBoostObjWrapper is needed here instead of using the existing TfPyObjWrapper? It looks like other code like VtValue is using TfPyObjWrapper, so it seems like it should be usable here too.

pxr/base/tf/pyBoostObjWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+
+/// \brief TfPyBoostObjWrapper
+///
+/// Wrapper of the boost::python::api::object
+///
+/// The wrapper provides a uniform type 
+/// usable when python extensions are accessible
+/// and when no python extensions are compiled
+#ifdef PXR_PYTHON_SUPPORT_ENABLED
+class TfPyBoostObjWrapper
```
sunyab: Is there a particular reason we need both TfPyBoostObjWrapper and TfPyObjWrapper? We weren't sure if there was something important we were missing here.

sunyab (changes requested):
Left questions about the need for TfPyBoostObjWrapper vs the existing TfPyObjWrapper. Thanks!

pxr/base/tf/anyWeakPtr.h
foisya-adsk: Thanks!  When we did the changes we wanted to keep things as separate as possible and we kept the return types the same as the ones we wrapped.  We will do as you suggest and unify under TfPyObjWrapper.  

 I think we we were too cautious when considering boost::python::object and boost::python::api::object... One is a "using ..." of the other!  We will simplify things.

pxr/base/tf/pyBoostObjWrapper.h
foisya-adsk: No, except that we we wanted to isolate the "proposed changes" as much as possible.  We will consolidate the code under TfPyObjWrapper...

pxr/base/tf/anyWeakPtr.h
bissonm: The other difference between the two uses is is that TfPyObjWrapper is holding a shared_ptr to an object.  However, object is basically a reference-counted holder as well, so it should have the same semantic.

gitamohr: > The other difference between the two uses is is that TfPyObjWrapper is holding a shared_ptr to an object. However, object is basically a reference-counted holder as well, so it should have the same semantic.

 The big important reason why this class exists and uses a shared_ptr is that in order to use boost::python::object you have to be holding the python GIL, since it invokes the python ref-counting operations under the hood, and boost::python::object does not take the GIL itself.  This is explained in the class doc for TfPyObjWrapper (https://graphics.pixar.com/usd/release/api/class_tf_py_obj_wrapper.html#details)

> Provides a wrapper around boost::python::object that works correctly for the following basic operations regardless of the GIL state: default construction, copy construction, assignment, (in)equality comparison, hash_value(), and destruction.
>
> None of those work correctly in the presence of an unlocked GIL for boost::python::object. This class only actually acquires the GIL for default construction, destruction and for some (in)equality comparisons. The other operations do not require taking the GIL.

foisya-adsk:
Martin and I just pushed an updated version with no extra TfPyBoostObjWrapper

pxr/base/tf/anyWeakPtr.h
foisya-adsk: It is worth noting that in the latest commit, in which we use only the wrapper class TfPyObjWrapper,  there will be a small overhead when creating the TfPyObjWrapper and deleting it.  For example:

    `TfPyObjWrapper. TfAnyWeakPtr::_PointerHolder<Ptr>::GetPythonObject() `

will return a `TfPyObjWrapper` and a subsequent call to the `Get() `method of a `TfPyObjWrapper` object will return the `boost::python::object`.

Thus, on the one hand, there is small cost when going through `TfPyObjWrapper`, and, on the other hand, it is protected because of the pointer management of `TfPyObjWrapper`. 

Is that OK?

foisya-adsk:
@spiffmon  Thanks to Gordon, my CLA status has been cleared.

spiffmon:
Yes!  We'll definitely get this in for the next release!

--spiff

MrKepzie:
Just to report that building on Apple clang version 13.0.0 (clang-1300.0.29.30)  I get a runtime crash where
the compile believes _getPyObj doesn't exist where, with this patch, it actually should. This leads to an offset in the function pointer table and makes the whole typeid system lead to undefined behavior.
I triple-checked it wasn't because of a mismatching includes and binaries and after rebuilds and verification the bug  was still happening. The only way around for me was to revert this patch.

```
    _EqualPtrFunc _equalPtr;
        _MakeMutableFunc _makeMutable;
-#ifdef PXR_PYTHON_SUPPORT_ENABLED
        _GetPyObjFunc _getPyObj;
-#endif // PXR_PYTHON_SUPPORT_ENABLED
        _StreamOutFunc _streamOut;
        _GetTypeidFunc _getTypeid;

```

sunyab:
@MrKepzie That sounds concerning. Were you running into this in a build with Python enabled or disabled? If you have more details it'd be great if you could file an issue so we can dig into it.

MrKepzie:
Build without python. I was paranoid and thought it was picking up another dylib at runtime, but no this was the one I compiled. Will file an issue

PR #1734, opened 2022-01-06
Register velocities, acclerations, and local constant-interpolation primvars on PIs

marktucker:
### Description of Change(s)

Register velocities, acclerations, and local constant-interpolation primvars
on point instancer primitives as hydra-accessible primvars. Otherwise
instancer-level settings applied as primvars can't be accessed by the
delegate, and constant-interpolation primvars can't be used to affect
prototypes marked as instanceable (where inheriting primvars doesn't
work).

This is a less ambitious version of #1731, which additoinally adds support for inherited primvars.

### DIFF
--- a/pxr/usdImaging/usdImaging/pointInstancerAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/pointInstancerAdapter.cpp
@@ -33,6 +33,7 @@
 
 #include "pxr/imaging/hd/perfLog.h"
 #include "pxr/imaging/hd/renderIndex.h"
+#include "pxr/imaging/hd/tokens.h"
 #include "pxr/usd/sdf/schema.h"
 #include "pxr/usd/usd/primRange.h"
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-7112

tcauchois:
For both constant primvars (here) and inherited primvars (in the other PR), you're changing the hydra schema in a way where we'll need to fix up our backends to now pull constant instancer data, which makes it harder to land.

Is it possible to get just velocities and accelerations by themselves?

marktucker:
I can certainly split outhe vel/accel parts of this PR into a separate reduced version of this PR. I'm not sure I see the downside to exposing the constant interp primvars though. If a given render delegate doesn't want to look at these values, it doesn't have to, and its behavior will continue to be the same as if hydra wasn't reporting these primvars.

Anyway, I'll create the PR once the 22.05 dust settles a bit. I'll be busy upgrading for the next little while.

PR #1737, opened 2022-01-10
Prevent double-creation of Bprims when using scene index emulation

marktucker:
### Description of Change(s)

When using scene index emulation, return after calling AddLegacyPrim when
inserting a new Bprim (otherwise it would be created twice).


jilliene:
Filed as internal issue #USD-7128

PR #1738, opened 2022-01-11
Update to MaterialX v1.38.3

kxl-adsk:
- Use CMake config directly from MaterialX install
- Update for API and library changes in v1.38.3
- Strengthen NodeDef discovery

This is the last update that is not version tagged in the C++ code.
Future updates will be bracketed using MATERIALX_MINOR_VERSION defines
introduced in v1.38.3.


pxr/imaging/hdSt/materialXShaderGen.cpp
```diff
@@ -384,6 +379,9 @@ HdStMaterialXShaderGen::_EmitMxFunctions(
         emitInclude(ShaderGenerator::T_FILE_TRANSFORM_UV, mxContext, mxStage);
     }
 
+    // Add light sampling functions
+    emitLightFunctionDefinitions(mxGraph, mxContext, mxStage);
+
```
bernardkwok, line 384: Adding a check here to see if needed: The albedo table initiation was affected by light refactoring and this fixes it. be0d3d65727ad343eaf469e9c179f772d31a9be2

pxr/usd/plugin/usdMtlx/CMakeLists.txt
```diff
@@ -19,10 +19,8 @@ pxr_plugin(usdMtlx
         usdShade
         usdUI
         usdUtils
-        ${MATERIALX_LIBRARIES}
-
-    INCLUDE_DIRS
-        ${MATERIALX_INCLUDE_DIRS}
+        MaterialXCore
```
bernardkwok, line 22: I'm curious why explicit library names are now used vs ${MATEIALX_LIBRARIES}. Will this affect users which use more than the
explicitly specified library currently ?
(This appears to be done in a few build configurations).

pxr/usd/plugin/usdMtlx/reader.cpp
```diff
@@ -1284,7 +1299,7 @@ _Context::AddShaderNode(const mx::ConstNodePtr& mtlxShaderNode)
 
     // Get the nodeDef for this shaderNode.
     mx::ConstNodeDefPtr mtlxNodeDef = mtlxShaderNode->getNodeDef();
-    if (mtlxShaderNode->getNodeDefString().empty()) {
+    if (!mtlxNodeDef) {
```
bernardkwok, line 1302: I wanted to double check here that this handles finding the <nodegraph>/<nodedef> associations found on <implementations>.

third_party/renderman-24/plugin/hdPrman/CMakeLists.txt
```diff
@@ -22,8 +22,14 @@ endif()
...
 set(optionalPublicClasses "")
 if (${PXR_ENABLE_MATERIALX_SUPPORT})
-    list(APPEND optionalLibs ${MATERIALX_LIBRARIES} hdMtlx)
-    list(APPEND optionalIncludeDirs ${MATERIALX_INCLUDE_DIRS})
+    list(APPEND optionalLibs
+        MaterialXCore
+        MaterialXFormat
+        MaterialXGenShader
+        MaterialXGenOsl
+        MaterialXRender
```
bernardkwok, line 30: I'm curious as to why there is a dependence on MaterialXRender ?
Generally I would assume this is not required.

pxr/usd/plugin/usdMtlx/reader.cpp
```diff
@@ -289,7 +290,11 @@ _FindMatchingNodeDef(
         }
 
         // Filter by types.
-        if (mtlxInterface && !mtlxInterface->isTypeCompatible(mtlxNodeDef)) {
+        if (mtlxInterface && !mtlxInterface->hasExactInputMatch(mtlxNodeDef)) {
```
bernardkwok, line 293: This is just a query: This assumes the latest API. I'm curious if code which handles multiple versions is not desirable e.g. if version preprocessor statements ?

build_scripts/build_usd.py
```diff
@@ -1345,20 +1345,43 @@ def InstallDraco(context, force, buildArgs):
...
-        # Currently MaterialX does not support shared builds on Windows
-        cmakeOptions = []
-        if Linux() or MacOS():
-            cmakeOptions += ['-DMATERIALX_BUILD_SHARED_LIBS=ON']
+        cmakeOptions = ['-DMATERIALX_BUILD_SHARED_LIBS=ON']
 
         cmakeOptions += buildArgs;
 
+        if "v1.38.3.zip" in MATERIALX_URL:
+            # This will be fixed in v1.38.4:
```
bernardkwok, line 1357: Perhaps comment that this is for when Python is not required.

pxr/usd/plugin/usdMtlx/utils.cpp
```diff
@@ -297,7 +297,7 @@ _ReadFromAsset(mx::DocumentPtr doc, const ArResolvedPath& resolvedPath,
                            newReadOptions);
         };
 
-    mx::readFromXmlString(doc, s, &readOptions);
+    mx::readFromXmlString(doc, s, searchPath, &readOptions);
```
bernardkwok, line 300: Is this purely to update the API, or will there be any behaviour change here.
If syntatical then should this instead be an empty search path ? Seems passing
`searchPath` is correct though. Just wondering.

JGamache-autodesk:
Github decided to fail all builds done in Windows 2016 VMs: https://github.com/actions/virtual-environments/issues/4312
Fixing this will require upgrading the azure-pipelines.yaml and azure-pypi-pipelines.yaml files to use the `windows-2019` VM which implies VS2019 instead of VS2017.

pxr/imaging/hdSt/materialXShaderGen.cpp
JGamache-autodesk: Storm has its own dome light functions (see line 531) and does not use the MaterialX albedo table. Adding this was necessary to get access to scene light functions. So no check required.

pxr/usd/plugin/usdMtlx/CMakeLists.txt
JGamache-autodesk: It allows linking only the libraries that are in use instead of all libraries.
Also, with this modern CMake implementation, the required include dirs (and also the required compiler definitions) gets added on a per-library basis.

pxr/usd/plugin/usdMtlx/reader.cpp
JGamache-autodesk: Context: at this exact point in the code, the MaterialX library has not been loaded. This means that a node that has a NodeDef string will fail to resolve, and previously would not be tried again. Changing the test to retry when the NodeDef was not found allows trying a second time to find the NodeDef using the string, but this time with libraries loaded.

third_party/renderman-24/plugin/hdPrman/CMakeLists.txt
JGamache-autodesk: I grepped for `#include <MaterialX` in the code and added what was found. In this case, I saw [one MaterialXRender include being accessed](https://github.com/PixarAnimationStudios/USD/blob/dev/third_party/renderman-24/plugin/hdPrman/matfiltMaterialX.cpp#L49).

pxr/usd/plugin/usdMtlx/reader.cpp
JGamache-autodesk: This requires knowing which version of MaterialX is in use and that can only be done starting with MaterialX v1.38.3.
I fully expect that submitting code that handles multiple versions will be the default for all future PRs.

build_scripts/build_usd.py
JGamache-autodesk: The default in USD is to build MaterialX without Python. This is how I found the fix was required.
I could possibly add a comment pointing to the fix if requested: https://github.com/autodesk-forks/MaterialX/pull/1344
And the fix for CMake relocability: https://github.com/autodesk-forks/MaterialX/pull/1350
That were both merged in ILM repo via: https://github.com/materialx/MaterialX/pull/805

pxr/usd/plugin/usdMtlx/utils.cpp
JGamache-autodesk: Just updating to [the latest API](https://github.com/materialx/MaterialX/pull/707). No behavior change.

jilliene:
Filed as internal issue #USD-7129

spiffmon:
Closing, per @seando-adsk , in favor of PR #1792

PR #1745, opened 2022-01-14
Add the concept of a "data sharing id" to assist render delegates in sharing data between prototype HdRprims

marktucker:
### Description of Change(s)
Implement a new GetDataSharingId method on HdSceneDelegate. This data sharing id, in USD terms, is the SdfPath of the implicitly create prototype primitive underlying an HdRprim. Hydra creates a unique HdRprim for every point instancer prototype, even if that prototype is an instaceable reference. It also creates a unique HdRprim for every unique set of primvars on instances, and based on a few other instance-variable attributes. But some render delegates could be much smarter about sharing data between HdRprims in their Sync calls if they had some hint that these HdRprims shared the same underlying data in the source scene delegate. Then the render delegate could decide for itself which primvars matter enough to create a new prototype rprim, or if many point instancer prims could share a common underlying mesh data structure between all the unique HdRprims.

Note that this change makes no attempt to alter the relationship between HdInstancers and HdRprims. It is just a hint that each render delegate can choose to use or ignore as it sees fit. So there should be no backward compatibility issues.

This change is made fully aware that it may (and hopefully will be) made obsolete sometime in the future by the new Hydra scene index framework, which may allow a single prototype HdRprim to be shared directly by multiple HdInstancers. But it is unclear how long that will take, and taking advantage of it would likely require a large rewrite of any render delegate that wanted to take advantage of any such new approach. So this still seems like it may be a worthwhile stopgap measure renderers could use right now.


marktucker:
I should add that the intention is for this GetDataSharingId method to be called in the HdRprim::Sync method, to decide whether the render delegate actually needs to query the geometry data for the prim, or can reuse some already-cached internal representation of the data fetched by another HdRprim with the same data sharing id. Hopefully that makes sense...

jilliene:
Filed as internal issue #USD-7146

tcauchois:
Hi Mark, to dive into this a bit more: are you hoping to use this in Storm, or Karma?  Storm does some deduplication at the topology and primvar buffer level already.

Also, are there specific types of scenes that are motivating this change?  I'm wondering if there's a usdImaging-side fix possible instead.

Finally, it's possible that sceneDelegate->GetScenePrimPath(rprimId, 0) will do what you want...

paoloemilioselva:
Hi Tom,

we are hoping to heavily rely on this on our side too, to avoid having to deduplicate a lot of prims and potentially have PrimAdapters to expand prims that we are not going to use anyway (still need to test it with PrimAdapters too thou).

This relates to a conversation we were having here: https://groups.google.com/g/usd-interest/c/QRWcrZKjaHQ

We'd like to give enough freedom to artists to be able to have layout-structures made of potentially a lot of PointInstancers and individual elements that could potentially point at the same asset, and we'd like to share the prototype across all PointInstancers and all individual elements too.

If we can basically access in Hydra the implicit prototypes created on the USD side, we can have a good way of informing the artists about how many renderer-exemplars they will be generating, by checking how many prototypes are actually generated in USD as implicit Prototypes.

I know that different primvar-descriptors will still make new prototypes on the Hydra side, but we can make primvar-edits in USD in a way that USD will generate new implicit prototypes, to ensure 1-to-1 matching of prototypes between USD and Hydra.

I don't know how this will change with the new Hydra and I don't know if the suggestion of using GetScenePrimPath is also possible, we'll need to investigate that, assuming that function is returning the path to the implicit prototype in the scenario I mentioned above.

This change is proving to be very useful so far, if used, and it doesn't change any existing behavior if not used, right ?

tcauchois:
Good to know you all are using it too.  I think UsdImagingDelegate could be smarter about instance aggregation (specifically, not breaking instances that are point instanced), and I'm hoping Hydra 2.0 instance aggregation is much more configurable.  Just for context, do you view this as a workaround for the current instance aggregation issues, or is this an API that you'd want around even when those are resolved?

Also: after thinking on it, ignore my comment about GetScenePrimPath.

paoloemilioselva:
I'm pretty sure by the time Hydra-2.0 comes, we have to review our delegates and rewrite a good part of them (both for Manuka and Gazebo).
But in the meanwhile (and for a while...) we'll be doing hydra offline renders in usd-21 (maybe, maybe... usd-22) and this API seems to be quite useful so far.

marktucker:
I don't know enough about Hydra 2.0 to know if it will provide an alternative solution to this problem, or if a render delegate would need a major rewrite before it could take advantage of any such new capabilities in hydra 2.0. However I think it's safe to say it will be on the order of years before everything get rewritten to directly use hydra 2.0. So I do see this (I hope) as a stopgap, but I expect it's a stopgap that will have to live on for a while.

Also, I realized I never answered your question (though I really thought I did). Karma does its own deduplication by looking at data arrays, similarly (I think) to what Storm does. So SideFX does not actually make use of this new API currently. We wanted to get some validation of it's usefulness (which we got from @paoloemilioselva ), and to hear Pixar's thoughts on the concept before we invest a bunch of effort into changing the way Karma does its de-duplication.

tcauchois:
For context: interpretation of USD already involves a lot of scene transformations, some of which can potentially break instancing.  I think the current set of instancing-breaking transformations mostly apply to storm; the big ones are purpose (applies to everybody if your renderer handles purpose) and inherited material binding/primvar flattening; both of the latter can vary per-instance in Prman, for example.  I expect there to be more instance-breaking transformations long-term, though, both to support new USD features and to handle scene index-based transformations.  Long term, I'm hoping to split instance aggregation into an op, so that different applications can drop in their own instance aggregators if they have different needs.  That'll take some time to land, of course.

That said, we'd like to take the API and unblock you all, but put in some caveats, since I think it'll be hard to promise that everything with the same sharing ID will always be aggregatable.  I'll comment on the sceneDelegate.h changes.

marktucker:
Thanks Tom. It's totally true that equal data sharing ids does not guarantee that two prototypes can be aggregated. Looking at the set of primvars, or even the primvar values on each prototype might make the renderer want to create separate prototypes. But the data sharing id can be hashed with these other factors to generate a renderer-specific "sharing id". This API is really just meant to be a good hint/starting point for making a custom decision about aggregating. At least that's how I see it.

paoloemilioselva:
I definitely agree on all of that.
I'd like to approach editing the scene from a user-standpoint as a "make sure you generate all the implicit prototypes you need" more than "make some changes and in Hydra you get some other prototypes".
If they can drive the implicit prototypes while editing their scenes, the artists will have a better understanding of how "complex" the scene is becoming too.
Thank you guys for this work!

tcauchois:
Mark: exactly.  So if we pull this in, I just want to document that it's a starting point and not authoritative.

tcauchois:
Paolo: this is a tricky area.  I'd definitely like to simplify the mapping, though.  And, as you call out, the workflows of editing instances vs editing prototypes are also tricky as well.

Issue #1520, opened 2021-05-17
Path resolution for mounted volumes in Windows does not work correctly


williamkrick:
### Description of Issue
Opening a USD file which uses anchored file paths from a mounted volume in windows doesn't work. The resolved volume GUID is not properly formatted and so the system cannot find the files. MayaUSD reports the volume GUID is something like `C:\??\Volume{009b65f5-0000-0000-0004-000000000000}\RTS\referenced_files\USD\barrelTests\Library\Assets\Barrels\Textures\repacked_BarrelWithLid\repacked_BarrelWithLid_UV_<UDIM>_BaseColor.png`.

According to the documentation on [volume GUIDs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-volume) the format of the volume GUID is not correct. The string should be something like `\\?\Volume{{009b65f5-0000-0000-0004-000000000000}\the\rest\of\the\path`.

Debugging the code, what I see happen is:

- There is a call to _WalkGraph in materialParamUtils.cpp for the shader node. That call finds the usd shade input attribute and calls UsdImaging_ResolveMaterialParamValue on it. The starting value of the assetPath looks like `../Textures/repacked_BarrelWithLid/repacked_BarrelWithLid_UV_<UDIM>_BaseColor.png`
- As a part of resolving the path there is a call to `SdfComputeAssetPathRelativeToLayer`. After this call the resolved path is correct. If I copy that string from the debugger I can get windows to open it. The resolved path is: `d:/MAYA-105710/RTS/referenced_files/USD/barrelTests/Library/Assets/Barrels/Textures/repacked_BarrelWithLid/repacked_BarrelWithLid_UV_1001_BaseColor.png`. The code still has to replace the 1001 with <UDIM> again, but that part of the code just hasn't execute yet. This path is correct and can be correctly opened.
- Before resolving finishes there is a call to _ResolveSymlinks. This discovers that d:\MAYA-105710 is a `IO_REPARSE_TAG_MOUNT_POINT` and returns the contents of the mount point reparse buffer, which are `\\??\\Volume{e7716c21-89be-11eb-bea8-50e085f32ec9}\\`
- Eventually, the incorrect volume GUID path gets concatenated with the rest of the path, and HdVP2Material tries to find the UDIM textures, but none of them work because the path is not correct.

[...]

PR #1746, opened 2022-01-14
Path resolution for mounted volumes in Windows does not work correctly

seando-adsk:
### Description of Change(s)

Some Windows reparse points represent drives mounted on an arbitrary directory. The data kept in the reparse point is an internal Windows object reference that cannot be used as a path. The change avoids trying to use it as a path by detecting its format: it starts with `\??` (backslash and double question marks).

### Fixes Issue(s)
- #1520


jilliene:
Filed as internal issue #USD-7147

PR #1755, opened 2022-01-28
Expand Python exception for debug output

kxl-adsk:
When converting an exception that originates in Python code, add the
error message and traceback to the TF_ERROR message to allow clients to
debug their Python callbacks.

### DIFF
--- a/pxr/base/tf/pyError.cpp
+++ b/pxr/base/tf/pyError.cpp
@@ -111,7 +111,29 @@ TfPyConvertPythonExceptionToTfErrors()
                     TfDiagnosticMgr::GetInstance().AppendError(*e);
             }
         } else {
-            TF_ERROR(exc, TF_PYTHON_EXCEPTION, "Tf Python Exception");
+            // Give as much info as possible to allow debugging
+            std::string msg = "Tf Python Exception";
+            extract<std::string> valueExtractor(exc.GetValue().get());
+            if (valueExtractor.check()) {
+                std::string errValue = valueExtractor();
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-7165

pmolodo:
Poking this MR, because it would be immensely helpful in some situations...

gitamohr:
As it is today, the entire python exception state travels in the TF_ERROR as the `DiagnosticInfo`.  This can be retrieved by calling `GetInfo<TfPyExceptionState>()` on the error.  Since callers could then easily produce this kind of formatted output themselves, I'm wondering why it's necessary to do this upfront here.  Is there something that's lost by leaving this to callers?

pmolodo:
Interesting - I didn't realize this was possible.
And looking into the code, it looks like we can already enable printing of the python stacktrace by simply setting `TF_LOG_STACK_TRACE_ON_ERROR`, since that will cause [TfDiagnosticMgr::PostError](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/diagnosticMgr.cpp#L222) to call [_PrintDiagnostic](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/diagnosticMgr.cpp#L236), which results in [TfPyExceptionState::GetExceptionString()](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/diagnosticMgr.cpp#L677) being printed.
So - yes, I think we have everything we need already.  Apologies!

PR #1757, opened 2022-02-02
Support shapes with blendshapes but no skinning when using the CPU computation

cameronwhite:
Currently, a shape that has blendshapes but no skinning isn't visible when using the CPU ExtComputation, but the GPU version in Storm behaves okay.
To reproduce, load the [attached file](https://github.com/PixarAnimationStudios/USD/files/7989033/blendshapes.usda.tar.gz) with `usdview blendshapes.usda --camera /cameras/camera1` and use a delegate such as Embree, or set `USDSKELIMAGING_FORCE_CPU_COMPUTE=1` in the environment to reproduce with Storm

### DIFF
--- a/pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
+++ b/pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
@@ -915,51 +915,54 @@ UsdSkelImagingSkeletonAdapter::InvokeComputation(
                             blendShapeWeights.UncheckedGet<VtFloatArray>(),
                             skinnedPoints);
 
-    if (!hasConstantInfluences.UncheckedGet<bool>()) {
+    if (numInfluencesPerComponent.UncheckedGet<int>() > 0) {
 
-        UsdSkelSkinPointsLBS(
-            geomBindXform.UncheckedGet<GfMatrix4f>(),
-            skinningXforms.UncheckedGet<VtMatrix4fArray>(),
-            influences.UncheckedGet<VtVec2fArray>(),
-            numInfluencesPerComponent.UncheckedGet<int>(),
-            skinnedPoints);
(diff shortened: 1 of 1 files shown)


cameronwhite:
(Github's diff isn't aligned very well, but it's actually just a one line change + indentation shift if you view with whitespace disabled)

spiffmon:
Thank you @cameronwhite - we'll try to get this looked at soon!  Could I ask you, though, to go as far as you can with a testcase?  I think that could be as simple as a source usda file (e.g. your attached file) with an embedded camera that demonstrates the problematic scenario, added in a suitably named subdir of usdImagingGL/testenv ?  We can take care of the rest of the harness... (and we're making progress on being able to publish the internal tests we have here, with NVidia's help) but having it be part of the PR will help!

cameronwhite:
Sure! I'll take a look at adding a test there

cameronwhite:
It looks like the existing test file for [testUsdImagingGLSkelBlendShapes](https://github.com/PixarAnimationStudios/USD/tree/dev/pxr/usdImaging/usdImagingGL/testenv/testUsdImagingGLSkelBlendShapes) does exhibit the issue if it's run with the CPU computation (`USDSKELIMAGING_FORCE_CPU_COMPUTE=1`).
So, I think it would just be a matter of having the test harness render versions for the CPU and GPU skinning, similar to what it looks like [testUsdImagingGLUsdSkel](https://github.com/PixarAnimationStudios/USD/tree/dev/pxr/usdImaging/usdImagingGL/testenv/testUsdImagingGLUsdSkel/baseline) does

spiffmon:
Haha - thanks @cameronwhite ; rookie move I was looking at the release branch, where we hadn't yet published any of our tests!

jilliene:
Filed as internal issue #USD-7173

PR #1767, opened 2022-02-11
Implement reverse lookup table in instanceAdapter to improve performance of getting proto prim

csyshing:
This PR is for the issue created earlier here to improve the Maya playback performance: https://github.com/PixarAnimationStudios/USD/issues/1740

### Description of Change(s)

- Implemented the "reverse lookup table" mentioned in _GetProtoPrim() method, which can improve the performance of UsdImagingInstanceAdapter::UpdateForTime().

### DIFF
--- a/pxr/usdImaging/usdImaging/instanceAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/instanceAdapter.cpp
@@ -334,6 +334,13 @@ UsdImagingInstanceAdapter::_Populate(UsdPrim const& prim,
                 instancerData.childPointInstancers.insert(protoPath);
             }
 
+            // Store cache path and instancer path mapping that
+            // helps to improve reversing lookup in `_GetProtoPrim()`
+            const auto& cacheIt = _protoPrimCacheMap.find(protoPath);
+            if (cacheIt == _protoPrimCacheMap.end()) {
+                _protoPrimCacheMap[protoPath] = instancerPath;
+            }
+
             TF_DEBUG(USDIMAGING_INSTANCER).Msg(
                 "[Add Instance NI] <%s>  %s (%s), adapter = %s\n",
                 instancerPath.GetText(), protoPath.GetText(),
@@ -1355,6 +1362,11 @@ UsdImagingInstanceAdapter::_ResyncPath(SdfPath const& cachePath,
         }
     }
 
+    // Clear proto prim cache
+    if (!_protoPrimCacheMap.empty()) {
+        _protoPrimCacheMap.clear();
+    }
+
     // Actually resync everything.
     for (SdfPath const& instancer : instancersToUnload) {
         _ResyncInstancer(instancer, index, reload);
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-7201

pxr/usdImaging/usdImaging/instanceAdapter.h
```diff
@@ -621,6 +621,12 @@ class UsdImagingInstanceAdapter : public UsdImagingPrimAdapter
     typedef TfHashMultiMap<SdfPath, SdfPath, SdfPath::Hash>
         _PrototypeToInstancerMap;
     _PrototypeToInstancerMap _prototypeToInstancerMap;
+
+    // Map from instance cache path to their instancer path.
+    // Note: this is for reducing proto prim lookup in _GetProtoPrim method.
+    typedef std::unordered_map<SdfPath, SdfPath, SdfPath::Hash>
+        _ProtoPrimCacheMap;
+    _ProtoPrimCacheMap _protoPrimCacheMap;
```
tcauchois: If you want to follow the naming convention you could call this _protoPrimToInstancerMap.

pxr/usdImaging/usdImaging/instanceAdapter.cpp
```diff
@@ -334,6 +334,13 @@ UsdImagingInstanceAdapter::_Populate(UsdPrim const& prim,
                 instancerData.childPointInstancers.insert(protoPath);
             }
 
+            // Store cache path and instancer path mapping that
+            // helps to improve reversing lookup in `_GetProtoPrim()`
+            const auto& cacheIt = _protoPrimCacheMap.find(protoPath);
+            if (cacheIt == _protoPrimCacheMap.end()) {
+                _protoPrimCacheMap[protoPath] = instancerPath;
+            }
```
tcauchois, line 342: One non-obvious note: this adapter adds proto prims that are either geometry or other instancers.  Geometry takes the form: /path/to/instancer.proto0_cube_id0.  _GetProtoPrim should be able to find these without using your new map (that's the first clause of _GetProtoPrim).  Instancers can't be re-rooted the same way, and they are probably what makes _GetProtoPrim slow.

All of this to say, you might be able to save some memory by wrapping this new code in `if (!isLeafInstancer) {`, since prims that fail that if statement should hit the (still there) first half of _GetProtoPrim.

pxr/usdImaging/usdImaging/instanceAdapter.cpp
```diff
@@ -1355,6 +1362,11 @@ UsdImagingInstanceAdapter::_ResyncPath(SdfPath const& cachePath,
         }
     }
 
+    // Clear proto prim cache
+    if (!_protoPrimCacheMap.empty()) {
+        _protoPrimCacheMap.clear();
+    }
```
tcauchois: This will break any kind of change processing, since _GetProtoPrim will start failing for all instancers as soon as you resync one of them.

What you want to do instead is go to _ResyncInstancer (called below, line ~2000); find the place where we iterate over the prim map removing those prims from hydra; and add a line removing those prim paths from the protoPrimCacheMap as well.  It'll be a little slower than clearing the map, but won't break change processing.

tcauchois:
We're super excited to merge this!  First, though, I left a few comments on the code: some suggestions, but one blocking issue with how you handle resyncs.  If you can send another iteration over we can pull this down and run it through our test suite.

csyshing:
Thanks @tcauchois , I can't seem to see those comments, maybe I am missing something?

pxr/usdImaging/usdImaging/instanceAdapter.cpp
csyshing: Code updated.

pxr/usdImaging/usdImaging/instanceAdapter.h
csyshing: Done.

pxr/usdImaging/usdImaging/instanceAdapter.cpp
csyshing: That's also been updated.

csyshing:
@tcauchois , I have updated the code (64e88db40), thanks for the feedback!

tcauchois:
Thanks! We'll send this through the test suites again and hopefully we can get it checked in soon.

PR #1771, opened 2022-02-15
[hdx] Expose secondary selection hilight color as a settable value.

marsupial:
### Description of Change(s)
Currently Hydra has a notion of two selection types: "selected" and "over".
While the color of "selected" is customizable, the color of "over" is hard-coded and locked to a blueish hue.
We'd like the ability to expose this so applications can customize the color for any style they may be using.

### Fixes Issue(s)
- "over" selected state always being rendered as a blue.


jilliene:
Filed as internal issue #USD-7219

PR #1775, opened 2022-02-18
Small fix to GfVec Slerp test

ablev-nv:
### Description of Change(s)

I was looking at the GfVec tests and noticed that one of the tests compares a vector to itself rather than the expected result of the Slerp call. Heavy stuff here :smile:


jilliene:
Filed as internal issue #USD-7223

Issue #1723, opened 2021-12-18
[Building Error] Failed to build USD - Arch Linux


adro79:
I'm failing to build USD in my Arch Linux system. My configuration is Python 3.9.9 and PySide2 5.15.2

Here's the last few lines before the error:
````
Consolidate compiler generated dependencies of target arch
[ 32%] Built target hdTiny_headerfiles
[ 32%] Built target usdRecursivePayloadsExample_headerfiles
[ 33%] Built target arch
Consolidate compiler generated dependencies of target tf
[ 39%] Built target tf
Consolidate compiler generated dependencies of target js
Consolidate compiler generated dependencies of target garch
[ 40%] Built target js
[ 41%] Built target garch
Consolidate compiler generated dependencies of target _garch
[ 41%] Built target _garch
Consolidate compiler generated dependencies of target gf
[ 44%] Built target gf
Consolidate compiler generated dependencies of target trace
Consolidate compiler generated dependencies of target cameraUtil
[ 45%] Built target trace
[ 45%] Built target cameraUtil
Consolidate compiler generated dependencies of target _tf
Consolidate compiler generated dependencies of target work
Consolidate compiler generated dependencies of target _trace
Consolidate compiler generated dependencies of target _cameraUtil
[ 45%] Building CXX object pxr/base/work/CMakeFiles/work.dir/singularTask.cpp.o
[ 45%] Built target _trace
[ 45%] Built target _cameraUtil
[ 46%] Built target _tf
Consolidate compiler generated dependencies of target vt
[ 46%] Built target vt
Consolidate compiler generated dependencies of target pxOsd
[ 47%] Built target pxOsd
[ 47%] Building CXX object pxr/imaging/pxOsd/CMakeFiles/_pxOsd.dir/wrapMeshTopology.cpp.o
[ 48%] Building CXX object pxr/imaging/pxOsd/CMakeFiles/_pxOsd.dir/module.cpp.o
[ 48%] Building CXX object pxr/imaging/pxOsd/CMakeFiles/_pxOsd.dir/wrapSubdivTags.cpp.o
[ 48%] Building CXX object pxr/imaging/pxOsd/CMakeFiles/_pxOsd.dir/wrapTokens.cpp.o
[ 48%] Building CXX object pxr/imaging/pxOsd/CMakeFiles/_pxOsd.dir/wrapMeshTopologyValidation.cpp.o
Consolidate compiler generated dependencies of target _vt
[...]

Issue #1721, opened 2021-12-17
USD fails to build with gcc-11


mont29:
https://github.com/PixarAnimationStudios/USD/blob/9b4cea6217bc49c166fd3686b786a2f18d3e6516/pxr/base/work/singularTask.h#L111

Current code builds fine on gcc-10, but this seems to require explicit `std` namespace with newer gcc: `std::size_t old = count;`

mont29:
Note that this issue also exists in recent releases, at least in 21.02 one.

 (in 21.02 release there is another similar one in `demangle.cpp`, where an [`#include <cxxabi.h>`](https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/base/arch/demangle.cpp#L141) happens inside of the `PXR_NAMESPACE_OPEN_SCOPE` namespace scope. But this seems to have been fixed in [current code](https://github.com/PixarAnimationStudios/USD/blob/faed18ce62c8736b02413635b584a2f637156bad/pxr/base/arch/demangle.cpp#L42) already.)

jilliene:
Filed as internal issue #USD-7093

mont29:
Hi, any update on this issue? Would submitting a pull request help getting it fixed faster?

This is a one-line fix, but not sure which of both suggested solutions (in the report and in #1723 one) would be the desired one here...

spiffmon:
Hi @mont29 and @sobotka - we do not build or validate against GCC's higher than our documented, but we would absolutely consider a PR against this repo!

mont29:
While generating the PR in own fork I found a few more similar issues (for Blender we typically build a minimal USD).

The root of these issues is that GCC 11 is no longer doing some implicit includes for some std files: https://www.gnu.org/software/gcc/gcc-11/porting_to.html#header-dep-changes

sybrenstuvel:
What is the status of this task? #1776 has been merged, so shouldn't this issue be resolved now?

spiffmon:
Hi @sybrenstuvel , I believe the main reason we have not closed this Issue or #1723 is that we do not build or test with gcc11, and as @mont29 mentioned in #1776 , there could be other, similar issues yet lurking in the codebase.  If someone confirms they can build a complete USD+Hydra+plugins build on gcc11, I think we'd be more comfortable closing this.

TheJackiMonster:
[...]

PR #1776, opened 2022-02-18
Fix compilation on GCC11.

mont29:
### Description of Change(s)

GCC 11 removed some 'implicit includes', see
https://www.gnu.org/software/gcc/gcc-11/porting_to.html#header-dep-changes
for reference.

NOTE: Could not build a full USD (with all optional modules enabled), so there may be a few more similar cases in the code base.

### Fixes Issue(s)
- #1721
- #1723


jilliene:
Filed as internal issue #USD-7224

Issue #1774, opened 2022-02-18
Value Clips memory usage grows with the square of the number of "times" entries


marktucker:
### Description of Issue
If you create a value clip with 15000 "times" entries, the Usd_Clip data structures consume about 5GB of data, regardless of the size of the data in the actual clip files.

This memory if consumed by the fact that each Usd_Clip object makes a copy of the 15000-entry timeMapping array created in the Usd_ClipSet constructor. I _think_ each Usd_Clip applies the same sorting procedure to the data, and makes it's own copy. Seems like the obvious solution would be to have the Usd_ClipSet constructor massage the data, then pass a shared pointer to each Usd_Clip constructor. I'd be happy to make a PR for this if you think this is the correct approach to fixing this.

### Steps to Reproduce
1. Unzip the attached file:
[value_clip_memory.zip](https://github.com/PixarAnimationStudios/USD/files/8094402/value_clip_memory.zip)
2. Load the box_loop.usd file into usdview. Memory usage will be 5-6GB, even though this is just a box flipping back and forth between two positions.

### Package Versions
Tested in 20.08, 21.08, and 21.11

jilliene:
Filed as internal issue #USD-7222

spiffmon:
Hey @marktucker - oof, we've never exercised the explicit form to that scale; good find!  Your approach seems to us like the right thing, and we'd be happy to get a PR.

LucaScheller:
Hey @jilliene / @spiffmon, 
hope you are doing well! 

We (at RISE | Visual Effects Studios) were wondering what the current state of the attached PR from Mark is? Has it been reviewed/are there plans to merge it soon (https://github.com/PixarAnimationStudios/USD/pull/1777)? Thanks again Mark for submitting the PR :)

We've been running into this issue quite often for heavy FX scenes, the fix would help us quite a bit if it gets released soon :)

Cheers,
Luca

spiffmon:
Hi @LucaScheller , although it hasn't been pushed out to gitHub yet, I believe we merged the PR just last week, so it will be in the dev branch soon, and in the next release, tentatively 22.08

LucaScheller:
Thanks for the update, that's great to hear :)

sunyab:
PR #1777 was merged and released in 22.08, closing this out. Thanks!

PR #1777, opened 2022-02-19
Change Usd_Clip to hold a shared pointer to the array of time mappings

marktucker:
### Description of Change(s)
Change Usd_Clip to hold a shared pointer to the array of time mappings.
Set up this array in Usd_ClipSet where all the Usd_Clip objects are
allocated. Prevents N^2 memory usage on the number of time mappings in
a clip set.

### Fixes Issue(s)
- #1774 

### DIFF
--- a/pxr/usd/usd/clip.cpp
+++ b/pxr/usd/usd/clip.cpp
@@ -69,23 +69,6 @@ UsdGetClipRelatedFields()
     };
 }
 
-struct Usd_SortByExternalTime
-{
-    bool 
-    operator()(const Usd_Clip::TimeMapping& x, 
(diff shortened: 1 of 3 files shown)


jilliene:
Filed as internal issue #USD-7226

pberto:
@marktucker nice improvement here.

Since you brought this up, how are you dealing with the other issues with memory limitation and clips? I mean when you actually do have a lot of data for each frame, and not just a box, it is possible to incur in going out of memory when usdStitchClip will go through all the data. At the present state the only solution that I could find was to make a clip with few frames and then edit (by scripting or even by hand) the timecodes in the clip file and in the topology file.

PR #1779, opened 2022-02-23
Add trace custom callback

aloysbaillet:
### Add optional custom callback for the trace library

This PR adds a completely optional way (both at compilation time and at execution time) to hook another profiling library to the trace library, leveraging the existing `TRACE_*` macros for use in other systems.

The `__FILE__` and `__LINE__` metadata was also added to the keys as it is quite nice to have, and most profiling system can display the information if present (I have not added that data in the chrome trace data though, but that could be done).

I've added an example [Tracy](https://github.com/wolfpld/tracy) wrapper client with rough implementation to provide an example on how the new callback API can be used, see gif of `usdview --tracy ...` running here: 
usdviewTracy

New `build_usd.py` options:
* `--tracecb`: to activate the new custom callback API
* `--tracy`: to activate the thin Tracy wrapper

One issue with the compilation option is that the binaries generated with and without custom callbacks are not ABI compatible, which can make life harder if one wanted to just replace the trace library without having to recompile everything... I'd love some input on that!

Notes:
* The `TRACE_CUSTOM_CALLBACK_DEFINE` macro needs to appear in at least one compilation unit for any library using the `TRACE_*` macros
* The new `TraceCustomCallback::RegisterCallbacks` API works with raw function pointers for speed and to avoid a cyclic dependency with the plug library (which already depends on the trace library).

Also, the compilation of the Tracy UI is not included in the build script as it requires a few packages that can be harder to install, but prebuilt binaries are easily found.


jilliene:
Filed as internal issue #USD-7232

spiffmon:
Thank you, @aloysbaillet - we think this is interesting and useful.  It may be a beat before we can get you good feedback, though, as we are a bit underwater, and this is meaty.

aloysbaillet:
Thanks Spiff! No worries, I'd love feedback and ideas when you get a chance, and I know everyone's busy. No rush!

PR #1787, opened 2022-02-26
Fix UDIM file resolving when no "1001" file exists

marktucker:
### Description of Change(s)
Correct problem with fix for #1329. When symlink resolution was removed, so
was any testing for whether or not the "Resolve" call succeeded. This change
restores the looping to look for an existing tile file rather than always
returning the attempted resolution of the first tile file.

### Fixes Issue(s)
- #1329 


jilliene:
Filed as internal issue #USD-7245

Issue #1733, opened 2022-01-05
Debug assertion when a material has no parameters


williamkrick:
### Description of Issue
USD throws a debug assertion opening a USD file which has a Shader which has no parameters. In release builds this error is not detected and the scene appears to load correctly.

### Steps to Reproduce
```
#usda 1.0
(
    defaultPrim = "GEO"
    metersPerUnit = 0.01
    upAxis = "Y"
)

def Xform "GEO"
{
    def Mesh "BILLBOARD" (
        prepend apiSchemas = ["MaterialBindingAPI"]
        kind = "component"
    )
    {
        uniform bool doubleSided = 1
        float3[] extent = [(-1.5, 0, -1.5), (1.5, 0, 1.5)]
        int[] faceVertexCounts = [4]
        int[] faceVertexIndices = [0, 1, 3, 2]
        rel material:binding = </GEO/Looks/TEXTUREDA>
        point3f[] points = [(-1.5, 0, 1.5), (1.5, 0, 1.5), (-1.5, 0, -1.5), (1.5, 0, -1.5)]
        texCoord2f[] primvars:map1 = [(0, 0), (1, 0), (0, 1), (1, 1)] (
            interpolation = "faceVarying"
        )
        int[] primvars:map1:indices = [0, 1, 3, 2]
    }

	def Scope "Looks"
	{
		def Material "TEXTUREDA"
		{
			token inputs:file1:varname = "map1"
			token outputs:surface.connect = </GEO/Looks/TEXTUREDA/TEXTUREDA.outputs:surface>

			def Shader "TEXTUREDA"
			{
				uniform token info:id = "UsdPreviewSurface"
				token outputs:surface
			}
		}
	}
}
```

1. Open the attached usda file with a debug version of USDView. Debug assertion & crash!
2. If I modify the given scene to add `color3f inputs:specularColor = (1, 1, 1)` to TEXTUREDA then USDView can successfully open the file in debug mode.

The debug assertion occurs when trying to call front on an empty vector. The error occurs in _ConvertHdMaterialNetworkToHdDataSources in dataSourceLegacyPrim.cpp of USD.

### System Information (OS, Hardware)
Windows

### Package Versions
USD 21.11 debug build!

spiffmon:
Nice find, @williamkrick !

jilliene:
Filed as internal issue #USD-7111

ikryukov:
[...]

PR #1791, opened 2022-02-28
Taking front of empty vector crashes in debug

seando-adsk:
### Description of Change(s)
Because MSVC will throw an unstoppable exception of a bad memory access.
Not an issue in release.

### Fixes Issue(s)
- #1733


jilliene:
Filed as internal issue #USD-7251

PR #1795, opened 2022-03-05
Fix usdview promoting render setting types from float to double

### LINKED ISSUE #913: [usdview] Float Render Delegate Render Settings are converted to double when using the Hydra Settings window
### Description of Issue
As of latest dev, usdview changes float render settings to double when using the Hydra Settings window. This happens, because usdview passes values from QDoubleSpinBox to VtValue through the python wrappers and it gets converted to double automatically (see Vt_ValueFromPython::_construct). Since UsdImagingGLEngine only allows int, float, bool and std::string render settings (see UsdImagingGLEngine::GetRendererSettingsList), the render delegates are forced to use float for floating point render settings, but they receive either float or double when used in usdview.

I was looking at a few possible workarounds and I think the easiest solution is convert double values to float in UsdImagingGLEngine::SetRendererSettings. Something like:

```
void
UsdImagingGLEngine::SetRendererSetting(TfToken const& id, VtValue const& value)
{
    if (ARCH_UNLIKELY(_legacyImpl)) {
        return;
    }
    TF_VERIFY(_renderIndex);
    if (value.IsHolding<double>()) {
        _renderIndex->GetRenderDelegate()->SetRenderSetting(
            id, VtValue(static_cast<float>(value.UncheckedGet<double>())));
    } else {
        _renderIndex->GetRenderDelegate()->SetRenderSetting(id, value);
    }
}
```
### Steps to Reproduce
1. add a float render setting to your render delegate
2. open the hydra settings window, press ok
3. the render delegate receives a double boxed in a vtvalue

### Package Versions
latest dev

pablode:
### Description of Change(s)

For render settings of type float, usdview passes a double to the render delegate.
To fix this, the type is corrected in UsdImagingGLEngine before it is passed to the delegate.

### Fixes Issue(s)

This fixes #913.

### DIFF
--- a/pxr/usdImaging/usdImagingGL/engine.cpp
+++ b/pxr/usdImaging/usdImagingGL/engine.cpp
@@ -1153,7 +1153,13 @@ UsdImagingGLEngine::SetRendererSetting(TfToken const& id, VtValue const& value)
     }
 
     TF_VERIFY(_renderDelegate);
-    _renderDelegate->SetRenderSetting(id, value);
+    // usdviewq passes a double as VtValue since Python has no 'float' type
+    if (value.IsHolding<double>()) {
+        _renderDelegate->SetRenderSetting(
+            id, VtValue(static_cast<float>(value.UncheckedGet<double>())));
+    } else {
+        _renderDelegate->SetRenderSetting(id, value);
+    }
 }
 
 void


pxr/usdImaging/usdImagingGL/engine.cpp
```diff
@@ -1153,7 +1153,13 @@ UsdImagingGLEngine::SetRendererSetting(TfToken const& id, VtValue const& value)
     }
 
     TF_VERIFY(_renderDelegate);
-    _renderDelegate->SetRenderSetting(id, value);
+    // usdviewq passes a double as VtValue since Python has no 'float' type
+    if (value.IsHolding<double>()) {
+        _renderDelegate->SetRenderSetting(
+            id, VtValue(static_cast<float>(value.UncheckedGet<double>())));
```
spiffmon, line 1159: There is builtin support for value casting:` VtValue::Cast<float>(value)`

Also, if we are going to do this here, instead of in the python bindings, then presumably the goal is to fully protecct render delegates such that they only receive the declared three value types.  If so, then should we not also add conversion-guards for `half`, `long`, `TfToken`, and issue a coding error if we see any other type?

pxr/usdImaging/usdImagingGL/engine.cpp
```diff
@@ -1153,7 +1153,13 @@ UsdImagingGLEngine::SetRendererSetting(TfToken const& id, VtValue const& value)
...
+    } else {
```
spiffmon, line 1160: I'd also make the conversion chain just about converting `value` or issuing an error and returning, so that there's only one call to SetRenderSetting() at the end.

jilliene:
Filed as internal issue #USD-7256

Issue #1797, opened 2022-03-07
Custom camera attributes aren't considered in `UsdImagingCameraAdapter::TrackVariability`


robpieke:
### Description of Issue

If you apply a custom schema to a camera (for example to attach a background image) and populate the custom attribute(s) with time-sampled data, Hydra doesn't trigger the appropriate syncing on frame changes. This is due to `UsdImagingCameraAdapter::TrackVariability` having a hard-coded list of attributes to consider.

In contrast, `UsdImagingLightAdapter::TrackVariability` iterates over *all* attributes.

PR for consideration inbound ... basically just copying what the lightAdapter does, at the end of the cameraAdapter :)

jilliene:
Filed as internal issue #USD-7258

sunyab:
PR #1798 was merged and released with v22.08. Closing, thanks!

PR #1798, opened 2022-03-07
Consider *all* attributes for `UsdImagingCameraAdapter::TrackVariability`

robpieke:
### Description of Change(s)

Copying behaviour from `lightAdapter.cpp` verbatim (other than changing `HdLight` to `HdCamera` and `light` to `camera`) to allow *all* attributes (e.g., custom attributes) to be recognised in `TrackVariability`.

This may become meaningless with future versions/refactors of Hydra, but wanted to submit for completeness as we're currently using this patch.

### Fixes Issue(s)
- #1797


jilliene:
Filed as internal issue #USD-7259

spiffmon:
@robpieke , I don't know whether pragmatically we'll want to accept this change, but I wanted to point out that the two cases are not actually equivalent, but you can't tell due to unfortunate code evolution.  I just did the archeology, and that lightAdapter code predates the change to make lights connectable.  I believe the **right** thing to do there is to consider only properties in the "inputs:" namespace.

The reason we don't want to just consider all custom attributes is because we might wind up paying dearly for custom pipeline data that is irrelevant to imaging.  The drum we've been beating is that any additional behaviors should be captured by Applied schemas, and new Hydra will allow you to attach behaviors for them that combine/extend the base adapters.

But since you can't do that yet, accepting this _may_ be a short-term compromise, but I'm not sure.

robpieke:
Thanks @spiffmon ... we needed *a* solution to handle animated background plates, so it was either extending the hardcoded list of attributes to check, or a catch-all. Whether we're unique, or whether this is a broader community issue ... I won't venture a guess. But, either way, we wanted to be transparent and share "this is what we're currently doing".

Looking forward to new Hydra :)

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -199,6 +199,24 @@ UsdImagingCameraAdapter::TrackVariability(UsdPrim const& prim,
                    false)) {
         return;
     }
+
+    if ((*timeVaryingBits & HdCamera::DirtyParams) != 0)
+        return;
+
```
tcauchois: Delete 92-205, since that's duplicate work we don't want to do.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -199,6 +199,24 @@ UsdImagingCameraAdapter::TrackVariability(UsdPrim const& prim,
...
+    // If any other camera attributes are time varying
+    // we will assume all camera params are time-varying.
+    const std::vector<UsdAttribute> &attrs = prim.GetAttributes();
+    for (UsdAttribute const& attr : attrs) {
+        // Don't double-count transform attrs.
+        if (UsdGeomXformable::IsTransformationAffectedByAttrNamed(
+                attr.GetBaseName())) {
+            continue;
+        }
```
tcauchois, line 133: `if (attr.GetBaseName() == UsdGeomTokens->clippingPlanes) { continue; }`
... otherwise, clipping planes changes will force you to recompute the projection matrix.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -199,6 +199,24 @@ UsdImagingCameraAdapter::TrackVariability(UsdPrim const& prim,
...
+        if (attr.GetNumTimeSamples() > 1) {
```
tcauchois: You should use attr.ValueMightBeTimeVarying() here instead.

tcauchois:
Hey Rob!  I don't know if you're all set in your tree; and like Spiff mentioned, the plan in the future is to add custom behaviors like yours via API schemas which get imaging adapters.  But until that lands, if you're interested in merging this I left a few notes.  After those are addressed this seems like a good short-term workaround.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
robpieke: It's not *exactly* duplicate work is it? The `_IsVarying` call takes a `perfToken` for example.

tcauchois: You could add the perf token increment below, if you want, or use _IsVarying on 215.  The main work that both you and _IsVarying are doing (aside from setting debug stats) is calling ValueMightBeTimeVarying and setting the dirty bit, though.  ValueMightBeTImeVarying (or GetNumTimeSamples()) can both be kind of expensive, which is why it's better if we don't call them twice per attribute.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
robpieke: Actually, I didn't understand this. Are you suggesting replacing the code with the clipping planes check? Adding the clipping planes check first? Why? :)

pxr/usdImaging/usdImaging/cameraAdapter.cpp
robpieke: Okay, but just before I delete the explicit checks against focalLegnth, fStop, etc, I'll paste this code:
```
    if (_IsVarying(prim,
                   UsdGeomTokens->horizontalAperture,
                   HdCamera::DirtyParams,
                   HdCameraTokens->horizontalAperture,
                   timeVaryingBits,
                   false)) {
        return;
    }
```
I can replace my new code with a generic
```
        if (_IsVarying(prim,
                       attr.GetName(),
                       HdCamera::DirtyParams,
                       attr.GetName(),
                       timeVaryingBits,
                       false)) {
            return;
        }
```
But I'll point out for the sake of absolute completeness that we're no longer explicitly using the UsdGeomTokens for the attribute name and the HdCameraTokens for the perf token but, rather, just the attribute name for both. Maybe in practice this is fine, but I want to highlight it.

tcauchois: Yeah, that's ok.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
tcauchois: Just like you "continue" if you run into an xform attribute, since you don't want it to affect "DirtyParams", similarly you don't want to process "clippingPlanes", since that is handled above as DirtyClipPlanes and shouldn't set "DirtyParams".

robpieke: Ah, gotcha

robpieke:
Hopefully I've addressed all the feedback now :)

tcauchois:
Yeah, thanks :).

FlorianZ:
Hi @robpieke. I wasn't able to find your signed CLA on our end. Did you submit one by chance? Thank you!!

robpieke:
> Hi @robpieke. I wasn't able to find your signed CLA on our end. Did you submit one by chance? Thank you!!

Hey @FlorianZ ... I thought I'd been added to the SideFX CLA but, in case not, I've just emailed over an individual CLA. Sorry for any confusion/delay!

FlorianZ:
You are right @robpieke! I updated our records to show that you're on the SideFX CLA. Sorry about that.

PR #1804, opened 2022-03-12
Reapply fix from #1465 that was reverted by #1580

marktucker:
Reapply fix from #1465 that was reverted by #1580 (presumably by accident).

### Description of Change(s)

Add a FILE_SHARE_READ flag to the CreateFile call that is used to open a symlink location on Windows. This failure was manifesting as a rare intermittent failure to properly resolve a file path inside a symlinked directory on Windows when running many simultaneous processes reading and writing USD files inside this symlinked directory.


jilliene:
Filed as internal issue #USD-7269

PR #1807, opened 2022-03-18
[usdImaging] fix _GetDirtyBitsForPrimvarChange for PrimvarChangeAdd

nvidia-jomiller:
### Description of Change(s)

It was observed that primvars such as widths for UsdGeomPoints would not
update in hdSt from the usdImagingDelegate. If the widths primvar was
not read from disk it would always display with it's default value. The
change here is that with a PrimvarChangeAdd it will set the incoming
dirty bit instead of just DirtyPrimvar

### Fixes Issue(s)
-


pxr/usdImaging/usdImaging/primAdapter.cpp
```diff
@@ -863,6 +862,7 @@ _GetDirtyBitsForPrimvarChange(
             dirty = HdChangeTracker::DirtyPrimvar;
```
nvidia-jomiller: Instead of moving the `PrimvarChangeAdd` case I did try to use `HdChangeTracker::DirtyComputationPrimvarDesc`. I wasn't sure if this was the bit that the comment might have been eluding to. However, that didn't seem to fix the issue for the `widths` attribute

jilliene:
Filed as internal issue #USD-7285

pxr/usdImaging/usdImaging/primAdapter.cpp
rajabala: Thanks for the PR. We're taking a look.
Re: the comment -- Hydra's change tracker dirty bits doesn't currently have a standalone bit to track changes to the list of primvars. The "descriptor changes" part refers to this. The DirtyPrimvar bit handles double duty of value changes as well as descriptor changes.

rajabala: Your PR made me examine the current code, and I believe we should return the dirty bit passed in (`valueChangeDirtyBit`) for all 4 scenarios (primvar added, primvar removed, primvar value change, primvar descriptor change).
We test against each or all of the primvar dirty bits in `HdChangeTracker::IsPrimvarDirty` and `HdChangeTracker::IsAnyPrimvarDirty` respectively. So, if we were to delete an authored widths primvar, instead of `DirtyPrimvar`, we ought to return `DirtyWidths`.

We can simplify and correct the logic in `primAdapter.cpp` as follows:
- Delete `_GetDirtyBitsForPrimvarChange`
- Remove `PrimvarChange` enum and return `void` in `_ProcessPrimvarChange` or fold this logic in the caller sites.
- Rename `valueChangeDirtyBit` to `primvarDirtyBit` and simply return it in `_ProcessNonPrefixedPrimvarPropertyChange` and `_ProcessPrefixedPrimvarPropertyChange`.

If that makes sense, could you please make these changes to the PR? Thanks!

nvidia-jomiller: Thank you for the notes and checking out the PR! Updated

PR #1810, opened 2022-03-22
Add the AppendDiscoveryResult to NdrRegistry

DuncanHopkinsFoundry:
This method allows the client to explicitly set additional discovery
results that would NOT be found through the plugin system. For example
to support lazily-loaded plugins which cannot be easily discovered in
advance.

### DIFF
--- a/pxr/usd/ndr/registry.h
+++ b/pxr/usd/ndr/registry.h
@@ -94,12 +94,14 @@ public:
     /// This method will not immediately spawn a parse call which will be
     /// deferred until a GetNode*() method is called.
     NDR_API
-    void AppendDiscoveryResult(NdrNodeDiscoveryResult&& discoveryResult);
+    void AddDiscoveryResult(NdrNodeDiscoveryResult&& discoveryResult);
 
     /// Copy version of the method above.
-    /// For performance reasons, one should prefer to use the move version.
+    /// For performance reasons, one should prefer to use the rvalue reference
+    /// form.
+    /// \overload
     NDR_API
-    void AppendDiscoveryResult(const NdrNodeDiscoveryResult& discoveryResult);
+    void AddDiscoveryResult(const NdrNodeDiscoveryResult& discoveryResult);
 
     /// Allows the client to set any additional parser plugins that would
     /// otherwise NOT be found through the plugin system.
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-7288

pxr/usd/ndr/registry.h
```diff
@@ -86,6 +86,21 @@ class NdrRegistry : public TfWeakBase
...
 
+    /// Allows the client to explicitly set additional discovery results that
+    /// would otherwise NOT be found through the plugin system. For example
+    /// to support lazily-loaded plugins which cannot be easily discovered
+    /// in advance.
+    ///
+    /// This method will not immediately spawn a parse call which will be
+    /// deferred until a GetNode*() method is called.
+    NDR_API
+    void AppendDiscoveryResult(NdrNodeDiscoveryResult&& discoveryResult);
```
spiffmon: "Append" implies knowledge that we're storing DR's in a list, which is an implementation detail we'd rather not be locked into.  Can we change this to "Add"?

pxr/usd/ndr/registry.h
```diff
@@ -86,6 +86,21 @@ class NdrRegistry : public TfWeakBase
...
+
+    /// Copy version of the method above.
```
spiffmon, line 99: Can we use the doxygen `\overload` directive, here?

pxr/usd/ndr/registry.h
```diff
@@ -86,6 +86,21 @@ class NdrRegistry : public TfWeakBase
...
+    /// For performance reasons, one should prefer to use the move version.
```
spiffmon: Can we say "prefer to use the rvalue reference form"?

pxr/usd/ndr/registry.h
DuncanHopkinsFoundry: Done

pxr/usd/ndr/registry.h
DuncanHopkinsFoundry: Done

pxr/usd/ndr/registry.h
DuncanHopkinsFoundry: Done

PR #1815, opened 2022-03-25
Improve _GatherDependencies performance by pre-caching path lookup in parallel

csyshing:
### Description of Change(s)

This PR is to improve the performance for `UsdImagingDelegate::SetTime` by pre-caching path dependencies in parallel, in our production shot, the time dropped from 1.2 min to be 10.x second in total.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1813

### DIFF
--- a/pxr/usdImaging/usdImaging/delegate.cpp
+++ b/pxr/usdImaging/usdImaging/delegate.cpp
@@ -907,13 +907,24 @@ UsdImagingDelegate::GetTimeWithOffset(float offset) const
 
 void
 UsdImagingDelegate::_GatherDependencies(SdfPath const& subtree,
-                                        SdfPathVector *affectedCachePaths)
+                                        SdfPathVector& affectedCachePaths)
 {
     HD_TRACE_FUNCTION();
 
-    if (affectedCachePaths == nullptr) {
+    const auto it = _dependenciesCacheMap.find(subtree);
+    if (it != _dependenciesCacheMap.end()) {
+        affectedCachePaths = it->second;
         return;
     }
+
+    _GatherDependenciesCache(subtree, affectedCachePaths);
+}
+
+void
+UsdImagingDelegate::_GatherDependenciesCache(SdfPath const& subtree,
+                                             SdfPathVector& affectedCachePaths)
+{
+    HD_TRACE_FUNCTION();
 
     // Binary search for the first path in the subtree.
     _DependencyMap::const_iterator start =
@@ -948,7 +959,7 @@ UsdImagingDelegate::_GatherDependencies(SdfPath const& subtree,
     // usd dependencies within subtree.
     std::sort(affectedPaths.begin(), affectedPaths.end());
     std::unique_copy(affectedPaths.begin(), affectedPaths.end(),
-            std::back_inserter(*affectedCachePaths));
+            std::back_inserter(affectedCachePaths));
 }
 
 void
@@ -975,6 +986,7 @@ UsdImagingDelegate::ApplyPendingUpdates()
     _coordSysBindingCache.Clear();
     _inheritedPrimvarCache.Clear();
     _pointInstancerIndicesCache.Clear();
+    _dependenciesCacheMap.clear();
 
     UsdImagingDelegate::_Worker worker(this);
     UsdImagingIndexProxy indexProxy(this, &worker);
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-7294

pxr/usdImaging/usdImaging/delegate.h
```diff
@@ -659,7 +659,13 @@ class UsdImagingDelegate : public HdSceneDelegate, public TfWeakBase {
     _DependencyMap _dependencyInfo;
 
     void _GatherDependencies(SdfPath const& subtree,
-                             SdfPathVector *affectedCachePaths);
+                             SdfPathVector& affectedCachePaths);
+
+    typedef TfHashMap<SdfPath, SdfPathVector, SdfPath::Hash> _DependenciesCacheMap;
+    _DependenciesCacheMap _dependenciesCacheMap;
+
+    void _GatherDependenciesCache(SdfPath const &subtree,
```
tcauchois: Can we change this to _CacheDependencies or _PrecacheDependencies?  It makes it more clear what the function is doing.

The pixar code convention is that parameters a function writes out to have to be a pointer, not a reference.  Since the SdfPathVector is owned by the _dependenciesCacheMap, you could change this to SdfPathVector**, or else make the return type "const SdfPathVector&" instead of void...

pxr/usdImaging/usdImaging/delegate.h
```diff
@@ -659,7 +659,13 @@ class UsdImagingDelegate : public HdSceneDelegate, public TfWeakBase {
     _DependencyMap _dependencyInfo;
 
     void _GatherDependencies(SdfPath const& subtree,
-                             SdfPathVector *affectedCachePaths);
```
tcauchois, line 662: As with _GatherDependenciesCache, our code convention is to not have data written out to reference parameters.

pxr/usdImaging/usdImaging/delegate.h
```diff
@@ -659,7 +659,13 @@ class UsdImagingDelegate : public HdSceneDelegate, public TfWeakBase {
...
+                             SdfPathVector& affectedCachePaths);
+
+    typedef TfHashMap<SdfPath, SdfPathVector, SdfPath::Hash> _DependenciesCacheMap;
+    _DependenciesCacheMap _dependenciesCacheMap;
```
tcauchois: I'm worried about this being confused with _dependencyInfo with the current name.  Some other names that could work: _flattenedDependencyCache, _recursiveDependencyCache, etc.

pxr/usdImaging/usdImaging/delegate.cpp
```diff
@@ -994,6 +1006,29 @@ UsdImagingDelegate::ApplyPendingUpdates()
...
+                sortedUsdPathsToResync.emplace(usdPath.GetParentPath());
+            } else if (usdPath.IsAbsoluteRootOrPrimPath()) {
+                sortedUsdPathsToResync.emplace(usdPath);
+            }
+        }
+
+        // Pre-cache dependencies in parallel
+        WorkDispatcher resyncPathsCacheDispatcher;
+        for (const auto& usdPath: sortedUsdPathsToResync) {
+            auto& affectedCachePaths = _dependenciesCacheMap[usdPath];
```
tcauchois: Building "sortedUsdPathsToResync" will be expensive.  You can delete the loop above by running through "usdPathsToResync", doing the GetPrimPath/GetParentPath/etc, and then replacing this line with:

auto pair = _dependenciesCacheMap.insert(usdPath);
if (pair.second) { continue; /* already inserted */ }
auto& affectedCachePaths = pair->first.second;

tcauchois: Note also that you might want to add caching for "usdPathsToUpdate" as well.  The paths to resync are a result of USD composition/type changes, whereas updates reflect property value changes; both call _GatherDependencies.

tcauchois (commented):
Thanks so much for the PR!  I've left a few comments for you, on some ideas for performance improvements and then some discussion of names/function signatures.  Would you mind taking a look?

csyshing:
Thanks for the review, I will update as soon as I can.

csyshing:
Hi @tcauchois , sorry for the delay, I have addressed the feedback in commit b8a78ea .
This PR was created based on v22.03, I could rebase the change and fast-forward to be based on latest `dev` instead, but let me know if you see any other problem, once it's ready, I could update again.
Please have a look and let me know, cheers!

PR #773, opened 2019-02-22
Adds UsdGeomPlane schema

chrisrydalch:
### Description of Change(s)
Extends UsdGeom with a simple Plane schema

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/772

screenshot-window-2019-02-19-222927

Issue #772, opened 2019-02-22
Add a Plane prim to UsdGeom


chrisrydalch:
### Description of Issue
It'd be great for UsdGeom to provide a simple plane primitive.

### Steps to Reproduce
n/a

### System Information (OS, Hardware)
n/a

### Package Versions
n/a

### Build Flags
n/a

jilliene:
Filed as internal issue #USD-5093

onurbingol:
It looks like it has been 3 years but I am still wondering if there are any updates on adding `UsdGeomPlane` to USD library.

spiffmon:
Hi @onurbingol - thanks for the ping!  @chrisrydalch is seeing if he can resuscutate PR #773 so that we get it moving.  We agree the case for having a Plane gprim has grown over time.

crydalch:
@spiffmon and @onurbingol, I should have a new PR ready shortly. Commit #773 needed updates because of the changes to USD since early 2019, and I no longer have access to that GitHub account (which was setup with my Blue Sky email). Thanks!

onurbingol:
Thanks, @crydalch!

onurbingol:
@spiffmon and @crydalch it looks like the plane prim has been merged with the USD library. I believe we could close this issue.

Thank you so much!

sunyab:
The UsdGeomPlane schema in #1819 was merged and released with 22.08. Closing this out, @onurbingol thanks for the heads-up!

PR #1819, opened 2022-03-28
Adds Plane primitive to UsdGeom.

crydalch:
### Description of Change(s)

Extends UsdGeom with a Plane primitive. This PR is an improvement on PR #773, which is out of date due to changes to the USD repo since Feb 2019. The original PR can't be updated, because that fork and user account are no longer accessible.

### Fixes Issue(s)
- #772 

Thanks!


jilliene:
Filed as internal issue #USD-7302

pxr/usd/usdGeom/schema.usda
```diff
@@ -803,6 +803,66 @@ class Cone "Cone" (
     )
 }
 
+class Plane "Plane" (
+    inherits = </Gprim>
+    doc = """Defines a primitive plane, centered at the origin, and is defined by
+    a cardinal axis, width, and length. It is also double-sided.
```
sunyab: Could we reword the last sentence to: "The plane is double-sided by default."

pxr/usd/usdGeom/schema.usda
```diff
@@ -803,6 +803,66 @@ class Cone "Cone" (
...
+    """
+    customData = {
+        dictionary extraPlugInfo = {
+            bool implementsComputeExtent = true
+        }
+    }
+) {
+    uniform bool doubleSided = true (
+        doc = """Planes are double-sided by default, clients may also supported
+        single-sided planes.
```
sunyab: Could we reword this to: "Planes are double-sided by default. Clients may also support single-sided planes."

pxr/usd/usdGeom/schema.usda
```diff
@@ -803,6 +803,66 @@ class Cone "Cone" (
...
+
+        \\sa UsdGeomGprim::GetDoubleSidedAttr()"""
+        )
+    double width = 2.0 (
```
sunyab, line 832: Was there a particular reason to choose 2 as the default length and width? Was it so the extents and other support code could use round numbers?

pxr/usd/usdGeom/plane.cpp
```diff
@@ -0,0 +1,353 @@
...
+_ComputeExtentMax(double width, double length, const TfToken& axis, GfVec3f* max)
+{
+    if (axis == UsdGeomTokens->x) {
+        *max = GfVec3f(0, length, width);
+    } else if (axis == UsdGeomTokens->y) {
+        *max = GfVec3f(width, 0, length);
+    } else if (axis == UsdGeomTokens->z) {
+        *max = GfVec3f(width, length, 0);
+    } else {
+      return false; // invalid axis
```
sunyab: Minor: could we fix the indentation here?

pxr/usd/usdGeom/plane.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+static bool
+_ComputeExtentMax(double width, double length, const TfToken& axis, GfVec3f* max)
+{
+    if (axis == UsdGeomTokens->x) {
+        *max = GfVec3f(0, length, width);
+    } else if (axis == UsdGeomTokens->y) {
+        *max = GfVec3f(width, 0, length);
+    } else if (axis == UsdGeomTokens->z) {
+        *max = GfVec3f(width, length, 0);
```
sunyab: I think the length and width all need to be divided by 2 here. In the computations below, extent gets set to [-max, max], which winds up giving an extent that's twice the width and length.

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.h
```diff
@@ -150,6 +150,33 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+/// Return a topology object for use with all generated "Plane" meshes.  This
+/// is constructed once and is identical for all planes.  The indices refer to
+/// the points array returned by UsdImagingGeneratePlaneMeshPoints().
+USDIMAGING_API
+const PxOsdMeshTopology&
+UsdImagingGetPlaneTopology();
+
+/// Generate an array of points describing a "plane". This is a mesh describing
+/// a double-sided plane aligned to an axis, with a width and length along the
+/// other tcardinal axes.
```
sunyab: Typo: "tcardinal" -> "cardinal"

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.h
```diff
@@ -150,6 +150,33 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+///
+/// Unlike most other primitives in this library, it's not possible to use a
+/// constant set of points and effect width and height adjustments by varying
```
sunyab: Should this be length instead of height?

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.h
```diff
@@ -150,6 +150,33 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+/// the transform matrix.  This function will generate the points with the
+/// requested parameters, and no additional transform is required.  The returned
+/// points are for use with the topology provided by
+/// UsdImagingGetPlaneTopology().
+USDIMAGING_API
+VtVec3fArray
+UsdImagingGeneratePlaneMeshPoints(
+    const double width,
+    const double height,
```
sunyab: Same question re: length here.

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.cpp
```diff
@@ -436,6 +436,59 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+    return planeTopo;
+}
+
+VtVec3fArray
+UsdImagingGeneratePlaneMeshPoints(double width,
+                        double length,
+                        TfToken const & axis)
+{
+    int numPoints = 4;
+    std::vector<GfVec3f> _points(numPoints);
```
sunyab: Style nitpick: we typically don't prepend locals with "_", that's reserved for class members.

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.cpp
```diff
@@ -436,6 +436,59 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+        _points = { GfVec3f( 0.0f,  0.5f * length,-0.5f * width ),
+                    GfVec3f( 0.0f,  0.5f * length, 0.5f * width ),
+                    GfVec3f( 0.0f, -0.5f * length, 0.5f * width ),
+                    GfVec3f( 0.0f, -0.5f * length,-0.5f * width ) };
+    } else if (axis == UsdGeomTokens->y) {
+        _points = { 
+                    GfVec3f( 0.5f * width, 0.0f,-0.5f * length ),
+                    GfVec3f(-0.5f * width, 0.0f,-0.5f * length ),
+                    GfVec3f(-0.5f * width, 0.0f, 0.5f * length ),
+                    GfVec3f( 0.5f * width, 0.0f, 0.5f * length ) };
```
sunyab: I wonder if this should be rotated 180-degrees, i.e. so that the "up" orientation of the corresponding texture coords is consistent for a stage with "Z" upAxis.

pxr/usd/usdGeom/schema.usda
```diff
@@ -803,6 +803,66 @@ class Cone "Cone" (
...
+        doc = """Extent is re-defined on Plane only to provide a fallback
+        value. \\sa UsdGeomGprim::GetExtentAttr()."""
+    )
+    texCoord2d[] primvars:st = [(1.0,1.0),(0.0,1.0),(0.0,0.0),(1.0,0.0)] (
+        doc = """The default st coordinates for texture maps.
+
+        \\sa UsdGeomGprim::GetPrimvarsStAttr()"""
+        interpolation = "varying"
+    )
+}
```
sunyab: We think the "primvars:st" attribute should be removed from this schema. We're uncomfortable hard-coding this name both because everybody has a different "uv" naming scheme; declaring "primvars:" values on base schema types has caused headaches in the past; and this is assuming a certain tessellation.

Our suggestion instead is that, somewhere in the prim comments, we mention that the plane is parameterized as s -> (0, 1) along width axis, t -> (0, 1) along length axis, and usdImaging will generate a primvars:st texcoord with this parameterization. And I guess since it's centered, that would map (0,1) -> (-width * 0.5, width * 0.5) on the appropriate axis.

If folks are ok with the primvar being hardcoded to "st", we're good; if they need to be able to name it, we could add a "token textureCoordinate" attribute in the future.

We could merge this PR without the primvars:st attribute and any of the suggestions above first, then look into implementing the additional behavior afterwards. If you (or anyone) would be interested in implementing the proposed changes in another PR after this landed, we'd gladly take a look. Otherwise it's something we can add to our queue though realistically we might not be able to get to that for a while.

sunyab (changes requested):
Thanks for your patience @chrisrydalch! I've added several notes from the team that we'd love to get your eyes on. We're aiming to get this in for the next release.

pxr/usd/usdGeom/schema.usda
crydalch: That makes sense. Will watch the discussion on usd-interest, but more than happy to remove the texture coordinates for now.

pxr/usd/usdGeom/schema.usda
crydalch: I honestly don't remember if there was a deliberate reason for 2; it's what I had in the original PR. @sunyab , would you prefer that I make it 1.0, so the default length/width is a"unit" plane?

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.h
crydalch: That's embarrassing... yep will replace height with length.

pxr/usd/usdGeom/schema.usda
spiffmon: Sorry I didn't respond to this on our internal review... 2.0 is in keeping with UsdGeomCube, whose `size` (length of each edge) fallback is 2.0, so I think it's a good choice!

crydalch: Ah that must be it. Sorry I missed your comment @spiffmon, I'll correct back to 2.0

pxr/usd/usdGeom/schema.usda
sunyab: Looks like this note was missed?

pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.cpp
```diff
@@ -436,6 +436,59 @@ UsdImagingGenerateCapsuleMeshPoints(
...
+    } else if (axis == UsdGeomTokens->y) {
+        points = { GfVec3f(-1.0f * width, 0.0f, 1.0f * length ),
+                   GfVec3f( 1.0f * width, 0.0f, 1.0f * length ),
+                   GfVec3f( 1.0f * width, 0.0f,-1.0f * length ),
+                   GfVec3f(-1.0f * width, 0.0f,-1.0f * length ) };
+    } else {
+        points = { GfVec3f( 1.0f * width, 1.0f * length, 0.0f ),
+                   GfVec3f(-1.0f * width, 1.0f * length, 0.0f ),
+                   GfVec3f(-1.0f * width,-1.0f * length, 0.0f ),
+                   GfVec3f( 1.0f * width,-1.0f * length, 0.0f ) };
```
sunyab: Not sure if this is correct, it seems like this'll create points for a plane that's 2x the desired width and length?

sunyab (changes requested):
Thanks for addressing the earlier notes! Just left two last notes/questions for you but otherwise everything else looks great.

crydalch:
@sunyab I've corrected the mistaken doubling of length and width, and restored the documentation note. Sorry again about yet another delay on this PR.

sunyab:
@crydalch No problem! I've merged this PR internally and it'll go out with the next push to the dev branch. Thanks for the quick iterations!

sunyab (approved):

PR #1820, opened 2022-03-30
[arch] use "aligned_alloc" not "std::aligned_alloc" on linux

pmolodo:
### Description of Change(s)

std::aligned_alloc is introduced in c++17, but USD only requires c++14

### Fixes Issue(s)

Without this, when building with GCC and using default build settings, will get this error when compiling:

```
/USD/pxr/base/arch/align.cpp: In function 'void* pxrInternal_v0_22__pxrReserved__::ArchAlignedAlloc(size_t, size_t)':
USD/pxr/base/arch/align.cpp:59:17: error: 'aligned_alloc' is not a member of 'std'
     return std::aligned_alloc(alignment, size);
```

### DIFF
--- a/pxr/base/arch/align.cpp
+++ b/pxr/base/arch/align.cpp
@@ -55,7 +55,7 @@ ArchAlignedAlloc(size_t alignment, size_t size)
 #elif defined(ARCH_OS_WINDOWS)
     return _aligned_malloc(size, alignment);
 #else
-    return std::aligned_alloc(alignment, size);
+    return aligned_alloc(alignment, size);
 #endif
 }
 
@@ -68,7 +68,7 @@ ArchAlignedFree(void* ptr)
 #elif defined(ARCH_OS_WINDOWS)
     _aligned_free(ptr);
 #else
-    std::free(ptr);
+    free(ptr);
 #endif
 }
 


pmolodo:
If you want me to add a check like this:

```cpp
#elif __cplusplus >= 201703L && defined(_GLIBCXX_HAVE_ALIGNED_ALLOC)
    return std::aligned_alloc(alignment, size);
#else
    return aligned_alloc(alignment, size);
#endif
```

...I can do that too.  I THINK no-namespace `aligned_alloc` should always be available, though...

pxr/base/arch/align.cpp
```diff
@@ -55,7 +55,7 @@ ArchAlignedAlloc(size_t alignment, size_t size)
 #elif defined(ARCH_OS_WINDOWS)
     return _aligned_malloc(size, alignment);
 #else
-    return std::aligned_alloc(alignment, size);
+    return aligned_alloc(alignment, size);
```
sunyab, line 58: I think we'll want to change the corresponding to call to std::free in ArchAlignedFree to regular free for consistency.

sunyab (changes requested):
Great catch! I think what you have here is fine, but I left one note I think we need to address.

pxr/base/arch/align.cpp
pmolodo: Can do - but just noticed that would make the first `#if` branch and the `#else` branch identical:

```cpp
#if defined(ARCH_OS_DARWIN) || (defined(ARCH_OS_LINUX) && defined(__GLIBCXX__) && !defined(_GLIBCXX_HAVE_ALIGNED_ALLOC))
    free(ptr);
#elif defined(ARCH_OS_WINDOWS)
    _aligned_free(ptr);
#else
    free(ptr);
#endif
```

...so I assume we can just do:

```cpp
#if defined(ARCH_OS_WINDOWS)
    _aligned_free(ptr);
#else
    free(ptr);
#endif
```

pmolodo: Unless I guess you'd rather preserve the `#if` check symmetry with above?

sunyab: I like preserving the symmetry, I think it makes it a bit more readable.

pmolodo: Got it. Done!

jilliene:
Filed as internal issue #USD-7303

PR #1823, opened 2022-04-01
Use SdfPathSet instead of StdPathVector for instancePaths in UsdImagingInstanceAdapter::_Populate

mtavenrath:
### Description of Change(s)
When having a high count of instances the runtime of std::vector::insert at 'random' positions in this function can potentially go up to O(N^2) whereas std::set::insert runs with O(N log(N)). I have profilings of scenes with 1M instances where std::vector::insert takes ~100s of total time (25% load time) whereas times for std::set::insert are mostly noise in the profiling.

Also when erasing an prototype avoid a copy by using std::swap to grab the data from the instancerData structure.

### Fixes Issue(s)
-


jilliene:
Filed as internal issue #USD-7305

Issue #1592, opened 2021-08-10
USD uses deprecated malloc hooks removed in glibc 2.34


musicinmybrain:
### Steps to Reproduce
1. Build on a platform with glibc 2.34 (such as Fedora Rawhide, the development version of Fedora Linux)

### System Information (OS, Hardware)

Fedora Rawhide

### Package Versions

21.08

### Build Flags

Discovered in distribution build. Should not be affected by build flags (except `PXR_MALLOC_LIBRARY`).

Worked around for now by enabling `jemalloc` with `PXR_MALLOC_LIBRARY`.

-----

```
/usr/bin/ld: ../../../../libusd_ms.so.0: undefined reference to `__memalign_hook'
/usr/bin/ld: ../../../../libusd_ms.so.0: undefined reference to `__malloc_hook'
/usr/bin/ld: ../../../../libusd_ms.so.0: undefined reference to `__free_hook'
/usr/bin/ld: ../../../../libusd_ms.so.0: undefined reference to `__realloc_hook'
```

From https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html:

```
* The deprecated memory allocation hooks __malloc_hook, __realloc_hook,
  __memalign_hook and __free_hook are now removed from the API.  Compatibility
  symbols are present to support legacy programs but new applications can no
  longer link to these symbols.  These hooks no longer have any effect on glibc
  functionality.  The malloc debugging DSO libc_malloc_debug.so currently
  supports hooks and can be preloaded to get this functionality back for older
  programs.  However this is a transitional measure and may be removed in a
  future release of the GNU C Library.  Users may port away from these hooks by
  writing and preloading their own malloc interposition library.
```

jilliene:
Filed as internal issue #USD-6832

hradec:
I was able to build USD 20.08 on fedora 35 (with the newer glibc), by patching  `pxr/base/arch/mallocHook.cpp` with this little `sed` line:
```
sed -i.bak  -e 's/.if !defined.ARCH_OS_WINDOWS./#if 0/' -e 's/.if defined.ARCH_COMPILER_GCC.*/#if 0/' -e 's/defined.ARCH_COMPILER_CLANG.//' -e 's/.if defined.ARCH_OS_LINUX./#if 0/' -e 's/.if !defined.ARCH_OS_LINUX./#if 1/' pxr/base/arch/mallocHook.cpp
```
[...]

PR #1830, opened 2022-04-10
Allow building against recent glibc with no malloc hooks (>= 2.34)

charlesfleche:
### Description of Change(s)

In version [2.34 glibc removed the deprecated](https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html) ```__malloc_hook```, ```__realloc_hook```, ```__memalign_hook``` and ```__free_hook```, preventing ```ArchMallocHook``` to link against glibc. This PR:
- Adds a CMake Option ```PXR_ENABLE_MALLOCHOOK_SUPPORT``` (default: ```ON```). When ```OFF```, the C symbol ```ARCH_MALLOC_HOOK``` is not defined and glibc ```__*_hook``` are not used
- Adds a build script option ```--enable-malloc-hook```
- Mention glibc>=2.34 and the CMake option in the build documentation
- Mention in the build doc that malloc overrides is only available for Linux systems

### Fixes Issue(s)
- #1592

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -413,7 +413,7 @@ There are certain optimizations that can be enabled in the build.
 ##### Malloc Library
 
 We've found that USD performs best with allocators such as [Jemalloc](https://github.com/jemalloc/jemalloc).
-In support of this, you can specify your own allocator through ```PXR_MALLOC_LIBRARY```.
+In support of this, for Linux systems you can specify your own allocator through ```PXR_MALLOC_LIBRARY```.
 This variable should be set to a path to a shared object for the allocator. For example,
 
 ```bash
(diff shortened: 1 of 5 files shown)


jilliene:
Filed as internal issue #USD-7317

musicinmybrain:
While working on updating USD to 22.05a in Fedora Linux, I switched to this PR. It works well for me.

It seems like we’ll have glibc 2.33 and older on RHEL and similar for a number of years, but otherwise they will quickly become less and less common. I wonder if it would make sense to *disable* these by default, and let those who are using them and have the necessary old glibc turn them back on.

spundun:
I think this pull request is obsoleted by other similar patch https://github.com/PixarAnimationStudios/USD/commit/8b19842ad49d6c3eb439318fefea19cde3e87781

gitamohr:
Oh yes, I meant to comment here and forgot, thanks! We opted for an approach where we just detect this case in the code and disable support automatically. That way there's no additional build-time configuration.

gitamohr:
Yes! I meant to comment here but I forgot, thanks! We went with an approach where we detect this case in the code and disable support automatically. This way we avoid additional build-time configuration.

PR #1839, opened 2022-04-20
[testing] output failed diffs (image or text) to a common dir

pmolodo:
### Description of Change(s)

Diagnosing issues with failed diffs (particularly image diffs) can be
difficult because the generated output files are in a randomly named
temporary directory, which can be difficult to locate.

With this change, for each failed diff, these files are created:

```
${CMAKE_BINARY_DIR}/Testing/Failed-Diffs/${TEST_NAME}/${filename}.result.${ext}
${CMAKE_BINARY_DIR}/Testing/Failed-Diffs/${TEST_NAME}/${filename}.baseline.${ext}
```


jilliene:
Filed as internal issue #USD-7338

cmake/defaults/Options.cmake
```diff
@@ -24,6 +24,9 @@
 option(PXR_STRICT_BUILD_MODE "Turn on additional warnings. Enforce all warnings as errors." OFF)
 option(PXR_VALIDATE_GENERATED_CODE "Validate script generated code" OFF)
 option(PXR_HEADLESS_TEST_MODE "Disallow GUI based tests, useful for running under headless CI systems." OFF)
+option(PXR_OUTPUT_FAILED_TEST_DIFFS
+    "Copy failed diff/image diff tests into \${CMAKE_BINARY_DIR}/Testing/Failed-Diffs/<PXR_CTEST_RUN_ID>/\${TEST_NAME}. <PXR_CTEST_RUN_ID> is automatically set per ctest-invocation, or read from an environment variable of the same name if set."
+    OFF)
```
tallytalwar: Is there a particular reason we wouldn't want to have this on all the time? We think it make sense to turn this on all the time for simplicity?

cmake/macros/testWrapper.py
```diff
@@ -244,7 +260,20 @@ def _copyTree(src, dest):
...
+                                      ".baseline")
+    resultName = _addFilenameSuffix(os.path.basename(resultFile), ".result")
+    baselineOutpath = os.path.join(failuresDir, baselineName)
+    resultOutpath = os.path.join(failuresDir, resultName)
+
+    if not os.path.isdir(failuresDir):
+        os.makedirs(failuresDir)
+
+    shutil.copy(baselineFile, baselineOutpath)
+    shutil.copy(resultFile, resultOutpath)
```
tallytalwar, line 277: Sunya noticed that it WBN to print out some kind of message here saying where the diff results are being copied to. That way someone looking at the test output can see the exact path they need to go to to look for more info.

cmake/macros/testWrapper.py
```diff
@@ -379,9 +408,14 @@ def _runCommand(raw_command, stdout_redir, stderr_redir, env,
 
     # If desired, diff the provided file(s) (must be generated by the test somehow)
     # with a file of the same name in the baseline directory
+    failuresDir = None
+    if args.failures_dir:
+        failuresDir = args.failures_dir.replace('<PXR_CTEST_RUN_ID>',
```
tallytalwar, line 421: If PXR_CTEST_RUN_ID isn't set in the environment this will generate an invalid path with "<", ">" in it. It was recommended to just drop "<", ">"?

tallytalwar (changes requested):
Thanks @pmolodo for the PR.

This generally looks fine to us. Could you please address few notes I am posting here from our internal review?

We also feel it will make sense to augment BUILDING.md to mention about where outputs from failed tests will be written.

cmake/defaults/Options.cmake
pmolodo: Sure, I'm all for that! The only argument against having it on by default is the it means the build directory might grow over time - each test run with a failed image diff will result in a new directory in `${CMAKE_BINARY_DIR}/Testing/Failed-Diffs`.  However, that's no different (and much less than) the filesystem usage growth that already occurs in the temp directories (it's just more easily visible / discoverable).  But I think the pros outweigh the cons - ideally you won't have failed tests too often, and only more "advanced" users are probably running the tests a lot anyway.

cmake/macros/testWrapper.py
pmolodo: Absolutely, will do!

cmake/defaults/Options.cmake
pmolodo: Removed:
https://github.com/pmolodo/USD/commit/c24838f63735082c515e7dc8b99032397c949c72#diff-9164722da603880b3c8f0383cdf4f24c3fd29963e4da223c9d6e7cb9c4119e4aR780

cmake/macros/testWrapper.py
pmolodo: Added:
https://github.com/pmolodo/USD/commit/c24838f63735082c515e7dc8b99032397c949c72#diff-20f6c1250283d4110c2cacd19f922472ce55350c2e19d29f9491db6b2e792793R279

cmake/macros/testWrapper.py
pmolodo: Done:
https://github.com/pmolodo/USD/commit/c24838f63735082c515e7dc8b99032397c949c72#diff-20f6c1250283d4110c2cacd19f922472ce55350c2e19d29f9491db6b2e792793R422

pmolodo:
Responded to the code review notes, and pushed.  To see just the incremental changes, look here:

https://github.com/pmolodo/USD/commit/53df6477702a00bab492f647b0b5ebfb2ff41857

PR #1842, opened 2022-04-26
[Tf] fix TfDenseHashMap test for windows debug builds

pmolodo:
### Description of Change(s)

the size of vectors changes in debug mode on windows - from `3*sizeof(void*)` to `4*sizeof(void*)`.

This causes an assertion of the sizeof TfDenseHashMap to fail on windows in debug mode.

### Fixes Issue(s)
- TfDenseHashMap test fails when building debug config on windows


jilliene:
Filed as internal issue #USD-7343

PR #1845, opened 2022-04-28
[arch] remove ArchTestCrash(bool) declaration

pmolodo:
### Description of Change(s)

From the looks of things, `ArchTestCrash(bool)` was the original signature; then at some point it was changed to `ArchTestCrash(ArchTestCrashMode)` and moved to `testArchUtil`.  However it seems that the old declaration was left dangling (though implementation was removed).

### Fixes Issue(s)
- ArchTestCrash(bool) declared but not implemented


jilliene:
Filed as internal issue #USD-7346

PR #1848, opened 2022-04-29
Allow building usdBakeMtlx on windows.

marktucker:
### Description of Change(s)
At the moment building the usdBakeMtlx binary is disabled on Windows. I made a few changes to the CMake file and added an api.h and marked a couple of functions to be exported from the library build in support of the python binary. With these changes I was able to build the library and binary. But I haven't had the chance to really test it, and I build USD in a custom environment, not using build_usd.py. And I was building against Mtls 1.38.4 using #1792.

So I'd really love some confirmation from anyone else building on Windows that this change actually works in isolation...

### DIFF
--- a/pxr/usdImaging/bin/usdBakeMtlx/CMakeLists.txt
+++ b/pxr/usdImaging/bin/usdBakeMtlx/CMakeLists.txt
@@ -1,8 +1,3 @@
-if (WIN32)
-    message(WARNING "UsdBakeMtlx is not supported on Windows")
-    return()
-endif()
-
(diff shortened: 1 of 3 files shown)


jilliene:
Filed as internal issue #USD-7357

sunyab:
Ack, sorry @marktucker -- we lost track of this PR and wound up making pretty much the exact same changes independently. Those changes should land with the next push to dev. We were able to successfully test this on a Windows machine in a "standard" build environment.

I'm going to close this out, sorry again and thanks!

PR #1849, opened 2022-05-02
Register velocities and acclerations on point instancer primitives as primvars

marktucker:
### Description of Change(s)

Register velocities and acclerations on point instancer primitives as hydra-accessible primvars.
I have not actually been able to test this change in isolation, other than confirming that it builds (since none of the standard USD render delegates have any interest in these attributes).

This is a trimmed down version of #1734 (which in turn is a trimmed down version of #1731).


jilliene:
Filed as internal issue #USD-7358

PR #1850, opened 2022-05-04
[arch] make testArchStackTrace work on windows debug even when relocated

pmolodo:
### Description of Change(s)

On windows, in debug mode, we need to copy testArchStaceTrace.pdb into the testing folder, so the test can find it (and print out the right stack), even if the build or test folder has been moved (ie, when the baked-in path in executable, where it also looks for the pdb, doesn't work)

### Fixes Issue(s)
- testArchStackTrace failed on windows debug if build folder was relocated


jilliene:
Filed as internal issue #USD-7359

PR #1856, opened 2022-05-10
Return False when image diff file not found

wrma6:
### Description of Change(s)
In testWrapper.py _imageDiff, the test will pass if glob does not find the file. This change would make the test return False (fail) if this happens, so that the problem is not silently passed over. Also, in the _diff function in the same file, there is an unnecessary second call to glob, which this change would remove.

### Fixes Issue(s)
- _imageDiff function in testWrapper.py returns True (passes) when the file cannot be found
- _diff function in testWrapper.py makes an unncessary repeat call to glob.glob
- two tests that reference the wrong images have been fixed
- one test (testUsdImagingGLInstancing_nestedInstance) has been disabled, the test itself is failing

### DIFF
--- a/cmake/macros/testWrapper.py
+++ b/cmake/macros/testWrapper.py
@@ -184,7 +184,7 @@ def _diff(fileName, baselineDir, verbose, failuresDir=None):
             "Error: could not files matching {0} to diff".format(fileName))
         return False
 
-    for fileToDiff in glob.glob(fileName):
+    for fileToDiff in filesToDiff:
         baselineFile = _resolvePath(baselineDir, fileToDiff)
         cmd = [diff, baselineFile, fileToDiff]
         if verbose:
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-7368

pmolodo:
Note - we've commented out one test, `testUsdImagingGLInstancing_nestedInstance`.  This actually isn't changing behavior, though, as the tests was effectively already being skipped, due to it looking for the wrong image to compare against (`testUsdImagingGLInstancing_instance3.png`, instead of `testUsdImagingGLInstancing_nestedInstance.png`.)  Prior to this change, that would just result in the test being silently skipped.

However, once the test was fixed to compare the correct file, it exposed what looks like a true test failure.  As a result, we've disabled the test for now.  If you'd like, we can make a separate issue to fix + re-enable this test.

Baseline:
testUsdImagingGLInstancing_nestedInstance baseline

Rendered:
testUsdImagingGLInstancing_nestedInstance rendered

wrma6:
It seems that a change to `testUsdImagingGLInstancing_nestedInstance` has been merged while this PR was open, so I have removed the changes in this PR for that test.

pmolodo:
Huzzah for the test getting fixed!  Can we get this merged so future test failures like that won't slip through unnoticed?

pmolodo:
Just checking in on this - it's not particularly urgent, but it's also a pretty simple fix, which helps ensure future correctness of tests.  As shown above, there was already one test failure which slipped through, and perhaps would have been caught + fixed sooner if this change had been in place...

sunyab:
Hey @pmolodo , sorry for the delay on this one. I had this staged but went on leave before I could commit it. This won't make it in for the upcoming 22.11 release but will be in the next one. Thanks!

pmolodo:
No worries - just didn't want it to drop off the back end. Thanks!

PR #1859, opened 2022-05-12
[HdSt] Remove excess parameters, change vega fix to member, support caps in pso

creijon:
### Description of Change(s)

- In `HdSt_PipelineDrawBatch` there were a couple of private functions which took the dispatch buffer as a param even though it is a local member of the class.
- The Vega fix in `HgiMetalGraphicsCmds` is neater if it is a member and can access the `_primitiveTypeChanged` directly.
- Enablement: In `HgiMetalGraphicsPipeline`, pass the hgi interface down to the internal functions so they can check the caps bits if necessary.  Not used now but useful in the future.

### Fixes Issue(s)
- Code cleaning.


jilliene:
Filed as internal issue #USD-7371

PR #1860, opened 2022-05-12
APPLE: [HdSt] CPU cull optim, only rebuild batch if visibility changes

creijon:
### Description of Change(s)

- Check whether the state of the visible flag has changed before rebuilding the draw batches.

### Fixes Issue(s)
- Performance for CPU culling


jilliene:
Filed as internal issue #USD-7372

PR #1861, opened 2022-05-12
Avoid python3 warning about not properly closing a file.

marktucker:
### Description of Change(s)

When running usdview in a python3 environment, a warning like the following could be output:
```
.....\python39\lib\site-packages\pxr\Usdviewq\appController.py:419: ResourceWarning: unclosed file <_io.TextIOWrapper name='...../python39/lib/site-packages/pxr/Usdviewq/usdviewstyle.q
ss' mode='r' encoding='cp65001'>
  self._setStyleSheetUsingState()
```

This change uses `with open as` to fix this problem.


jilliene:
Filed as internal issue #USD-7373

Issue #1862, opened 2022-05-13
Storm shader compile error with a primvar reader connected to opacity


marktucker:
### Description of Issue
Usdview rendering with Storm outputs a shader compiler error when loading an asset with a USD Preview Surface material with a Primvar Reader prim connected to the opacity input of the material. This seems to be a regression as the same material works in USD 21.11, but produces an error in USD 22.05.

### Steps to Reproduce
1. Open the attached asset in usdview 22.05: 
[torus.zip](https://github.com/PixarAnimationStudios/USD/files/8683743/torus.zip)
2. Error is printed to terminal, and the material doesn't display properly (the torus should be green). The error message looks like:
```
Error in 'pxrInternal_v0_22__pxrReserved__::HdSt_DrawBatch::_GetDrawingP
rogram' at line 279 in file F:\dev_tools\src\usd\usd-22.05\USD-py3.9\pxr\imaging
\hdSt\drawBatch.cpp : 'Failed to compile shader for prim /torus1/mesh_0.'
```

### Package Versions
22.05

jilliene:
Filed as internal issue #USD-7374

PR #1863, opened 2022-05-13
Fix Storm code generation from USD PreviewSurface for GLSL bindings to primvars

marktucker:
### Description of Change(s)
Fix code generation for GLSL bindings to primvars. This type of binding was missing the definition of HdGetScalar_name to call HdGet_name.

I'm fairly out of my depth with this change, so I'm not sure if this is the right approach. But in the specific case of opacity reported in issue #1862, this change fixes the code generation and the material works as expected.

### Fixes Issue(s)
- #1862 


jilliene:
Filed as internal issue #USD-7375

davidgyu:
Thanks, Mark! This fix looks good to me, and thank you for the test case in #1862 We'll make sure to add a case for this to our test suite and get the fix merged.

PR #1866, opened 2022-05-17
usdImagingGL idiff tests should be able to detect the difference between visibly distinct reference images 

wrma6:
### Description of Change(s)
With the current parameters used for the image diff tests in usdImagingGL, the reference images cannot be distinguished from each other: idiff passes many image pairs as "similar enough" even though they are (visibly) distinctly different. Specifically, the Fail parameter of 1 (the most lenient value possible) usually passes every image pair, so tests using that rely solely on the additional Perceptual test. Using a Python script, all pairs of images referenced by each test were compared using idiff to determine reasonable Fail and FailPercent parameters. The proposed change would ensure that for the parameters and reference images of each test, the parameters can at least differentiate between any pair of reference images that differ.

### Fixes Issue(s)
- 22 tests have their Fail parameter changed to .05 and their FailPercent parameter changed to .03. This means to fail a test between two given images, if more than 0.03% of their pixels differ by more than 0.05 on the 1-255 colour scale (colour difference of about 12.5).

### DIFF
--- a/pxr/usdImaging/usdImagingGL/CMakeLists.txt
+++ b/pxr/usdImaging/usdImagingGL/CMakeLists.txt
@@ -1100,8 +1100,8 @@ pxr_register_test(testUsdImagingGLComplexityOsd3
         testUsdImagingGLComplexity_1.png
         testUsdImagingGLComplexity_1.1.png
         testUsdImagingGLComplexity_1.2.png
-    FAIL 1
-    FAIL_PERCENT 1
+    FAIL .05
+    FAIL_PERCENT .03
     PERCEPTUAL
     EXPECTED_RETURN_CODE 0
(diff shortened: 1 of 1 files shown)


pmolodo:
Noticed that this covers similar territory as this commit:
b5aa21242f6a18e5f4b16ba8e119a9cbe3233beb
...but seems to fix different tests.  Should the error threshholds we're altering here be updated to match the more stringent ones in that commit?

jilliene:
Filed as internal issue #USD-7393

pmolodo:
Just poking to see about getting this merged...

spiffmon:
@pmolodo , this is in our queue, but our recent focus for 22.08, oss-imaging-wise, has been native silicon build support.

pmolodo:
Got it, thanks for the info!

Issue #1870, opened 2022-05-20
Windows .usd file creation on network drives can fail when preceded by specific file access checks


marktucker:
### Description of Issue
Windows file existence checking has the possibility to leave an error set, that can be erroneously flagged as an error in a subsequent file existence/permission check. For some reaosn this seems to only affet network drives, and only very specific sequences of file checks triggers this problem.

### Steps to Reproduce
1. On a Windows machine, where the Z: drive is a network mount, and z:/bugs exists as a directory, run the following code:
```
from pxr import Sdf
l1 = Sdf.Layer.CreateAnonymous()
l2 = Sdf.Layer.Find("z:/bugs/12345/foo/bar/thing.usd")
l1.Export("z:/bugs/12345/foo/bar/thing.usd")
```
2. The Export call will fail with the following exception:
Traceback (most recent call last):
```
  File "<console>", line 1, in <module>
pxr.Tf.ErrorException: 
        Error in 'pxrInternal_v0_22__pxrReserved__::TfSafeOutputFile::Replace' at line 138 
in file F:\dev_tools\src\usd\usd-22.05\USD-py3.9\pxr\base\tf\safeOutputFile.cpp : 'U
nable to determine the real path for 'z:/bugs/121495/foo/bar/thing.usd': The system 
cannot find the path specified.

'Error in 'pxrInternal_v0_22__pxrReserved__::Usd_CrateFile::CrateFile::StartPacking'
 at line 2506 in file F:\dev_tools\src\usd\usd-22.05\USD-py3.9\pxr\usd\usd\crateFile
.cpp : 'Unable to open z:/bugs/121495/foo/bar/thing.usd for write'
```

### System Information (OS, Hardware)
Windows only.

jilliene:
Filed as internal issue #USD-7395

sunyab:
This fix was released in v22.08. Closing, thanks!

PR #1871, opened 2022-05-20
Fix Windows file existence check to prevent errors when creating new .usd files

marktucker:
### Description of Change(s)
When testing for the existance of a file on Windows, we want to clear the
error state if the system complains with either ERROR_FILE_NOT_FOUND _or_
ERROR_PATH_NOT_FOUND. These two errors together are equivalent to ENOENT
on Linux, which is the equivalent check that happens on that OS when
calling TfPathExists.

### Fixes Issue(s)
- #1870 

### DIFF
--- a/pxr/base/tf/fileUtils.cpp
+++ b/pxr/base/tf/fileUtils.cpp
@@ -88,7 +88,9 @@ Tf_HasAttribute(
     const DWORD attribs =
         GetFileAttributesW(ArchWindowsUtf8ToUtf16(path).c_str());
     if (attribs == INVALID_FILE_ATTRIBUTES) {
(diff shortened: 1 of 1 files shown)


asluk:
Looks like this would fix https://github.com/PixarAnimationStudios/USD/issues/1759 as well? Thanks @marktucker !

marktucker:
Yes! Just ran the code from that issue (with and without this change applied), and it is also fixed by this PR.

nvidia-jomiller:
Testing things out with network mapped drives as in https://github.com/PixarAnimationStudios/USD/issues/1759 works great! Thanks @marktucker

jilliene:
Filed as internal issue #USD-7396

PR #1872, opened 2022-05-20
Crash when reporting invalid instance

seando-adsk:
### Description of Change(s)
When an invalid prim spec is used for an instance on a stage that is no longer
valid, the prototype-for-instance pointer would be null yet would be traversed.
So instead of a an exception being thrown, we would get a crash. The fix checks
the pointer before using it.
### Fixes Issue(s)
Crash when reporting invalid instance


jilliene:
Filed as internal issue #USD-7397

sunyab:
Revisiting some older PRs -- I was able to reproduce the issue here and confirmed that this PR fixes it. However, I discovered that the problem extends beyond just instances: instance proxies, prototypes, and prims in prototypes from expired stages are also susceptible to a similar bug. I wound up implementing a slightly larger change that fixes all of these issues together. 

I'm going to close this PR out in favor of the more comprehensive fix. This fix is checked in locally but won't land on GitHub until after the upcoming USD release. Thanks for filing the report and PR!

PR #1875, opened 2022-05-21
Apple Silicon Support

creijon:
### Description of Change(s)

Builds on the work in https://github.com/PixarAnimationStudios/USD/pull/1898 which adds PySide6 support.

Enables native Apple Silicon support for Hydra-Storm.  This includes applications which usd PySide bindings such as usdview as well as applications that link in USD directly.

To build usdview on Apple Silicon it is necessary to install PySide6, the changes for this have now been isolated to https://github.com/PixarAnimationStudios/USD/pull/1898

Embree upgraded to 3.13.2 to support Apple Silicon

Thanks to the rest of the team at Apple for this work.

Tested On:
Built with the following options: `--ptex --openvdb --openimageio --opencolorio --alembic --draco --hdf5 --embree --tests`
- macOS Catalina, Big Sur and Monterey on Intel architecture.  With and without PySide2 installed.
- macOS Big Sur and Monterey on M1 and M1Pro architectures.  With and without PySide2 installed next to PySide6.
- Ubuntu 21.04 with and without PySide2 installed next to PySide6.

Known Limitations:
- No support for cross compilation between Intel and Apple Silicon.
- No support for universal binaries.

### Fixes Issue(s)
- Native Apple Silicon support

### DIFF
--- a/pxr/usdImaging/usdviewq/stageView.py
+++ b/pxr/usdImaging/usdviewq/stageView.py
@@ -34,7 +34,7 @@ from math import tan, floor, ceil, radians as rad, isinf
 import os, sys
 from time import time
 
-from .qt import QtCore, QtGui, QtWidgets, QtOpenGL
+from .qt import QtCore, QtGui, QtWidgets, QtOpenGL, QGLWidget, QGLFormat
 
 from pxr import Tf
 from pxr import Gf
@@ -628,9 +628,7 @@ class HUD():
 
         for name in self._groups:
             group = self._groups[name]
-
-            tex = qglwidget.bindTexture(group.qimage, GL.GL_TEXTURE_2D, GL.GL_RGBA,
-                                        QtOpenGL.QGLContext.NoBindOption)
+            tex = qglwidget.BindTexture(group.qimage)
             GL.glUniform4f(self._glslProgram.uniformLocations["rect"],
                            2*group.x/width - 1,
                            1 - 2*group.y/height - 2*group.h/height,
@@ -638,10 +636,8 @@ class HUD():
                            2*group.h/height)
             GL.glUniform1i(self._glslProgram.uniformLocations["tex"], 0)
             GL.glActiveTexture(GL.GL_TEXTURE0)
-            GL.glBindTexture(GL.GL_TEXTURE_2D, tex)
             GL.glDrawArrays(GL.GL_TRIANGLE_STRIP, 0, 4)
-
-            GL.glDeleteTextures(tex)
+            qglwidget.ReleaseTexture(tex)
 
         GL.glBindTexture(GL.GL_TEXTURE_2D, 0)
         GL.glBindBuffer(GL.GL_ARRAY_BUFFER, 0)
@@ -675,7 +671,7 @@ def _ComputeCameraFraming(viewport, renderBufferSize):
 
     return CameraUtil.Framing(displayWindow, dataWindow)
 
-class StageView(QtOpenGL.QGLWidget):
+class StageView(QGLWidget):
     '''
     QGLWidget that displays a USD Stage.  A StageView requires a dataModel
     object from which it will query state it needs to properly image its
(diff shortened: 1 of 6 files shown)


pxr/usdImaging/usdviewq/stageView.py
```diff
@@ -628,20 +639,18 @@ def draw(self, qglwidget):
...
+                tex.bind()
+                self.drawWidgetTexture(width, height, group)
+                tex.release()
+                tex.destroy()
+            else:
+                tex = qglwidget.bindTexture(group.qimage, GL.GL_TEXTURE_2D, GL.GL_RGBA,
+                                            QtOpenGL.QGLContext.NoBindOption)
+                GL.glBindTexture(GL.GL_TEXTURE_2D, tex)
+                self.drawWidgetTexture(width, height, group)
+                GL.glDeleteTextures(tex)
```
spiffmon: @jonny-apple , we'd really like to keep all Qt/pyside version logic in qt.py.  I think this one could be quite nicely wrapped up in a BindTexture context object implemented appropriately for the two different versions, maybe injected into QtOpenGL ?

pxr/usdImaging/usdviewq/stageView.py
```diff
@@ -838,14 +847,18 @@ def rendererAovName(self):
...
+        if PySideModule == 'PySide6':
+            if msaa == "1":
+                glFormat.setSamples(4)
+            super(StageView, self).__init__()
+            self.setFormat(glFormat)
+        else:
+            if msaa == "1":
+                glFormat.setSampleBuffers(True)
+                glFormat.setSamples(4)
+            super(StageView, self).__init__(glFormat, parent)
```
spiffmon: Seems like a context object would also allow you to vary the timing of the call to super's `__init__`, for the two impl's here?  Naming is not as obvious to me, though...

pxr/usdImaging/usdviewq/stageView.py
```diff
@@ -1488,12 +1501,18 @@ def updateGL(self):
         playback.  The client driving playback at a particular rate should
         instead call updateForPlayback() to image the next frame."""
         if not self._dataModel.playing:
-            super(StageView, self).updateGL()
+            if PySideModule == 'PySide6':
+                super(StageView, self).update()
```
spiffmon: Just inject `updateGL()` into QGLWidget ?  Not entirely satisfying as it's being deprecated, but I'm not sure atm whether there's a good reason we're calling `updateGL()` here rather than straight-up `update()`

pxr/usdImaging/usdviewq/stageView.py
```diff
@@ -2101,9 +2120,14 @@ def mousePressEvent(self, event):
...
-                self.switchToFreeCamera()
-                self._cameraMode = "truck"
+            if PySideModule == 'PySide6':
+                if event.button() == QtCore.Qt.MiddleButton:
+                    self.switchToFreeCamera()
+                    self._cameraMode = "truck"
+            else:
+                if event.button() == QtCore.Qt.MidButton:
+                    self.switchToFreeCamera()
+                    self._cameraMode = "truck"
```
spiffmon: Seems we are already using `MiddleButton` in frameSlider.py, so it is available in pyside2, so you can just "upgrade" here unconditionally!

creijon: Cheers

pxr/usdImaging/usdviewq/stageView.py
creijon: Good point Spiff, I'll move the logic back to qt.py.

creijon: I've injected a BindTexture and ReleaseTexture to the QGLWidget class.

pxr/usdImaging/usdviewq/stageView.py
creijon: Thanks, done.

pxr/usdImaging/usdviewq/stageView.py
creijon: Moved this to an injected function `InitQtWidget`

jilliene:
Filed as internal issue #USD-7381

PR #1876, opened 2022-05-23
[HGI] Cleanup of vertex bindings, split off HgiMetal step functions

creijon:
### Description of Change(s)

General code cleanup around the HgiMetal implementation of graphicsCmds.  The vertex buffer bindings were being passed around as two parallel arrays of buffer handles and byte offsets with a 'firstBinding' parameter too.  Moved this to a new struct `HgiVertexBufferBinding`.

This allows the `HdStPipelineDrawBatch` to be cleaned up too.

There was a lot of logic around the calculation and setting of the vertex buffer step functions in `HgiMetalGraphicsCmd` this has been split out to a separate class `HgiMetalStepFunctions` which can be used by other parts of the engine.

### Fixes Issue(s)
- Code maintainability and clarity

### DIFF
--- a/pxr/imaging/hdSt/pipelineDrawBatch.cpp
+++ b/pxr/imaging/hdSt/pipelineDrawBatch.cpp
@@ -1137,51 +1137,42 @@ _GetVertexBuffersForDrawing(_BindingState const & state)
 }
 
 uint32_t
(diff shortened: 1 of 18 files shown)


jilliene:
Filed as internal issue #USD-7382

PR #1883, opened 2022-05-24
Optimize Usd_LinearInterpolator. 

mtavenrath:
### Description of Change(s)
Swapping the lowerValue with the _result and calling _result.data() afterwards results in a memmove in the data() function call. Later on the Usd_Lerp loop overwrites the all values which have been copied before with the memmove call making the memmove an expensive nop.

Also using operator[] on each loop iteration on upperValue results in a function call to uperValue.data() on each loop iteration which can easily be avoided by fetching the data pointer once before.

### Fixes Issue(s)
Improves performance of Usd_LinearInterpolator.
-

### DIFF
--- a/pxr/usd/usd/interpolators.h
+++ b/pxr/usd/usd/interpolators.h
@@ -291,30 +291,35 @@ private:
             upperValue = lowerValue;
         }
 
-        _result->swap(lowerValue);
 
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #USD-7389

PR #1888, opened 2022-06-02
Enable Storm for Apple Intel GPUs

slingthor:
Update: Covers primitive ID as well and is everything needed for Intel GPU enablement from only being barycentric and base vertex offset fixes.

Enable Storm for Apple Intel GPUs

### Description of Change(s)
- Use post tessellation vertex shader on Intel in most cases to support hardware barycentric coordinates on Intel Apple machines using metal ( needs hardware barycentric as no geometry shader is presented ). Uses the tessellation coordinate as a barycentric coordinate.
- Use HdStVBOSimpleMemoryManager instead of HdStVBOMemoryManager for nonUniformAggregationStrategy and nonUniformImmutableAggregationStrategy in the resourceRegistry on Apple intel machines to work around a Metal driver bug on Apple Intel Macs which prevents us from emulating base vertex via setVertexoffset. This change has a performance overhead, but correctness can not be achived through simple means otherwise. This change cause the baseVertex to always be 0. This bug has been fixed internally at Apple and is expected to be fixed in MacOS 13.0
- Pass patch_id between stages to work around a driver issue where the hardware primitive id is wrong if patch_id is not passed as an interstage.
- Fix the primitive id calculation for post tessellation vertex shader triquads ( division by two )

### Fixes Issue(s)
- Missing barycentric coordinates on Apple Intel Macs
- Incorrect base vertex workaround for patch drawing on Apple Intel Macs

### DIFF
--- a/pxr/imaging/hdSt/codeGen.cpp
+++ b/pxr/imaging/hdSt/codeGen.cpp
@@ -1392,7 +1392,8 @@ HdSt_CodeGen::Compile(HdStResourceRegistry*const registry)
         capabilities->IsSet(HgiDeviceCapabilitiesBitsShaderDoublePrecision);
     const bool minusOneToOneDepth =
         capabilities->IsSet(HgiDeviceCapabilitiesBitsDepthRangeMinusOnetoOne);
-
+    const bool useTessellationBarycentric =
+        capabilities->IsSet(HgiDeviceCapabilitiesBitsTessellationBarycentric);
     bool const useHgiResourceGeneration =
         IsEnabledHgiResourceGeneration(capabilities);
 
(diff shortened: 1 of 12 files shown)


jilliene:
Filed as internal issue #USD-7408

jonny-apple:
Also enables support for Intel-GPU based Macs!

jonny-apple:
Found these issues in testing:

The function `HgiMetal::IsBackendSupported` should have the following conditional:
```
  return 
    _capabilities->IsSet(HgiDeviceCapabilitiesBitsBuiltinBarycentrics) ||
    _capabilities->IsSet(HgiDeviceCapabilitiesBitsTessellationBarycentric);
```

The check for `hasIntel` in `HgiMetalCapabilities::HgiMetalCapabilities` doesn't return a string on a Intel HD 630.  Should test against `[device isLowPower]` instead.

PR #1894, opened 2022-06-07
Add color4 type to Sdr

seando-adsk:
### Description of Change(s)
As discussed in https://groups.google.com/g/usd-interest/c/G8BSo8jzfyY

### Fixes Issue(s)
This allows differentiating between color4 and vector4 input types at the Sdr
level when inspecting MaterialX shader nodes.


jilliene:
Filed as internal issue #USD-7413

spiffmon:
@seando-adsk and @JGamache-autodesk , we're reviewing this, and I have a followup question, which is whether there is need to update any HdMtlx and/or Storm code to handle color4 for passing data back into MaterialX codegen?

That could be followup work, for sure, along with untangling the OSL component.

JGamache-autodesk:
Hi @spiffmon, the reconversion back to MaterialX was working before (and after) this fix. The hdMtlx code [uses the type found in the MaterialX NodeDef](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/imaging/hdMtlx/hdMtlx.cpp#L231) and since the C++ representation of both float4 and color4 is GfVec4f, the values were converted correctly. I did not see any Hydra/Storm issues before, and none were introduced by this fix.

spiffmon:
Thank you, @jgamache-autodesk - great to hear!

--spiff

PR #1895, opened 2022-06-07
Parse MaterialX metadata into SdrProperty 

### LINKED ISSUE #1874: MaterialX metadata is not parsed into Sdr metadata
### Description of Issue

We would like to offer rich parameter editing in our DCC, which requires some metadata information to be present in Sdr.

Let's start with Metadata that has a direct Sdr counterpart:

1. MaterialX inputs of type "string" that have "enum" metadata should submit the options via NdrOptionVec.
2. "uiname" metadata maps directly to SdrPropertyMetadata->Label
3. "uifolder" metadata maps directly to SdrPropertyMetadata->Page

This should allow us to split the standard_surface parameters into their proper pages, offer a dropdown for enum-like string inputs, and use nicer labels on these controls.

This leaves a long list of metadata with no Sdr counterpart: `uimin`, `uimax`, `uistep`, `uisoftmin`, `uisoftmax`, `unit`, `unittype`, `defaultgeomprop`, `enumvalues`, `colorspace`. We would like to transfer them as-is into the SdrProperty metadata dictionary so DCCs that can make use of them will find them without having to reparse MaterialX definitions.

Would you be OK with that?

### Steps to Reproduce
1. Build with MaterialX enabled
2. Run this script:
```
>>> n = Sdr.Registry().GetShaderNodeByIdentifier("ND_image_color3")
>>> n.GetInput("uaddressmode").GetOptions()
[]  # Expected: [('constant', ''), ('clamp', ''), ('periodic', ''), ('mirror', '')]
>>> n = Sdr.Registry().GetShaderNodeByIdentifier("ND_standard_surface_surfaceshader")
>>> n.GetInput("base_color").GetLabel()
''  # Expected: 'Base Color'
>>> n.GetInput("base_color").GetPage()
'' # Expected: 'Base'
>>> n.GetInput("specular_IOR").GetMetadata()
{'widget': 'default'} # Expected: {'widget': 'default', 'uimin': '0.0', 'uisoftmax': '3.0', 'label': 'Index of Refraction', 'page': 'Specular'}
```

### Build Flags
Enable MaterialX.

seando-adsk:
### Description of Change(s)
Fixes #1874 

- "enum" and "enumvalues" are parsed as options
- "doc" is parsed as help
- "uiname" is parsed as label
- "uifolder" is parsed as page
- all other MaterialX metadata is added as hints


jilliene:
Filed as internal issue #USD-7414

pxr/usd/usdMtlx/parser.cpp
```diff
@@ -207,14 +296,13 @@ ShaderBuilder::AddProperty(
...
     // Record builtin primvar references for this node's inputs.
-    static const std::string defaultgeompropName("defaultgeomprop");
     if (!isOutput && primvars != nullptr) {
 
         // If an input has "defaultgeomprop", that means it reads from the
         // primvar specified unless connected. We mark these in Sdr as
         // always-required primvars; note that this means we might overestimate
         // which primvars are referenced in a material.
-        const auto& defaultgeomprop = element->getAttribute(defaultgeompropName);
+        const auto& defaultgeomprop = element->getAttribute(_tokens->defaultgeomprop.GetString());
```
spiffmon: TfToken should auto-convert to std::string

pxr/usd/usdMtlx/parser.cpp
```diff
@@ -239,6 +327,45 @@ ShaderBuilder::AddProperty(
         metadata[SdrPropertyMetadata->ImplementationName] = j->second;
     }
 
+    if (!isOutput) {
+        ParseMetadata(metadata, SdrPropertyMetadata->Label, element, "uiname");
+        ParseMetadata(metadata, SdrPropertyMetadata->Help, element, "doc");
+        ParseMetadata(metadata, SdrPropertyMetadata->Page, element, "uifolder");
```
spiffmon: Any reason why these three literals, plus "enum" and "enumvalues" above aren't also added to the private tokens?

pxr/usd/usdMtlx/parser.cpp
```diff
@@ -239,6 +327,45 @@ ShaderBuilder::AddProperty(
...
+        ParseMetadata(metadata, _tokens->uisoftmin, element);
+        ParseMetadata(metadata, _tokens->uisoftmax, element);
+        ParseMetadata(metadata, _tokens->uistep, element);
+        ParseMetadata(metadata, _tokens->unit, element);
+        ParseMetadata(metadata, _tokens->unittype, element);
+        ParseMetadata(metadata, _tokens->defaultgeomprop, element);
+
+        if (!metadata.count(SdrPropertyMetadata->Help) &&
+             metadata.count(_tokens->unit)) {
+            // The unit can be helpful if there is no documentation.
```
spiffmon, line 357: Not sure what `unittype` is, but would it also be useful?  Seems like `defaultgeomprop` would also be handy to provide?

spiffmon:
Just a couple questions/sugs, but looks good overall, @JGamache-autodesk .  Let us know if you plan on updating the PR?

pxr/usd/usdMtlx/parser.cpp
JGamache-autodesk: Yes, but not necessarily a good one. These strings are only used as std::string on the MaterialX side of things. Would be best to have them as tokens as well since that prevents an unnecessary malloc/strcpy.

pxr/usd/usdMtlx/parser.cpp
JGamache-autodesk: I see only two places in MaterialX where unittype is defined. There are a few angles defined as being "unittype = radian" and one thin_film thickness being tagged with "unittype = nanometer". I suspect ST coordinates could be in meters in some future, but we are not there yet.

spiffmon: Oh hmmm...   these will want to be **very prominently** documented, as they diverge from USD norms.  All angles are degrees in USD schemas, and linear units are generally relative to stage `metersPerUnit` .  So yeah, I'd double-down on wanting to fold unittype into the description, if the nodes that use them don't already provide documentation that mentions them!

spiffmon:
@JGamache-autodesk , we might be able to get this into 23.02 if you can update the PR this week/early next week?

JGamache-autodesk:
Hi @spiffmon, I have a patch in my local repo. I will try to get it merged into the PR early next week.

JGamache-autodesk (approved):

pxr/usd/usdMtlx/parser.cpp
```diff
@@ -48,6 +48,26 @@ TF_DEFINE_PRIVATE_TOKENS(
 
     ((discoveryType, "mtlx"))
     ((sourceType, ""))
+
+    (colorspace)
+    (defaultgeomprop)
+    (defaultinput)
+    (doc)
+    (enum)
```
spiffmon: @JGamache-autodesk , did this compile for you?  `enum` is not a legal token/identifier name in C++.  YOu can do something like:
```
((enum_, "enum"))
```

Thanks!

seando-adsk: @spiffmon I added the patch (from Jerry).

spiffmon: Looks great - thanks, @seando-adsk !

PR #1898, opened 2022-06-10
Pyside6 Support for usdView

creijon:
### Description of Change(s)

Allow usdView to run with PySide6, which is used in preference to PySide2 if both are present.

### Fixes Issue(s)
- usdView Pyside6 support


sunyab:
Filed as internal issue #USD-7421

PR #1909, opened 2022-06-17
tf: account for Python GIL explicit initialization deprecation from Python 3.7 to 3.9

mattyjams:
### Description of Change(s)

Starting with Python 3.7, the GIL is initialized as part of Py_Initialize() and PyEval_InitThreads() no longer needs to be called separately. In Python 3.9, PyEval_InitThreads() no longer does anything and was marked deprecated.

See here for more detail:
https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads

This change avoids the use of PyEval_InitThreads() and PyEval_ThreadsInitialized() when building against Python 3.7 or later and cleans up the deprecation warnings that are otherwise issued. Those functions are slated for removal in Python 3.11.

### DIFF
--- a/pxr/base/tf/pyInterpreter.cpp
+++ b/pxr/base/tf/pyInterpreter.cpp
@@ -67,6 +67,9 @@ TfPyInitialize()
 
     if (!Py_IsInitialized()) {
 
+        // Starting with Python 3.7, the GIL is initialized as part of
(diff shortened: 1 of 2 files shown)


jilliene:
Filed as internal issue #USD-7439

PR #1910, opened 2022-06-18
Permit usdView debugging in Xcode

creijon:
### Description of Change(s)

Xcode builds its dylibs into a different tree than that of the command line, so when debugging usdView it is necessary to set the PXR_SHARED_LIBRARY_LOCATION env variable to the lib path in the build folder.

### Fixes Issue(s)
- Xcode debugging.


jilliene:
Filed as internal issue #USD-7440

PR #1913, opened 2022-06-22
Allow for Metal Storm applications without openGL

creijon:
### Description of Change(s)
This PR is provided for reference to anyone that needs a Metal-only application of USD Storm without OpenGL.

It only sets up the openGL state if there is a valid contexts, which removes the need to set up a dummy openGL context for Metal applications.

The behaviour of openGL applications like usdview isn't affected.

### Fixes Issue(s)
- Dummy openGL context in Metal applications.


jilliene:
Filed as internal issue #USD-7442

PR #1915, opened 2022-06-23
APPLE: macOS Cross Compilation

creijon:
### Description of Change(s)

This is a smaller change than the full Universal build support put up here: https://github.com/PixarAnimationStudios/USD/pull/1893

Allows for building ARM64 images on x86 machines and x86 ones on Apple Silicon by specifying the `--build-target` option.

It simplifies the changes into build_usd.py, avoiding the complexity of creating two images from TBB.

### Fixes Issue(s)
- Cross compilation on macOS


sunyab:
Filed as internal issue #USD-7454

jclosure:
+1 needed

meshula:
Are the version changes to OIIO, Embree, Alembic, and OpenEXR required to achieve cross compilation?

I thought Embree got the bump 3.13 to get M1 compilation going, I guess the reversion here is because the issues have been addressed in a different way?

Some of these changes, like bumps to OpenEXR would seem to hinge on USD moving to a higher vfxplatform level.

creijon:
Merged into dev in commit 4d3fa46ff269196555135e20b10089c4fba1d0ce

PR #1918, opened 2022-06-24
Enable Storm for Apple Intel GPUs

slingthor:
Note: Updates since last PR, mainly with regards to 13.0 and higher enablement

Description of Change(s)

- Use post tessellation vertex shader on Intel in most cases to support hardware barycentric coordinates on Intel Apple machines using metal ( needs hardware barycentric as no geometry shader is presented ). Uses the tessellation coordinate as a barycentric coordinate.
- Use HdStVBOSimpleMemoryManager instead of HdStVBOMemoryManager for nonUniformAggregationStrategy and nonUniformImmutableAggregationStrategy in the resourceRegistry on Apple intel machines to work around a Metal driver bug on Apple Intel Macs which prevents us from emulating base vertex via setVertexoffset. This change has a performance overhead, but correctness can not be achived through simple means otherwise. This change cause the baseVertex to always be 0. This bug has been fixed internally at Apple and is expected to be fixed in MacOS 13.0
- Pass patch_id between stages to work around a driver issue where the hardware primitive id is wrong if patch_id is not passed as an interstage.
- Fix the primitive id calculation for post tessellation vertex shader triquads ( division by two )
- Only enable Intel if version is 13.0 or more of MacOS
- Emulate primitive id with patch id


sunyab:
Filed as internal issue #USD-7456

PR #1928, opened 2022-06-28
Do not access PyFrameObject fields directly on Python 3.9+

musicinmybrain:
Fixes a Python 3.11 incompatibility. Still accesses `PyCodeObject` fields directly.

### Fixes Issue(s)
- (No issue filed.)

### DIFF
--- a/pxr/base/tf/pyTracing.cpp
+++ b/pxr/base/tf/pyTracing.cpp
@@ -35,8 +35,9 @@
 
 #include <tbb/spin_mutex.h>
 
-// This is from python, needed for PyFrameObject.
+// These are from python, needed for PyFrameObject.
 #include <frameobject.h>
+#include <patchlevel.h>
 
 #include <list>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


pxr/base/tf/pyTracing.cpp
```diff
@@ -110,11 +111,24 @@ static int _TracePythonFn(PyObject *, PyFrameObject *frame,
...
+#if PY_VERSION_HEX >= 0x030a00f0
+    // PyFrame_GetCode introduced in 3.9, added to Stable ABI in 3.10;
+    // PyFrameObject is opaque in 3.11
+    PyCodeObject * code = PyFrame_GetCode(frame);
+#else
+    // Use direct access to PyFrameObject fields.
+    PyCodeObject * code = frame->f_code;
+    // Create a strong reference as PyFrame_GetCode() would, so we do not have
+    // to conditionalize decrementing the reference count.
+    (void) Py_NewRef((PyObject *) code);
```
hroncok: My Python C API skills are rusty, but why not `Py_INCREF((PyObject *) code)`?

musicinmybrain: > My Python C API skills are rusty, but why not `Py_INCREF((PyObject *) code)`?

And I have spent a ton of time in Python and in C but have hardly touched Python’s C API at all.

Anyway, looking at the relevant documentation, it seems like `Py_INCREF` is a better fit, given that I don’t need the `PyObject` reference that `Py_NewRef` would return. Plus, it looks like `Py_NewRef` wasn’t in the older versions of Python I’m trying to cover here anyway.

musicinmybrain:
Adjusted based on @hroncok’s good suggestion. Also adjusted the C-style cast style to match others in the codebase (`(Foo *) var` → `(Foo *)var`).

musicinmybrain:
Now the patch defines a `static inline` compatibility version of `PyFrame_GetCode` on older Pythons that don’t have it, as suggested in https://docs.python.org/3.11/whatsnew/3.11.html#id6. I think the result is a bit cleaner.

Now `PyFrame_GetCode` is used exactly where it is available: Python 3.9.0b1 and later, again based on example in the CPython documentation.

I’ve made the change in a separate commit for clarity, but it can of course be squashed.

sunyab:
Filed as internal issue #USD-7464

PR #1930, opened 2022-06-30
Perform translations of DirtyCategories dirty bit

marktucker:
Perform dirty bit translations so that changes to rprim category
membership result in sync calls when using the UsdImaging scene delegate
with the scene index emulation layer.

### Description of Change(s)
This is sort of an amendment to #1653. Even after applying that patch, light linker updates didn't work in Houdini using USD 22.05 when a new light linker collection was added to the scene. Tracking this down revealed that the DirtyCategories bit was being set on the right rprims, but the translation of this dirty bit through the emulation scene index was losing this dirty bit. So this change tracks the dirty bit in HdDirtyBitsTranslator.


sunyab:
Filed as internal issue #USD-7466

PR #1932, opened 2022-07-04
Don't crash in UsdImagingInstanceAdapter::GetScenePrimPaths when passing invalid instanceIndices

mtavenrath:
### Description of Change(s)
Added index checking for each passed instance index to avoid out of bounds accesses.

### Fixes Issue(s)
Crash when passing invalid instance indices

### DIFF
--- a/pxr/usdImaging/usdImaging/instanceAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/instanceAdapter.cpp
@@ -2538,31 +2538,46 @@ UsdImagingInstanceAdapter::GetScenePrimPaths(
         if (!instanceIndices.empty()) {
             int minIdx = std::numeric_limits<int>::max();
             int maxIdx = 0;
+            int validIndices = 0;
 
             // determine the min/max index to determine how many bits have to be
             // allocated in the requestIndicesMap.
             for (size_t i = 0; i < instanceIndices.size(); i++) {
-                int remappedIndex = indices[instanceIndices[i]];
-                minIdx = std::min(minIdx, remappedIndex);
-                maxIdx = std::max(maxIdx, remappedIndex);
+                int instanceIndex = instanceIndices[i];
+
+                // skip invalid indices
+                if (instanceIndex < indices.size()) {
+                    int remappedIndex = indices[instanceIndex];
+                    minIdx = std::min(minIdx, remappedIndex);
+                    maxIdx = std::max(maxIdx, remappedIndex);
+                    ++validIndices;
+                }
             }
 
-            // For each requested index provide a mapping into the result vector
-            // Indices in the map set to std::numeric_limits<int>::max()
-            // are not being requested.
-            std::vector<int> requestedIndicesMap(
-                maxIdx - minIdx + 1, std::numeric_limits<int>::max());
+            // at least one index was valid, get the prim paths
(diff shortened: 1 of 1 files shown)


asluk:
Hi @sunyab  / @tcauchois -- I can do a review pass on this before you take a look later this week, if that works. This fixes some critical crashes in Omniverse that rely on GetScenePrimPaths. I'll also look to get a developer on my team to look into making a unit test if possible. Thanks!

sunyab:
Filed as internal issue #USD-7470

pxr/usdImaging/usdImaging/instanceAdapter.cpp
```diff
@@ -2542,11 +2542,20 @@ UsdImagingInstanceAdapter::GetScenePrimPaths(
...
+                    int remappedIndex = indices[instanceIndex];
+                    minIdx = std::min(minIdx, remappedIndex);
+                    maxIdx = std::max(maxIdx, remappedIndex);
+                }
             }
 
+            // not a single index valid valid, make empty requestedIndicesMap
+            if (minIdx > maxIdx)
+                minIdx = maxIdx;
+
```
sunyab: The comment says that `requestedIndicesMap` created below will be empty in this case, but it appears it will be a vector of size 1 instead.

But as a simplification: if there isn't a single valid index, can't we just return immediately?

pxr/usdImaging/usdImaging/instanceAdapter.cpp
```diff
@@ -2555,8 +2564,13 @@ UsdImagingInstanceAdapter::GetScenePrimPaths(
...
+                int instanceIndex = instanceIndices[i];
+
+                // skip invalid indices
+                if (instanceIndex < indices.size()) {
+                    int remappedIndex = indices[instanceIndex];
+                    requestedIndicesMap[remappedIndex - minIdx] = i;
+                }
             }
 
             result.resize(instanceIndices.size());
```
sunyab: This `resize` call doesn't seem to take into account whether there were any invalid entries in `instanceIndices`, so in that case `result` will be too large and contain empty paths, which clients likely won't expect. Do we need to do some extra handling here?

sunyab (changes requested):

sunyab:
@asluk We're getting ready to publish 22.08-rc1 in the next few days and since this is a regression introduced in this release we'd like to merge this as soon as possible. I've left some initial review notes and will update if there are additional notes from the team.

asluk:
> @asluk We're getting ready to publish 22.08-rc1 in the next few days and since this is a regression introduced in this release we'd like to merge this as soon as possible. I've left some initial review notes and will update if there are additional notes from the team.

Thanks @sunyab ; I don't have any further notes on the PR on top of yours.

pxr/usdImaging/usdImaging/instanceAdapter.cpp
mtavenrath: Yes, I updated the code to skip the whole loop if there's no valid instance index.

pxr/usdImaging/usdImaging/instanceAdapter.cpp
mtavenrath: I changed the code to count the number of valid indices for a correct resize.

PR #1936, opened 2022-07-07
APPLE: OCIO 2.1.1 with native Metal support

creijon:
### Description of Change(s)

Add OCIO 2 with GPU path for openGL and Metal.

Thanks to the team at Apple for the work on implementing this.

### Fixes Issue(s)
- Full OCIO support

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -21,7 +21,7 @@ Our test machines have the following software versions installed
 | Intel TBB     | 2019 Update 6        | 2018 Update 1, 2019 Update 6 | 2019 Update 6                  |
 | OpenSubdiv    | 3.5.0                | 3.5.0                        | 3.5.0                          |
 | OpenImageIO   | 2.1.16.0             | 2.3.15.0                     | 2.1.16.0                       |
-| OpenColorIO   | 1.1.0                | 1.1.0                        | 1.1.0                          |
+| OpenColorIO   | 1.1.0 (default)      | 2.1.2                        | 2.1.2                          |
 | OSL           | 1.10.9               |                              |                                |
 | Ptex          | 2.3.2                | 2.1.33                       | 2.1.33                         |
 | Qt for Python | PySide2 5.14.1       | PySide6 6.3.1                | PySide2 5.14.1                 |
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1338,10 +1338,15 @@ OPENIMAGEIO = Dependency("OpenImageIO", InstallOpenImageIO,
 
 ############################################################
 # OpenColorIO
-
-OCIO_URL = "https://github.com/imageworks/OpenColorIO/archive/v1.1.0.zip"
+OCIO_VERSION_1 = "1.1.0"
+OCIO_VERSION_2 = "2.1.2"
 
 def InstallOpenColorIO(context, force, buildArgs):
+    if context.versionOCIO == 2:
+        OCIO_URL = "https://github.com/AcademySoftwareFoundation/OpenColorIO/archive/refs/tags/v{version}.zip".format(version=OCIO_VERSION_2)
+    else:
+        OCIO_URL = "https://github.com/imageworks/OpenColorIO/archive/v{version}.zip".format(version=OCIO_VERSION_1)
+
     with CurrentWorkingDirectory(DownloadURL(OCIO_URL, context, force)):
         extraArgs = ['-DOCIO_BUILD_TRUELIGHT=OFF',
                      '-DOCIO_BUILD_APPS=OFF',
@@ -1350,7 +1355,8 @@ def InstallOpenColorIO(context, force, buildArgs):
                      '-DOCIO_BUILD_TESTS=OFF',
                      '-DOCIO_BUILD_PYGLUE=OFF',
                      '-DOCIO_BUILD_JNIGLUE=OFF',
-                     '-DOCIO_STATIC_JNIGLUE=OFF']
+                     '-DOCIO_STATIC_JNIGLUE=OFF',
+                     '-DOCIO_USE_OPENEXR_HALF=ON']
 
         # OpenImageIO v1.1.0 fails to build on Windows with the RelWithDebInfo
         # build type because it doesn't set up the correct properties for the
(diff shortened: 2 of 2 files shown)


pxr/imaging/hdx/colorCorrectionTask.cpp
```diff
@@ -151,78 +275,200 @@ HdxColorCorrectionTask::_CreateOpenColorIOResources()
...
-#else
         OCIO::DisplayViewTransformRcPtr transform =
             OCIO::DisplayViewTransform::Create();
         transform->setDisplay(display);
         transform->setView(view);
         transform->setSrc(inputColorSpace.c_str());
-#endif
+    
+        if (!_looksOCIO.empty()) {
+            transform->setDisplay(_looksOCIO.c_str());
```
remia: Looks cannot be specified in `setDisplay()`, if we want to keep the look override feature, we can use LegacyViewingPipeline pass it the DisplayTransform and use `setLooksOverrideEnabled()` / `setLooksOverride()`.

remia (commented):
That's a great contribution! As it is removing support of OpenColorIO 1.x, if this is approved, I guess we should also ask for a minimal version 2 when looking for OpenColorIO in CMake `find_package()` call.

pxr/imaging/hdx/colorCorrectionTask.cpp
```diff
@@ -111,20 +122,133 @@ bool
...
+typedef
+  std::vector<UniformBufferData> UniformBuffersData;
+
+template<typename T, int nElements = 1>
+void _SetConstantValue(UniformBuffersData& uniformData,
+                       std::string uniformType, std::string uniformName,
+                       T const* values, uint32_t count = 1)
+{
+    // Set the dummy value to 123456789
+    // so dummy value it is recongizable in a buffer
```
meshula:
```suggestion
    // that is easily recognizable in a buffer
```

pxr/imaging/hdx/colorCorrectionTask.cpp
```diff
@@ -275,13 +517,93 @@ HdxColorCorrectionTask::_CreateShaderResources()
...
+                ++bindingIdx;
+                if(_hgi->GetAPIName() == HgiTokens->Metal)
+                {
+                    fsCode += "textureBind_" + texInfo.texName + ",";
+                    fsCode += "samplerBind_" + texInfo.texName + ",";
+                }
+                else
+                {
+                    // For OpenGL case:
+                    // Since OCIO texture don't have binding index, we tend to use
```
meshula:
```suggestion
                    // Since OCIO textures don't have a binding index, we tend to use
```

pxr/imaging/hdx/colorCorrectionTask.cpp
```diff
@@ -275,13 +517,93 @@ HdxColorCorrectionTask::_CreateShaderResources()
...
+                    // the declaration provided by Hgi, that has proper binding and
+                    // layout. Therefore, we subsitute sampler name in the shader
```
meshula:
```suggestion
                    // layout. Therefore, we substitute sampler name in the shader
```

pxr/imaging/hdx/colorCorrectionTask.cpp
creijon: Fixed the grammar

pxr/imaging/hdx/colorCorrectionTask.cpp
creijon: Fixed the grammar

pxr/imaging/hdx/colorCorrectionTask.cpp
creijon: Ah, I see that there is an excellent example of this in the OCIO docs pages. I'll put the change in.

remia:
Tested on CentOS 7 and it worked nicely. I had to comment out the Metal language part because I was compiling with OCIO 2.0.4, which brings me to raising again the question about VFX Reference platform compatibility. I read not so long ago USD was just in the process of moving to VFX 2020 as the minimum supported platform. This still feature OCIO 1.1.1 which this PR drops support for, this PR actually needs VFX 2022 for Metal support which could easily be downgraded to VFX 2021 by having a conditional path for Metal. But actually supporting VFX 2020 would require re-introducing the conditional path that were present before the changes in this PR. That's something for USD leads to answer I assume.

Another point, unrelated to this PR, I found the new UI control for OCIO display view selection hard to understand, I'm not exactly sure what's going on but it will probably need another look.

sunyab:
Filed as internal issue #USD-7476

spiffmon:
Yes, this PR will need to be reworked to *not* drop support for OCIO 1.x,
or we will not even be able to build with it internally to Pixar.

--spiff

hodoulp:
> Yes, this PR will need to be reworked to *not* drop support for OCIO 1.x,
or we will not even be able to build with it internally to Pixar.

@spiffmon The OCIO v1 has limited GPU support (i.e. LUT approximation in GPU), no Metal support and OCIO v2 is fully backward compatible. But just out of curiosity, could you explain why using v2 is a blocker for Pixar?

spiffmon:
@hodoulp , Pixar does not currently have OCIO 2.x installed in our software suite, and we do have code outside of the USD project (that must cohabitate with USD) that uses OCIO.  For sure we could (and plan to eventually) update to using 2.x everywhere, and if as you say 2.x is 100% backwards compatible, that should be straightforward.  But it is still a reasonable-sized task, and we would not want the acceptance of this PR to be predicated on us doing that work.

Additionally, as already mentioned, we are trying to be good Platform citizens to the degree possible, and build_usd.py currently tracks CY2020 except where absolutely incompatible on specific platforms.    We're OK with accepting support for 2.x alongside 1.x in the code and having build_usd.py pull in 2.x only on Mac for the Metal support.   We understand that switching fully to 2.x will considerably simplify parts of build_usd.py, but we can't justify it quite yet.  We have verified that we cannot build against 1.1.1 with the PR as it stands currently.

creijon:
> @hodoulp , Pixar does not currently have OCIO 2.x installed in our software suite, and we do have code outside of the USD project (that must cohabitate with USD) that uses OCIO. For sure we could (and plan to eventually) update to using 2.x everywhere, and if as you say 2.x is 100% backwards compatible, that should be straightforward. But it is still a reasonable-sized task, and we would not want the acceptance of this PR to be predicated on us doing that work.
> 
> Additionally, as already mentioned, we are trying to be good Platform citizens to the degree possible, and build_usd.py currently tracks CY2020 except where absolutely incompatible on specific platforms. We're OK with accepting support for 2.x alongside 1.x in the code and having build_usd.py pull in 2.x only on Mac for the Metal support. We understand that switching fully to 2.x will considerably simplify parts of build_usd.py, but we can't justify it quite yet. We have verified that we cannot build against 1.1.1 with the PR as it stands currently.

Hi Spiff, we'll update the PR to include 1.1.1 support.  Is it acceptable to have this code in blocks protected by `#ifdef`?

spiffmon:
@jonny-apple , thank you - a smallish number of `#ifdef`s do seem appropriate here!

creijon:
> 

@spiffmon , we've re-introduced OCIO 1.1.1.  But we need a mechanism to control which version of OCIO is downloaded by `build_usd.py`.  Would an additional parameter into be suitable, or switch based on platform?  Currently macOS only supports OCIO 2 so we can select that.

spiffmon:
Thanks for asking, @jonny-apple ! Though it may eventually make sense to provide a switch in advance of vfxreference adoption, we'd like to be complexity-conservative for now, so would request you predicate the version on MacOS.

Thanks again!

creijon:
> 

Thanks @spiffmon , I'll make the change.

hodoulp:
I have a concern about having two different versions depending of the platform (by default, OCIO v1 for Windows and Linux, and OCIO v2 for macOS) because I think it could add confusion to users/devs/maintainers.

The OCIO v2 added numerous new features and greatly improved GPU (and CPU) processing. So, an OCIO v1 configuration file will work great with the OCIO v2 library (i.e. macOS) but the other way around is not true (i.e. an OCIO v2 configuration file does not work with the OCIO v1 library (i.e. Windows & Linux)).

I suggest to have a compilation flag (applicable to all platforms) to select the OCIO version which could be set to take OCIO v1 by default for now i.e. leaving Pixar ecosystem safe.

creijon:
> I have a concern about having two different versions depending of the platform (by default, OCIO v1 for Windows and Linux, and OCIO v2 for macOS) because I think it could add confusion to users/devs/maintainers.
> 
> The OCIO v2 added numerous new features and greatly improved GPU (and CPU) processing. So, an OCIO v1 configuration file will work great with the OCIO v2 library (i.e. macOS) but the other way around is not true (i.e. an OCIO v2 configuration file does not work with the OCIO v1 library (i.e. Windows & Linux)).
> 
> I suggest to have a compilation flag (applicable to all platforms) to select the OCIO version which could be set to take OCIO v1 by default for now i.e. leaving Pixar ecosystem safe.

Thanks, it sounds like a really good change.  I'll put it in.

creijon:
I’ve updated the branch to handle both OCIO 1.1.0 or 2.1.1 based on the user selection in build_usd.py, with defaults to 1.1.0 on Linux and 2.1.1 elsewhere.

I’ve moved the version-specific code in colorCorrectionTask.cpp into separate functions because the inline blocks of #ifdef made it very hard to understand the flow of the main functions.

Hope this helps,

Jon

PR #1949, opened 2022-07-14
[usdImaging] Preserved excludes paths on prim resync

fabal:
### Description of Change(s)
[usdImaging] Preserved excluded paths on prim resync.

We came across this issue in Maya with an application making some modification with the edit target set to the session layer. This would trigger the display of geometries meant to be excluded from hydra's render.

### Fixes Issue(s)
- Excluded geometries rendered upon a stage's local stack modification


sunyab:
Filed as internal issue #USD-7492

PR #1950, opened 2022-07-15
Add missing <algorithm> include for std::find and std::transform

charlesfleche:
### Description of Change(s)

Building with clang-13 was failing because of missing <algorithm> include. This PR adds the missing include.

### Fixes Issue(s)
-


charlesfleche:
The build issue looks like a fluke / timeout on the CI. Is there a way to re-trigger a run manually ?

sunyab:
Filed as internal issue #USD-7493

FlorianZ:
@charlesfleche, agreed! Don't worry about the build issue. We can see it built fine on Linux and macOS.

PR #1953, opened 2022-07-17
[tests] fixes for tests if paths have spaces

pmolodo:
### Description of Change(s)

pxr_register_test command takes (PRE_/POST_)COMMAND strings, which are
parsed into arguments via shlex.split.  However, path variables (which)
might contain spaces) were not quoted in these commands.

For references, all the changes in this commit were made via this regex
(executed in VSCode):

Find:
```regex
^((?: *#)?( *)pxr_register_test\((?:.|\n(?!\2\)))*?(?:PRE_|POST_)?COMMAND ")\$([^ "]*)
```
Replace:
```regex
$1'$$$3'
```

### Fixes Issue(s)
- Fixes test failures for usdedit if the path to `chmod` or `head` have a space in them

Note that the majority of the changes here are to fix issues with spaces in `CMAKE_INSTALL_PREFIX`.  While this PR may not completely resolve all build issues for spaces in `CMAKE_INSTALL_PREFIX` (in my test, boost had problems... hopefully fixed in newer releases?), it is least a step in the right direction, and a better example to folow in future.

### DIFF
--- a/extras/imaging/examples/hdTiny/CMakeLists.txt
+++ b/extras/imaging/examples/hdTiny/CMakeLists.txt
@@ -37,7 +37,7 @@ if (APPLE)
         ENV
             DYLD_INSERT_LIBRARIES=${CMAKE_INSTALL_PREFIX}/share/usd/examples/plugin/hdTiny.dylib
             ${PXR_PLUGINPATH_NAME}=${CMAKE_INSTALL_PREFIX}/share/usd/examples/plugin/hdTiny/resources
-        COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testHdTiny" 
+        COMMAND "'${CMAKE_INSTALL_PREFIX}/tests/testHdTiny'" 
         STDOUT_REDIRECT output.txt
         DIFF_COMPARE output.txt
     )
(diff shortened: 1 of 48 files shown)


sunyab:
Filed as internal issue #USD-7496

pmolodo:
Just noticed that [this commit](https://github.com/PixarAnimationStudios/OpenUSD/commit/b7a3793db20334b1b61d339b2427056ffd0e2801) addressed builds with paths with whitespace, but notes that it doesn't fix tests - which reminded me of this old PR.

If nothing else, the above regex might be handy...

sunyab:
@pmolodo Yep, I was working on the issue from this PR which led me to fixing the build issue in that commit. 

I was poking around at alternate fixes to avoid having to touch all of the test registrations and came up with a change that uses a similar regex, but in the implementation of pxr_register_test itself. That way, commands are automatically quoted. It seems to be working but I need to do a few more tests. One way or another, I'm hoping to put this issue to bed some time in the next release.

pmolodo:
Oh, awesome, good to hear!

nvmkuruc:
@sunyab @pmolodo Is this still an issue?

sunyab:
@nvmkuruc Yes it is, I haven't had a chance to come back to this yet and might not get to it for a bit in lieu of the python work.

PR #1956, opened 2022-07-19
APPLE: Fix CPU culling in PipelineDrawBatch

creijon:
### Description of Change(s)

When using `HdSt_PipelineDrawBatch`, the logic in `HdSt_RenderPass` was not checking that it had GPU culling disabled and therefore the CPU frustum culling was not being used.

We now check the flag for both `HdSt_IndirectDrawBatch` and `HdSt_PipelineDrawBatch` depending on what is enabled.

### Fixes Issue(s)
- There was no frustum culling on the HdSt_PipelineDrawBatch, since the GPU culling isn't implemented yet.  CPU culling was not being enabled in its place.
- Tested by loading KitchenSet, moving camera into the scene and then using the `HD_FREEZE_CULL_FRUSTUM` debug option in usdView.  Trucking the camera back should show that only the primitives within the frozen frustum are visible.


sunyab:
Filed as internal issue #USD-7503

PR #1958, opened 2022-07-21
[cmake] remove unused PXR_PYTHON_ENABLED and PXR_PYTHON_MODULES_ENABLED

pmolodo:
### Description of Change(s)

these are compile definitions set by the _pxr_library cmake function;
however, they are:

- never used anywhere
- set unconditionally (ie, regardless of PXR_ENABLE_PYTHON_SUPPORT)
- unneeded, as c++ code can include `pxr.h`, and use `#ifdef PXR_PYTHON_SUPPORT_ENABLED`


sunyab:
Filed as internal issue #USD-7516

PR #1967, opened 2022-07-28
APPLE: glslfx metadata extensions

slingthor:
### Description of Change(s)
- Extend glslfx to support
  - Able to express "centroid" and "sample" for shader data
  - Able to express spacings for tessellation. This is important so that we can express that for Metal in the future inside of the glslfx file instead of expressing it on the host code side.
  - Consume the parsed centroid and sample qualifiers for shaders in the shader generator and write out to the shader
- Introduce HgiTessellationSpacing which will be used in post tessellation control and basis curve Metal support

### DIFF
--- a/pxr/imaging/hd/tokens.h
+++ b/pxr/imaging/hd/tokens.h
@@ -185,6 +185,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 #define HD_SHADER_TOKENS                        \
     (alphaThreshold)                            \
+    (centroid)                                  \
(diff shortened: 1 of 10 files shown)


sunyab:
Filed as internal issue #USD-7525

slingthor:
Just made a minor change
fractional_even_spacing ->fractional_even
fractional_odd_spacing ->fractional_odd

when parsing qualifiers in codegen

slingthor:
Brought in as part of basis curve / mesh curve work

PR #1975, opened 2022-07-30
GetShader function in shader program

slingthor:
### Description of Change(s)
Adds a convenient GetShaderFunction for ShaderProgram.
Found it justified as it landed itself well as a member function and it could prevent repetition from users of the class


sunyab:
Filed as internal issue #USD-7530

slingthor:
This was brought in as part of the mesh/basis curve work

PR #1978, opened 2022-08-01
APPLE: Fix crash in skinned animation on Metal

creijon:
### Description of Change(s)

The function `HdStCopyComputationGPU::Execute()` had a call to `GLF_POST_PENDING_GL_ERRORS` in it.  This macro expands out to include openGL-specific calls such as `glGetError()` etc.  On Metal applications these cause an exception if the GL context wasn't initialised.

This crash was found when performing skinning calculations.  See the HumanFemale test model for a reproduction, running in the Apple HydraPlayer sample application from WWDC 2022.

Also removed the include to `pxr/imaging/glf/diagnostic.h` in `pxr/imaging/hdSt/extCompGpuComputation.cpp` where it isn't needed.

Suggestion: The file `pxr/imaging/glf/diagnostic.h` is included in various other places in HdSt files, but it is only used for the `GLF_GROUP_FUNCTION` macro.  I think we should move this macro to the Hd library to remove the dependancy from HdSt onto Glf entirely.  It's dangerous to have the glf code included, even if it isn't used.  I'm happy to make this change but it should be in a separate PR.

### Fixes Issue(s)
- Crash in macOS applications

### DIFF
--- a/pxr/imaging/hdSt/copyComputation.cpp
+++ b/pxr/imaging/hdSt/copyComputation.cpp
@@ -21,8 +21,6 @@
 // KIND, either express or implied. See the Apache License for the specific
 // language governing permissions and limitations under the Apache License.
 //
-#include "pxr/imaging/glf/diagnostic.h"
-
 #include "pxr/imaging/hdSt/copyComputation.h"
 #include "pxr/imaging/hdSt/bufferArrayRange.h"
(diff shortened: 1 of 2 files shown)


sunyab:
Filed as internal issue #USD-7538

PR #1980, opened 2022-08-03
Apple: Intel graphics enablement

slingthor:
### Description of Change(s)
Enables Intel on Apple
- Works around primitive ID and hardware barycentric limitations by always taking the post tessellation vertex path and uses the patch_id and tessellation coordinate instead
- Works around a texture writing limitation on Intel, where we would see a glitch with the format specified while in an argument buffer.
- Enable the requiresBasePrimitiveOffset workaround for Intel
- Works around a limitation on Intel where the indirect drawing patch buffer's per stride contents's size has to be a multiple of 32
### Fixes Issue(s)
-  Hydra Storm not working on Intel Macs

### DIFF
--- a/pxr/imaging/hdSt/codeGen.cpp
+++ b/pxr/imaging/hdSt/codeGen.cpp
@@ -2316,16 +2316,6 @@ HdSt_CodeGen::_CompileWithGeneratedHgiResources(
             HgiShaderKeywordTokens->hdPosition);
         
         if (_hasPTVS) {
-            std::string vecSize = std::string();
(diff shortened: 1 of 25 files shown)


slingthor:
Will be looking at simplifying the interstage declaration of ptvsBarycentrics coord - had to remove it initially because of glslfx metadata limitations. Looking into improving upon that limitation to simplify even further.
Functionally I believe it is sound.

sunyab:
Filed as internal issue #USD-7539

creijon:
Looks like this has been merged now as part of the changes from @davidgyu

Issue #1983, opened 2022-08-05
Running USD from build directory without installing


jdumas:
Currently USD relies on a very specific folder structure to load plugin information at runtime:

https://github.com/PixarAnimationStudios/USD/blob/3abc46452b1271df7650e9948fef9f0ce602e3b2/cmake/macros/Private.cmake#L195-L204

Given the current CMake setup, this means that one needs to *install* the USD target into the final destination folder in order to run anything that is compiled against the USD library. This is problematic when building USD from source via [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html), since one requires an additional `install()` step to run things like unit tests.

My suggestion to remedy this would be as follows:
- Build everything under the right folder structure from the start (e.g. under `${PROJECT_BINARY_DIR}/install/lib/usd`)
- Do not rely on CMake's `install()` command to move/rename resources files (`plugInfo.json` and `generatedSchema.usda`). If everything is built with the right folder structure, the `install()` command will simply copy the whole `${PROJECT_BINARY_DIR}/install` as is.

In order to achieve this I would suggest the following changes:
1. Set the `LIBRARY_OUTPUT_DIRECTORY` on USD CMake targets to change their output folder when built:
  ```cmake
  set_target_properties(<usd_target> PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/install/lib)
  ```
2. In addition to the `install()` rule for resource files, provide an additional `configure_file()` that takes care of copying the resource to the right location:
  https://github.com/PixarAnimationStudios/USD/blob/3abc46452b1271df7650e9948fef9f0ce602e3b2/cmake/macros/Private.cmake#L254-L258
3. Change the following `configure_file()` to create the `plugInfo.json` at the right location:
  https://github.com/PixarAnimationStudios/USD/blob/3abc46452b1271df7650e9948fef9f0ce602e3b2/cmake/macros/Private.cmake#L106-L109
  
  
I opened #1984 which contains the proposed changes. I haven't looked in details at the python side of things, but for a basic build on my machine it performs as intended.

sunyab:
Filed as internal issue #USD-7551

PR #1984, opened 2022-08-06
Build resources with the folder structure to avoid install() step

jdumas:
### Description of Change(s)

- Set LIBRARY_OUTPUT_DIRECTORY on USD CMake targets
- Add `configure_file()` to copy `plugInfo.json` and other resources to the correct location in the build folder

The proposed changes do no affect the install structure of USD, but allow running USD from the build folder without installing the files. See #1983 for more explanations.

### Fixes Issue(s)
- #1983


sunyab:
Filed as internal issue #USD-7552

PR #1987, opened 2022-08-09
[arch, plug] fix paths that contain unicode char for plugInfo.json

nvidia-jomiller:
### Description of Change(s)
arch: ArchGetAddressInfo was using GetModuleFileName instead of GetModuleFileNameW on Windows. This prevented the objectPath to be correctly returned should the module found at the address containing a unicode character.

plug: _ReadPlugInfoObject did not properly handle the case of a UTF-8 encoded string on Windows. This would result in the plugInfo.json file not being loaded if it resided at a path containing a unicode character

### Fixes Issue(s)

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -97,6 +97,19 @@ def Python3():
     return sys.version_info.major == 3
 
 def GetLocale():
+    if Windows():
+        # Windows handles encoding a little differently then Linux / Mac
+        # For interactive streams (isatty() == True) it will use the
+        # console codepage, otherwise it will return an ANSI codepage.
+        # To keep things consistent we'll force UTF-8 for non-interactive
+        # streams (which is recommended by Microsoft). See:
+        # https://docs.python.org/3.6/library/sys.html#sys.stdout
+        # https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
+        if sys.stdout.isatty():
+            return sys.stdout.encoding
+        else:
+            return "UTF-8"
+
     return sys.stdout.encoding or locale.getdefaultlocale()[1] or "UTF-8"
 
 def GetCommandOutput(command):
--- a/pxr/base/plug/CMakeLists.txt
+++ b/pxr/base/plug/CMakeLists.txt
@@ -166,6 +166,21 @@ pxr_create_test_module(TestPlugModuleIncomplete
     INSTALL_PREFIX PlugPlugins
 )
 
+if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    # Create a test module that has a unicode character in it's path
+    # for Windows
+    pxr_create_test_module(TestPlugModuleUnicode
+        INSTALL_PREFIX PlugPlugins
+        DEST_DIR TestPlugMödüleÜnicöde
+    )
+else()
+    # Create a test module without a unicode character in the path
+    # to make this test consistent on other OSes
+    pxr_create_test_module(TestPlugModuleUnicode
+        INSTALL_PREFIX PlugPlugins
+    )
+endif()
+
 pxr_register_test(testPlug
     PYTHON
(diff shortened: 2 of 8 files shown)


nvidia-jomiller:
It looks like the build for Windows on Azure is hanging. It's working correctly on Windows 10 but will investigate when time permits.

The line where this seems to be hanging is here:
https://github.com/PixarAnimationStudios/USD/blob/release/build_scripts/build_usd.py#L266

sunyab:
Filed as internal issue #USD-7554

build_scripts/build_usd.py
```diff
@@ -97,6 +97,19 @@ def Python3():
...
 
 def GetLocale():
+    if Windows():
+        # Windows handles encoding a little differently then Linux / Mac
+        # For interactive streams (isatty() == True) it will use the
+        # console codepage, otherwise it will return an ANSI codepage.
+        # To keep things consistent we'll force UTF-8 for non-interactive
+        # streams (which is recommended by Microsoft). See:
+        # https://docs.python.org/3.6/library/sys.html#sys.stdout
+        # https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
```
nvidia-jomiller, line 107: I was able to track down the build issue with Windows Azure and it has something to do with what default encoding is set for the terminal running `build_usd.py`. I made a small change to [GetLocale() in build_usd.py](https://github.com/PixarAnimationStudios/USD/pull/1987/files#diff-88086ae4493b825e88542e8e0f512ad651197ae226692311f4e15699c9275ad8R101) to force `UTF-8`  on non-interactive terminals for Windows.

Specifically [sys.stdout.encoding](https://docs.python.org/3.6/library/sys.html#sys.stdout) on Windows will use the ANSI codepage for non-interactive streams and [Microsoft recommends to use UTF-8 or UTF-16 over ANSI codepages](https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers)

meshula: nice find

pxr/base/plug/CMakeLists.txt
```diff
@@ -166,6 +166,21 @@ pxr_create_test_module(TestPlugModuleIncomplete
     INSTALL_PREFIX PlugPlugins
 )
 
+if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    # Create a test module that has a unicode character in it's path
+    # for Windows
+    pxr_create_test_module(TestPlugModuleUnicode
+        INSTALL_PREFIX PlugPlugins
+        DEST_DIR TestPlugMödüleÜnicöde
```
nvidia-jomiller, line 174: It's worth pointing out that `DEST_DIR` will be a created directory which contains unicode characters to verify the fix

nvidia-jomiller:
I know this isn't a very high priority bug, but it would be great if this could get reviewed and merged into `dev`. At least a few users have ran into this same problem and one has verified that it fixes the issue:
https://groups.google.com/g/usd-interest/c/Lit5oKhejoE/m/2aKR_8h-BQAJ

Thanks!

spiffmon:
We're scheduling this for the next release, @nvidia-jomiller - thanks!

meshula:
@nvidia-jomiller I was wondering if you've tried this under Windows 11? Although the fix seems to address the specific case in the issue title, I haven't been able to get "unicode user name in path" case mentioned in the usd-interest message working without further work in addition to the changes you propose here, and I'm wondering if it's a Windows 10 vs 11 thing.

nvidia-jomiller:
@meshula I haven't tried under Windows 11 as my Windows machine is still on Windows 10. I assume you're testing on Windows 11? What additional work did you need to do with the usd-interest case to get it working?

meshula:
TfGlob failed to discover the pluginfo.json files and I made a fix that I'm not particularly excited about. If you are interested in seeing some of the issues around unicode in filenames versus toolchains (not USD specifically), I have been collecting findings here... https://github.com/meshula/m-n

PR #1988, opened 2022-08-10
[ar] Allow a plugin resolver to declare itself as primary

nvidia-jomiller:
### Description of Change(s)

A plugin resolver could only be considered a primary resolver if it does not specify uriSchemes in plugInfo.json. This prevents a "fallback" sort of behavior when multiple plugin resolvers that support URI schemes need to live in the same environment but one is intended to be the primary resolver

This change adds a "primary" flag for a plugin resolver in plugInfo.json to explicitly set. The "primary" flag indicates that the plugin resolver could be a primary resolver but also supports specific URI schemes

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1989

### DIFF
--- a/pxr/usd/ar/resolver.cpp
+++ b/pxr/usd/ar/resolver.cpp
@@ -73,6 +73,9 @@ TF_DEFINE_PRIVATE_TOKENS(_tokens,
     // Plugin metadata key for resolver URI schemes.
     (uriSchemes)
 
+    // Plugin metadata key for specifying primary resolver functionality
+    (primary)
+
     // Plugin metadata keys for specifying resolver functionality.
     (implementsContexts)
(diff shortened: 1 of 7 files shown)


pxr/usd/ar/resolver.cpp
```diff
@@ -73,6 +73,9 @@ TF_DEFINE_PRIVATE_TOKENS(_tokens,
     // Plugin metadata key for resolver URI schemes.
     (uriSchemes)
 
+    // Plugin metadata key for specifying primary resolver functionality
+    (primary)
```
nvidia-jomiller, line 77: This might be better named as `canBePrimary`

sunyab:
Filed as internal issue #USD-7555

PR #1995, opened 2022-08-16
Fix Markdown syntax.

waywardmonkeys:
Triple backticks aren't required for inline code spans when there
aren't backticks within the inline code spans.

Some Markdown tools don't handle these as well as they could.


sunyab:
Filed as internal issue #USD-7560

PR #1997, opened 2022-08-17
Use literal `true` instead of `1` for bools.

waywardmonkeys:
This fixes one class of warnings for the `pxr/base` code from clang-tidy.

### DIFF
--- a/pxr/base/tf/pxrLZ4/lz4.cpp
+++ b/pxr/base/tf/pxrLZ4/lz4.cpp
@@ -956,7 +956,7 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
                     break;   /* match found */
                 }
 
-            } while(1);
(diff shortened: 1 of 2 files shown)


spiffmon:
Hi @waywardmonkeys ,
   Can you please submit your changes to lz4.cpp to the [LZ4 gitHub repo](https://github.com/lz4/lz4)?  We are simply inlining it, and would not want to deviate except for a critical fix.  Once it's been accepted there, it would make sense for us to incorporate it here.  Thanks!

sunyab:
Filed as internal issue #USD-7561

sunyab:
I've merged the fix to the getenv.cpp internally and that'll get into the dev branch with the next push. I've  skipped the change to LZ4 since that fix is more appropriate for the LZ4 repo itself.

PR #2007, opened 2022-08-23
feat: add VS 2022 & boost 1.78 & python 3.10 support

sssooonnnggg:
feat: add VS 2022 & boost 1.78 & python 3.10 support

CMake version >= 3.24 required

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -145,6 +145,10 @@ def IsVisualStudioVersionOrGreater(desiredVersion):
         return version >= desiredVersion
     return False
 
+def IsVisualStudio2022OrGreater():
+    VISUAL_STUDIO_2022_VERSION = (17, 0)
+    return IsVisualStudioVersionOrGreater(VISUAL_STUDIO_2022_VERSION)
+
 def IsVisualStudio2019OrGreater():
     VISUAL_STUDIO_2019_VERSION = (16, 0)
     return IsVisualStudioVersionOrGreater(VISUAL_STUDIO_2019_VERSION)
@@ -376,7 +380,9 @@ def RunCMake(context, force, extraArgs = None):
     # building a 64-bit project. (Surely there is a better way to do this?)
     # TODO: figure out exactly what "vcvarsall.bat x64" sets to force x64
     if generator is None and Windows():
-        if IsVisualStudio2019OrGreater():
+        if IsVisualStudio2022OrGreater():
+            generator = "Visual Studio 17 2022"
+        elif IsVisualStudio2019OrGreater():
             generator = "Visual Studio 16 2019"
         elif IsVisualStudio2017OrGreater():
             generator = "Visual Studio 15 2017 Win64"
@@ -681,10 +687,11 @@ ZLIB = Dependency("zlib", InstallZlib, "include/zlib.h")
 if MacOS():
     # This version of boost resolves Python3 compatibilty issues on Big Sur and Monterey and is
     # compatible with Python 2.7 through Python 3.10
-    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.76.0/source/boost_1_76_0.tar.gz"
+    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz"
     BOOST_VERSION_FILE = "include/boost/version.hpp"
 elif Linux():
-    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.70.0/source/boost_1_70_0.tar.gz"
+    # We should use different boost version depending on python version on Linux, 
+    # and detecting python version need context to be initialized, so we set `BOOST_URL` later.
     BOOST_VERSION_FILE = "include/boost/version.hpp"
 elif Windows():
     # The default installation of boost on Windows puts headers in a versioned 
@@ -694,10 +701,17 @@ elif Windows():
     #
     # boost 1.70 is required for Visual Studio 2019. For simplicity, we use
     # this version for all older Visual Studio versions as well.
-    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.70.0/source/boost_1_70_0.tar.gz"
-    BOOST_VERSION_FILE = "include/boost-1_70/boost/version.hpp"
+    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz"
+    BOOST_VERSION_FILE = "include/boost-1_78/boost/version.hpp"
 
 def InstallBoost_Helper(context, force, buildArgs):
+    
+    if Linux():
+        # We're using 1.70 to adhere to the VFX Reference Platform but are forced to use 1.78 to support 3.10.
+        pythonInfo = GetPythonInfo(context)
+        boostVersion = "1_78_0" if pythonInfo[3] == "3.10" else "1_70_0"
+        BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_{}.tar.gz".format(boostVersion)
+    
     # Documentation files in the boost archive can have exceptionally
     # long paths. This can lead to errors when extracting boost on Windows,
     # since paths are limited to 260 characters by default on that platform.
@@ -806,6 +820,8 @@ def InstallBoost_Helper(context, force, buildArgs):
                 b2_settings.append("toolset=msvc-14.1")
             elif context.cmakeToolset == "v140":
                 b2_settings.append("toolset=msvc-14.0")
+            elif IsVisualStudio2022OrGreater():
+                b2_settings.append("toolset=msvc-14.3")
             elif IsVisualStudio2019OrGreater():
                 b2_settings.append("toolset=msvc-14.2")
             elif IsVisualStudio2017OrGreater():
@@ -2160,9 +2176,9 @@ if not isPython64Bit:
 if which("cmake"):
     # Check cmake requirements
     if Windows():
-        # Windows build depend on boost 1.70, which is not supported before
-        # cmake version 3.14
-        cmake_required_version = (3, 14)
+        # Windows build depend on boost 1.78, which is not supported before
+        # cmake version 3.24
+        cmake_required_version = (3, 24)
     else:
         cmake_required_version = (3, 12)
     cmake_version = GetCMakeVersion()


e4lam:
Which VS2022 version did you build it in? As a heads up, I've also been building USD with VS2022 v17.3.1 and found a serious bug in the compiler that made USD unusable. The first symptom of it being that the `testSdfLayer` unit test fails. I still have other test failures but I haven't gone far enough down to road to confirm which ones are legitimate failures or not. (Some of the ones I get are because the unit test can't find PySide for some reason.)

For reference, the bug I reported is here:
https://developercommunity.visualstudio.com/t/cl-19321933-ternary-operator-bug-call/10126287

In the meantime, I got things going by applying this patch in my local build to workaround it:
```
diff -urN --strip-trailing-cr USD-py3.9/pxr/usd/sdf/layer.cpp USD-py3.9.new/pxr/usd/sdf/layer.cpp
--- USD-py3.9/pxr/usd/sdf/layer.cpp     2022-08-16 17:00:31.924779700 -0400
+++ USD-py3.9.new/pxr/usd/sdf/layer.cpp 2022-08-19 23:24:06.588275300 -0400
@@ -727,8 +727,12 @@
     // If we're trying to open an anonymous layer, do not try to compute the
     // real path for it.
     ArAssetInfo assetInfo;
-    string resolvedLayerPath = isAnonymous ? layerPath :
-        Sdf_ResolvePath(layerPath, computeAssetInfo ? &assetInfo : nullptr);
+    string resolvedLayerPath;
+    if (isAnonymous)
+        resolvedLayerPath = layerPath;
+    else
+        resolvedLayerPath = Sdf_ResolvePath(layerPath,
+                                    computeAssetInfo ? &assetInfo : nullptr);

     // Merge explicitly-specified arguments over any arguments
     // embedded in the given identifier.
```

sssooonnnggg:
> Which VS2022 version did you build it in? As a heads up, I've also been building USD with VS2022 v17.3.1 and found a serious bug in the compiler that made USD unusable. The first symptom of it being that the `testSdfLayer` unit test fails. I still have other test failures but I haven't gone far enough down to road to confirm which ones are legitimate failures or not. (Some of the ones I get are because the unit test can't find PySide for some reason.)
> 
> For reference, the bug I reported is here: https://developercommunity.visualstudio.com/t/cl-19321933-ternary-operator-bug-call/10126287
> 
> In the meantime, I got things going by applying this patch in my local build to workaround it:
> 
> ```
> diff -urN --strip-trailing-cr USD-py3.9/pxr/usd/sdf/layer.cpp USD-py3.9.new/pxr/usd/sdf/layer.cpp
> --- USD-py3.9/pxr/usd/sdf/layer.cpp     2022-08-16 17:00:31.924779700 -0400
> +++ USD-py3.9.new/pxr/usd/sdf/layer.cpp 2022-08-19 23:24:06.588275300 -0400
> @@ -727,8 +727,12 @@
>      // If we're trying to open an anonymous layer, do not try to compute the
>      // real path for it.
>      ArAssetInfo assetInfo;
> -    string resolvedLayerPath = isAnonymous ? layerPath :
> -        Sdf_ResolvePath(layerPath, computeAssetInfo ? &assetInfo : nullptr);
> +    string resolvedLayerPath;
> +    if (isAnonymous)
> +        resolvedLayerPath = layerPath;
> +    else
> +        resolvedLayerPath = Sdf_ResolvePath(layerPath,
> +                                    computeAssetInfo ? &assetInfo : nullptr);
> 
>      // Merge explicitly-specified arguments over any arguments
>      // embedded in the given identifier.
> ```

I use VS2022 17.1.1，it seems that this version has right output for your rvalue test

sunyab:
Filed as internal issue #USD-7582

build_scripts/build_usd.py
```diff
@@ -386,7 +392,7 @@ def RunCMake(context, force, extraArgs = None):
     if generator is not None:
         generator = '-G "{gen}"'.format(gen=generator)
 
-    if IsVisualStudio2019OrGreater():
+    if IsVisualStudio2022OrGreater() or IsVisualStudio2019OrGreater():
```
seando-adsk: Since you are making a change here I have a fix for a problem I found where on Windows you use a generate other than Visual Studio (such as Ninja).
`if generator and 'Visual Studio' in generator and IsVisualStudio2019OrGreater():`

Also is the or really needed? The function is IsVisualStudio2019**OrGreater**() so shouldn't it apply to VS2022 already?

sssooonnnggg: Yes, I think you are right, this is not needed

seando-adsk (commented):
I would suggest changing line 395 of build_usd.py from:
`    if IsVisualStudio2019OrGreater():`
to:
`    if generator and 'Visual Studio' in generator and IsVisualStudio2019OrGreater():`
The code now is assuming that on Windows the generator being used is Visual Studio, but there are other generators (such as Ninja) that work on Windows.

Otherwise all the changes look fine.

seando-adsk (approved):

build_scripts/build_usd.py
```diff
@@ -681,10 +687,10 @@ def InstallZlib(context, force, buildArgs):
 if MacOS():
     # This version of boost resolves Python3 compatibilty issues on Big Sur and Monterey and is
     # compatible with Python 2.7 through Python 3.10
-    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.76.0/source/boost_1_76_0.tar.gz"
+    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz"
     BOOST_VERSION_FILE = "include/boost/version.hpp"
 elif Linux():
-    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.70.0/source/boost_1_70_0.tar.gz"
+    BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz"
```
sunyab: We're currently trying to adhere to the [VFX Reference Platform](https://vfxplatform.com/) for CY2020 as much as possible, especially on Linux. That produces builds that are the same as our internal builds, which may be helpful for troubleshooting or debugging issues.(*)

So, I'd ask that we only use boost 1.78 on Linux if we detect that we are running Python 3.10 or later. It'd be great to put a comment in here too saying that we're using 1.70 to adhere to the VFX Reference Platform but are forced to use 1.78 to support 3.10.

sunyab (changes requested):
Thank you for looking into this! I left one note below but otherwise this seems good to me.

Also, I think @seando-adsk 's comment about the code at line 395 is valid but it's a pre-existing issue and I think doesn't need to be addressed in this change. I'm actually not sure if specifying the platform name causes any problems for generators that don't support them -- the CMake docs imply that the name is just made available for generators to use only if they want to.

Thanks!

build_scripts/build_usd.py
sssooonnnggg: Done, boost version is depending on python version on Linux now.

sssooonnnggg:
Thanks for reviewing, boost version is depending on python version on Linux now.

seando-adsk (approved):

Tisten:
You ought to bump the required cmake version to 3.24 as well: https://github.com/PixarAnimationStudios/USD/blob/203806dfa0c9bf66c34d5cc58697101849de6fe3/build_scripts/build_usd.py#L2181

meshula:
@Tisten Good note on the cmake version bump, that is required for sure or people will run into problems. The support for VS2022 was added in 3.21. (https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html) Are you aware of fixes after 3.21 for VS2022 that might be critical? I'm not arguing for one version or another, just checking what the minimum viable is.

Tisten:
Well, the initial comment by [sssooonnnggg](https://github.com/sssooonnnggg) was:
> CMake version >= 3.24 required

And that has been my experience as well, 3.22 didn't work but 3.24 did. Can't really say which fix it is that makes or breaks it.

sssooonnnggg:
> You ought to bump the required cmake version to 3.24 as well:
> 
> https://github.com/PixarAnimationStudios/USD/blob/203806dfa0c9bf66c34d5cc58697101849de6fe3/build_scripts/build_usd.py#L2181

Done.

seando-adsk (approved):

build_scripts/build_usd.py
```diff
@@ -2160,9 +2176,9 @@ def ForceBuildDependency(self, dep):
 if which("cmake"):
     # Check cmake requirements
     if Windows():
-        # Windows build depend on boost 1.70, which is not supported before
-        # cmake version 3.14
-        cmake_required_version = (3, 14)
```
Danielmelody, line 2165: Requiring newest cmake while not force using 1.78 seems unreasonable. A user would have to install cmake 3.24 for all windows build, even with VS2019 or VS2017

build_scripts/build_usd.py
```diff
@@ -2160,9 +2176,9 @@ def ForceBuildDependency(self, dep):
...
+        # Windows build depend on boost 1.78, which is not supported before
+        # cmake version 3.24
+        cmake_required_version = (3, 24)
```
Danielmelody, line 2181: Requiring newest cmake while not force using 1.78 seems unreasonable. A user would have to install cmake 3.24 for all windows build, even with VS2019 or VS2017

sssooonnnggg: > A user would have to install cmake 3.24 for all windows build, even with VS2019 or VS2017

Yes, I agree that it seems unreasonable.

In this PR, boost 1.78 is forced on Windows, CMake lower than 3.24 will not find boost 1.78 correctly, so CMake 3.24 is required.

Danielmelody: So, in my opinion, we should not force using 1.78, only if we use python > 3.8, the cmake and boost's version requirement should be present

SimonBoorer:
Is there a reason why we can't just use -A x64 in USD/build_usd.py and not explicitly specify a generator? The CMAKE_GENERATOR_PLATFORM documentation that the -A option refers to says that this is supported by VS 2005 and above:
https://cmake.org/cmake/help/v3.14/manual/cmake.1.html#options
https://cmake.org/cmake/help/v3.14/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM

Is this not the "better way" that the TODO comment refers to?

pmolodo:
Pinging this because we're looking to add Python 3.10 and VS2022 support as well.  Are there still outstanding issues to be addressed for this MR?

meshula:
It's good as is! We'll update this issue after the holidays!

seando-adsk:
FYI, at Autodesk we are building USD with python 3.10 binding support. The one thing I had to do was update boost version to 1.76 on all platforms.

pmolodo:
Saw the recent commits related to this, which is great!  Think the only thing still left from this PR is the branching of boost + cmake versions if using python 3.10.

Also - I noticed a typo, and filed an issue, though you're probably already on it:

https://github.com/PixarAnimationStudios/USD/issues/2143

Thanks again!

meshula:
@pmolodo Thanks for the notes. I see the bit you are referring to to get 3.10 support on Linux.

pmolodo:
> @pmolodo Thanks for the notes. I see the bit you are referring to to get 3.10 support on Linux.

Since we're deviating slightly from this PR anyway... I actually think we need to branch based on python version for all platforms, not just Linux.  As @seando-adsk noted [above](https://github.com/PixarAnimationStudios/USD/pull/2007#issuecomment-1369846697), boost needs to be updated to at least 1.76 for any platform that's compiling against python-3.10.  I'm thinking we should just force to boost 1.78 and cmake 3.24 for all platforms, for consistency, if python >= 3.10.

I'm happy to make a PR for this myself if that would be easier.

meshula:
It makes sense to have consistent versions on all platforms py310's selected. I'm working it through now.

PR #2008, opened 2022-08-23
Remove macOS experimental warning

dgovil:
### Description of Change(s)
This PR just removes the CMake warning that states the macOS build is experimental. Since USD Metal is now part of the repo, and we regularly build USD for macOS with the regular components, I think it's safe to remove the disclaimer.


sunyab:
Filed as internal issue #USD-7583

sunyab:
We've removed the "experimental" message from both macOS and Windows builds and cleaned up some associated docs in a separate change that'll be part of the next dev push. Closing this out, thanks for the nudge!

Issue #2014, opened 2022-08-27
Value Clips break if the stage layer stack is modified before reading values


marktucker:
### Description of Issue
When a value clip is loaded, a USD_Clip object is created per-clip. But the layer for that clip is not immediately loaded. Instead, the clip layer loading is deferred until a value is needed from that layer. Because of the way Usd_Clip stores the layer that owns the clip metadata, the calculation of the clip file path will be wrong if the stage's layer stack is modified after the Usd_Clip object is created, but before the value is read.

### Steps to Reproduce
1. Unzip this file: 
[cubeclip.zip](https://github.com/PixarAnimationStudios/USD/files/9437156/cubeclip.zip)
2. From the directory containing cube.usda, run this python script:
```
from pxr import Sdf, Usd
s = Usd.Stage.Open('cube.usda')
l = Sdf.Layer.CreateAnonymous()
s.GetSessionLayer().subLayerPaths.append(l.identifier)
p = s.GetPrimAtPath('/geo/cube')
a = p.GetAttribute('size')
print(a.Get(3))
```
3. The output value will be 2.0 instead of the correct value of 1.0833333. This is because the individual clip files are not being found.

tallytalwar:
Filed as internal issue #USD-7603

PR #2015, opened 2022-08-27
Record the absolute path to the clip file when constructing a Usd_Clip

marktucker:
### Description of Change(s)
Record the absolute path to the clip file when constructing the Usd_Clip
object, rather than re-calculating the clip file path when data is first
requested from the clip file. This prevents incorrect clip file path
calculations if the layer stack has been modified since the Usd_Clip
object was created.

I'm not sure if this is the correct way to fix this (the other thought I had was responding to change notifications affecting the layer stack, but that seemed like it would be far more complicated, with no real benefit.

### Fixes Issue(s)
- #2014

### DIFF
--- a/pxr/usd/usd/clip.cpp
+++ b/pxr/usd/usd/clip.cpp
@@ -122,9 +122,14 @@ Usd_Clip::Usd_Clip(
     if (TF_VERIFY(sourceLayerIndex < sourceLayerStack->GetLayers().size())) {
         const ArResolverContextBinder binder(
             sourceLayerStack->GetIdentifier().pathResolverContext);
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-7604

sunyab:
Hey @marktucker, thanks for this PR, I confirmed this does fix the repro case you posted in #2014. However, I found a related issue where the order of entries returned by `UsdProperty::GetPropertyStack` will also be affected by the original bug, and that doesn't get fixed by this change.

I have an alternate fix that solves both problems. I'm going to close this PR out and submit the fix against #2014. Thanks!

Issue #2013, opened 2022-08-25
Hydra fails to update primvars on connection notification


JGamache-autodesk:
### Description of Issue

When connecting the first file texture to a UsdPreviewSurface shader, the viewport does not refresh correctly.

I have traced this back to [UsdImagingDelegate::ApplyPendingUpdates()](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdImaging/delegate.cpp#L1041) where a property path is not strong enough to cause a call to [_ResyncUsdPrim(usdPath, &indexProxy)](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdImaging/delegate.cpp#L1050) which correctly repopulates the mesh and fetches the new "st" primvar requirement.

Would this work as a potential fix? In [UsdImagingDelegate::_OnUsdObjectsChanged()](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usdImaging/usdImaging/delegate.cpp#L1111), loop on the `_infoChanges` and resync the prim path of any prim that has a `didChangeAttributeConnection` flag active in the SdfChangeList::Entry vector.

### Steps to Reproduce
1. Load the attached scene un usdview
2. Run the following script in the interpreter window:
```
file_output = UsdShade.Shader(usdviewApi.stage.GetPrimAtPath("/Looks/usdPreviewSurface1SG/file1")).CreateOutput("rgb", Sdf.ValueTypeNames.Color3f)
surf_input = UsdShade.Shader(usdviewApi.stage.GetPrimAtPath("/Looks/usdPreviewSurface1SG/usdPreviewSurface1")).CreateInput("diffuseColor", Sdf.ValueTypeNames.Color3f)
surf_input.ConnectToSource(file_output)
```

EXPECTED: Viewport refreshes and shows texture on plane
BUG: Viewport refreshes, but plane turns yellow, which is the color at UV (0,0)
WORKAROUND: Deactivating then reactivating the plane prim will show correct texturing

The scene is a quad with a textured UsdPreviewSurface material that only requires connecting the file texture output to the surface to be complete. A graph editor component attached to that scene would expect the viewport to correctly refresh when doing the final connection.

### System Information (OS, Hardware)

OS: Windows

### Package Versions

USD: 22.08

JGamache-autodesk:
[...]

PR #2017, opened 2022-08-30
Resync prim path on connection changed

seando-adsk:
### Description of Change(s)

A connection changed notification implies a shader rebuild, which should
force a rescan of cached data for the affected material/light/camera.

In the case of a material, this will also force a rescan of the dependent
geometries, which is exactly what is needed to fix the refresh issue.

### Fixes Issue(s)
- #2013


tallytalwar:
Filed as internal issue #USD-7606

PR #2020, opened 2022-09-02
fix: avoid using cpp keywords as tokens after `camelCased`

sssooonnnggg:
If we define a token such as "Default", the generated cpp code will not compile.
Because the cpp keyword check is performed before camelCased, so the token "default" is used as a variable name in generated code, which causes compile issues.


tallytalwar:
Filed as internal issue #USD-7609

PR #2022, opened 2022-09-04
[sdf, usd] Fix crash when URL-encoded characters are used to generate session layer identifiers

asluk:
[sdf, usd] Fix crash when URL-encoded characters are used to generate session layer identifiers

When opening a stage whose root layer has URL-encoded characters, sequences such as %20 for spaces could get misinterpreted as format strings when generating session layer identifiers on the fly via TfStringPrintf. This change uses stringstreams to ensure that URL-encoded characters never get misinterpreted on this codepath, which was previously leading to crashes such as the one in the testUsdStage case added here.

Thanks to Seema Kulkarni and Edward Slavin for earlier work on this PR.

### DIFF
--- a/pxr/usd/sdf/assetPathResolver.cpp
+++ b/pxr/usd/sdf/assetPathResolver.cpp
@@ -258,6 +258,12 @@ Sdf_GetAnonLayerIdentifierTemplate(
     const string& tag)
 {
     string idTag = tag.empty() ? tag : TfStringTrim(tag);
+
+    // Ensure that URL-encoded characters are not misinterpreted as
+    // format strings to TfStringPrintf in Sdf_ComputeAnonLayerIdentifier.
+    // See discussion in https://github.com/PixarAnimationStudios/USD/pull/2022
+    idTag = TfStringReplace(idTag, "%", "%%");
+
     return _Tokens->AnonLayerPrefix.GetString() + "%p" +
         (idTag.empty() ? idTag : ":" + idTag);
 }
--- a/pxr/usd/usd/CMakeLists.txt
+++ b/pxr/usd/usd/CMakeLists.txt
@@ -458,6 +458,11 @@ pxr_install_test_dir(
 )
 
 pxr_install_test_dir(
+    SRC testenv/testUsdStage
+    DEST testUsdStage
+)
+
+pxr_install_test_dir(
     SRC testenv/testUsdAppliedAPISchemas
     DEST testUsdAppliedAPISchemas_AutoApplyDisabled
 )
--- a/pxr/usd/usd/testenv/testUsdStage.py
+++ b/pxr/usd/usd/testenv/testUsdStage.py
@@ -29,6 +29,10 @@ from pxr import Sdf,Usd,Tf
 allFormats = ['usd' + c for c in 'ac']
 
 class TestUsdStage(unittest.TestCase):
+    def test_URLEncodedIdentifiers(self):
+        stage = Usd.Stage.Open("Libeccio%20LowFBX.usda")
+        assert stage
+
     def test_Repr(self):
         stage = Usd.Stage.CreateInMemory()
 
--- a/pxr/usd/usd/testenv/testUsdStage/Libeccio%20LowFBX.usda
+++ b/pxr/usd/usd/testenv/testUsdStage/Libeccio%20LowFBX.usda
@@ -1,0 +1,17 @@
+#usda 1.0
+(
+    defaultPrim = "hello"
+)
+
+def Xform "hello"
+{
+    custom double3 xformOp:translate = (4, 5, 6)
+    uniform token[] xformOpOrder = ["xformOp:translate"]
+
+    def Sphere "world"
+    {
+        float3[] extent = [(-2, -2, -2), (2, 2, 2)]
+        color3f[] primvars:displayColor = [(0, 0, 1)]
+        double radius = 2
+    }
+}


asluk:
Looks like there are some newly failing tests after I rebased to latest dev; will try to look into those over the next couple weeks.

pxr/usd/sdf/assetPathResolver.cpp
```diff
@@ -223,7 +223,22 @@ Sdf_ComputeAnonLayerIdentifier(
...
 {
     TF_VERIFY(layer);
-    return TfStringPrintf(identifierTemplate.c_str(), layer);
+
+    // Identifier template is either "anon:" or "anon:SOME_TAG"
+    // and we respectively return "anon:LAYER_ADDRESS" or "anon:LAYER_ADDRESS:SOME_TAG".
+    // We use stringstream instead of TfStringPrintf so that URL-encoded characters such as
+    // %20 do not get misinterpreted as format strings, which triggers crashes in the
+    // string library.
+    std::stringstream strstream;
```
meshula: std::stringstream brings in further behaviors, and can be slow compared to plain old concatenation. Since the form of the template is known, and no locale info and so on should be considered, using TfStringJoin here instead would avoid any new unexpected behaviors and be more performant than either stringstream or TfStringPrintf to boot.

asluk: Thanks @meshula ! Will rework accordingly.

tallytalwar:
Filed as internal issue #USD-7612

pxr/usd/sdf/assetPathResolver.cpp
asluk: Hi @meshula -- dumb question-- without TfStringPrintf or stringstream, how would we get the layer address into the string via TfStringJoin?

Secondly, SdfLayer::SdfLayer has

`    // If the identifier has the anonymous layer identifier prefix, it is a
    // template into which the layer address must be inserted. This ensures
    // that anonymous layers have unique identifiers, and can be referenced by
    // Sd object reprs.
    string layerIdentifier = Sdf_IsAnonLayerIdentifier(identifier) ?
        Sdf_ComputeAnonLayerIdentifier(identifier, this) : identifier;`
        
whereas testSdfLayer has 

`    def test_FindOrOpenNoAssetAnonLayers(self):
        """
        This test case confirms assumptions about how anonymous layers behave 
        for fileformats that can read/generate layers without requiring a 
        resolved asset to read. These assumptions allow anonymous layer 
        identifiers along with file format arguments to be used to refer to 
        fully dynamically generated layers without requiring a placeholder real 
        asset file that is never read.
        """`

which deliberately does not want the layer address inserted into the identifier for those cases. I'm not sure how best to manage the handshaking here between Sdf_GetAnonLayerIdentifierTemplate and Sdf_ComputeAnonLayerIdentifier, as it's possible for an input to come in with something like ":%p:" inside of a tag to Sdf_GetAnonLayerIdentifierTemplate  .

In the test_FindOrOpenNoAssetAnonLayers cases, Sdf_GetAnonLayerIdentifierTemplate is not called, so maybe it's possible to have Sdf_GetAnonLayerIdentifierTemplate prefix its return values with some sentinel other than "anon:", which Sdf_GetAnonLayerIdentifierTemplate then interprets as requiring the layer address in the identifier. But I'm not sure that's the most elegant and clear approach.

Thanks for reading!

meshula: Hi Aaron, thanks for your patience!

We followed the rabbit down the hole on the `:%p:` business you note above!  Ultimately, just before  `SdfComputeAnonLayerIdentifier` is invoked, this routine is called, to poke in the layer identifier:

```cpp
string
Sdf_GetAnonLayerIdentifierTemplate(
    const string& tag)
{
    string idTag = tag.empty() ? tag : TfStringTrim(tag);
    return _Tokens->AnonLayerPrefix.GetString() + "%p" +
        (idTag.empty() ? idTag : ":" + idTag);
}
```

We think the most robust way to handle the issue you report, and preserve the important `:%p`  behavior, is to modify the `Sdf_GetAnonLayerIdentifierTemplate` routine so that `string idTag` is computed not by white space trimming tag, but by white space trimming the tag AND also doubling any percents in the input tag in order to safely escape them. (By "robust", here I mean that your URL-containing strings will be bottlenecked reliably through `Sdf_GetAnon...` making it a better fix site than `Sdf_ComputeAnon...`.) 

Side note: Ideally, a single loop might construct `idTag` by simultaneously trimming white space and duplicating the percent signs, but given the amount of string copying already exhibited in the current implementation of `Sdf_GetAnonLayerIdentifierTemplate`, I don't think it would make a difference to overly optimize it.

asluk: Thanks @meshula ! I'll take a pass at it.

meshula:
Looks good, thanks!

meshula:
Hi @asluk ! I haven't debugged yet to see what's going wrong, but I wanted to give you a heads up that I haven't had a successful test run ~

```
====================================================================== 
ERROR: test_URLEncodedIdentifiers (__main__.TestUsdStage)
 ---------------------------------------------------------------------- 

File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/tests/testUsdStage", 
line 33, in test_URLEncodedIdentifiers
  stage = Usd.Stage.Open("Libeccio%20LowFBX.usda") 
pixar.Tf.ErrorException:  Error in 'UsdStage::Open' at line 879 in file pxr/usd/usd/stage.cpp : 
   'Failed to open layer @Libeccio%20LowFBX.usda@'
```

asluk:
> Hi @asluk ! I haven't debugged yet to see what's going wrong, but I wanted to give you a heads up that I haven't had a successful test run ~
> 
> ```
> ====================================================================== 
> ERROR: test_URLEncodedIdentifiers (__main__.TestUsdStage)
>  ---------------------------------------------------------------------- 
> Traceback (most recent call last): 
> 
> File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/tests/testUsdStage", 
> line 33, in test_URLEncodedIdentifiers
>   stage = Usd.Stage.Open("Libeccio%20LowFBX.usda") 
> pixar.Tf.ErrorException:  Error in 'UsdStage::Open' at line 879 in file pxr/usd/usd/stage.cpp : 
>    'Failed to open layer @Libeccio%20LowFBX.usda@'
> ```

Thanks @meshula ... it's passing for me on Windows; will try it on Ubuntu via WSL2 next week (are the Azure pipelines running tests? They are all passing, but I forget if they are only compiling). 🙏

tallytalwar:
@asluk yeah azure pipeline doesn't run tests at present. We want to get to it but I think timeouts are a concern there! https://github.com/PixarAnimationStudios/USD/blob/release/azure-pipelines.yml

asluk:
> > Hi @asluk ! I haven't debugged yet to see what's going wrong, but I wanted to give you a heads up that I haven't had a successful test run ~
> > ```
> > ====================================================================== 
> > ERROR: test_URLEncodedIdentifiers (__main__.TestUsdStage)
> >  ---------------------------------------------------------------------- 
> > Traceback (most recent call last): 
> > 
> > File "/scratch/build_local/dev/src/inst/fedora-gcc64-opt/pxr/tests/testUsdStage", 
> > line 33, in test_URLEncodedIdentifiers
> >   stage = Usd.Stage.Open("Libeccio%20LowFBX.usda") 
> > pixar.Tf.ErrorException:  Error in 'UsdStage::Open' at line 879 in file pxr/usd/usd/stage.cpp : 
> >    'Failed to open layer @Libeccio%20LowFBX.usda@'
> > ```
> 
> Thanks @meshula ... it's passing for me on Windows; will try it on Ubuntu via WSL2 next week (are the Azure pipelines running tests? They are all passing, but I forget if they are only compiling). 🙏

Thanks @meshula  and @tallytalwar -- the test is passing for me on Windows and on Ubuntu via WSL2; not sure what might be happening differently on Fedora :(

PR #2024, opened 2022-09-06
Prevent invalid binding relationship access in UsdShadeMaterialBindingAPI

tkchanat:
### Description of Change(s)
The validity of the binding relationship is not verified before accessing the stage, which then throws a `Usd_ThrowExpiredPrimAccessError` while dereferencing the prim data handle. Adding an extra check in the if-condition prevents such invalid access and returns an empty material if the binding relationship is invalid.

### DIFF
--- a/pxr/usd/usdShade/materialBindingAPI.cpp
+++ b/pxr/usd/usdShade/materialBindingAPI.cpp
@@ -252,7 +252,7 @@ UsdShadeMaterialBindingAPI::DirectBinding::DirectBinding(
 UsdShadeMaterial
 UsdShadeMaterialBindingAPI::DirectBinding::GetMaterial() const
 {
-    if (_bindingRel && !_materialPath.IsEmpty()) {
+    if (_bindingRel.GetPrim() && !_materialPath.IsEmpty()) {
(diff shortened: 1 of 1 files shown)


pxr/usd/usdShade/materialBindingAPI.cpp
```diff
@@ -252,7 +252,7 @@ UsdShadeMaterialBindingAPI::DirectBinding::DirectBinding(
 UsdShadeMaterial
 UsdShadeMaterialBindingAPI::DirectBinding::GetMaterial() const
 {
-    if (!_materialPath.IsEmpty()) {
+    if (_bindingRel && !_materialPath.IsEmpty()) {
```
spiffmon: Thank you, @tkchanat !  Can you change this pattern to `if (_bindingRel.GetPrim() ... ` ?  It is sufficient, and cheaper than a full Relationship validity test.  Thank you!

tkchanat: Thanks for the review @spiffmon! I changed the pattern in the latest commit and it's looking good now.

tallytalwar:
Filed as internal issue #USD-7614

PR #2030, opened 2022-09-08
Fix find opensubdiv condition check

Danielmelody:
### Description of Change(s)
Cmake would define `OPENSUBDIV_${UPPERCOMPONENT}_LIBRARY` to `OPENSUBDIV_${UPPERCOMPONENT}_LIBRARY-NOTFOUND`  if find opensubdiv failed, this variable is **definded**, but evals to false, so we should check the variable itself, not whether it is defined.
Ref: https://cmake.org/cmake/help/latest/command/find_library.html
### Fixes Issue(s)
- None


tallytalwar:
Filed as internal issue #USD-7629

PR #2040, opened 2022-09-13
Thor/dev/openxr debugging apple

slingthor:
### Description of Change(s)
Currently building and debugging opener and thus also openimageio using generator Xcode and debug is not possible due to the USD findopenexr cmake stage can't find openexr due to the openexrconfig.h is not referenced in the openexr cmakelist. This patch fixes that. For full debugging support using openimageio, also requires a PR outstanding for libpng and improvement to AR to find shared debug libraries.

### Fixes Issue(s)
- Can't build USD with openexr/openimageio because findopenxr won't find openexr on Generator Xcode


meshula:
Hi @slingthor, this seems like a reasonable change. Did you close it because you found a different work around, or is it not actually a problem?

The reason I ask is this could be a reasonable PR upstream on the OpenEXR project, and could be patched into a point release of OpenEXR.

PR #2042, opened 2022-09-15
APPLE: Texture and buffer writability in Metal/HGI along with settings where appropriate

slingthor:
### Description of Change(s)
Allows HGI to define readonly buffers and textures
Allow HGIMetal to consume readonly buffers and textures
Take advantage of that in the resource binder and as well in hdx where we manually set up resources
Sets buffers we don't index into to live in constant space, and pass them as const reference instead of as a pointer in device space on Metal

### DIFF
--- a/pxr/imaging/hgi/resourceBindings.h
+++ b/pxr/imaging/hgi/resourceBindings.h
@@ -85,6 +85,7 @@ struct HgiBufferBindDesc
     HgiBindResourceType resourceType;
     uint32_t bindingIndex;
     HgiShaderStage stageUsage;
+    bool writable;
 };
 using HgiBufferBindDescVector = std::vector<HgiBufferBindDesc>;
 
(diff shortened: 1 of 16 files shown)


tallytalwar:
Filed as internal issue #USD-7640

pxr/imaging/hgi/resourceBindings.h
```diff
@@ -130,6 +131,7 @@ struct HgiTextureBindDesc
     HgiBindResourceType resourceType;
     uint32_t bindingIndex;
     HgiShaderStage stageUsage;
+    bool readOnly;
```
creijon: In shadersection we use `_writable` while here we use `readOnly`.  For consistency shouldn't they be the same?

creijon (changes requested):

creijon (changes requested):
Make sure that the `readOnly` flag is initialised in every instance of the `HgiBufferBindDesc` creation.  There's a few instances in `flatNormals.cpp`, `quadrangulate.cpp`, `resourceBinder.cpp`, `smoothNormals.cpp`, `subdivision.cpp`, `boundingBoxTask.cpp`,  that need to be covered.

Also the comparator in `resourceBindings.cpp` needs to include a test for the flag.

slingthor:
PR updated! Will be sharing some metrics on MacOS as well

creijon (approved):

PR #2045, opened 2022-09-17
APPLE: GPU culling on PipelineDrawBatch

creijon:
### Description of Change(s)

Implementation of GPU culling using compute shaders in the PipelineDrawBatch.  The vertex-shader transform feedback approach that was used previously in the IndirectDrawBatch has not been supported in HGI since it would require multiple command buffers for synchronisation of the stages.  The compute shader approach does not require a separate reset pass for instance culling since it can expand out the instances in a loop within the compute shader and populate the drawCommand buffers directly.

Indexing into the drawCommands in the shader is done through a small emulation function in the codegen, `GetDrawingCoordField()`, which has to be implemented in the kernel.  This uses the thread ID to offset into the drawCommand buffer and access the encoded instance count, transform etc directly.

Also, the `g_instanceID` parameter needs to be set in the culling kernel so that the codegen can access it in `GetInstanceIndexCoord()`. 

For performance reasons, I've added a flag to specify that the compute encoder is created with concurrent dispatch.

To use the PipelineDrawBatch on OpenGL the option `HDST_ENABLE_HGI_RESOURCE_GENERATION` has been changed to default `true`

There was also an issue around how the garbage collector was releasing resources outside of the HGI BeginFrame::EndFrame pair. This was causing a crash on OpenGL when picking a prim in the viewport. I've put a potential fix for this in HdxPickTask.

### Fixes Issue(s)
- Culling on instanced and non-instanced primitives on the GPU.

Tested on:
- MacBook Pro M1 Max, with macOS Ventura
- MacBook Pro i7 Radeon 560, with macOS Ventura
- HP Xeon nVidia 3080, with Ubuntu 20.02


tallytalwar:
Filed as internal issue #USD-7642

slingthor:
Looks sound! Some ideas for minor improvements
Edit: Actually has problems on Linux

slingthor:
Requires more testing on Linux with PipelineDrawBatch and HGI resource generation enabled, so that it runs the new culling but not the one in indirect draw batch.

The culling program needs to compile and run on Linux under HGI.

Some starting points are fixing the implicit casts and ensuring the drawing coordinates are correctly defined for compute shaders

creijon:
Thanks for finding this.  I've fixed the code gen.

There was also an issue around how the garbage collector was releasing resources outside of the HGI BeginFrame::EndFrame pair.  This was causing a crash on OpenGL when picking a prim in the viewport.  I've put a potential fix for this in HdxPickTask.

creijon:
Accidentally closed with a force-push.  Reopened as https://github.com/PixarAnimationStudios/USD/pull/2053

PR #2046, opened 2022-09-18
APPLE: Single GPU callback for freeing all argument buffers on a command buffer

creijon:
### Description of Change(s)

Previously, when each argument buffer was created there would be a callback added to the command buffer to return it to the free list once it was used.  This requires a mutex to be locked for each callback, since the free list is read on the CPU thread.

For complicated scenes this could result in hundreds of callbacks and mutex locks, which makes the debugger hard to follow and introduces the possibility of stalls between the CPU and GPU.

### Fixes Issue(s)
- Performance instability.


tallytalwar:
Filed as internal issue #USD-7643

PR #2048, opened 2022-09-21
APPLE: Fix for crash on loading non-filterable textures

creijon:
### Description of Change(s)
In Metal, several of the texture formats don't support filtering, so the call to `generateMipmapsForTexture` with fail with an exception on pipeline validation.

### Fixes Issue(s)
- Crashing when loading U16 textures etc.


tallytalwar:
Filed as internal issue #USD-7655

PR #2055, opened 2022-10-05
Add displayName Metadata to Prim Spec

erslavin:
- Added `displayName` metadata field to prims
- Adjusted tests to test for and modify authored `displayName` metadata

It is important that we keep ABI compatibility on this change, so Get/Set DisplayName methods were not added to `UsdPrim`.  This addition will be added to the PR later - all access to `displayName` on a prim at the moment should be via Get/Set Metadata.


asluk:
This is part of the Omniverse work for UTF-8 support-- alongside @erslavin's implementation of Unicode identifiers which is not ABI-compatible, we are also offering displayName support so that stakeholders can use it for breadcrumbs back to their UTF-8 encoded source data, and view the displayNames in DCCs that support them. Thanks!

tallytalwar:
Filed as internal issue #USD-7681

asluk:
I know folks are probably busy putting the final touches on v22.11, and we don't expect to make this into there, but please let us know for now if there is any major objection to this PR in principle, as we'll need to deploy it ahead of an official USD release to unblock some key customer workflows-- thanks!

spiffmon:
Hi @asluk and @erslavin , without having done a proper review yet, we don't have any major objections **but** we feel it is important to include a minimum of support in usdview, something like:
* An option in the PrimView browser (in the "Show" menu) to enable use of displayNames, which defaults to on.
* Ensure that when this option _is_ on:
  * Using the search bar will search display names since that's what the user sees
  * the prim tooltips will indicate the prim's True Name.
* The path selection bar at the top of usdview would continue to show true names, even when the "Show display names" option is on, since it must be copy/pastable into relationship target paths, etc.

I know we have yet to provide a similar treatment for property displayNames in the property browser, but we feel the potential for confusion is greater at the level of scene hierarchy... but also that especially anyone who is using the displayName feature would really expect to be able to navigate their scene in usdview using them.

Thanks!

asluk:
> Hi @asluk and @erslavin , without having done a proper review yet, we don't have any major objections **but** we feel it is important to include a minimum of support in usdview, something like:
> 
> * An option in the PrimView browser (in the "Show" menu) to enable use of displayNames, which defaults to on.
> * Ensure that when this option _is_ on:
>   
>   * Using the search bar will search display names since that's what the user sees
>   * the prim tooltips will indicate the prim's True Name.
> * The path selection bar at the top of usdview would continue to show true names, even when the "Show display names" option is on, since it must be copy/pastable into relationship target paths, etc.
> 
> I know we have yet to provide a similar treatment for property displayNames in the property browser, but we feel the potential for confusion is greater at the level of scene hierarchy... but also that especially anyone who is using the displayName feature would really expect to be able to navigate their scene in usdview using them.
> 
> Thanks!

Agreed - we'll work on the usdview side over the next week or so-- thanks!

pxr/usd/sdf/schema.cpp
```diff
@@ -872,6 +872,8 @@ SdfSchemaBase::_RegisterStandardFields()
                        SdfMetadataDisplayGroupTokens->core)
         .MetadataField(SdfFieldKeys->DisplayGroupOrder,
                        SdfMetadataDisplayGroupTokens->core)
+        .MetadataField(SdfFieldKeys->DisplayName,
```
asluk, line 875: We'll add this to field keys in wrapPrimSpec.cpp as well

erslavin:
What if the displayName metadata is empty and the option is on - I would expect in this case the prim name would be used, what do you think?

spiffmon:
Absolutely, @erslavin - that is what we do in Presto, where `displayName` is a _computed_ quantity, but if the computational inputs are empty, we fall back to prim name.  I also wanted to clarify what I said about tooltip... we want it to provide useful information regardless of whether the new mode is on or off, so it should probably **always** tell you both the prim name and displayName (if it has one).

Thanks!

erslavin:
Work in progress - no automated tests yet but modifications were made in `usdview` to conform to the request

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -982,6 +982,13 @@ def __init__(self, parserData, resolverContextFn):
             self._ui.actionShow_Abstract_Prims.triggered.connect(
                 self._toggleShowAbstractPrims)
 
+            # nv begin #prim-display-name
```
asluk: These comment blocks will get removed on the next push. Thanks for the quick turnaround, Ed!

erslavin: Removed on latest commit

erslavin:
Augmented the existing automated tests for `usdview` to account for `displayName` and made some modifications to account for `displayName` not being authored.  Let me know if the functionality is what you expect - in particular the way I implemented search was to search both the `displayName` (if authored) and prim identifier when the setting is on, and only the identifier when it's off.

pxr/usd/sdf/schema.cpp
erslavin: added on the last commit

spiffmon:
@erslavin , can you please migrate all of the "DisplayName"-related methods on `UsdProperty` up to `UsdObject`, now that it is applicable to all object types?

spiffmon:
... and we'll also want to test that in testUsdMetadata

pxr/usd/usd/testenv/testUsdSchemaRegistry/resources/schema.usda
```diff
@@ -35,6 +35,7 @@ class "AbstractTest" (
 
 class MetadataTest "MetadataTest" (
     inherits = </AbstractTest>
+    displayName = "Display Name"
```
spiffmon: I don't think we should allow `displayName` to be authored into schema definitions - it's too much of a landmine, that your **type** can effectively override the name you've given to a prim!  Accordingly, `usdGenSchema` should detect and flag this as an invalid schema definition, which we should also test in `testUsdSchemaGen` .

pxr/usd/usd/testenv/testUsdMetadata.py
```diff
@@ -284,6 +284,13 @@ def test_DisplayName(self):
             attr = foo.CreateAttribute("attr", Sdf.ValueTypeNames.String)
             rel = foo.CreateRelationship("rel")
 
+            # verify display name metadata on prim
+            self.assertEqual(foo.GetMetadata("displayName"), None)
+            self.assertEqual(foo.HasAuthoredMetadata("displayName"), False)
+            self.assertEqual(foo.SetMetadata("displayName", "foo"), True)
+            self.assertEqual(foo.GetMetadata("displayName"), "foo")
+            self.assertEqual(foo.HasAuthoredMetadata("displayName"), True)
+
```
spiffmon: A good place to test the methods moved to `UsdObject()`

pxr/usdImaging/usdviewq/utils.h
```diff
@@ -59,6 +59,10 @@ class UsdviewqUtils {
         bool visVaries;
         std::string name;
         std::string typeName;
+
+        // nv begin #prim-display-name
+        std::string displayName;
+        // nv end
```
spiffmon: Can we remove the decorations?

pxr/usdImaging/usdviewq/utils.cpp
```diff
@@ -104,6 +104,19 @@ UsdviewqUtils::PrimInfo::PrimInfo(const UsdPrim &prim, const UsdTimeCode time)
     else
         name = _tokens->root.GetString();
     typeName = prim.GetTypeName().GetString();
+
+    // nv begin #prim-display-name
```
spiffmon: Can we remove the decorations?

pxr/usdImaging/usdviewq/utils.cpp
```diff
@@ -104,6 +104,19 @@ UsdviewqUtils::PrimInfo::PrimInfo(const UsdPrim &prim, const UsdTimeCode time)
...
+
+    if (prim.HasAuthoredMetadata(SdfFieldKeys->DisplayName))
+    {
+        prim.GetMetadata<std::string>(SdfFieldKeys->DisplayName, &displayName);
```
spiffmon: Can use the `UsdObject` API you'll be adding.

pxr/usdImaging/usdviewq/wrapUtils.cpp
```diff
@@ -48,6 +48,7 @@ _GetPrimInfo(UsdPrim const &prim, UsdTimeCode time)
 {
     UsdviewqUtils::PrimInfo info = UsdviewqUtils::GetPrimInfo(prim, time);
     
+    // nv begin #prim-display-name
```
spiffmon: decorations...

pxr/usdImaging/usdviewq/primViewItem.py
```diff
@@ -256,6 +260,10 @@ def _nameData(self, role):
                 toolTip = 'Inactive ' + toolTip
             elif self.isInstance:
                 toolTip = 'Instanced ' + toolTip
+        
+            # tooltip should always show both name and display name
+            toolTip = toolTip + "<br>Name: " + self.name + "<br>Display Name: " + self.displayName
```
spiffmon: Thanks, @erslavin - I was not very clear in my request.  If a prim *has* a displayName, then we should show both name and displayName regardless of the "Show Display Names" setting.  But if a prim has **no** displayName, then I don't think we need to add anything here.

pxr/usdImaging/usdviewq/mainWindowUI.ui
```diff
@@ -2660,6 +2661,14 @@
...
    </property>
   </action>
+  <action name="actionShow_Prim_DisplayName">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Use Display Name (Prims)</string>
+   </property>
+  </action>
```
spiffmon, line 2671: Can't immediately tell whether you did or not, but sometimes devs like to hand-edit the `.ui` files, which can cuase puzzling extraneous diffs for the next developer who actually uses `qdesigner`.  So we request that, if you **do** hand-edit a `.ui` file, you then open it in qdesigner and Save it back out so that Qt puts its own idea of ordering into the file.

🙏

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2185,16 +2188,28 @@ def _findPrims(self, pattern, useRegex=True):
...
 
-        matches = [prim.GetPath() for prim
-                   in Usd.PrimRange.Stage(self._dataModel.stage,
-                                             self._displayPredicate)
-                   if isMatch(prim.GetName())]
+        if self._dataModel.viewSettings.showPrimDisplayNames:
+            matches = [prim.GetPath() for prim
+                    in Usd.PrimRange.Stage(self._dataModel.stage,
+                                                self._displayPredicate)
+                    if ((prim.HasAuthoredMetadata("displayName") and isMatch(prim.GetMetadata("displayName"))) or isMatch(prim.GetName()))]
```
spiffmon: Discussed this with our designers, and they feel it may be too confusing to match against names you can't see, so can we only match against a non-empty displayName, when present?  And if that change doesn't bring us under 80 columns, can we do a continuation, also, please?

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2185,16 +2188,28 @@ def _findPrims(self, pattern, useRegex=True):
...
         if self._dataModel.viewSettings.showAllPrototypePrims:
-            for prototype in self._dataModel.stage.GetPrototypes():
-                matches += [prim.GetPath() for prim
-                            in Usd.PrimRange(prototype, self._displayPredicate)
-                            if isMatch(prim.GetName())]
+            if self._dataModel.viewSettings.showPrimDisplayNames:
+                for prototype in self._dataModel.stage.GetPrototypes():
+                    matches += [prim.GetPath() for prim
+                                in Usd.PrimRange(prototype, self._displayPredicate)
+                                if ((prim.HasAuthoredMetadata("displayName") and isMatch(prim.GetMetadata("displayName"))) or isMatch(prim.GetName()))]
```
spiffmon: As above...

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -3189,6 +3204,11 @@ def _toggleShowAbstractPrims(self):
         self._dataModel.selection.removeAbstractPrims()
         self._resetPrimView()
 
+    def _toggleShowPrimDisplayName(self):
+        self._dataModel.viewSettings.showPrimDisplayNames = (
+            self._ui.actionShow_Prim_DisplayName.isChecked())
+        self._resetPrimView()
```
spiffmon, line 3243: Doesn't affect this code, but reminds me to ask for verification: the Prim View browser should already be handling the `signalSettingChanged` signal, and we should make sure that its response includes resetting any search we've previously performed, since its results might differ depending on whether we're showing displayNames.

pxr/usdImaging/bin/testusdview/testenv/testUsdviewPrimTreeExpansion/testUsdviewPrimTreeExpansion.py
```diff
@@ -128,6 +128,19 @@ def _testAllExpanded(appController):
...
     _expandPrims(appController, ["/spheres", "/A", "/A/B", "/A/B/C"])
     initialExpandedPrims = _getExpandedPrims(appController)
+
+    # test to see if the display name of the capsule is showing
+    prim = appController._dataModel.stage.GetPrimAtPath("/Shapes/Pill")
+    item = appController._primToItemMap.get(prim)
+    assert item._nameData(QtCore.Qt.DisplayRole) == "CapsuleDisplayName"
+
+    # clear the a-variant and just select the b-variant
+    _selectVariant(appController, 0, FIRST_VARIANT)
```
spiffmon: It's confusing to see a zero, here, and since no such variantSet was being set previously, makes me just a tad concerned that the clause below your new code is getting its initial conditions whisked out from under it?

Can we add a tuple up top like we do for the other VariantSets so it's more consistent here?

pxr/usdImaging/usdviewq/utils.h
erslavin: yep, sorry I thought those were already removed, I'll take another pass

pxr/usdImaging/usdviewq/mainWindowUI.ui
erslavin: I probably did - I likely don't have the qdesigner installed but I'll take care of it.

pxr/usdImaging/usdviewq/utils.h
erslavin: ah I have a bunch of changes I never pushed that's why, let me integrate the latest comments and I'll get those over

erslavin:
> @erslavin , can you please migrate all of the "DisplayName"-related methods on `UsdProperty` up to `UsdObject`, now that it is applicable to all object types?

Methods are now moved to `UsdObject` to apply to both prims and properties.  Also adjusted tests to use both generic metadata methods and specific display name methods

pxr/usd/usd/testenv/testUsdSchemaRegistry/resources/schema.usda
erslavin: Removed

pxr/usd/usd/testenv/testUsdMetadata.py
erslavin: Adjusted this test to use both generic methods and specific methods for `displayName`

pxr/usdImaging/usdviewq/utils.h
erslavin: Removed

pxr/usdImaging/usdviewq/utils.cpp
erslavin: Removed

pxr/usdImaging/usdviewq/utils.cpp
erslavin: Adjusted to use specific `displayName` methods

pxr/usdImaging/usdviewq/wrapUtils.cpp
erslavin: Removed

pxr/usdImaging/usdviewq/primViewItem.py
erslavin: Modified to conform to this behavior

pxr/usdImaging/usdviewq/mainWindowUI.ui
erslavin: This one still needs to be addressed

pxr/usdImaging/usdviewq/appController.py
erslavin: Adjusted behavior here, but one note - since we defaulted the show prim display name option to true, a couple of tests had to be modified because they were no longer "intuitive".  Specifically, the `usdview` tests for property search and prim tree view expansion needed to have the `displayName` metadata added to the test files so that the appropriate prim was found because the new behavior no longer allows it to find those using the prim name as a backup if `displayName` isn't authored.  After working with it a little, I am leaning toward the default behavior for `usdview` to show prim names rather than prim display names (simply due to the fact that none of the usd files currently out there will have `displayName` authored, so they might be surprised when they search and don't find anything)

pxr/usdImaging/usdviewq/appController.py
erslavin: Adjusted

pxr/usdImaging/bin/testusdview/testenv/testUsdviewPrimTreeExpansion/testUsdviewPrimTreeExpansion.py
erslavin: Added a tuple to represent the empty combo box selection.  Note that this isn't a variant - it's using the combo box to clear the variant selection for the `a_variant` (hence the zero to represent the user clearing the combo box selection)

pxr/usdImaging/usdviewq/appController.py
spiffmon: Sorry, I think I'm misunderstanding your concern, @erslavin ?  When "Show Display Names" is on, then search should consult displayName, but **only** for prims on which it is authored, i.e. we should just match based on what you see in the browser.  I don't think that will be surprising, and I would still advocate for having the default be "on" because authoring displayNAme and _not_ seeing it reflected in the browser is surprising, I'd argue (acknowledging again that's already the boat we're in with the property browser).

erslavin: ok that was my misunderstanding, let me update again thanks!

erslavin: Ok hopefully we are in agreement now.

pxr/usdImaging/usdviewq/mainWindowUI.ui
erslavin: Saved from Qt Designer

pxr/usdImaging/bin/testusdview/testenv/testUsdviewPrimSearch/testUsdviewPrimSearch.py
```diff
@@ -48,6 +48,46 @@ def _search(appController, searchTerm, expectedItems):
...
+    # won't produce the prim name as a result
+    _search(appController, 'g', [])
+
+    # On an invalid search, the results will be
+    # equivalent to what was previously there (which in
+    # this case is nothing...)
+    _search(appController, 'xxx', [])
+
+    # Do a regex based search - again, this is
+    # not looking at prim names here not display names
```
spiffmon: Wonderful tests and deeply appreciative of the great descriptions for the test cases!
Minor typo, here: double negative... I think the first "not" is incorrect?

pxr/usdImaging/bin/testusdview/testenv/testUsdviewPrimTreeExpansion/testUsdviewPrimTreeExpansion.py
```diff
@@ -24,6 +24,9 @@
 #
 
 # positions and names of our variants
+# special case entry for 0, which represents a non-selection
+# in the variant selection combo box
+EMPTY = (0, '')
```
spiffmon, line 29: 🙏

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2185,16 +2188,34 @@ def _findPrims(self, pattern, useRegex=True):
...
-                                             self._displayPredicate)
-                   if isMatch(prim.GetName())]
+        matches = []
+        if self._dataModel.viewSettings.showPrimDisplayNames:
+            for prim in Usd.PrimRange.Stage(self._dataModel.stage, self._displayPredicate):
+                if prim.HasAuthoredDisplayName():
+                    if isMatch(prim.GetDisplayName()):
+                        matches.append(prim.GetPath())
+                elif isMatch(prim.GetName()):
+                    matches.append(prim.GetPath())
```
spiffmon: What would you think about enhancing the `isMatch` lambda to take the prim and `useDisplayNames` as input ?  I think it would simplify the logic here, and reduce/eliminate code duplication with handling prototypes, below?

spiffmon (commented):
Thanks, so much, @erslavin - just a couple last things for you.

pxr/usdImaging/usdviewq/appController.py
erslavin: This block of code already there should do it I believe, specifically the `self._primSearchResults = []` portion:
```
# This appears to be "reasonably" performant in normal sized pose caches.
    # If it turns out to be too slow, or if we want to do a better job of
    # preserving the view the user currently has, we could look into ways of
    # reconstructing just the prim tree under the "changed" prim(s).  The
    # (far and away) faster solution would be to implement our own TreeView
    # and model in C++.
    def _resetPrimView(self, restoreSelection=True):
        expandedPrims = self._getExpandedPrimViewPrims()

        startingDepth = 3
        with self._makeTimer("reset Prim Browser to depth %d" %
                             startingDepth), BusyContext():
            self._computeDisplayPredicate()
            with self._primViewSelectionBlocker:
                self._ui.primView.setUpdatesEnabled(False)
                self._ui.primView.clear()
                self._primToItemMap.clear()
                self._itemsToPush = []
                # force new search since we are blowing away the primViewItems
                # that may be cached in _primSearchResults
                self._primSearchResults = []
                self._populateRoots()
                # it's confusing to see timing for expand followed by reset with
                # the times being similar (esp when they are large)
                if not expandedPrims:
                    self._expandToDepth(startingDepth, suppressTiming=True)

                if restoreSelection:
                    self._refreshPrimViewSelection(expandedPrims)
                self._ui.primView.setUpdatesEnabled(True)
            self._refreshCameraListAndMenu(preserveCurrCamera = True)
```

pxr/usdImaging/usdviewq/appController.py
erslavin: I refactored the code into a true function (that still used the same lambdas underneath).  Trying to fit it all into a single lambda expression made it a little unreadable.

pxr/usdImaging/bin/testusdview/testenv/testUsdviewPrimSearch/testUsdviewPrimSearch.py
erslavin: Fixed typo issue

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2172,29 +2175,66 @@ def setFrameField(self, frame):
...
+        the match is always performed against the prim's name.
+
+        Args:
+            pattern (str): The pattern to use to match the name.  Pattern
+                           is either a sequence of characters or a regex
+                           expression.  If it is a regex expression, the
+                           isRegex parameter should be set to True.
+            isRegex (bool): True if the given pattern is a regex expression
+                            or False if just a sequence of characters.
+            prim (object): A python facing UsdPrim object on whos properties
```
spiffmon: whos -> whose
(only noting this because of more important note below)

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2172,29 +2175,66 @@ def setFrameField(self, frame):
...
+            matchLambda = re.compile(pattern, re.IGNORECASE).search
+        else:
+            pattern = pattern.lower()
+            matchLambda = lambda x: pattern in x.lower()
+
+        if useDisplayName:
+            if prim.HasAuthoredDisplayName():
+                return matchLambda(prim.GetDisplayName())
+            else:
+                return matchLambda(prim.GetName())
```
spiffmon: Seeing the logic so nicely laid out here reminds me that to avoid angering the python performance gods on this set of tests that could run millions of times in many scenes (since we usually process all instance proxies, as well!), we'd likely be considerably better off _just_ calling `GetDisplayName()` up front and comparing switching on a non-empty result, rather than making the extra `HasAuthoredDisplayName()` API call (which, internally, does _just_ about the same amount of work as `GetDisplayName()`, incidentally).  If you use the `--timing` usdview option, I believe that each search will print out how long it takes, so if you're curious, you could do that profiling.  I might be worrying needlessly here, but this kind of thing is exactly why we created the PrimInfo structure/dance to service PrimViewItems, to substantial benefit.

spiffmon (commented):
Just one quick performance note, @erslavin , otherwise this looks fantastic!

pxr/usdImaging/usdviewq/appController.py
erslavin: Fixed typo

pxr/usdImaging/usdviewq/appController.py
erslavin: You are correct - running `usdview` with the `--timing` option does show a slight decrease in search speed if we get the display name and compare to `None` / `""` vs. checking for authored first.  These are small scenes, but this slight difference will add up in larger ones.  Made the change.

spiffmon (commented):
Corey on the USD team noted a few things that I missed - passing them on to you, @erslavin

pxr/usd/usd/object.h
```diff
@@ -626,6 +626,30 @@ class UsdObject {
     USD_API
     bool HasAuthoredDocumentation() const;
 
+    /// Return this property's display name (metadata).  This returns the
```
spiffmon: Here and below, appearances of "property" should now be "object"

pxr/usdImaging/usdviewq/primViewItem.py
```diff
@@ -256,6 +260,12 @@ def _nameData(self, role):
                 toolTip = 'Inactive ' + toolTip
             elif self.isInstance:
                 toolTip = 'Instanced ' + toolTip
+        
+            # tooltip should always show both name and display name
+            toolTip = toolTip + "<br>Name: " + self.name
+            if self.displayName != "":
```
spiffmon: simply `if displayName:` is simpler and should yield the same results?

pxr/usdImaging/usdviewq/utils.cpp
```diff
@@ -104,6 +104,15 @@ UsdviewqUtils::PrimInfo::PrimInfo(const UsdPrim &prim, const UsdTimeCode time)
...
     typeName = prim.GetTypeName().GetString();
+
+    if (prim.HasAuthoredDisplayName())
+    {
+        displayName = prim.GetDisplayName();
+    }
+    else
+    {
+        displayName = "";
+    }
```
spiffmon: The HasAuthoredDisplayName check is redundant here; the fallback for GetDisplayName should be the empty string.

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2172,29 +2175,74 @@ def setFrameField(self, frame):
 
     # Prim/Attribute search functionality =====================================
 
+    def _isMatch(self, pattern, isRegex, prim, useDisplayName):
```
spiffmon, line 2178: A _suggestion_ that if `_isMatch` were defined inside of `_findPrims`, then it would only need a `prim` parameter.  Perhaps you were anticipating other uses outside of _findPrims, and there's also something to be said for having documented parameters, but thought we'd mention it for your consideration.

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2172,29 +2175,74 @@ def setFrameField(self, frame):
...
+        if useDisplayName:
+            # typically we would check prim.HasAuthoredDisplayName()
+            # rather than getting the display name and checking
+            # against the empty string, but HasAuthoredDisplayName
+            # does about the same amount of work of GetDisplayName
+            # so we'd be paying twice the price for each prim
+            # search, which on large scenes would be a big performance
+            # hit, so we do it this way instead
+            displayName = prim.GetDisplayName()
+            if displayName is not None and displayName != "":
```
spiffmon: why isn't this just: `if displayName:` ?

spiffmon (commented):
OK, that's all - thanks, Ed!

pxr/usdImaging/usdviewq/primViewItem.py
erslavin: I tend to prefer explicit checks to implicit ones for the reader but we can change it to `if self.displayName:` if you prefer.

pxr/usdImaging/usdviewq/utils.cpp
erslavin: I ran into an issue early on where it was actually returning `None` when it was not authored.  I can test this again, but there were some tests failing prior to one of the initial commits because of this.

pxr/usdImaging/usdviewq/appController.py
erslavin: I would leave this to your preference, as I cannot attest to whether it might be used elsewhere.  I can move it if you prefer.

pxr/usdImaging/usdviewq/appController.py
erslavin: see comment above about encountering `None` being returned when it wasn't authored at one point

pxr/usdImaging/usdviewq/primViewItem.py
spiffmon: I don't know if you'd consider it python programming "style", but my assessment of coding practice here, which I think is what Corey was getting at, is that needing to be aware of whether any particular python function is going to return None or an empty string or an empty list is a source of fragility, but they all evaluate to `False`, so doing the simple check `if displayName:` leads to more resilient code?  I am not a python expert or style-proselytizer,  but I think this relates to all of the other things you call out below as having encountered...

pxr/usdImaging/usdviewq/appController.py
spiffmon: Let's leave it as it is, for now, then.

pxr/usdImaging/usdviewq/primViewItem.py
erslavin: this is one of the things I find extremely irritating about Python.  The docs state:
```
In the context of Boolean operations, and also when expressions are used by control flow statements, the following values are interpreted as false: False, None, numeric zero of all types, and empty strings and containers (including strings, tuples, lists, dictionaries, sets and frozensets). All other values are interpreted as true. User-defined objects can customize their truth value by providing a __bool__() method.
```
While it may be more "pythonic" to do something like `if x:`, the result is not what I would expect intuitively.  I will change this (and the one below in `appController.py` ) such that it reads `if displayName:`, but without inherently being aware of this statement in the docs it reads awkwardly for me.

pxr/usdImaging/usdviewq/appController.py
erslavin: see above, will change to `if displayName:`

pxr/usdImaging/usdviewq/utils.cpp
erslavin: Sorry, noticed this was the C++ code, in this case it is redundant.  Will change.

pxr/usd/usd/object.h
erslavin: Modified comments to read `object's` instead of `property's`

pxr/usdImaging/usdviewq/primViewItem.py
erslavin: Changed to `if displayName:`

pxr/usdImaging/usdviewq/utils.cpp
erslavin: Changed to `displayName = prim.GetDisplayName();`

pxr/usdImaging/usdviewq/appController.py
erslavin: Changed to `if displayName:`

spiffmon:
We won't be doing any more dev-pushes to github this year, but just wanted to let you know this one _did_ get merged internally in  time for 23.02, @erslavin - thanks for your patience!

asluk:
> We won't be doing any more dev-pushes to github this year, but just wanted to let you know this one _did_ get merged internally in time for 23.02, @erslavin - thanks for your patience!

Thanks @spiffmon and @erslavin !

PR #2057, opened 2022-10-06
APPLE: Fix for Alembic HDF5 on universal builds

creijon:
### Description of Change(s)

HDF5 needs to set the size of a long double based on architecture, but the cmake logic doesn't support multiple targets.  For x64 and arm64 this is 16 bytes so we can force it.

### Fixes Issue(s)
- HDF5 not building when `--build-target universal` defined.


tallytalwar:
Filed as internal issue #USD-7683

PR #2058, opened 2022-10-06
Fix bug in HdExtComputationUtils::_GenerateDependencyMap() with multiple input computations

cameronwhite:
### Description of Change(s)

The last element in the queue was used instead of the front element that is subsequently popped from the queue, which caused some input computations to be skipped if the queue ever has multiple entries.

Extended testHdExtComputationUtils to add code coverage for this by using multiple input computations (the skinning computation only has a single input aggregator computation, so it didn't encounter this bug).


tallytalwar:
Filed as internal issue #USD-7684

PR #2060, opened 2022-10-11
[gf] make vec*.h use GfSqrt instead of sqrt

pmolodo:
### Description of Change(s)

with gcc, at least, this avoids using a double for vec3f::GetLength, which avoids a compiler warning when using -Wconversion

### Fixes Issue(s)
- warning when using -Wconversion flag on gcc

### DIFF
--- a/pxr/base/gf/vec.template.h
+++ b/pxr/base/gf/vec.template.h
@@ -252,8 +252,7 @@ public:
 {% if IS_FLOATING_POINT(SCL) %}
     /// Length
     {{ SCL }} GetLength() const {
-        // TODO should use GfSqrt.
-        return sqrt(GetLengthSq());
+        return GfSqrt(GetLengthSq());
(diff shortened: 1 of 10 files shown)


tallytalwar:
Filed as internal issue #USD-7707

pxr/base/gf/vec.template.h
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2016 Pixar
```
sunyab: Our policy has been not to update copyright years when making changes to existing files. Could I ask you to revert this bit and regenerate the code? Thanks!

sunyab (changes requested):

pxr/base/gf/vec.template.h
pmolodo: Done!

PR #2064, opened 2022-10-14
[vt] fix for crash when creating VtArray from large python buffer

ablev-nv:
### Description of Change(s)

The third parameter of `std::accumulate` will determine the return type for `std::accumulate`. If the type is not explicit, then the compiler will infer it. More info here:

https://en.cppreference.com/w/cpp/algorithm/accumulate

The `Vt_ArrayFromBuffer` helper uses `std::accumulate` to determine the number of items to be created in the `VtArray`. Because the third parameter given here is just `1`, the compiler assumes this is an `int` and uses that for the return type from `std::accumulate`. This is fine for most sane uses of `VtArray` that are initialized from a Python buffer with less than 2 billion-ish items. But if your Python buffer has more than maxint items, then the `int` return type overflows, and the process usually crashes.

This change makes the third parameter explicitly `Py_ssize_t`, which is the type of the other inputs to `std::accumulate` and was likely the intended output type all along.

### Fixes Issue(s)

Crash or incorrect behavior when creating a `VtArray` from a python buffer with more than maxint elements.

### DIFF
--- a/pxr/base/vt/arrayPyBuffer.cpp
+++ b/pxr/base/vt/arrayPyBuffer.cpp
@@ -416,12 +416,12 @@ Vt_ArrayFromBuffer(TfPyObjWrapper const &obj,
     // Check that the number of elements matches.
     auto multiply = [](Py_ssize_t x, Py_ssize_t y) { return x * y; };
     auto numItems = std::accumulate(
-        view.shape, view.shape + view.ndim, 1, multiply);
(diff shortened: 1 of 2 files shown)


gitamohr:
Nice find and fix, thanks Alan!

tallytalwar:
Filed as internal issue #USD-7719

PR #2068, opened 2022-10-20
Make timing tests compare to upper / lower bounds

pmolodo:
### Description of Change(s)

When running tests in parallel, we get intermittent failures of
timing related tests due to the elapsed time being outstide of the
allowed range.

However, this could just be due to thread scheduling issues, as
an indeterminate amount of time could pass between the recorded trace
and the standard library call which was used as "ground truth".

Now, I bracket the trace records on both sides with standard library
timing calls, to provide expected upper and lower bounds, which should
be more forgiving of randomness from thread scheduling.

A follow up to changes in:

cadce6fff973a97d2c83f1a8609a3baa85720b7b - Avoid hard-coded comparisons for timing tests

### Fixes Issue(s)
- Intermittent failures of testTrace and TfStopwatch tests when running tests with many processes

### DIFF
--- a/pxr/base/tf/testenv/stopwatch.cpp
+++ b/pxr/base/tf/testenv/stopwatch.cpp
@@ -44,6 +44,13 @@ IsClose(double a, double b, double epsilon=1e-3)
 }
 
 static bool
+IsCloselyBounded(double value, double lower, double upper, double epsilon=1e-3)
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-7727

PR #2071, opened 2022-10-25
APPLE: Enable ICB on Intel

slingthor:
### Description of Change(s)
Enable Indirect draw command buffers on Intel as they are working with all latest changes to dev and run through Apple's tests now


tallytalwar:
Filed as internal issue #USD-7729

PR #2075, opened 2022-10-29
APPLE: Explicitly enable fast math and O3 optimisation for release builds for Metal on Mac

slingthor:
### Description of Change(s)
Fast math and O3 enablement on release for Metal on Macs yields in significant performance improvements. Fast math was very likely implicitly on for Metal, but O3 on Shaders will proof useful and we explicitly state Os for the host side.


spiffmon:
Hi @slingthor - thanks for keeping an eye out for performance!  I have some concerns with this change, though.

Firstly, I'm pretty sure we have been bitten by the fastmath subnormal virus before, and would not want to make USD a vector for it on any platform (unless you have determined that on Apple Silicon and clag, it is not actually a virus?  If so, that would be great to see in the notes, as [fastmath is known to be problematic](https://simonbyrne.github.io/notes/fastmath/).  I am also quite shocked that tests pass with `-ffinite-math-only` because USD makes explicit use of NaNs in UsdTimeCode, and performs isnan checks several other places in the code.  More broadly, though, it might be better to test individual sub-features of the fastmath umbrella to see which make the biggest difference so that the tradeoffs can be discussed?  As USD gets incorporated into more disciplines, the accuracy tradeoffs of some of the optimizations may be a bigger deal than others.  It would be great if you can connect with @AlexSchwank about perf test harness so this could be rigorously determined and documented?

Secondly, are your use of `-O3` and `-Os` antagonistic?  Given that `-O3` can increase object code size and greatly increase build times (already a complaint with USD), it again would be good to document what this buys us in performance over `-O2` (and/or even `-Os` by itself?), and if you do continue with math optimizations, those should be split out into a separate change from the `-O` changes.

Finally, while I understand clang and gcc are going to yield different gains, based on your findings it seems like it would be worth having a discussion on usd-interest about enabling the final set of optimizations on all platforms?

slingthor:
Hey Spiff! So these turn on fast math and O3 for Metal only, not the host side code -> Host side code is being explicitly set to Os and only for release builds and fastmath not turned on for host side ( I'll add a comment for that - I'm not sure if there's a single google result about the usage of this attribute in CMAKE, had to use https://cmake.org/cmake/help/latest/prop_tgt/XCODE_ATTRIBUTE_an-attribute.html and then the name of the attribute from the Xcode project file in text format to find how to set it ).

The build size increase would be for the size of the compiled metal shaders which are made on runtime.
We have been using fast math extensively for Metal here at Apple and also on USD builds and think we don't see issues with it for the MSL code.

To make the case though, I'll go and measure the improvements from making each of these changes and come back to you with the results!

I am aware that fast math does make a big difference (I noticed that no existing compiler performance settings were coming through when using generator Xcode, but it probably was on already though)

I'll also verify the subnormal issue, and also especially check if it's affecting any of our numerical approximation results in shaders

 I think the tests were passing just fine because this is only affecting the MSL shaders and related code - I believe O3 is quite reasonable on the shaders as they are built on runtime and assembly size then is not a part of the shipped application and perhaps we don't want to spend loading time resources on optimising that, which -OS, Fast and OZ would for example do, and some of them would be too risky as well.

Most likely fast math was already on as well for Metal, as the default for Metal is having it on, but I simply felt a bit concerned as CMAKE generated Xcode projects have no settings field for it, unlike Xcode generated new projects.

spiffmon:
Ah, thanks for that info, @slingthor - I did wonder a bit about the `_MTL_` in those settings, and if the shader compilation code is truly sandboxed, it seems like even the subnormal setting wouldn't leak out?

I do still wonder about `-O2` vs `-Os` for all release builds!

slingthor:
@spiffmon Yes agree! The subnormal setting should not leak out to the user's file system or give the user the ability to do anything with the OS!
I've removed the -Os setting for the host side on release as a part of the PR, agree that it's too opinionated and user specific to include like that.

Perhaps it's too user specific what the optimisation level is for including it as a parameter, but it comes to mind a parameter to add in the future, something like "--optimisation-level"

sunyab:
Filed as internal issue #USD-7738

PR #2079, opened 2022-11-01
APPLE: Fix OpenImageIO and OpenEXR debug libraries not found

slingthor:
### Description of Change(s)
Modify build script to indicate to CMAKE we are building a debug build
If we are doing a debug build and we are doing a Mac build, append a postfix to correctly find the debug library files.

### Fixes Issue(s)
Debug OpenImageIO / OpenEXR builds cannot be made on Apple platforms

### DIFF
--- a/cmake/modules/FindOpenEXR.cmake
+++ b/cmake/modules/FindOpenEXR.cmake
@@ -68,10 +68,16 @@ foreach(OPENEXR_LIB
 
     # OpenEXR libraries may be suffixed with the version number, so we search
     # using both versioned and unversioned names.
+    set(DEBUG_POSTFIX )
+    if(DEFINED PXR_USE_DEBUG_BUILD)
+        if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND ${PXR_USE_DEBUG_BUILD} MATCHES ON)
+            set(DEBUG_POSTFIX _d)
+        endif()
+    endif()
     find_library(OPENEXR_${OPENEXR_LIB}_LIBRARY
         NAMES
-            ${OPENEXR_LIB}-${OPENEXR_MAJOR_VERSION}_${OPENEXR_MINOR_VERSION}
-            ${OPENEXR_LIB}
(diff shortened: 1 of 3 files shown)


slingthor:
Attempted to re-trigger the tests (initial run was a fault on my part), but after re-triggering they seem to be failing on startup. Is there any way to retry this?

meshula:
@slingthor I gave it a kick

sunyab:
Filed as internal issue #USD-7742

cmake/modules/FindOpenEXR.cmake
```diff
@@ -68,10 +68,16 @@ foreach(OPENEXR_LIB
 
     # OpenEXR libraries may be suffixed with the version number, so we search
     # using both versioned and unversioned names.
+    set(DEBUG_POSTFIX )
+    if(DEFINED PXR_USE_DEBUG_BUILD)
+        if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND ${PXR_USE_DEBUG_BUILD} MATCHES ON)
+            set(DEBUG_POSTFIX _d)
+        endif()
+    endif()
```
seando-adsk, line 76: We are building a full debug and also building OpenEXR (and Alembic) statically. Locally I've added extra NAMES lines for debug and static. But I like your approach. When building OpenExr statically the libs are appended with "_s" and "_s_d". Could that be added here as well?

meshula:
Per this issue https://github.com/PixarAnimationStudios/USD/issues/2231 the problem may extend into OpenVdb as well, in terms of needing to patch the OpenVdb find script, and also checking with the OpenVdb project to discover if the problem is fixed in a later revision.

meshula:
Please note @Ben1138's feedback from #2231, quoted below. This patch is necessary for Windows, and probably Linux as well for debug builds.

Patching FindOpenEXR.cmake and build_usd.py as described in PR https://github.com/PixarAnimationStudios/USD/pull/2079 solved the issue of USD not finding OpenEXR. About OpenVDB, I just left it out (--no-openvdb). Now I got a successfull debug build!

When I patched FindOpenEXR.cmake I left out the if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin" part, since I'm on Windows.
I didn't patch FindOpenImageIO.cmake though, since it built fine for me.

Issue #1438, opened 2021-01-29
Support for python3 command/shebang


bdeda:
https://github.com/PixarAnimationStudios/USD/blob/7ad0e51d4d0d37a090c18c8a8f6e57509d62141c/cmake/macros/shebang.py#L44

In the shebang.py code, it uses "python" as the command for both python 2 and python 3 environments, which causes problems where the site uses the "python3" command name. Locally I have changed this for our builds to check sys.version_info.major and use "python3" instead of "python" when building in a python 3 environment. Can we make a similar change in the Pixar code, or at least make this more configurable via -DPYTHON_SHEBANG and -DPYTHON_COMAND_NAME cmake variables?

https://github.com/PixarAnimationStudios/USD/blob/7ad0e51d4d0d37a090c18c8a8f6e57509d62141c/cmake/macros/shebang.py#L50
We have a similar issue with line 50, where we do not want to write the command used during build to be used in the shebang lines of the command files. Making this more configurable would be ideal.

jilliene:
Filed as internal issue #USD-6554

c64kernal:
Hi @bdeda -- that sounds good to us, making this more configurable seems very reasonable. If you have a PR for us to consider, we'd be happy to have a look.

rstelzleni:
Which scripts wind up with python in the `#!` line when building with python 3? I haven't ever run into that with python 3 builds, but I might not be looking at the right scripts. `usdcat` and `usdview` have the full path to the python interpreter used to build USD. I can definitely see that not being useful though.

bdeda:
In our build system, we have wrapper calls to set up env for executables. That means we have a simple wrapper call to python executables as a shim to use during build, and not meant to be embedded into usdcat and usdview in the bin dir. Also, the shebang.py script does not differentiate between "python3" and "python" as the command names. I'm not sure what other studios are doing, but we are going with "python3" as our interpreter name, and want that written into the shebang lines.

tanant:
@bdeda not sure if you're still eyeballing this but I did up a PR that would implement some of your old PR's suggestions

Issue #2082, opened 2022-11-04
Windows cmake install error due to backslash in Python3_INCLUDE_DIRS


pmolodo:
### Description of Issue
When running cmake install on windows, we get a cmake syntax error:

```
CMake Error at cmake/macros/Public.cmake:367 (pxr_library):
  Syntax error in cmake code at

    W:/f651207570849f20/USD/cmake/macros/Public.cmake:367

  when parsing string

    LIBRARIES;tf;gf;sdf;usd;usdGeom;W:/f651207570849f20/_install/py37.windows-x86_64.usdview.release/lib/draco.lib;INCLUDE_DIRS;W:/f651207570849f20/_install/py37.windows-x86_64.usdview.release/include/boost-1_70;W:/f651207570849f20/_install/py37.windows-x86_64.usdview.release/include;W:\f651207570849f20\_build\target-deps\python\Include;CPPFILES;attributeDescriptor.cpp;attributeFactory.cpp;exportTranslator.cpp;fileFormat.cpp;importTranslator.cpp;writer.cpp;PYTHON_CPPFILES;moduleDeps.cpp;PYMODULE_CPPFILES;module.cpp;wrapDraco.cpp;PYMODULE_FILES;__init__.py;RESOURCE_FILES;plugInfo.json;DOXYGEN_FILES;overview.dox

  Invalid character escape '\f'.
```

The change seems to be related to dbabdaf1460cc6e765b17bf776b54644462749ee, which altered the way that python is found in cmake.  It seems that now windows paths are not properly escaped / do not have backslashes converted to forward slashes.

### Steps to Reproduce

1. Run `usd_build.py` with the `--build-python-info` flag on windows, where the paths have backslashes.  (Note - note actually sure if using `--build-python-info` is required to reproduce, but that's we we did.)

### System Information (OS, Hardware)
Windows 10

### Package Versions
USD - dev branch - d27b1be4bc357c7c025b2e10339605282d8cadc1

### Build Flags
`--build-python-info` - ie, we used:

```
--build-python-info W:\f651207570849f20\_build\target-deps\python\python.exe W:\f651207570849f20\_build\target-deps\python\Include W:\f651207570849f20\_build\target-deps\python\libs\python37.lib 3.7
```

sunyab:
@pmolodo We're seeing this on our Azure builds too. I'm not sure why we didn't see this in our local Windows tests builds, though. What version of CMake are you running?

sunyab:
[...]

PR #2083, opened 2022-11-04
USD-6554 - add in shebang and cmd controls

tanant:
### Description of Change(s)
This change defines a new variable - `PYTHON_COMMAND_NAME` - which can be used to specifically target the windows .cmd wrappers, and exposes `PXR_PYTHON_SHEBANG` to allow build time override (e.g. if you're using a wrapped python instance to build, and that's meant to be build-time only or similar.)

Both will fall back to `${PYTHON_EXECUTABLE}` matching current behaviour.

This follows on from the discussion in #1438, I've just extend the closed/unmerged PR (#1445) so it's an explicit selection rather than guessing on the interpreter name (i.e. if you're doing this, you're in control of your build environment so you probably know what to specify). 

Split out to a new file instead of creating a complex arg-counting-flag-whatevering-signature for `shebang.py`

**Note:** I wasn't 100% sure what license to apply for the new `wincmd.py` file so have tagged it with the unmodified Apache 2.0 lisc. 

### Fixes Issue(s)
#1438

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,11 +24,21 @@ include(Packages)
 
 # This has to be defined after Packages is included, because it relies on the
 # discovered path to the python executable.
-set(PXR_PYTHON_SHEBANG "${PYTHON_EXECUTABLE}" 
-    CACHE 
-    STRING
-    "Replacement path for Python #! line."
-)
+if (NOT PXR_PYTHON_SHEBANG)
+    set(PXR_PYTHON_SHEBANG "${PYTHON_EXECUTABLE}"
+        CACHE
+        STRING
+        "Replacement path for Python #! line."
+    )
+endif()
+
+if (NOT PYTHON_COMMAND_NAME)
+    set(PYTHON_COMMAND_NAME "${PYTHON_EXECUTABLE}"
+        CACHE
+        STRING
+        "Python command to launch usd tools (e.g. usdview) with."
+    )
+endif()
 
 # CXXDefaults will set a variety of variables for the project.
 # Consume them here. This is an effort to keep the most common
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ if (NOT PXR_PYTHON_SHEBANG)
 endif()
 
 if (NOT PYTHON_COMMAND_NAME)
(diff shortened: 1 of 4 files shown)


tanant:
Wasn't sure the reason why the mentioned PR was closed, but saw the underlying issue was still open so have quickly drafted this up. @sunyab I think this would address your last comment there around the problem of defaulting to python3 on windows due as there's no script-level auto-guessing, it's all explicitly controlled by the user when building.

tanant:
Wanting to check - those failing CI builds, a lot (all?) of the failures look like missing deps on the build host - is that something resulting from my PR? 

- linux build, runs fine 
- macOS build fails on missing `uic` 
- windows build is complaining about a `\h` escape, which i think is from the mixed-backslash string that cmake is being passed (`...RS;D:/USDinst/include/boost-1_70;C:\hostedtoolcache\windows\Python\3.7.9\x64\Include;CPPFILES;fi...`)
  -  EDIT: looks like will be fixed at same time #2082 is done
- failing PyPI packaging on linux is complaining about not being able to find python3

sunyab:
Hi @tanant, we're seeing these CI failures across our builds so I don't believe this is anything specific to your change. We're looking into fixes, hopefully we can re-run the builds for this PR once we get that in. Thanks!

sunyab:
Filed as internal issue #USD-7748

CMakeLists.txt
```diff
@@ -24,11 +24,21 @@ include(Packages)
...
+if (NOT PXR_PYTHON_SHEBANG)
+    set(PXR_PYTHON_SHEBANG "${PYTHON_EXECUTABLE}"
+        CACHE
+        STRING
+        "Replacement path for Python #! line."
+    )
+endif()
+
+if (NOT PYTHON_COMMAND_NAME)
+    set(PYTHON_COMMAND_NAME "${PYTHON_EXECUTABLE}"
```
tanant: Ah. Slight nuance/pickup by @seando-adsk - this actually _isn't_ the existing behaviour. Existing behaviour would be to use the literal string 'python'. This has the effect of putting in a hardpath to the python executable used to build this. 

TL;DR unexpected/unintended behaviour change. I'll add another commit to this branch to change that to better match 1:1

tanant: (aaaaand done)

PR #2085, opened 2022-11-09
APPLE: Fixes for small leaks

creijon:
### Description of Change(s)

Need to explicitly release the compiler options and descriptors.
Found leaks in the `MTLSamplerDescriptor` since they needed release too.

### Fixes Issue(s)
- Small memory leaks


sunyab:
Filed as internal issue #USD-7770

PR #2086, opened 2022-11-10
USD Crate format Documentation

dgovil:
### Description of Change(s)

This PR adds an initial pass of the Crate format specification. Where possible it tries to describe things at a byte or bit level.
This is a fairly large undertaking to reverse out the code into natural language, so requires extra scrutiny.

This adds a few features:

* Custom sphinx directive to read marked comments within the C++ source code. This allows the majority of the documentation to live beside the implementation. It also allows for ordering to change as needed which is handy. 
This enables Alex's request to have something akin to literate programming approach within the code.
* Custom sphinx directive to generate byte layouts, inspired by the [PKZip documentation.](https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html)

Sphinx was chosen to include this documentation because:

* The other specifications are in Sphinx already
* Doxygen has limited support for custom documentation extensions

The document includes some warnings for now. I think it would be best to leave the warnings in place till we are certain that the descriptions are accurate. Beyond that, it might be good to take a leaf out of other project specifications books where the official implementation is the gold truth and warnings explicitly state that.

Please note that this PR also includes https://github.com/PixarAnimationStudios/USD/pull/2084 because it's easier on my eyes than the default RTD theme, and provides better visual sectioning differences as well, which are necessary for how deep the nesting levels get.

A PDF preview is attached but the PDF formatting gets messed up so it's still recommended to sphinx-build the docs and view it that way. 
[USDC File Format Specification — Universal Scene Description 22.11 documentation.pdf](https://github.com/PixarAnimationStudios/USD/files/9984656/USDC.File.Format.Specification.Universal.Scene.Description.22.11.documentation.pdf)

### DIFF
--- a/docs/spec_usdc.rst
+++ b/docs/spec_usdc.rst
@@ -386,6 +386,8 @@ Bool
 
 The **Bool** type is a basic Binary boolean that can be checked by taking any non-zero number as True, and 0 as False.
 
+Booleans in an array are stored as an array of bytes (8 bits per entry).
+
 UChar
 ^^^^^
 
--- a/docs/spec_usdc.rst
+++ b/docs/spec_usdc.rst
@@ -1,0 +1,757 @@
+==============================
+USDC File Format Specification
+==============================
+
+.. include:: rolesAndUtils.rst
+.. include:: <isonum.txt>
+
+Copyright |copy| 2022, Pixar Animation Studios,  *version 0.9.0*
+
+.. contents:: :local:
+
+Introduction
+=============
+
+The **USD Crate** format is binary encoding of the USD scene graph.
+This document aims to document the layout of the this format.
+
+.. warning::
+    This is a Work in Progress documentation of the USD Crate format. It is not guaranteed to
+    be an exact description of the format right now. Contributions to improve accuracy are welcome.
+
+    If areas are unspecified, or if a discrepancy occurs, the implementation in the official USD library
+    should always take precedence as being the canonical implementation of the format.
+
+    It is **NOT** yet recommended to make your own implementation based on this document
+    , and we encourage developers to use the official implementation.
+
+    This document does not describe the composition elements of USD. Implementing this document
+    will only allow for reading a single USD layer.
+
+
+Versions
+---------
+
+This document only aims to document version `0.9.0` of the Crate format and higher.
+For older versions of the Crate format, please refer to the USD code implementations.
+
+These older crate files are exceedingly rare outside of Pixar and very early USD adopters.
+
+Version History
+^^^^^^^^^^^^^^^
+
+Even though the spec only starts with 0.9.0, it is useful to understand the high level history of versions
+
+
+* 0.9.0: Added support for the timecode and timecode[] value types.
+* 0.8.0: Added support for SdfPayloadListOp values and SdfPayload values with layer offsets.
+* 0.7.0: Array sizes written as 64 bit ints.
+* 0.6.0: Compressed (scalar) floating point arrays that are either all ints or can be represented efficiently with a lookup table.
+* 0.5.0: Compressed (u)int & (u)int64 arrays, arrays no longer store '1' rank.
+* 0.4.0: Compressed structural sections.
+* 0.3.0: (broken, unused)
+* 0.2.0: Added support for prepend and append fields of SdfListOp.
+* 0.1.0: Fixed structure layout issue encountered in Windows port.
+* 0.0.1: Initial release.
+
+
+Order of Reads
+--------------
+
+The Crate format is designed for minimal parsing on file load.
+
+To achieve this, the general structure of the file is set up in the preamble section below that must be read first.
+
+Value types are parsed on demand from there on out.
+
+Integers
+--------
+
+Integers are stored with the least significant bit first to allow for fast loading on `Little Endian <https://en.wikipedia.org/wiki/Endianness>`_ systems.
+
+Index
+-----
+
+Many fields reference into an another section to get their value. USD uses unsigned 32 bit integers as an index unless
+otherwise specified.
+
+
+Compression
+===========
+
+LZ4 Compression
+----------------
+
+Various section and data blocks use `LZ4 Compression <https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md>`_.
+
+USD vendors the LZ4 library with modifications and can be found `here <https://github.com/PixarAnimationStudios/USD/tree/release/pxr/base/tf/pxrLZ4>`_.
+It is currently using `version 1.9.2 <https://github.com/lz4/lz4/releases/tag/v1.9.2>`_ of the library.
+
+.. usdcppdoc:: pxr/base/tf/fastCompression.cpp lz4
+
+Integer Compression
+-------------------
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp compressed_int
+
+
+Preamble
+========
+
+The Crate format has a preamble that must be read prior to parsing the rest of the file.
+
+This contains all the structural information that is required to identify Prim specifications and
+their resulting attributes.
+
+
+Bootstrap
+---------
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp bootstrap
+
+Table of Contents
+-----------------
+
+The Table of Contents are found at the byte offset given by the bootstrap above.
+
+It consist of named sections , as documented below.
+
+Sections
+--------
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp section
+
+There are several sections with their own specific internal structures.
+
+Sections should be parsed in the order below since many sections depend on their predecessors.
+
+Tokens
+^^^^^^
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp tokens_section
+
+Strings
+^^^^^^^
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp strings_section
+
+Fields
+^^^^^^
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp fields_section
+
+Field Sets
+^^^^^^^^^^
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp field_sets_section
+
+Paths
+^^^^^
+
+.. usdcppdoc:: pxr/usd/usd/crateFile.cpp paths_section
+
+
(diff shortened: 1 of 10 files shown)


pxr/base/tf/fastCompression.cpp
```diff
@@ -101,6 +101,24 @@ TfFastCompression::CompressToBuffer(
     return compressed - origCompressed;
 }    
 
+/* usddoc lz4
+ * The first byte of any buffer passed to be decrypted stores the number of chunks used.
```
meshula:
```suggestion
 * The first byte of any buffer passed to be decompressed stores the number of chunks used.
```

pxr/usd/usd/crateFile.cpp
```diff
@@ -1861,6 +2014,18 @@ _ReadPossiblyCompressedArray(
 
 struct _CompressedIntsReader
 {
+    /* usddoc compressed_int
+     *  Compressed integers are stored as a contiguous , homogenous array of
```
meshula:
```suggestion
     *  Compressed integers are stored as a contiguous, homogenous array of
```

pxr/usd/usd/crateFile.cpp
```diff
@@ -3235,6 +3448,16 @@ CrateFile::_ReadStructuralSections(Reader reader, int64_t fileSize)
     if (m.IsClean()) _ReadSpecs(reader);
 }
 
+/* usddoc bootstrap
+ * Every USD file starts with a Bootstrap that tells us information to parse the file.
+ *
+ * Files must start with an identifier (PXR-USDC), the version number and the offset to the Table of Contents.
+ * There is additional reserved space at the end of the bootstrap for future use.
+ *
+ * Version numbers are stored as Major , Minor and Patch followed by unused bytes.
```
meshula:
```suggestion
 * Version numbers are stored as Major, Minor and Patch followed by unused bytes.
```
the table shows 8  bytes for version, how are major, minor, and patch, packed into those 8 bytes?

pxr/usd/usd/crateFile.h
```diff
@@ -79,13 +79,36 @@ template <class T> struct _IsBitwiseReadWrite;
...
-// Value in file representation.  Consists of a 2 bytes of type information
-// (type enum value, array bit, and inlined-value bit) and 6 bytes of data.
-// If possible, we attempt to store certain values directly in the local
-// data, such as ints, floats, enums, and special-case values of other types
-// (zero vectors, identity matrices, etc).  For values that aren't stored
-// inline, the 6 data bytes are the offset from the start of the file to the
-// value's location.
+/* usddoc value_rep
+ * Crate stores Value Representations as data blobs that are read on demand.
+ * This allows large USD scenes to be read incredibly quickly by deferring all
```
meshula:
```suggestion
 * This allows large USD scenes to be read as quickly as possible by deferring all
```
avoiding subjective language

pxr/usd/usd/crateFile.h
```diff
@@ -79,13 +79,36 @@ template <class T> struct _IsBitwiseReadWrite;
...
+ * data reads to the latest point possible.
+ *
+ * Value reps are stored as an unsigned 64 bit integer.
```
meshula:
```suggestion
 * Value representations are stored as an unsigned 64 bit integer.
```
prefer full words to aid full text indexing and search

pxr/usd/usd/crateFile.cpp
```diff
@@ -1936,6 +2107,21 @@ _ReadPossiblyCompressedArray(
...
+ * Compressed float arrays start with an unsigned 64 bit integer to represent their element count.
+ *
+ * Arrays that are 16 bytes or smaller are not compressed.
+ *
+ * Following this, a char is used to represent the array encoding scheme:
+ *
+ * * **i** for integer compression (follow the compressed integer array logic)
+ * * **t** for lookup tables (described below)
+ *
+ * The size of the lookup table (LUT) is a 32bit unsigned integer.
```
meshula:
```suggestion
 * The size of the lookup table (LUT) is a 32 bit unsigned integer.
```

pxr/usd/usd/crateFile.cpp
```diff
@@ -1179,16 +1213,58 @@ class CrateFile::_Reader : public _ReaderBase
...
+     * stored as contiguous key:value pairs.
+     */
     SdfVariantSelectionMap Read(SdfVariantSelectionMap *) {
         return ReadMap<SdfVariantSelectionMap>();
     }
+    /* usddoc layer_offset
+     * A layer offset represents a time offset and scale between layers.
+     * It is represented by two doubles.
+     *
+     * The first represents the time offset to be used , and the second the scale.
```
meshula:
```suggestion
     * The first represents the time offset to be used, and the second the scale.
```

pxr/usd/usd/crateFile.cpp
```diff
@@ -430,6 +430,20 @@ struct _PathItemHeader {
...
 struct _IsBitwiseReadWrite<_PathItemHeader> : std::true_type {};
 
+/* usddoc listop_header
+ * ListOps have an 8-byte bitmasked header that determine what type of operation
+ * is being run.
+ *
+ * In order, from least to most significant bit:
+ *
+ *  * **IsExplicit** : Removes all items and changes the list to be explicit
+ *  * **HasExcplicitItems** : Fills the list with the items included in this ListOp
```
meshula:
```suggestion
 *  * **HasExplicitItems** : Fills the list with the items included in this ListOp
```

meshula (commented):
Exciting! This review has my notes

not sure where the string in the pdf, "Integers are stored with the least significant bit", comes from, but should read "least  significant byte".

Can the "Certain types" "Internal to Pixar" be flagged in some way in the table that follows?

"For multidimensional type like Matrix2D".... The explanation and diagram that follows suggest the diagonal only is stored as three entries? Very confusing as one expects to find nine values, and one would want to know something like, "matrix rows are stored consecutively in memory such that they can be addressed  by the C syntax,  matrix[row][column]" Some note about whether these are mathematically column major matrices stored in C row major order is probably helpful.

In Base Types, Bool should note if it is stored as an 8-bit value where the values of 0 or 1 are valid (needs double checking, I'm not sure). Also, some notes on bool array ~ are they bytes, or bit packed (I haven't checked)

dgovil:
Thanks, addressed some of your notes.

With regards to some of the remaining notes:

* That specific Matrix diagram  you mentioned is for inlined types, which only seem to store the diagonal values. I do mention the row major nature of Matrix values in the Matrix section lower down, but I added an additional note up here about it for clarity since Inlined types show up first. Since there'd only be 6 bytes left, only a matrix2D would fit fully inlined and there's no specialization for it in the code.

* For binary arrays, I'm not sure and I can't seem to find the exact implementation reader in the code. Maybe @gitamohr can shed some light? I believe it's not a bit packed vector since I don't see a specialization anywhere.

dgovil:
Also @asluk had asked for a zip of the sphinx doc, so here's the zip. You'll want to open `specs_usdc.html`
[usddocs.zip](https://github.com/PixarAnimationStudios/USD/files/9985693/usddocs.zip)

meshula:
Thanks for the update.

@gitamohr The specific question was the encoding of bools, and arrays of bools.

syoyo:
> @gitamohr The specific question was the encoding of bools, and arrays of bools.

Array of bools `bool[]` is encoded as array of bytes(8bit), since bool array is represented as `VtArray<bool>`  and  `VtArray<bool>` uses `sizeof(bool) = 1` as an element size to allocate memory: https://github.com/PixarAnimationStudios/USD/blob/0c7b9a95f155c221ff7df9270a39a52e3b23af8b/pxr/base/vt/array.h#L854

For scalar `bool` type, datasize could be arbitrary, since it is represented as inlined value and fits to ValueRep's 6bytes payload.

dgovil:
Thanks @syoyo , I updated the section on Bool with that note, and also thanks because I used both USD and your TinyUSDZ project as a reference when writing these docs. Having the two was very helpful.

syoyo:
@dgovil 

>  Having the two was very helpful.

🤩👍

sunyab:
Filed as internal issue #USD-7771

docs/spec_usdc.rst
```diff
@@ -0,0 +1,760 @@
...
++--------------------+------------+----------------------------------------------------+
+| customLayerData    | dictionary | Custom user specified data                         |
++--------------------+------------+----------------------------------------------------+
+| defaultPrim        | token      | The prim to use when this layer is referenced      |
++--------------------+------------+----------------------------------------------------+
+| documentation      | string     | Top level documentation for this layer             |
++--------------------+------------+----------------------------------------------------+
+| metersPerUnit      | double     | The scale unit for this layer                      |
++--------------------+------------+----------------------------------------------------+
+| primChildren       | token[]    | A list of top level children to limit the layer to |
```
syoyo, line 233: `primChildren` and `properties` in Layer and Prim meta uses TokenVector(41), not array of `token`

```
name = "primChildren", value = ty: 41, isArray: 0, isInlined: 0, isCompressed: 0 
```

syoyo:
@dgovil Added comment on type for `primChildren` and `properties` https://github.com/PixarAnimationStudios/USD/pull/2086/files#r1027088026

TokenVector and `token[]`  have different meaning in pxrUSD implementation.
For example, you cannot use TokenVector to encode `token[]` attribute. When I encode a Prim attribute with `token[]` type as a TokenVector, pxrUSD apprently loads it finely but prints corrupted info in ASCII like this:

```
uniform token[] jointNames = <'vector<TfToken, allocator<TfToken> >' @ 0x55a0ab36c450>
```

spiffmon:
Hi @dgovil - have been meaning to provide some high level feedback that @gitamohr and I agreed on.  We think it would be overall clarifying and more economical to **first** provide a specification for the Sdf Data Model, which defines all terms, concepts, types, etc.  Then, each of the text-format and crate specifications can reference and build off of that shared data model specification.  Having the data model pulled out on its own would also be a boon to anyone trying to adapt a foreign data format to USD through an `SdfFileFormat` plugin.

What do you think?

dgovil:
I think that makes sense. Would just a Sphinx rst page make sense for something like that? Since it wouldn't be tied to any code etc

spiffmon:
Yes, absolutely - though it might someday become part of something larger
hosted/interred elsewhere, a new page under our current "Specifications"
section makes sense to me for now?

--spiff

Issue #1749, opened 2022-01-19
HdStorm fails to render a mesh with many material GeomSubsets


jufrantz:
### Description of Issue

When rendering a mesh with many material subsets in USD 21.11/HdStorm some subsets fail to render.
I attached a simple [usd file](https://github.com/PixarAnimationStudios/USD/files/7897362/manyGeomSubsets.usda.zip) that shows the issue; it contains a mesh with 100 material subsets; 41 last subsets fail to render on our side : 

![many subsets in usdview 21.11](https://user-images.githubusercontent.com/40573039/150141857-b5c55d42-664b-4107-9535-24864e36331d.png)

And we get bunch of errors, like these:
```
Warning: in _ValidateCompilation at line 182 of USD-21.11/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (GL_FRAGMENT_SHADER): 0(2299) : error C1008: undefined variable "HdGet_edgeIndices"

ERROR: Usdview encountered an error while rendering.
        Error in 'pxrInternal_v0_21__pxrReserved__::HdBufferArrayRangeContainer::Set' at line 49 in file USD-21.11/pxr/imaging/hd/bufferArrayRange.cpp : 'Index negative in HdBufferArrayRangeContainer::Set()'
        Error in 'pxrInternal_v0_21__pxrReserved__::HdSt_DrawBatch::_GetDrawingProgram' at line 273 in file USD-21.11/pxr/imaging/hdSt/drawBatch.cpp : 'Failed to compile shader for prim /World/Mesh.'
        Error in 'pxrInternal_v0_21__pxrReserved__::HdSt_DrawBatch::_GetDrawingProgram' at line 292 in file USD-21.11/pxr/imaging/hdSt/drawBatch.cpp : 'Failed verification: ' res ' -- Failed to compile with fallback material network'
        Error in 'pxrInternal_v0_21__pxrReserved__::HdSt_IndirectDrawBatch::ExecuteDraw' at line 1099 in file USD-21.11/pxr/imaging/hdSt/indirectDrawBatch.cpp : 'Failed verification: ' glslProgram ''
```

### Steps to Reproduce

1. Open the [usd file](https://github.com/PixarAnimationStudios/USD/files/7897362/manyGeomSubsets.usda.zip) in usdview and select GL renderer.

### System Information (OS, Hardware)
- Linux CentOS
- GeForce GTX 1080, Driver Version: 390.87

### Package Versions
[...]

PR #2089, opened 2022-11-14
APPLE: Fix for Issue #1749

creijon:
### Description of Change(s)
The _topology field in the drawingcoord was overflowing the 8-bit range for assets with a large number of geomsubsets.  This has been promoted to a 32-bit value to avoid this.

The instancePrimvar field could also overflow, so this has been promoted to a 32 bit value too.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/1749 

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -780,7 +780,6 @@ def InstallBoost_Helper(context, force, buildArgs):
             'threading=multi', 
             'variant={variant}'.format(variant=boostBuildVariant),
             '--with-atomic',
-            '--with-program_options',
             '--with-regex'
         ]
 
--- a/cmake/defaults/Packages.cmake
+++ b/cmake/defaults/Packages.cmake
@@ -43,7 +43,7 @@ find_package(Boost REQUIRED)
 
 # Boost provided cmake files (introduced in boost version 1.70) result in 
 # inconsistent build failures on different platforms, when trying to find boost 
-# component dependencies like python, program options, etc. Refer some related 
+# component dependencies like python, etc. Refer some related
 # discussions:
 # https://github.com/boostorg/python/issues/262#issuecomment-483069294
 # https://github.com/boostorg/boost_install/issues/12#issuecomment-508683006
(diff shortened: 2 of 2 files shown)


spiffmon:
I'm not up to date on current architecture, but am curious, @jonny-apple , if this changes the size of the structure, when packed?  If so, it would be great to see some memory numbers with large numbers of meshes?

creijon:
> I'm not up to date on current architecture, but am curious, @jonny-apple , if this changes the size of the structure, when packed?  If so, it would be great to see some memory numbers with large numbers of meshes?

Very good point spiff, I'll run some numbers on the large stages that we have as test assets.

sunyab:
Filed as internal issue #USD-7774

creijon:
@spiffmon , I've run a few variations of the field sizes in the `HdDrawingCoord` structure.
I've used the TS4 assets that you gave us, which has 456,356 prims.
These figures were from running usdview on a Mac with latest OS.  It takes a little bit of time for the memory to settle down, so there is some variability.

Firstly, the original 8 bits per field
(9 bytes total, probably rounded to 12)

Physical footprint:         4.7G
Physical footprint (peak):  5.6G

Physical footprint:         4.8G
Physical footprint (peak):  5.6G

Physical footprint:         4.6G
Physical footprint (peak):  5.6G


Then 16 bits for `_topology` and `_instancePrimvar`
(11 bytes total, probably rounded to 12)

Physical footprint:         4.7G
Physical footprint (peak):  5.8G

Physical footprint:         4.6G
Physical footprint (peak):  5.6G

Physical footprint:         4.6G
Physical footprint (peak):  5.8G

Then 32 bits for `_topology` and `_instancePrimvar`
(15 bytes total, probably rounded to 16)
Physical footprint:         4.7G
Physical footprint (peak):  5.5G

Physical footprint:         4.8G
Physical footprint (peak):  5.5G

Physical footprint:         4.8G
Physical footprint (peak):  5.6G

It looks like this has a small impact on the overall memory size, but if there is a concern then 16 bits for these fields gives us enough headroom and shouldn't change the memory footprint.

I'll amend the PR to use 16 bits.

spiffmon:
Thanks so much, @jonny-apple - that seems like a good choice, for now.  I'll note that we can double the value of our money by using `uint_16t` for those two fields that we specifically want headroom with?

drwave:
And to be clear - this came from a real bug we found with importing OBJ files that we overflowed how many geom subsets we needed to create import the data to USD - so this isn’t academic - it was a weird bug we ran into!

PR #2092, opened 2022-11-15
Two small improvements to the OBJ plugin

nicolaspopravka:
### Description of Change(s)
The proposed changes slightly improve the example plugin to convert between the OBJ file format and USD, and hopefully make it even more useful ;-)

### Fixes Issue(s)
- OBJ files with no `g` statements, or a `g` statement not followed by a group name, could not be referenced in. With this change, one is now able to do so, like this:
```
def Mesh "default" (
    prepend references = @mesh.obj@</default>
 )
 {
 }
```
- OBJ files with `vt` statements could not be rendered with textures when referenced in. This is fixed.

### DIFF
--- a/extras/usd/examples/usdObj/stream.cpp
+++ b/extras/usd/examples/usdObj/stream.cpp
@@ -186,7 +186,7 @@ UsdObjStream::AddFace(Face const &face)
 {
     // If there aren't any groups, add one first.
     if (_groups.empty()) {
-        AddGroup(string());
+        AddGroup(string("default"));
     }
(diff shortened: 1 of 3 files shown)


sunyab:
Filed as internal issue #USD-7777

nicolaspopravka:
@sunyab I see you've filed this PR as an internal issue. What is generally to be expected from this ? Thanks.

sunyab:
Hi @nicolaspopravka, the internal issue number is just for bug tracking purposes on our side and indicates that we've queued this PR for triage and prioritization.

PR #2093, opened 2022-11-15
Add Python version to pxrConfig.cmake

nicolaspopravka:
### Description of Change(s)
These proposed changes address the problem of how to configure another downstream cmake-based project that depends on USD, on systems with multiple versions of Python installed. The solution is to stash in the generated `pxrConfig.cmake` file the exact version used at build time, for the other projects to find, if possible. The build script can now help with that.  

### Fixes Issue(s)
- `find_package(Python)` in `pxrConfig.cmake` may not find the correct version of Python for another downstream project to successfully build. To be correct, `find_package(Python <version> EXACT)` is needed, with `<version>` the version of Python that USD was built with.


sunyab:
Filed as internal issue #USD-7778

PR #2096, opened 2022-11-16
Compare HdPrimvarDesrciptor equality by name and role

tylerm-nv:
### Description of Change(s)
Removes interpolation from the `HdPrimvarDescriptor` `==` overload.
### Fixes Issue(s)
Fixes an issue where `UsdImagingPrimAdapter::_MergePrimvar` will insert a duplicate descriptor into the vector if the interpolation is changed.

We observe this issue when calling `SetWidthsInterpolation` on basis curves after the widths attribute has been created. The equality check always returns false because the interpolation was changed, so it inserts a duplicate. Our delegate queries each interpolation type, and the last one found wins. This often results in the wrong type, and we can't proceed with some logic.

I tried to write tests for this, but could not reproduce with testusdview. The shader used for curves requires the primvar to be defined, it emits a shader compilation error if only the WidthsAttr is present. And it will render in some invalid states, but error in others, like this:

Initial render state:
- widths value [3, 3, 5]
- widths interp varying

Storm renders the expected image

Second render state:
- widths value [1]
- widths interpolation varying

Storm renders the expected image despite not changing interpolation.

If you reverse that order, so that it starts with a valid constant state and change to varying, it will fail with a shader error. Even with the operations in a changeblock.

### DIFF
--- a/pxr/usdImaging/usdImaging/primAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/primAdapter.cpp
@@ -768,12 +768,16 @@ UsdImagingPrimAdapter::_MergePrimvar(
     bool indexed) const
 {
     HdPrimvarDescriptor primvar(name, interp, role, indexed);
-    HdPrimvarDescriptorVector::iterator it =
-        std::find(vec->begin(), vec->end(), primvar);
-    if (it == vec->end())
-        vec->push_back(primvar);
-    else
-        *it = primvar;
+
+    for (HdPrimvarDescriptorVector::iterator it = vec->begin();
+        it != vec->end(); ++it) {
+        if (it->name == name) {
+            *it =  primvar;
+            return;
+        }
+    }
+
+    vec->push_back(primvar);
 }
 
 void
--- a/pxr/imaging/hd/sceneDelegate.h
+++ b/pxr/imaging/hd/sceneDelegate.h
@@ -191,8 +191,7 @@ struct HdPrimvarDescriptor {
         : name(name_), interpolation(interp_), role(role_), indexed(indexed_)
     { }
     bool operator==(HdPrimvarDescriptor const& rhs) const {
-        return name == rhs.name && role == rhs.role
-            && interpolation == rhs.interpolation;
+        return name == rhs.name && role == rhs.role;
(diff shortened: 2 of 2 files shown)


sunyab:
Filed as internal issue #USD-7781

tcauchois:
Hi!

Thanks for the submission but I don't think changing the equality operator here is safe.  Is there a way to fix the bug by changing UsdImagingPrimAdapter::_MergePrimvar instead?

Thanks,
Tom

tylerm-nv:
It could be fixed with a custom `HdPrimvarDescriptor` equality comparison exclusively used by the `std::find` in `UsdImagingPrimAdapter::_MergePrimvar`, is that an acceptable fix? The comparison for that function would be just name and role.

tcauchois:
Yeah, that sounds good.  Although maybe we only want to check name?  Unique name = unique primvar, and "interpolation" and "role" are just modifiers on how we access that information.  Thanks!

tcauchois:
New version looks a lot better!  On github it looks like the PR still has the sceneDelegate.h changes but I'm not sure if I'm just looking at the wrong thing.

Anyway, we'll pull this down and run it through the test suite and then we can merge it!

tylerm-nv:
Oops, I forgot to revert the original changes. Pushed a commit for that.

pxr/usdImaging/usdImaging/primAdapter.cpp
```diff
@@ -768,12 +768,16 @@ UsdImagingPrimAdapter::_MergePrimvar(
...
 {
     HdPrimvarDescriptor primvar(name, interp, role, indexed);
-    HdPrimvarDescriptorVector::iterator it =
-        std::find(vec->begin(), vec->end(), primvar);
-    if (it == vec->end())
-        vec->push_back(primvar);
-    else
-        *it = primvar;
+
+    for (HdPrimvarDescriptorVector::iterator it = vec->begin();
```
nvmkuruc, line 772: If you're open to the suggestion, I think you can preserve the existing structure and the usage of the STL with `std::find_if`.

Either way, it's great to see this fix!

PR #2101, opened 2022-11-18
Apple: Convert UsdZip to C++

dgovil:
This PR depends on https://github.com/PixarAnimationStudios/USD/pull/2090 , and follows up by converting usdzip to C++.
The files that are changed from that PR are:

* `pxr/usd/bin/usdzip/CMakeLists.txt`
* `pxr/usd/bin/usdzip/usdzip.cpp`

additionally a new `GetFileSize` function was added to the `pxr/usd/bin/common.h/cpp` files.

Some notes:

* Compliance Checking results in an error being surfaced to the user since those have yet to be ported to C++. I was torn on whether to leave the argument in place or not, but either way an error seems best for now.
* `testUsdZipInputFiles2` fails because the order of the files listed from the UDSZ are in a different order. I'm not sure what to do here. I'm not sure what causes the files to be listed in a different order, but I don't see anything obvious. I'm getting the filelist in the same order as the `UsdZipFile` provides it, but it looks like `UsdZipFile` gives it to me in a different order? specifically `src/sub/c.png` and `src/sub/d.txt` are returned in flipped positions.

- [X] I have verified that all unit tests pass with the proposed changes **with one minor exception**

### DIFF
--- a/pxr/usd/bin/usdzip/usdzip.cpp
+++ b/pxr/usd/bin/usdzip/usdzip.cpp
@@ -1,0 +1,368 @@
+//
+// Copyright 2022 Apple
+//
+// Licensed under the Apache License, Version 2.0 (the "Apache License")
+// with the following modification; you may not use this file except in
+// compliance with the Apache License and the following modification to it:
+// Section 6. Trademarks. is deleted and replaced with:
+//
+// 6. Trademarks. This License does not grant permission to use the trade
+//    names, trademarks, service marks, or product names of the Licensor
+//    and its affiliates, except as required to comply with Section 4(c) of
+//    the License and to reproduce the content of the NOTICE file.
+//
+// You may obtain a copy of the Apache License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the Apache License with the above modification is
+// distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the Apache License for the specific
+// language governing permissions and limitations under the Apache License.
+//
+
+
+#include <pxr/pxr.h>
+#include "pxr/base/tf/pxrCLI11/CLI11.h"
+#include <pxr/usd/ar/resolver.h>
+#include <pxr/usd/ar/resolverContextBinder.h>
+#include <pxr/base/arch/fileSystem.h>
+#include <pxr/usd/sdf/assetPath.h>
+#include <pxr/base/tf/pathUtils.h>
+#include <pxr/base/tf/fileUtils.h>
+#include <pxr/usd/usd/zipFile.h>
+#include <pxr/base/tf/debug.h>
+#include <pxr/usd/usdUtils/dependencies.h>
+
+
+PXR_NAMESPACE_USING_DIRECTIVE
+using namespace pxr_CLI;
+
+struct Args {
+    std::string usdzFile;
+    std::vector<std::string> inputFiles = {};
+    bool recurse = false;
+    std::string asset;
+    std::string arkitAsset;
+    bool checkCompliance = false;
+    std::string listTarget;
+    std::string dumpTarget;
+    bool verbose = false;
+};
+
+void Configure(CLI::App *app, Args &args) {
+    app->add_option("usdzFile", args.usdzFile,
+                    "Name of the .usdz file to create or to inspect the contents of.");
+    app->add_option("inputFiles", args.inputFiles,
+                    "Files to include in the .usdz files");
+    app->add_flag("-r,--recurse", args.recurse,
+                  "If specified, files in sub-directories are recursively added to the package");
+    app->add_option("-a,--asset", args.asset,
+                    "Resolvable asset path pointing to the root layer" \
+                  "of the asset to be isolated and copied into the package.");
+    app->add_option("--arkitAsset", args.arkitAsset,
+                    "Similar to the --asset option, the --arkitAsset "\
+                  "option packages all of the dependencies of the named scene file.\n"\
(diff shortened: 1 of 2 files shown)


pxr/usd/bin/common.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <iostream>
+#include <fstream>
+
+
+std::string GetVersionString() {
+    return std::to_string(PXR_MAJOR_VERSION) + "." + std::to_string(PXR_MINOR_VERSION) + "." + std::to_string(PXR_PATCH_VERSION);
+}
+
+int GetFileSize(const std::string &path) {
```
meshula: Probably `ArchGetFileLength(...)` would work instead of adding a new function?

It uses GetFileSize on Windows, and fstat for linux/mac. I think that's okay here.

sunyab:
Filed as internal issue #USD-7784

pxr/usd/bin/common.cpp
dgovil: Ah good call, I was looking for `FileSize` and couldn't find a function.
Swapped it out in the latest commit

dgovil:
Refactored to base on https://github.com/PixarAnimationStudios/USD/pull/2107 
I adjusted the failing recursive test to pass since I don't think it's actually an issue, just a difference in order of how the Python zipfile works compared to the USD zip reader, but both are correct.

dgovil:
@meshula I wanted to revisit this PR and see what you think would be a good approach to get it in.

1. I can update it to be built on top of the latest dev, and following the conventions of the latest USD dev commit. That's not an issue.
2. Compliance checker is only available in Python which gates the `--checkCompliance` flag. I'm not sure on the ETA of a C++ rewrite of the compliance checks, but I was wondering what you thought about calling the Python compliance checker if compiled with Python, and otherwise having the flag raise an error if used. When the C++ rewrite of the checks lands, we can swap that out fairly easily.

Anyway would love your thoughts since I'd love to have usdzip in my non-Python builds of USD for selfish reasons.

meshula:
Other non-python-centric groups such as gamedev are keen to have non-pythonic versions of the tools, so you're not the only one looking for C++ versions of these tools. Since you tagged me I feel compelled to have a cunning thought.

What if you make an intermediate stage for your project, that doesn't lose functionality?

name the c++ tool something else, like I dunno, `usdziptool`
`usdzip.py` checks for `--complianceCheck` and does the thing. If it passes, it passes all the arguments to `usdziptool` to actually do the work of creating the archive.

now you can run `usdziptool` in a python-less environment, and the existing functionality of `usdzip` is unaffected if one has a python-present environment.

Maybe it's practical to have usdzip and usdzip.py live side by side, so maybe the subterfuge of a renamed c++ variant is not needed. Not sure.

dgovil:
Hmm I can give that a try and see what feels better, though I'm a little hesitant to introduce an extra file to the mix. The last round of C++ rewrites I did, the Pixar folks had said to go with a hard switchover for simplicit.

Anyway, I'll try your suggestion out in addition to mine. The changes are minimal either way, so should be quick to try out.

meshula:
Re-reading your #2 and your response, I now understand your option #2 (I'd read it wrong the first time). That sounds roughly equivalent to what I suggested but with one less published file, so that suggestion sounds reasonable to me.

dgovil:
Okay great, that simplifies things

dgovil:
@meshula okay donesies. This is rebased to the latest dev commit as of this morning.
It will either fail the compliance checker function if compiled without Python, or it uses CPython to call the Python code.

I didn't end up using the Tf Python functions because I was getting random segfaults when trying to use it, that I couldn't unwind. Since this is a leaf utility, I ended up just using standard CPython functions. 

Let me know if there's any changes y'all recommend. Hopefully it's in a state that's acceptably mergeable.

dgovil:
Excellent, all the tests passed. I always forget msvc doesn't support all the C keywords that clang and gcc do.

nvmkuruc:
Hi all.  Just curious if this PR is still under consideration.

Thanks!

pxr/usd/bin/usdzip/usdzip.cpp
```diff
@@ -0,0 +1,445 @@
...
+
+    if (closeAfterUse) {
+        static_cast<std::ofstream *>(out)->close();
+        delete out;
+    }
+}
+
+
+/// Deal with symlinks in case AssetResolver doesn't know how to anchor it properly given the current context.
+std::string ResolveAllInputs(Args &args) {
```
matthewcpp, line 258: I am working on getting this ready to merge.  It looks like this function / logic is not preset in the python version of usdzip.  Are you able to shed some light on the type of paths that this function is resolving which otherwise do not work?

dgovil: Ah my memory is slightly rusty given the age of this PR but I think it was when you had a symlink that was relative but invoked usdzip on it from a path that wasn't the current directory of the symlink. That would cause it to be treated as relative to your current directory instead.

matthewcpp: Hey @dgovil , thanks for getting back to me!

I tried reproducing this on both my mac and linux workstation without any luck.  Just to make sure I got the setup right: I made an asset, symlinked the root with a relative path.  then from a different directory I ran usdzip on the symlinked root.

I wasn't able to observe any noticeable  failures with the python version or the cpp port with this function removed.  Do you happen to have details for a repro case?  Give the age of the PR perhaps it is possible this was actually fixed by a change that went in between then and now?

dgovil: Yeah I'm guessing some of the other resolution stuff elsewhere might have fixed whatever edge case I was handling. 

I'd be in favour of taking this out and then we can always add it back if we discover it's necessary.

PR #2107, opened 2022-11-28
Add CLI11 to Tf

dgovil:
### Description of Change(s)
This PR adds [CLI11](https://github.com/CLIUtils/CLI11) to `Tf` under the `pxr/base/tf/pxrCLI11` in the  `PXR_NS` namespace.

This splits out the addition of the library in https://github.com/PixarAnimationStudios/USD/pull/2090  per the notes there.


sunyab:
Filed as internal issue #USD-7796

sunyab:
Hey @dgovil, just a heads-up: we've merged this PR internally but during review wound up making some changes to more fully isolate the CLI11 in Tf by adding extra compile-time checks and a hardcoded `pxr_CLI` namespace along with the normal pxr namespace. These are to guard against the unlikely-but-possible cases where USD's embedded CLI11 might conflict with another version of CLI11 in user code.

Practically, this means the other PRs that build against Tf's CLI11 will need to have something like `using namespace pxr_CLI` added. I'm adding that in as I merge the other PRs, but wanted to let you know FYI.

dgovil:
Thanks for the headsup, Sunya! It's much appreciated.

PR #2108, opened 2022-11-28
Port USDTree to C++

dgovil:
### Description of Change(s)

This is a port of usdtree to C++, depends on https://github.com/PixarAnimationStudios/USD/pull/2107.

Edit: Nevermind regarding the issues. I was expecting the test directories to use the same `output.txt` but they had different names which was tripping me up.

All tests are now passing on my end 


sunyab:
Filed as internal issue #USD-7797

PR #2109, opened 2022-11-29
Port sdfdump to CLI11

dgovil:
### Description of Change(s)

This PR ports `sdfdump` to CLI11 (requires https://github.com/PixarAnimationStudios/USD/pull/2107)
Since I was removing `boost` usage anyway, I also changed the uses of `boost::lexical_cast<double>` to `std::stod`, so now there's no boost dependency within this command. 


sunyab:
Filed as internal issue #USD-7798

PR #2115, opened 2022-11-29
Test fixes for float precision issues (ie, on Linux aarch64)

pmolodo:
### Description of Change(s)

A collection of float precision tolerance fixes, to resolve test failures on Linux aarch64 platforms (tested on an Arm Neoverse N1).

- [gf] testGfMatrix: increase tolerance for float AssertDeterminant test
- [gf] testGfRotation: add a small tolerance for associativity test
- [examples] testUsdDancingCubeExample - only do diff if on x86_64 (not aarch64)

- https://github.com/PixarAnimationStudios/USD/pull/2131
- https://github.com/PixarAnimationStudios/USD/pull/2154

### Fixes Issue(s)
- testGfMatrix failure on Linux aarch64
- testGfRotation failure on Linux aarch64
- testUsdDancingCubesExample failure on Linux aarch64

### DIFF
--- a/extras/usd/examples/usdDancingCubesExample/CMakeLists.txt
+++ b/extras/usd/examples/usdDancingCubesExample/CMakeLists.txt
@@ -34,14 +34,15 @@ pxr_install_test_dir(
 )
 
 # XXX:
-# Baselines for this test were generated on a build on Linux
+# Baselines for this test were generated on a build on Linux x86_64
 # using GCC. The results may differ on other platforms and compilers
 # due to floating point precision issues. So, we only perform the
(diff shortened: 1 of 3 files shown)


sunyab:
Filed as internal issue #USD-7816

tallytalwar:
We are fine with this change, but are curious if there are any specific compiler flags being used, like "-ffast-math" or others which break IEEE-754 compliance, as the differences with arm numbers seem a bit large than expected.

Thanks

pmolodo:
We're not setting that, so not unless build_usd.py sets it...

PR #2116, opened 2022-11-29
Allow external configuration of TBB Build Type

dgovil:
### Description of Change(s)

This PR allows external configuration of which TBB build type to look for.
This helps in setups like ours where we only distribute the Release TBB build in our Rez environment, so USD builds fail unless we override this variable. Typically we've just force set it internally, but this change allows it to be configured as a parameter to CMake.

The default is the same as it is today,  but it allows configurations like ours to work around it without needing to also distribute the debug TBB builds within Rez. I know it's logical for most people to build Debug USD against Debug USD, but in a lot of cases, I find people aren't interacting with TBB symbols.

### DIFF
--- a/cmake/modules/FindTBB.cmake
+++ b/cmake/modules/FindTBB.cmake
@@ -108,16 +108,17 @@ if(NOT TBB_FOUND)
   
   if(NOT DEFINED TBB_USE_DEBUG_BUILD)
     if(CMAKE_BUILD_TYPE MATCHES "(Debug|DEBUG|debug|RelWithDebInfo|RELWITHDEBINFO|relwithdebinfo)")
-      set(TBB_BUILD_TYPE DEBUG)
+      set(_TBB_BUILD_TYPE DEBUG)
     else()
-      set(TBB_BUILD_TYPE RELEASE)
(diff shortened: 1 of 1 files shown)


meshula:
Are you aiming to (1) eliminate the debug symbols, for size (or uselessness) or (2) force an optimized build?

Looking at this diff, I realize that there might be a pre-existing bug. RelWithDebInfo, on Windows, sets TBB_DEBUG, which might cause TBB to link with the debug CRT, and USD itself to link with the release CRT. Malloc'ing on one, and free'ing on the other causes difficult to analyze heap corruption bugs. It's unfortunate that cmake conflates a Debug build with "optimization off + debug symbols on + use the debug CRT on windows". I guess we need to audit the TBB build script, and possibly fix another aspect of this Find script.

dgovil:
We were eliminating the debug symbols due to size of distribution.
Though acutally I'll rescind this PR because I realized that I can just module `TBB_USE_DEBUG_BUILD` based on some changes elsewhere in our internal repo, and don't need to control `TBB_BUILD_TYPE` directly.
Previously, setting `TBB_USE_DEBUG_BUILD` alone was causing other issues that I thought were coming from USD but were coming from an internal change that slipped by me.

PR #2120, opened 2022-12-02
Support for Unicode UTF-8 Identifiers

erslavin:
### Description of Change(s)

Initial draft for review session with Pixar based on submitted white paper with additional changes to support separating out original ASCII parser and implementation of new Unicode UTF-8 parser to fully retain original logic.

- Added additional lexer / parser for UTF-8 Identifiers
- Added TF setting to scope changes behind TF_UTF8_IDENTIFIERS
- Made first pass at Tf string methods to switch on setting and use ASCII vs. Unicode validation rules
- Enabled different rules for identifiers vs. prim names (validation for prim names falls back to identifiers when under ASCII rules)
- Added UCA sorting for UTF-8 strings and tests against Unicode conformance tests
- Pulled common parse layer methods out of textFileFormat parser to enable use by both ASCII and Unicode flavors
- Added support for Unicode UTF-8 identifiers / prim names in usdGenSchema

Caveats:

- There are 4 tests that do not pass when the new `TF_UTF8_IDENTIFIERS` setting is on; this is due to the TF environment banner being written out, which makes the output differ from the files checked in to compare against
- Original logic should be retained in all cases.  There will be a small performance hit to check against the runtime switch TF_UTF8_IDENTIFIERS to determine which code paths to take (ASCII vs. Unicode).  This is something that needs further investigation.
- Needs testing against very large stages to understand impact of both performance hit to original logic (because of the runtime switch) as well as impact of UTF-8 character processing.  In particular, the UCA (Unicode Collation Algorithm) for sorting is likely much slower than the dictionary ordering algorithm currently in place.  Since this affects ordering of child prims and property names, it's something that needs to be measured on large stages.
- There are a number of files that are added for unicode character class caching (to try and speed up some of the character class checks) that are generated by scripts in the pxr/base/tf/unicode directory.  There is currently no support in the build process to build and run these scripts to generate these files.  This should be added for completeness.
- The new lexer / parser generated for Unicode path and textFileFormat processing need to be added to the portion of the build script that allows users to generate the lexers / parsers and verified that the complete generate --> build --> run sequence works for all of the new changes.
- Flex 2.6.4 and bison 2.7.12 were used to generate the new lexer / parser and the existing textFileFormat ASCII parser was hand modified.  Need to generate these files completely using these known versions.  These versions were chosen to also support win_flex and win_bison in the future for those generating on windows outside of WSL.
- The tests now have to be run with TF_UTF8_IDENTIFIERS off (the default case) and on (the new case) for complete verification of additional changes.

### Fixes Issue(s)
-


pxr/base/tf/stringUtils.h
```diff
@@ -683,18 +727,113 @@ std::string TfStringCatPaths( const std::string &prefix,
...
-    unsigned x = *p;
-    if (!x || number(x)) {
-        return false;
+    if (UseUTF8Identifiers())
+    {
+        // use unicode utils to validate the identifier
+        return TfUnicodeUtils::IsValidUTF8Identifier(identifier, 
+            identifier.begin(), identifier.end());
+    }
+    else
```
dgovil, line 753: Minor nit (and I defer to others on style), but it would be nice to not have the `else` statement here since the `if` has an early return.

Has a few benefits:
* Smaller diff since the insertion of the new branch would be the only thing that changed.
* Easier to catch if there were any unintended changes in the existing code (since the diff highlighting doesn't delineate between indented code and modified code very well on Github)
* Reduces code indentation which (IMHO) is nice for readability

pxr/base/tf/stringUtils.h
```diff
@@ -683,18 +727,113 @@ std::string TfStringCatPaths( const std::string &prefix,
...
+    {
+        // use unicode utils to validate the identifier
+        return TfUnicodeUtils::IsValidUTF8Identifier(identifier, 
+            begin, end);
+    }
+    else
+    {
+        size_t count = (end - begin);
+        char const *p = &(*begin);
+        auto letter = [](unsigned c) { return ((c-'A') < 26) || ((c-'a') < 26); };
```
dgovil, line 798: Since some of these checks are shared with TfIsValidIdentifier above, it might be nice to break them out into a shared inlined function (like `IsALetter` or something)? I get paranoid about drift in the future in case someone edited one and forgot the other etc...

erslavin: The original change did exactly that, and the white paper describes that issue.  For this PR, I intentionally left all code paths that were there the same, so as not to introduce anything extra.

spitzak:
I would stay away from using any usd or locale specific collating. You do not want the sort to change depending on the locale. Also (at least on Linux) the libraries have serious errors or bugs, such as ignoring punctuation, so that this is considered the correct sorted order:
```
    fooa.jpg
    foo.jpg
    fook.jpg
```
(in case you can't figure out what happened, 'a'<'j'<'k')

It's ok if you use information about USD code points to choose collating order, but it has to be exactly the same order for all users, and with no rude surprises like the above. Also it is far more important to fix numerical ordering (so `foo9` < `foo10`) and that usually means you can't use any stock collator other than gnu versionsort.

pxr/base/tf/stringUtils.h
erslavin: I think these are reasonable points, and once the PR has been approved in principle this seems like a change worth having.  The reason it is structured this way at the moment is so that it is abundantly clear to the reviewer that the original code paths are intact in almost all instances.  Once of the biggest concerns with a change like this is "will it introduce subtle errors or differences into my applications / workflows".  For example, in a first pass I had modified the original lexer / parser to support both the original and new workflows, but it had the potential to introduce subtle errors because the new lexer was more permissive than the old and relied on additional validation methods to be sure what we lexed was the right token type.  I had hoped that the way it is structured now makes it much clearer that the original code paths are untouched, with the additional overhead of the runtime check for the new switch.  The only place where I diverged from this philosophy was the introduction of separate validation rules for prim names, where the child policy on prims was changed to use the  new prim name validation rules (albeit when the switch is off it actually falls back to the same behavior, but it's not necessarily obvious to the reviewer).

erslavin:
@spitzak The implementation here is a (unoptimized) hand-rolled implementation of the Unicode Collation Algorithm, which you can find here: https://unicode.org/reports/tr10/.  This does not depend on any locale specific functions (and in fact, ignores the very few use cases where locale context is required to ascertain correct ordering), nor does it depend on additional libraries to not introduce additional dependencies into USD.  

The resulting ordering will be exactly the same for all users - this is the beauty of UCA.  However, and specifically with regard to numerical sorting as you might be used to in ASCII dictionary orderings, the default collation tables (which are quite extensive) do not order latin numerics the way you might think (`foo1`, `foo2`, and `foo17` should sort as `foo1`, `foo17`, and `foo2` according to UCA).  It is possible (and this is one case where it is explicitly stated in TR10 that you might want to have) to hand tweak the provided collation tables to get the specialized behavior you want.  So while it is possible to implement a UCA variant that provides the exact same sorting order as USD does today, it is not trivial, the collation tables for this are not provided by the Unicode standard, and it would take extra effort to do so (in addition with coming up with a new set of conformance tests, which are quite extensive and provided as part of this PR in TR10).

I expect there to be some hesitance in regard to the different sorting, and that's ok (however do note that the sort order is still exactly the same for every user while still taking into account the complexities of characters outside the non-extended Latin range).  The tradeoffs to having support for Unicode vs. not until there is a complete agreement in sort order to me weigh in favor of having, especially since support is behind a run-time switch that defaults to existing behavior.  Where the sort order is different in the existing tests should be clear from the changes in the PR.    Furthermore, USD provides functionality to customize property ordering, so there is a mitigation that can be used to preserve today's functionality, if that's specifically what is needed in your workflows.

spitzak:
That is good if the sort order is the same for everybody in all cases. This includes when LC_COLLATE is set to "" or "C"?

The others are just annoyances. PLEASE fix the ASCII punctuation so it always sorts before any letters. I have to run with LC_COLLATE set to C to fix this as it makes the names average users write on files not sort in any logical way. Make absolutely sure that "foo.jpg" sorts less than "fooa.jpg". This breaks the sorting of variable-sized numbers far more than the need for strverscmp. 

As for the numbers, well failure to support that is really really common. But it is hugely annoying to users who do not understand why they have to zero-pad all their numbers.

My recommendation is that the sort algorithm purposely split the strings at ASCII punctuation and numbers and sort only the character substrings with the unicode collating algorithm. The other substrings should be considered less than the characters and compared with each other using strverscmp.

spitzak:
Note that LC_COLLATE="C" is not a good solution, it does not sort characters in the way users expect. But the problem with ASCII puncutation (especially '.' and '-' and '/') is so overwhelmingly bad that I have to run it this way. A hybrid solution that does this correctly but also uses Unicode for the hard part of sorting letters would be ideal.

pxr/base/tf/stringUtils.cpp
```diff
@@ -1150,37 +1187,58 @@ TfStringCatPaths( const string &prefix, const string &suffix )
...
+                result.push_back(*p);
+            }
+        }
+        return result;
     }
-    return result;
+}
+
+std::string
+TfMakeValidPrimName(const std::string& in)
```
gitamohr: The Tf library does not use the term "prim" (it doesn't know what prims are) and this change should not introduce that, thanks!

pxr/base/tf/stringUtils.h
```diff
@@ -683,18 +727,113 @@ std::string TfStringCatPaths( const std::string &prefix,
...
 TfIsValidIdentifier(std::string const &identifier)
 {
-    char const *p = identifier.c_str();
-    auto letter = [](unsigned c) { return ((c-'A') < 26) || ((c-'a') < 26); };
-    auto number = [](unsigned c) { return (c-'0') < 10; };
-    auto under = [](unsigned c) { return c == '_'; };
-    unsigned x = *p;
-    if (!x || number(x)) {
-        return false;
+    if (UseUTF8Identifiers())
```
gitamohr, line 747: I think that the general tactic we'll need to go with in order to avoid performance fallout is to optimistically take the ASCII path, and only if we see a byte with the high-order bit set do we hop off onto a slower UTF-8 path.

pxr/usd/sdf/testenv/testSdfPath.py
```diff
@@ -140,17 +157,80 @@ def BasicTest(path, elements):
...
+                    testProperty, testIndex)
         
         print('\tPassed')
         
+        # ========================================================================
+        # Test UTF-8 encoded paths
+        # ========================================================================
+        utf8_good_paths = [
+            "123",
+            "123test",
```
gitamohr, line 178: The Nd category (decimal digits) is not part of xid_start, so these paths should _not_ be considered good, correct?  Or am I missing something here?

pxr/base/tf/stringUtils.h
```diff
@@ -683,18 +727,113 @@ std::string TfStringCatPaths( const std::string &prefix,
...
+}
+
+/// Tests whether \a primName is valid.
+///
+/// A prim name is valid if it's at least one character long and all characters
+/// are valid UTF-8 sequences.  This method ignores the value of 
+/// TF_UTF8_IDENTIFIERS and instead uses either UTF-8 or ASCII rules for 
+/// validation based on the value of \a utf8.
+///
+inline bool
```
gitamohr: I was under the impression that we were going to expand prim & property names to be UTF-8 style identifiers (xid_start + xid_continue), not any arbitrary sequence of utf-8 encoded code points.  That seems a bit bananas to me, a single space being a valid prim name for example?  Or allowing control codes, or SdfPath syntax characters?

pxr/usd/usd/testenv/testUsdPrims.py
```diff
@@ -614,13 +614,19 @@ def test_PropertyOrder(self):
         assert po
         attrs = po.GetAttributes()
         # expected order:
-        expected = ['A0', 'a1', 'a2', 'A3', 'a4', 'a5', 'a10', 'A20']
+        if Tf.GetEnvSetting('TF_UTF8_IDENTIFIERS'):
+            expected = ['A0', 'a1', 'a10', 'a2', 'A20', 'A3', 'a4', 'a5']
```
gitamohr: This will need to be addressed -- it's important to us that sequences of numerals sort as their numeric value.

pxr/base/tf/stringUtils.h
erslavin: Yes, this comment leaves out the part below "Valid UTF-8 sequences here means all characters must be in the XID_Continue character class".  This additional comment should be there.

pxr/base/tf/stringUtils.cpp
erslavin: I think that this can probably be lofted up into Sdf without a huge amount of effort, I can look into it.

pxr/usd/usd/testenv/testUsdPrims.py
erslavin: I think that we could minimally (without doing DUCET tailoring) figure out subsequences of Nd character classes and treat those special if required.

pxr/usd/sdf/testenv/testSdfPath.py
erslavin: This went along with the prim name expansion, such that prim paths have elements that can have numeric start characters in Nd by virtue of expanding the first character to be XID_Continue (rather than XID_Start) so since these are valid prim names, the path is valid in this test.

spitzak: I think it is ok if non-ASCII digits sort incorrectly. However it is pretty important that numbers sort as before, and also that punctuation effect the sort exactly as though the words separated by them were individually sorted. USD relies on this, putting '/' and '.' into strings and expecting them to sort in the same order as the same strings truncated at those punctuation marks.
I think a method that would work is to seqment the strings being compared into alternating sequences of non-alphabetic ASCII (N) and all other characters (all UTF-8 and all ASCII letters) (L). (N) strings are compared using the normal sorting (which puts numbers in order). (L) strings are compared using UTF-8 collating (putting words in alphabetical order and mixing case). A string that starts with (N) is strictly less than a string that starts with (L).

pxr/usd/usd/testenv/testUsdPrims.py
spitzak: See comment above. I think it would be sufficient to only work with ASCII digits this way and it could be merged with necessary changes so that ASCII punctuation is considered more significant than later letters. This is done by extracting the ASCII non-alphabetic sequences and comparing them with our own code.

sunyab:
Filed as internal issue #USD-7818

pxr/usd/usd/testenv/testUsdPrims.py
erslavin: Implementation now uses current sorting algorithm with the addition of UTF-8 byte sorting when there is a mismatch.

pxr/usd/sdf/testenv/testSdfPath.py
erslavin: See comment on sorting below, this part has to do with prim names.

pxr/base/tf/stringUtils.h
erslavin: Implementation now uses current sorting algorithm with the addition of UTF-8 byte sorting when there is a mismatch.

pxr/base/tf/stringUtils.cpp
erslavin: Lofted these methods up to Sdf

pxr/base/tf/stringUtils.h
erslavin: Added this comment when lofting those methods up to Sdf

pxr/base/tf/stringUtils.h
```diff
@@ -68,6 +79,7 @@ class TfToken;
 /// as opposed to just passing \c caller).
 ///
 /// \note \c TfStringPrintf() is just a wrapper for \c ArchStringPrintf().
+///
```
nvmkuruc: This can probably be removed.

pxr/base/tf/stringUtils.cpp
```diff
@@ -67,6 +67,28 @@ TF_DEFINE_ENV_SETTING(TF_UTF8_IDENTIFIERS,
...
 
+namespace {
+    bool _IsASCIIValue(const char& c)
+    {
+        return(static_cast<int>(c) >=0 && static_cast<int>(c) <=127 );
+    }
+
+    bool _IsInASCIIValueRange(const std::string& str)
+    {
+        for(std::string::const_iterator it = str.begin();
```
nvmkuruc: Can a ranged based for loop or `std::all_of` be used here?

pxr/base/tf/stringUtils.cpp
```diff
@@ -63,6 +63,32 @@ using std::vector;
...
 PXR_NAMESPACE_OPEN_SCOPE
 
+TF_DEFINE_ENV_SETTING(TF_UTF8_IDENTIFIERS, 
+    false, 
+    "Allow UTF8 strings as identifiers and prim names");
+
+namespace {
+    bool _IsASCIIValue(const char& c)
+    {
+        return(static_cast<int>(c) >=0 && static_cast<int>(c) <=127 );
```
nvmkuruc: Could this be done just by checking the leading bit?

pxr/base/tf/stringUtils.cpp
```diff
@@ -801,7 +833,8 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
         }
         l = *lcur, r = *rcur;
         // If they are letters that differ disregarding case, we're done.
-        if (((l & ~0x20) != (r & ~0x20)) & bool(l & r & ~0x3f)) {
+        // but only if they are ASCII (i.e., the high bit is not set)
+        if ((((l & (1<<7)) == 0) && ((r & (1<<7)) == 0)) && (((l & ~0x20) != (r & ~0x20)) & bool(l & r & ~0x3f))) {
```
nvmkuruc: Should / could `_IsASCIIValue` be used here?

pxr/base/tf/stringUtils.h
```diff
@@ -46,6 +48,15 @@
...
 PXR_NAMESPACE_OPEN_SCOPE
 
+// environment setting for defaulting back to standard ASCII rules for 
+// identifiers and prim names by default this value is false indicating 
+// that normal USD processing paths will parse and determine identifier
+// validity.  If this value is set to true, Unicode processing paths
+// will be used on UTF-8 character sequences to determine identifier
+// validity.  In addition, if this value is true, slightly different
+// rules will be used to evaluate generic identifiers vs. prim names.
+TF_API extern TfEnvSetting<bool> TF_UTF8_IDENTIFIERS;
```
nvmkuruc: Do you need to expose this in the header?  Can't `TfUseUTF8Identifiers` be used downstream?

pxr/base/tf/stringUtils.h
```diff
@@ -517,7 +583,7 @@ struct TfDictionaryLessThan {
         // different cases, or numerical comparisons, so we special-case this
         // first.
         char l = lhs.c_str()[0], r = rhs.c_str()[0];
-        if (((l & ~0x20) != (r & ~0x20)) & bool(l & r & ~0x3f)) {
+        if ((((l & (1<<7)) == 0) && ((r & (1<<7)) == 0)) && (((l & ~0x20) != (r & ~0x20)) & bool(l & r & ~0x3f))) {
```
nvmkuruc, line 576: Could / should `_IsAsciiValue` be used here?

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
```diff
@@ -0,0 +1,75 @@
...
+                del stage
+
+                # reopen the binary content and ensure that the string 
+                # representations have content - note they won't be the same
+                # because the Export added an additional Generated from line
+                text_stage = Usd.Stage.Open("utf8_content.usda")
+                binary_stage = Usd.Stage.Open(f.name)
+
+                text_layer_as_string = text_stage.ExportToString()
+                self.assertTrue(len(text_layer_as_string) > 0)
```
nvmkuruc, line 67: You can just `assertTrue(text_layer_as_string)` since an empty string will evaluate to `False`.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
```diff
@@ -0,0 +1,75 @@
...
+
+                binary_layer_as_string = binary_stage.ExportToString()
+                self.assertTrue(len(binary_layer_as_string) > 0)
```
nvmkuruc, line 70: You can just `assertTrue(binary_layer_as_string)` since an empty string will evaluate to `False`.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_LoadUTF8BinaryFile(self):
+        """Tests whether UTF-8 content can be appropriately loaded from a usdc file."""
+        ################################################################################
+        if Tf.GetEnvSetting('TF_UTF8_IDENTIFIERS'):
+            # loading the content should succeed under this test
+            stage = Usd.Stage.Open("utf8_content.usda")
+
+            # export the content to binary
+            with tempfile.NamedTemporaryFile(suffix='.usdc') as f:
```
nvmkuruc, line 55: It looks like closing and reopening the file might be system dependent.

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

I'd advocate using `TemporaryDirectory` instead.

It would be cool if Usd could export to file like objects at some point in the future.

pxr/usd/usd/prim.h
```diff
@@ -322,7 +322,7 @@ class UsdPrim : public UsdObject
     /// Return this prim's property names (attributes and relationships) that
     /// have authored scene description, ordered according to the strongest
     /// propertyOrder statement in scene description if one exists, otherwise
-    /// ordered according to TfDictionaryLessThan.
+    /// ordered according to TfDictionaryLessTha.
```
nvmkuruc: Should this be `TfDictionaryLessThan`?

pxr/usd/sdf/textParserContext.h
```diff
@@ -47,8 +47,6 @@
 // Lexical scanner type.
 typedef void *yyscan_t;
 
-PXR_NAMESPACE_OPEN_SCOPE
```
nvmkuruc: Why was it necessary to remove the `PXR_NAMESPACE_OPEN_SCOPE`?

pxr/usd/sdf/textParserContext.cpp
```diff
@@ -25,20 +25,16 @@
 #include "pxr/pxr.h"
 #include "pxr/usd/sdf/textParserContext.h"
 
-PXR_NAMESPACE_OPEN_SCOPE
```
nvmkuruc: Why was it necessary to remove this?

pxr/usd/sdf/path.h
```diff
@@ -806,6 +806,29 @@ class SdfPath : boost::totally_ordered<SdfPath>
     /// This returns \c true if IsValidIdentifier() does.
     SDF_API static bool IsValidNamespacedIdentifier(const std::string &name);
 
+    /// Returns whether \p name is a legal prim name for any path component.
+    SDF_API static bool IsValidPrimName(const std::string& name);
+
+    /// Returns whether \p name is a legal prim name for any path component.
+    /// This method uses the value of \p utf8 to determine which rules to
+    /// use to validate, regardless of the value of TF_UTF8_IDENTIFIERS.
+    SDF_API static bool IsValidPrimName(const std::string& name, bool utf8);
```
nvmkuruc: I think I'd leave the `utf8` flag out of the Path API.  It seems like something that would just have to be deprecated once the UTF8 env setting is removed.

pxr/base/tf/unicodeUtils.cpp
```diff
@@ -0,0 +1,150 @@
+//
+// Copyright 2016 Pixar
```
nvmkuruc: Is this the right date?

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,617 @@
...
+
+    void AppendUTF8Char(uint32_t codePoint, std::string& result);
+
+    /// Defines an iterator over a UTF-8 encoded string that extracts unicode code point values.
+    ///
+    /// UTF-8 is a variable length encoding, meaning that one Unicode character can be encoded in UTF-8 as
+    /// 1, 2, 3, or 4 bytes.  This iterator takes care of iterating the necessary characters in a string
+    /// and extracing the Unicode code point of each UTF-8 encoded character in the sequence.
+    ///
+    class utf8_const_iterator : public std::iterator <std::forward_iterator_tag, uint32_t> {
```
nvmkuruc: `std::iterator` is deprecated in C++17.

https://en.cppreference.com/w/cpp/iterator/iterator

nvmkuruc: Having this iterator is great.

pxr/base/tf/stringUtils.h
erslavin: are you referring to the empty line?

pxr/base/tf/stringUtils.cpp
erslavin: sure, can change, probably only saving the multiple evaluations of the `end` condition

pxr/base/tf/stringUtils.cpp
erslavin: it can, I can make that change and use it in `_LessImpl` below

pxr/base/tf/stringUtils.cpp
erslavin: It could

pxr/base/tf/stringUtils.h
erslavin: I would argue against that in this instance - I wanted the `_IsAsciiValue` to be useable only in the translation unit inside the anonymous namespace.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: I want to say these were boilerplate copied from another test and adjusted, but I'll have to go back and verify.  If so, I would recommend against changing it as it would have to be changed in all tests containing it.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: I want to say these were boilerplate copied from another test and adjusted, but I'll have to go back and verify.  If so, I would recommend against changing it as it would have to be changed in all tests containing it.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: I want to say these were boilerplate copied from another test and adjusted, but I'll have to go back and verify.  If so, I would recommend against changing it as it would have to be changed in all tests containing it.

pxr/base/tf/unicodeUtils.cpp
erslavin: No will change

pxr/usd/sdf/textParserContext.cpp
erslavin: There was some refactoring done to make this object useable from both lexers.  If I remember correctly, removing the namespace here made the resulting `.lex` files much simpler (because they don't live in the namespace) but I'd have to go back and verify.

pxr/usd/sdf/textParserContext.h
erslavin: There was some refactoring done to make this object useable from both lexers. If I remember correctly, removing the namespace here made the resulting .lex files much simpler (because they don't live in the namespace) but I'd have to go back and verify.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: For example `testUsdAbcAlembicData.py`

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: For example, `testUsdFileFormats.py`

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: For example, testUsdFileFormats.py

pxr/base/tf/stringUtils.h
erslavin: no, I suppose not

pxr/base/tf/unicodeUtils.h
erslavin: Removed reliance on `std::iterator` by adding `using` statements internal to class

pxr/base/tf/unicodeUtils.cpp
erslavin: Fixed

pxr/usd/sdf/path.h
erslavin: Removed

pxr/base/tf/stringUtils.cpp
erslavin: Changed to range based for

pxr/base/tf/stringUtils.cpp
erslavin: Modified and used in `_LessImpl`

pxr/base/tf/stringUtils.cpp
erslavin: Modified and used in _LessImpl

pxr/base/tf/stringUtils.h
erslavin: Removed from header

pxr/usd/usd/prim.h
erslavin: Fixed

pxr/usd/sdf/textParserContext.cpp
erslavin: Placed them back under the namespace and verified behavior.

pxr/usd/sdf/textParserContext.h
erslavin: Placed them back under the namespace and verified behavior.

pxr/base/tf/stringUtils.h
nvmkuruc: Yup!

erslavin: Fixed

pxr/base/tf/stringUtils.cpp
```diff
@@ -63,6 +63,37 @@ using std::vector;
...
+// to determine identifier validity.  In addition, if this value is true, 
+// slightly different rules will be used to evaluate generic identifiers 
+// vs. prim names.  Setting this value to false will fallback to original
+// ASCII identifier rules.
+TF_DEFINE_ENV_SETTING(TF_UTF8_IDENTIFIERS, 
+    true, 
+    "Allow UTF8 strings as identifiers and prim names");
+
+namespace {
+    bool _IsASCIIValue(const char& c)
```
nvmkuruc: The C++ core guidelines recommend cheap to copy types be passed by value instead of by reference.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

pxr/base/tf/unicode/tfGenCharacterClasses.py
```diff
@@ -0,0 +1,193 @@
...
+#include <utility>
+
+"""
+
+xid_start_class = []
+xid_continue_class = []
+xid_start_range_pairs = []
+xid_continue_range_pairs = []
+
+def _write_cpp_files(destination_directory):
```
nvmkuruc: An explicit type annotation might be helpful here.

pxr/base/tf/unicode/tfGenCharacterClasses.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    Writes the C++ code file that will initialize character class
+    sets with the values read by this script.
+
+    Args:
+        destination_directory: A string defining the path at which the generated cpp file will be written to.
+                               If the specified directory does not exist, it will be created.
+    """
+    if not os.path.exists(destination_directory):
```
nvmkuruc, line 91: Given that Python 2 support is being removed in the next USD release, consider if the more modern `pathlib` is worth using instead for new code.

pxr/base/tf/unicodeUtils.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <tuple>
+#include <algorithm>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+namespace TfUnicodeUtils {
+    ///
+    /// Appends the UTF-8 byte representation of \c codePoint to \c result.
+    /// 
+    void AppendUTF8Char(uint32_t codePoint, std::string& result)
```
nvmkuruc, line 42: This block of code is really useful in succinctly encapsulating the UTF-8 encoding.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    /// The XID_Continue class of characters include those in XID_Start plus characters having the Unicode General Category of nonspacing marks,
+    /// spacing combining marks, decimal number, and connector punctuation.
+    /// That is, the character must have a category of XID_Start | Nd | Mn | Mc | Pc
+    ///
+    /// UTF-8 characters are variable encoded, so \a sequenceStart defines the first byte in the UTF-8 character sequence. This method can be used for entire
+    /// strings by passing identifier.begin() and identifier.end(), but also on sub sequences defined by the given iterator ranges to avoid copying
+    /// the subsequence to a temporary string for evaluation.
+    ///
+    inline TF_API bool IsValidUTF8Identifier(const std::string& identifier, const std::string::const_iterator& sequenceStart, const std::string::const_iterator& end)
```
nvmkuruc: I don't think `TF_API` is required for `inline` functions.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+                // e.g. the byte sequence C2 41 would be detected as a 2-byte 
+                // sequence, but it is invalid because of the second byte signature
+                // by the standard, this should process the C2 as invalid but consume
+                // 41 as a valid 1-byte UTF-8 character
+                codePoint = 0xFFFD;
+            }
+
+            return codePoint;
+        }
+
```
nvmkuruc, line 117: Should you provide an `operator->` implementation?  See earlier note about needing a proxy pointer type.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+                // but we don't like exceptions in USD, so we'll signal this is bad
+                // by setting the code point to 0xFFFD (this mostly happens
+                // when a high / low private surrogate is used)
+                // TODO: note that this isn't precisely conformant, as we likely
+                // consumed an entire sequence when a subset would have been invalid
+                // e.g. the byte sequence C2 41 would be detected as a 2-byte 
+                // sequence, but it is invalid because of the second byte signature
+                // by the standard, this should process the C2 as invalid but consume
+                // 41 as a valid 1-byte UTF-8 character
+                codePoint = 0xFFFD;
```
nvmkuruc: It would be helpful to make a `constexpr` alias for `0xFFFD` to give it a name in order to explain why it's the return value in an error case and make it easier for users to check for the error code point value.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+            }
+        }
+
+        ///
+        /// Retrieves the Unicode code point of the next character in the UTF-8 encoded sequence (defined by
+        /// \a begin) and returns the value in \a codePoint.  This method will return \a true if the encoded
+        /// sequence is valid.
+        /// If the encoding is invalid, this method will return \a false and \a codePoint will be set to 0.
+        ///
+        inline bool GetNextUTF8Character(std::string::const_iterator& begin, uint32_t& codePoint) const
```
nvmkuruc: I'd advocate for returning by value instead of by reference when possible (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-out-multi).

For example, this could return `std::pair<uint32_t, bool>`, similar to `std::map` insertion.

My other observation is that the naming of `GetNextUTF8Character` does not suggest that the input iterator is being modified.  `AdvanceUTF8Character` or `IncrementUTF8Character` might be clearer.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+        // at least one character
+        if (sequenceStart == end)
+        {
+            return false;
+        }
+
+        // first character in the XID_Start character class
+        utf8_const_iterator utf8Iterator(sequenceStart, end);
+        uint32_t codePoint = *utf8Iterator;
+        bool found = xidStartClass.find(codePoint) != xidStartClass.end();
```
nvmkuruc: I'm going to argue that the reuse of `found` potentially complicates the maintenance and readability of the function.  (See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-reuse)

Consider this alternative formulation.
```
if (xidStartClass.find(codePoint) != xidStartClass.end() &&
    std::none_of(xidStartRangeClass.cbegin(), xidStartRangeClass.cend(), [codePoint](const auto& range) { return codePoint >= range.first && codePoint <= range.second; }) {
    return false;
}
```

This reduces the number of branches and avoids sharing `found` between the two parts of the algorithm.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+            // wasn't in the XID start class single or range classes
+            return false;
+        }
+
+        // remaining characters
+        utf8Iterator++;
+        for (; utf8Iterator != end; utf8Iterator++)
+        {
+            codePoint = *utf8Iterator;
+            bool found = (xidStartClass.find(codePoint) != xidStartClass.end()) || (xidContinueClass.find(codePoint) != xidContinueClass.end());
```
nvmkuruc: I would advocate trying to use `std::none_of` to simplify this logic of comparing against classes.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+///
+/// Used by \ref page_tf_StringUtils to determine validity of identifiers and prim names.
+///
+namespace TfUnicodeUtils {
+
+    // 21-bits are required to store the maximum Unicode code point value
+    // so we use an unsigned integer to represent the value
+    extern TF_API std::unordered_set<uint32_t> xidStartClass;
+    extern TF_API std::unordered_set<uint32_t> xidContinueClass;
+    extern TF_API std::vector<std::pair<uint32_t, uint32_t>> xidStartRangeClass;
```
nvmkuruc, line 61: Are the range classes disjoint and ordered?  Can that be used to optimize some of the lookups (ie. binary search)?

I'm wondering if it is possible to avoid exposing the data structures as part of the public API and instead hide the specific data structure choices in wrapper utilities?  Depending on the size of the class, profiling could reveal that a `vector` with binary search outperforms the `unordered_set` (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon2-prefer-using-stl-vector-by-default-unless-you-have-a-reason-to-use-a-different-container) and it would be nice if that was abstracted away.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    /// The XID_Continue class of characters include those in XID_Start plus characters having the Unicode General Category of nonspacing marks,
+    /// spacing combining marks, decimal number, and connector punctuation.
+    /// That is, the character must have a category of XID_Start | Nd | Mn | Mc | Pc
+    ///
+    /// UTF-8 characters are variable encoded, so \a sequenceStart defines the first byte in the UTF-8 character sequence. This method can be used for entire
+    /// strings by passing identifier.begin() and identifier.end(), but also on sub sequences defined by the given iterator ranges to avoid copying
+    /// the subsequence to a temporary string for evaluation.
+    ///
+    inline TF_API bool IsValidUTF8PrimName(const std::string& identifier, const std::string::const_iterator& sequenceStart, const std::string::const_iterator& end)
```
nvmkuruc: Should prim name references be moved to `Sdf`?

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    /// That is, the character must have a category of Lu | Ll | Lt | Lm | Lo | Nl | '_'
+    ///
+    inline TF_API bool IsUTF8CharXIDStart(const std::string& identifier, const std::string::const_iterator& sequenceStart)
+    {
+        // extract the unicode code point value from the UTF-8 encoding
+        utf8_const_iterator utf8Iterator(sequenceStart, identifier.end());
+
+        // check to see whether the code point is in any of the valid singular sets
+        uint32_t codePoint = *utf8Iterator;
+        bool found = xidStartClass.find(codePoint) != xidStartClass.end();
```
nvmkuruc: See earlier note about potentially using `std::none_of`.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    /// UTF-8 characters are variable encoded, so \a sequenceStart defines the first byte in the UTF8 character sequence.
+    /// This method will advance \a sequenceStart iterator to the first byte of the next UTF8 character and evaluate
+    /// if the UTF8 character it read is part of the XID_Start character class.
+    ///
+    /// The XID_Start class of characters are derived from the Unicode General_Category of uppercase letters, lowercase letters, titlecase letters,
+    /// modifier letters, other letters, letters numbers, plus Other_ID_Start, minus Pattern_Syntax and Pattern_White_Space code points.
+    /// That is, the character must have a category of Lu | Ll | Lt | Lm | Lo | Nl | '_'
+    ///
+    inline TF_API bool IsUTF8CharXIDStart(const std::string& identifier, const std::string::const_iterator& sequenceStart)
```
nvmkuruc: Would this function be more generally useful if it took in a `utf8_const_iterator`?  The user wouldn't have to pass in both the string and the sequence start.  I suspect this would make the function usable in utilities like `IsValidUTF8Identifier`, reducing duplicate code.

nvmkuruc: I don't think `inline` functions require `TF_API`.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    /// UTF8 characters are variable encoded, so \a sequenceStart defines the first byte in the UTF8 character sequence.
+    /// This method will advance \a sequenceStart iterator to the first byte of the next UTF8 character and evaluate
+    /// if the UTF8 character it read is part of the XID_Continue character class.
+    ///
+    /// The XID_Continue class of characters include those in XID_Start plus characters having the Unicode General Category of nonspacing marks,
+    /// spacing combining marks, decimal number, and connector punctuation.
+    /// That is, the character must have a category of XID_Start | Nd | Mn | Mc | Pc
+    ///
+    inline TF_API bool IsUTF8CharXIDContinue(const std::string& identifier, const std::string::const_iterator& sequenceStart)
```
nvmkuruc: See earlier notes about `std::none_of`, using `utf8_const_iterator` as the iterator, and `TF_API` / `inline`.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    std::string TF_API MakeValidUTF8Identifier(const std::string& identifier);
+
+    ///
+    /// Constructs a valid result prim name from \a primName.
+    /// If \a primName is already valid, then return value of this method should be value stored in \a primName.
+    ///
+    /// A prim name is valid according to the rules associated with \a IsValidUTF8PrimName.
+    ///
+    std::string TF_API MakeValidUTF8PrimName(const std::string& primName);
```
nvmkuruc: See earlier note about references to prim name.

pxr/base/tf/unicode/tfGenCharacterClasses.py
```diff
@@ -0,0 +1,193 @@
...
+    arguments = _parseArguments()
+
+    # parse the UnicodeDatabase.txt file
+    # each line represents a single code point with information about the character
+    # represented by that code point
+    # for now we are only interested in the code point itself and
+    # the character class, which reside in columns 0 and 2 respectively
+    unicode_database_file_name = os.path.join(arguments.srcDir, UNICODE_DATABASE_FILE)
+    if not os.path.exists(unicode_database_file_name):
+        raise RuntimeError("Error in script: Could not find 'UnicodeDatabase.txt' at path %s!" % (arguments.srcDir))
```
nvmkuruc: Given that Python 2 is being deprecated in the next release, consider if `f` strings are worth using here instead of `%` formatting.  They're usually more readable and apparently more performant.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    ///
+    class utf8_const_iterator {
+
+    public:
+
+        using iterator_category = std::forward_iterator_tag;
+        using value_type = uint32_t;
+        using difference_type = ptrdiff_t;
+        using pointer = uint32_t*;
+        using reference = uint32_t&;
```
nvmkuruc: `operator*` returns a `uint32_t`, so I believe the `reference` type should be `uint32_t`.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+    /// and extracing the Unicode code point of each UTF-8 encoded character in the sequence.
+    ///
+    class utf8_const_iterator {
+
+    public:
+
+        using iterator_category = std::forward_iterator_tag;
+        using value_type = uint32_t;
+        using difference_type = ptrdiff_t;
+        using pointer = uint32_t*;
```
nvmkuruc: I suspect you're going to need to create a proxy type for `pointer` since `operator*` returns a temporary.  It also should probably be explicitly `const` since the type advertises itself as a `const_iterator`.

pxr/base/tf/unicodeUtils.h
```diff
@@ -0,0 +1,625 @@
...
+        }
+
+    private:
+
+        ///
+        /// Retrieves the variable encoding length of the UTF-8 character currently pointed to by the iterator.
+        /// This can be 1, 2, 3, or 4 depending on the encoding of the UTF-8 character.
+        /// If the encoding cannot be determined, this method will return 0.
+        ///
+        inline std::iterator_traits<std::string::const_iterator>::difference_type GetEncodingLength() const
```
nvmkuruc: It's convention in `pxr` for private methods to have `_` as a prefix.

pxr/base/tf/unicodeUtils.h
nvmkuruc: It's convention in `pxr` for private methods to have `_` as a prefix.

pxr/base/tf/unicodeUtils.h
nvmkuruc: It does not appears as if `identifier` is used in this implementation.

pxr/base/tf/unicodeUtils.h
nvmkuruc: It does not appear as if `identifier` is used in this implementation.

pxr/usd/sdf/textFileFormat.cpp
```diff
@@ -374,7 +374,6 @@ SdfTextFileFormat::ReadFromString(
             TfDynamic_cast<SdfDataRefPtr>(data), &hints)) {
         return false;
     }
-
```
nvmkuruc, line 377: Is the white space change necessary?

pxr/usd/sdf/testenv/testSdfPathParser.cpp
```diff
@@ -233,11 +232,27 @@ int main()
         NULL
     };
 
+    char const* goodUtf8[] = {
```
nvmkuruc, line 239: Is the reason that this is "goodUtf8" because the changes that allow for UTF-8 change the class of valid prim identifiers to just the continue class?

pxr/base/tf/unicodeUtils.h
nvmkuruc: Alternatively, since the rules for prim names are just valid sequences of XID_Continue / XID_Start, you could probably come up with a more generic  utility name that doesn't explicitly reference prims.

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: Recommend leaving this as is because it's consistent with current tests and nothing is lost by the explicit semantics here

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: Recommend leaving this as is because it's consistent with current tests and nothing is lost by the explicit semantics here

pxr/usd/usd/testenv/testUsdLoadUtf8Content.py
erslavin: I'm not sure what the issue is here, the documentation states `Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).`  The `f.close` statement here makes sure it works across both platforms, but doesn't take anything away on Unix-based systems.

pxr/base/tf/unicode/tfGenCharacterClasses.py
erslavin: The removal of Python 2 does not deprecate the methods present in the `os.path` module.  I think that this method is semantically clear in what it is trying to do and no further path manipulation is being done to make `pathlib` a clearer option in this case.

pxr/base/tf/unicodeUtils.h
erslavin: While it isn't technically required, it does allow clients to get a pointer to the method that resides in the module if desired.  99% of use cases won't require this so will remove.

pxr/base/tf/unicodeUtils.h
erslavin: For the first part, there are methods that are constructed this way all over the USD code base (e.g., the `Has` method of `SdfAbstractData`).  I personally find this more readable than returning a `std::pair` in this case.  Recommend leaving as is.

For the second part, I would agree, will adjust naming.

pxr/base/tf/unicodeUtils.h
erslavin: I'm just going to remove the "Prim" part of this to turn it into `IsValidUTF8Name`

pxr/base/tf/unicodeUtils.h
erslavin: see above

pxr/base/tf/unicodeUtils.h
erslavin: see above

pxr/base/tf/unicodeUtils.h
erslavin: removed "Prim" part of method name

pxr/base/tf/unicodeUtils.h
erslavin: not sure what you mean here.  Here is the definition of `std::iterator` in msvc:
```
template <class _Category, class _Ty, class _Diff = ptrdiff_t, class _Pointer = _Ty*, class _Reference = _Ty&>
struct _CXX17_DEPRECATE_ITERATOR_BASE_CLASS iterator { // base type for iterator classes
    using iterator_category = _Category;
    using value_type        = _Ty;
    using difference_type   = _Diff;
    using pointer           = _Pointer;
    using reference         = _Reference;
};
```
All I did was replace the use of the iterator derivation with the appropriate `using` statements as defined here.

pxr/base/tf/unicodeUtils.h
erslavin: see above

pxr/usd/sdf/testenv/testSdfPathParser.cpp
erslavin: Yes, this is correct

pxr/base/tf/unicodeUtils.h
erslavin: Refactored and simplified logic between the two methods using this approach

pxr/base/tf/unicodeUtils.h
erslavin: Refactored and simplified logic between the two methods using this approach

pxr/usd/sdf/testenv/testSdfPathParser.cpp
nvmkuruc: It might be helpful to leave a comment here noting it can be removed once the utf8 env setting is phased out.

pxr/base/tf/unicodeUtils.h
erslavin: Changed to `uint32_t`

pxr/base/tf/unicodeUtils.h
erslavin: No `operator->` is necessary here as we discussed as the requirement is removed going forward.  Changed `pointer` alias to `void`

pxr/base/tf/unicodeUtils.h
erslavin: Refactored this method to take a code point and refactored callers appropriately

nvmkuruc:
@erslavin I think we can close this PR now that it's been broken up and integrated in 24.03.

erslavin:
Old PR

PR #2124, opened 2022-12-07
APPLE: Rename MemoryBarrier func to InsertMemoryBarrier to avoid Windows issue

creijon:
### Description of Change(s)

In `Windows.h` there is the statement: `#define MemoryBarrier __faststorefence`. Depending on whether this header is included and in which order it is, this results in the `MemoryBarrier` symbol on `HgiGraphicsCmds` `HgiComputeCmds` and `HgiBlitCmds` being renamed as `__faststorefence`

We had to work around this in the GPU frustum culling PR, https://github.com/PixarAnimationStudios/USD/pull/2053, by using an `#undef`, but this is fragile since re-ordering the includes could cause a regression.

Perhaps a more effective solution is to rename the function?  In this case to `InsertMemoryBarrier`.

### Fixes Issue(s)
- Potential build issues on Windows platforms.

### DIFF
--- a/pxr/imaging/hdSt/indirectDrawBatch.cpp
+++ b/pxr/imaging/hdSt/indirectDrawBatch.cpp
@@ -1423,7 +1423,7 @@ HdSt_IndirectDrawBatch::_ExecuteFrustumCull(
 
         // Make sure the reset-pass memory writes
         // are visible to the culling shader pass.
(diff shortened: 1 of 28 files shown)


sunyab:
Filed as internal issue #USD-7820

PR #2126, opened 2022-12-07
Draft: Spec and Grammar for USDA 

dgovil:
### Description of Change(s)

This is a draft PR to create a specification for the USDA file format.
It includes a PEG grammar for USDA, and explanations of various elements.
The PEG grammar is fairly vanilla and geared towards the parsimonious Python library.

@meshula had asked for railroad diagrams, however I wanted to handle that as a V2 of this. It should be straightforward to setup later, but I didn't want to make this larger than it had to be

I also stubbed in a `unicode_identifier` based on a C# identifier regex. It is definitely not correct, and I'd appreciate if I could get some advice/help from  @erslavin on that front to line up with his PR: https://github.com/PixarAnimationStudios/USD/pull/2120

A preview is included here:
[USDA File Format Specification — Universal Scene Description 23.02 documentation.pdf](https://github.com/PixarAnimationStudios/USD/files/10180841/USDA.File.Format.Specification.Universal.Scene.Description.23.02.documentation.pdf)

### DIFF
--- a/docs/_ext/usda.peg
+++ b/docs/_ext/usda.peg
@@ -1,0 +1,84 @@
+# Top Level Parsing Element
+layer = format
+        whitespace
+        (whitespace_newline metadata_block?)?
+        (whitespace_newline prim_spec*)?
+        whitespace
+
+
+# Base Components
+space = ~"([ \s])"
+any_space = (space/"\t")
+newline = "\n"/"\r\n"/"\r"/ "\u2028"/"\u2029"
+whitespace = (any_space/newline/comment)*
+whitespace_newline = whitespace # Indicates requiring a newline
+comment_start = "#"
+comment = comment_start ~"[^\n]*"
+boolean = "true"/"false"
+digits = ~"\d+"
+integer = "-"? digits
+exponent = "e" digits
+float = (integer ("." digits exponent?)?)/("." (digits exponent?)?)
+numerical = float
+identifier = ascii_identifier/unicode_identifier
+unicode_identifier = ~"^(?:((?!\d)\w+(?:\.(?!\d)\w+)*)\.)?((?!\d)\w+)"
+ascii_identifier = ~"[a-zA-Z_]+[a-zA-Z_0-9]*"
+string = string_double_quotes/string_single_quotes/string_triple_quotes
+string_double_quotes = ~"\".*\""
+string_single_quotes = ~"'.*'"
+string_triple_quotes = ~"\"\"\".*\"\"\""
+
+tuple = "(" any_space* (value any_space* "," any_space*)* value any_space*")"
+assignment = any_space* "=" any_space*
+value = (boolean/numerical/string/tuple/array/reference/dictionary/timesamples)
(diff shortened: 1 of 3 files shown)


dgovil:
If you want to try parsimonious directly with the parser, here is a snippet (and a file with a Cube exported from Maya)
[foo.usda.zip](https://github.com/PixarAnimationStudios/USD/files/10181327/foo.usda.zip)

```
from parsimonious import Grammar

with open("usda.peg", "r") as fp:
    grammar = Grammar(fp.read())

with open("/Users/dhruvgovil/Downloads/foo.usda", "r") as fp:
    parsed = grammar.parse(fp.read())

print(parsed)
```

Replace the file as needed.
I ran this through a bunch of the USDA files in the usd repo by finding and looping through them all. 
Note there are some issues, that I think are issues with parsimonious's packrat implementation where I believe the PEG definition is correct but parsimonious fails.

It doesn't seem to like falling back to the secondary option if the first option fails to capture everything after it. So for example if you had `integer/float` and hit `1.0` it would error because integer fails, but it should have fallen back to float. This is why I handle all numericals as float. I think the PEG is correct but their lib isn't falling back, but if anyone has corrections to the grammar I'd be happy to see how to resolve that.

While this is solvable for numericals, one instance where I couldn't solve it is arrays with a trailing `,`. Since parsimonious is matching `(value + ",")* + (value + ","?)` , the first group catches the last group and then sees that the last value is missing. This should be fine in a standard PEG as I understand it, but maybe there's a way to describe this in a grammar appropriately for the library.

sunyab:
Filed as internal issue #USD-7822

syoyo:
@dgovil `varying` Variability keyword is missing in `property_specifiers`(and `varying` only appears in relationship). Also, there is a precedence in `property_specifiers` items in pxrUSD's USDA parser.
 https://github.com/PixarAnimationStudios/USD/blob/0c7b9a95f155c221ff7df9270a39a52e3b23af8b/pxr/usd/sdf/textFileFormat.yy#L2815

Some examples:

```
#usda 1.0

def "bora"
{
  varying append rel myrel = </bora> # usdcat NG
  append varying rel myrel = </bora> # usdcat OK
  custom append varying rel myrel = </bora> # usdcat NG
  append custom varying rel myrel = </bora> # usdcat OK
}
```

dgovil:
Thanks! Not sure how I missed that. The folks at NVidia also had a few notes too that I'll wait for and then take another pass at this.

mchaput:
The PEG uses simple regexes for strings, implying that usda doesn't allow any escaping inside strings. Is that true, or is it just a shortcut? I don't know enough C/Bison to check if it's the same in the .yy file...

syoyo:
USDA has somewhat special treatment for escaping characters in a string(especially triple quoted string and control sequence character). 

Take a look at TinyUSDZ's string escaping/unescaping function

https://github.com/syoyo/tinyusdz/blob/a01aa7d73de375767d1629ff8bd2c0f181ccedbb/src/str-util.cc#L9

and files with `quotes` or `escape` in its filename: https://github.com/syoyo/tinyusdz/tree/dev/tests/usda

It'd be better that string escapement rule is included in the Spec.

PR #2127, opened 2022-12-08
Fix windows debug build assertion in std::stl

AdamFelt:
On a windows debug build we were faced with few crashes due to asserts in std::stl. The crash is tripped inside pxr/usdImaging/usdImaging/delegate.cpp#L1100 .  In there, the alg. is multithreaded with a final wait for completion in line 1105.  This problem appears to be related to threading. In each thread, one is passing usdPath by reference which is not thread safe.
This change makes it thread safe by passing a const reference.


sunyab:
Filed as internal issue #USD-7823

PR #2131, opened 2022-12-12
[usdShade] GetPrimvarNamesMetadataString - only append non-empty string

pmolodo:
### Description of Change(s)

on some compilers / architectures, if you do:

    metadata["primvars] = GetPrimvarNamesMetadataString(...)

...this results in a default / empty string being inserted into metadata["primvars"] BEFORE GetPrimvarNamesMetadataString is called. Therefore, we only insert existing non-empty strings into the list that GetPrimvarNamesMetadataString builds. (And it's a good check in general anyway.)

- https://github.com/PixarAnimationStudios/USD/pull/2115
- https://github.com/PixarAnimationStudios/USD/pull/2154

### Fixes Issue(s)
- failure of testUsdShadeShaderDef  on linux aarch64 using gcc 7.5.


sunyab:
Filed as internal issue #USD-7830

PR #2133, opened 2022-12-13
[UsdUtils] Add UDIM support for dependencies/usdz and switch material…

lbiasco:
…ParamUtils to use shared functions

This PR is a continuation of the work from #1477, which is closed since the related accounts are no longer accessible. I tried to vet these changes as much as possible, but I wouldn't be surprised if there are some uses of UDIMs that I missed, so definitely let me know of any uncommon cases to include tests for.

### Description of Change(s)

- Added three UDIM utility functions to UsdUtils: `UsdUtilsIsUdimIdentifier`, `UsdUtilsGetUdimFiles`, and `UsdUtilsResolveUdimPath`. These are also Python-wrapped.
- Switched UsdImaging's materialParamUtils to use these functions (the previous UDIM utilities were moved and are used by the functions).
- Changed _AssetLocalizer to use these functions and include UDIM files with dependency computation and USDZ creation.
- Added a test module for these UDIM functions and an additional dependency computation test. I didn't add a new USDZ test since the dependency test should account for the changes.

### DIFF
--- a/pxr/usd/usdUtils/CMakeLists.txt
+++ b/pxr/usd/usdUtils/CMakeLists.txt
@@ -82,6 +82,7 @@ pxr_test_scripts(
     testenv/testUsdUtilsStitch.py
     testenv/testUsdUtilsStitchClips.py
     testenv/testUsdUtilsTimeCodeRange.py
+    testenv/testUsdUtilsUdims.py
     testenv/testUsdUtilsUpdateSchemaWithSdrNode.py
     testenv/testUsdUtilsVarSelsSessionLayer.py
 )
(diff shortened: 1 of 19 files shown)


sunyab:
Filed as internal issue #USD-7832

sunyab:
Hi @lbiasco, just wanted to mention that we iterated on your change internally and wound up with something a bit larger than your PR here. We've committed that separately and will close out this PR when that commit hits the dev branch shortly. Thank you!

lbiasco:
Awesome, thank you for the update!

sunyab:
Closing per previous comment, thanks!

PR #2138, opened 2022-12-27
[tf] fix docstring of TfGetPathName and TfGetBaseName

pmolodo:
### Description of Change(s)

The docstring for TfGetPathName stated that:

    TfGetPathName(s)+TfGetBaseName(s) == s

"as long as \c s doesn't end with multiple adjacent slashes".  This is incorrect - the equation holds only if s doesn't end with ANY trailing slashes, as TfGetBaseName first strips all trailing slashes.  Fixed it to have the correct condition.

Also added further clarification to the docstring of TfGetBaseName to note it's strailing-slash stripping behavior.

### Fixes Issue(s)
- Incorrect docstring for TfGetPathName


sunyab:
Filed as internal issue #USD-7837

PR #2144, opened 2023-01-06
APPLE: Fix for building boost 1.78 on macOS universal

creijon:
### Description of Change(s)

With the latest dev push, boost has been updated to 1.78. 

With this version b2 no longer requires the `architecture=combined` option for universal builds and throws an error if it is defined.

### Fixes Issue(s)
- Universal builds on macOS

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -707,7 +707,7 @@ elif Windows():
     # boost 1.70 is required for Visual Studio 2019. For simplicity, we use
     # this version for all older Visual Studio versions as well.
     # boost 1.78 is required for Visual Studio 2022.
-    if VisualStudio2022OrGreater():
+    if IsVisualStudio2022OrGreater():
         BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.tar.gz"
         BOOST_VERSION_FILE = "include/boost-1_78/boost/version.hpp"
     else:


sunyab:
Filed as internal issue #USD-7845

build_scripts/build_usd.py
```diff
@@ -762,6 +758,10 @@ def InstallBoost_Helper(context, force, buildArgs):
                                 " linkflags=\"{0}\"".format(macOSArch)
             bootstrapCmd += " --with-toolset=clang"
 
+            # If Python2 is also installed in the system then it is sometimes
+            # found first, so we force the bootstrap to Python3
+            bootstrapCmd += " --with-python-version=3"
+
```
sunyab: It seems like this ought to only be needed if `context.buildPython` is enabled. I think it could also use the information returned by `GetPythonInfo` to specify the full version, like:

```
if context.buildPython:
    pythonInfo = GetPythonInfo(context)
    bootstrapCmd += " --with-python-version={ver}".format(ver=pythonInfo[3])
```

In theory this seems like this should be independent of macOS but I'm OK with leaving it in this block for now and just adding a comment saying something to that effect.

sunyab (changes requested):

build_scripts/build_usd.py
creijon: After testing this on the latest public release of macOS Ventura it seems that the `--with-python-version=3` option isn't necessary.  It only causes a problem for our internal builds.  So I think I can remove it.  The other change, to remove the `combined` option is needed with 1.78 to build universal binaries.

creijon:
Sorry, I force-pushed this by accident.  Will re-open under a new PR.

creijon:
Re-created as https://github.com/PixarAnimationStudios/USD/pull/2150

PR #2154, opened 2023-01-11
[usdSkel] testUsdSkelBakeSkinning - only do usda diff on x86_64 arch

pmolodo:
### Description of Change(s)

...as floating point precision issues may result in differences on other architectures.

(ie, the test would fail on linux aarch64 hardware)

This change follows the pattern from testUsdDancingCubesExample:

https://github.com/PixarAnimationStudios/USD/blob/4ee62b99c91afbea38f2afd855fbd21cb23d6e64/extras/usd/examples/usdDancingCubesExample/CMakeLists.txt#L37

- https://github.com/PixarAnimationStudios/USD/pull/2115
- https://github.com/PixarAnimationStudios/USD/pull/2131

### Fixes Issue(s)
- Failure of testUsdSkelBakeSkinning on aarch64 hardware

### DIFF
--- a/pxr/usd/usdSkel/CMakeLists.txt
+++ b/pxr/usd/usdSkel/CMakeLists.txt
@@ -142,12 +142,25 @@ pxr_register_test(testUsdSkelAnimQuery
     EXPECTED_RETURN_CODE 0
 )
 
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-7867

PR #2157, opened 2023-01-13
PVS-Studio static analyzer fixes for base/arch

dimhotepus:
### Description of Change(s)

Run PVS-Studio static analyzer and fix found issues.

### DIFF
--- a/pxr/base/arch/assumptions.cpp
+++ b/pxr/base/arch/assumptions.cpp
@@ -68,7 +68,7 @@ Arch_ObtainCacheLineSize()
     size_t total = bufferSize / sizeof(INFO);
 
     // Allocate the array of processor INFOs.
-    std::unique_ptr<INFO[]> buffer(new INFO[total]);
+    std::unique_ptr<INFO[]> buffer = std::make_unique<INFO[]>(total);
 
     size_t lineSize = 0;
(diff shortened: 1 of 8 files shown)


pxr/base/arch/systemInfo.cpp
```diff
@@ -89,10 +89,7 @@ _DynamicSizedRead(
     const std::function<bool(char*, size_t*)>& callback)
 {
     // Make a buffer for the data.
-    // We use an explicit deleter to work around libc++ bug.
```
meshula: It appears the fix associated with the linked bug has not yet been released?

https://github.com/llvm/llvm-project/commit/85d3e7a7299f0db3286cf0b611bdfac79f39e967

dimhotepus: My bad, thought closed bug means fix is merged.

Dropped my commit, thanks!

meshula: It's hard to track old LLVM bugs because they transitioned to a new system, and that bug is from the old system. As best as I can tell the fix is in a preview release of an upcoming llvm, but that is, as best as I can tell ;)

sunyab:
Filed as internal issue #USD-7869

Issue #2170, opened 2023-01-17
testUsdBugs fails with float valued file truncation


nvmkuruc:
### Description of Issue

When the passed value to `truncate` is a `float`, the following error message is produced

```
======================================================================
ERROR: test_USD_5196 (__main__.TestUsdBugs)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mkuruc/code/inst/USD/tests/testUsdBugs", line 433, in test_USD_5196
    fobj.truncate(size / 2)
TypeError: 'float' object cannot be interpreted as an integer

----------------------------------------------------------------------
Ran 13 tests in 0.992s
```

Converting to an explicit integer division avoids the issue.

### Steps to Reproduce

```
ctest -C Release -R testUsdBugs --output-on-failure
```

### System Information (OS, Hardware)

Ubuntu 22.04.1 LTS

### Package Versions

Python 3.10
Boost 1.78
TBB 2020.3

### Build Flags
`--no-imaging --tests`

nvmkuruc:
I have a fix.

sunyab:
Filed as internal issue #USD-7881

nvmkuruc:
@jesschimein I think this can be closed.  The fix was merged last year.

jesschimein:
Thanks Matt!  Closing this issue related to #2171

PR #2171, opened 2023-01-17
Use explicit integer division as argument to truncate in testUsdBugs

nvmkuruc:
### Description of Change(s)

Explicit integer division avoids passing a float to truncate.

### Fixes Issue(s)
#2170


sunyab:
Filed as internal issue #USD-7882

Issue #2172, opened 2023-01-18
Prefer `TfHash` to `boost::hash`


nvmkuruc:
### Description of Issue

A large number of the non-python boost dependencies in the USD code base are dependencies on `boost/functional/hash`.  `pxr/base/tf` has its own hashing mechanism that has already been successfully integrated across the USD code base.  Switching to it reduces the number of dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

sunyab:
Filed as internal issue #USD-7883

nvmkuruc:
This is done.

```grep -r boost::hash pxr/``` only returns references to comments about `hash_value` ADL.

PR #2173, opened 2023-01-18
Replace boost hash usage with TfHash in pxr/base/tf

nvmkuruc:
### Description of Change(s)

To remove the dependency of `pxr/base/tf` on boost's hashing functions
* `fileUtils` replaces `boost::hash_combine` with `TfHash::Combine`
* `TfRefPtr` and `TfWeakPtrFacade` now implement `hash_value` in terms of the existing `TfHashAppend`
* Hash test coverages is added to `TfRefPtr` and `TfWeakPtr` tests
* Until support for `std::hash` is enabled, an overload for `std::type_index` to `TfHashAppend` is now provided
* Support is added for `std::set` and `std::map` to `TfHash`.  This will facilitate changes changes in other downstream libraries.

### Fixes Issue(s)
-#2172 (first of several forthcoming PRs)

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -33,6 +33,9 @@
 
 #include <cstring>
 #include <string>
+#include <map>
+#include <set>
+#include <typeindex>
 #include <type_traits>
 #include <utility>
 #include <vector>
@@ -92,6 +95,35 @@ inline void
 TfHashAppend(HashState &h, std::vector<T> const &vec)
 {
     h.AppendContiguous(vec.data(), vec.size());
+}
+
+// Support std::set.
+// NOTE: Supporting std::unordered_set is complicated because the traversal
+// order is not guaranteed
+template <class HashState, class T, class Compare>
+inline void
+TfHashAppend(HashState& h, std::set<T, Compare> const &elements)
+{
+    h.AppendRange(std::begin(elements), std::end(elements));
+}
+
+// Support std::map.
+// NOTE: Supporting std::unordered_map is complicated because the traversal
+// order is not guaranteed
+template <class HashState, class Key, class Value, class Compare>
+inline void
+TfHashAppend(HashState& h, std::map<Key, Value, Compare> const &elements)
+{
+    h.AppendRange(std::begin(elements), std::end(elements));
+}
+
+// Support std::type_index.  When TfHash support for std::hash is enabled,
+// this explicit specialization will no longer be necessary.
+template <class HashState>
+inline void
+TfHashAppend(HashState& h, std::type_index const &type_index)
+{
+    return h.Append(type_index.hash_code());
 }
 
 // Support for hashing std::string.
--- a/pxr/base/tf/testenv/hash.cpp
+++ b/pxr/base/tf/testenv/hash.cpp
@@ -294,6 +294,12 @@ Test_TfHash()
     std::vector<int> vint = {1, 2, 3, 4, 5};
     printf("hash(vector<int>): %zu\n", h(vint));
 
+    std::set<int> sint = {1, 2, 3, 4, 5};
+    printf("hash(set<int>): %zu\n", h(sint));
+
+    std::map<int, uint32_t> mint = {{-1, 1}, {2, 3}, {-4, 5}};
+    printf("hash(map<int, uint32_t>): %zu\n", h(mint));
+
     std::pair<int, float> intfloat = {1, 2.34};
     printf("hash(pair<int, float>): %zu\n", h(intfloat));
 
@@ -305,6 +311,9 @@ Test_TfHash()
 
     printf("combine hash of the 3: %zu\n",
            TfHash::Combine(vint, intfloat, vp));
+
+    // Validate support for std::type_index
+    printf("hash(type_index): %zu\n", h(std::type_index(typeid(int))));
 
     TfHasher tfh;
     //BoostHasher bh;
--- a/pxr/base/tf/testenv/weakPtr.cpp
+++ b/pxr/base/tf/testenv/weakPtr.cpp
@@ -124,6 +124,19 @@ static void _TestComparisons()
     TF_AXIOM( !(NULL == x) );
 }
 
+static void _TestHash()
+{
+    TF_AXIOM(TfHash()(MonkeyInterfaceWeakPtr()) ==
+             TfHash()(MonkeyInterfaceWeakPtr(nullptr)));
+    {
+        Human h;
+        MonkeyInterfaceWeakPtr p(&h);
+        TF_AXIOM(TfHash()(p) == TfHash()(p));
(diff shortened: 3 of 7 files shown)


sunyab:
Filed as internal issue #USD-7884

pxr/base/tf/testenv/hash.cpp
```diff
@@ -294,6 +295,12 @@ Test_TfHash()
     std::vector<int> vint = {1, 2, 3, 4, 5};
     printf("hash(vector<int>): %zu\n", h(vint));
 
+    std::set<int> sint = {1, 2, 3, 4, 5};
+    printf("hash(set<int>): %zu\n", h(sint));
+
+    std::map<int, uint> mint = {{-1, 1}, {2, 3}, {-4, 5}};
```
spiffmon: Not sure whether it will affect subsequent pr’s but noting we had to replace `uint` with `uint32_t` for windows, @nvmkuruc

nvmkuruc: PR updated to use `uint32_t` and dependent PRs have all been updated.

pxr/base/tf/hash.h
```diff
@@ -33,6 +33,9 @@
 
 #include <cstring>
 #include <string>
+#include <set>
+#include <map>
+#include <typeindex>
```
spiffmon: WBN to maintain headers in alphabetical order here and in the other files in this change.

pxr/base/tf/hash.h
```diff
@@ -94,6 +97,33 @@ TfHashAppend(HashState &h, std::vector<T> const &vec)
     h.AppendContiguous(vec.data(), vec.size());
 }
 
+// Support std::set
```
spiffmon: tiniest nit: the other comments for the other "Support <type>." have '.' at the end. might as well keep this consistent.

pxr/base/tf/hash.h
```diff
@@ -94,6 +97,33 @@ TfHashAppend(HashState &h, std::vector<T> const &vec)
...
+template <class HashState, class T>
+inline void
+TfHashAppend(HashState& h, std::set<T> const &elements)
```
spiffmon: Why do we not support custom comparators for std::set but we do for std::map?

pxr/base/tf/hash.h
```diff
@@ -94,6 +97,33 @@ TfHashAppend(HashState &h, std::vector<T> const &vec)
...
+{
+    h.AppendRange(std::begin(elements), std::end(elements));
+}
+
+// Support std::map
+// NOTE: Supporting std::unordered_map is complicated because the traveral
```
spiffmon: Typo: "traveral" -> "traversal"

It might be worth adding a similar comment above std::set regarding std::unordered_set.

pxr/base/tf/hash.h
```diff
@@ -94,6 +97,33 @@ TfHashAppend(HashState &h, std::vector<T> const &vec)
...
+// order is not guaranteed
+template <class HashState, class Key, class Value, class Compare>
+inline void
+TfHashAppend(HashState& h, std::map<Key, Value, Compare> const &elements)
+{
+    h.AppendRange(std::begin(elements), std::end(elements));
+}
+
+// Support std::type_index.  When TfHash support for std::hash is enabled,
+// this expliclit specialization will no longer be necessary.
```
spiffmon: Typo: "expliclit" -> "explicit"

pxr/base/tf/testenv/weakPtr.cpp
```diff
@@ -124,6 +124,24 @@ static void _TestComparisons()
     TF_AXIOM( !(NULL == x) );
 }
 
+static void _TestHash(){
+    TF_AXIOM(TfHash()(MonkeyInterfaceWeakPtr()) == TfHash()(MonkeyInterfaceWeakPtr(nullptr)));
```
spiffmon: WBN to follow formatting in this file and put brace on new line. Also, I think the first line probably violates our 80 char width guideline?

pxr/base/tf/testenv/weakPtr.cpp
```diff
@@ -124,6 +124,24 @@ static void _TestComparisons()
...
+        Human h;
+        MonkeyInterfaceWeakPtr p(&h);
+        TF_AXIOM(TfHash()(p) == TfHash()(p));
+        TF_AXIOM(TfHash()(p) == TfHash()(MonkeyInterfaceWeakPtr(p)));
+    }
+    {
+        // Verify collisions are unlikely
+        Human h1, h2;
+        MonkeyInterfaceWeakPtr p1(&h1), p2(&h2);
+        TF_AXIOM(TfHash()(p1) != TfHash()(p2));
```
spiffmon: It doesn't seem robust/possible to verify that something is "unlikely".  Unless we are 100% sure (or are making a defining statement) that equal hashes are **impossible** here, then the test could fail randomly, and we suggest removing it.

spiffmon (changes requested):
Hey @nvmkuruc , passing on a few notes from Alex, Ed, and Sunya, mostly identifying a few patterns for this and other PR's.  With these addressed, the other PR's in this sequence should go much more quickly.  Thank you!!

pxr/base/tf/hash.h
nvmkuruc: No reason other than it became necessary for `std::map` in other downstream changes but not `std::set`.  I'm updating `std::set`.

pxr/base/tf/testenv/weakPtr.cpp
nvmkuruc: This makes sense to remove.  Running that test I think was necessarily for validation, but I agree it's not obvious how to include it in the unit test.

pxr/base/tf/testenv/hash.cpp
```diff
@@ -28,6 +28,7 @@
 #include "pxr/base/tf/stopwatch.h"
 #include "pxr/base/tf/token.h"
 #include "pxr/base/tf/weakPtr.h"
+#include "pxr/base/tf/hash.h"
```
spiffmon: can we sort this up?

pxr/base/tf/weakPtrFacade.h
```diff
@@ -29,10 +29,10 @@
 #include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/refPtr.h"
 #include "pxr/base/tf/weakBase.h"
+#include "pxr/base/tf/hash.h"
```
spiffmon: also sort up?

spiffmon (commented):
Just two little cleanliness things - thanks, @nvmkuruc !

pxr/base/tf/weakPtrFacade.h
nvmkuruc: Done

PR #2176, opened 2023-01-19
Replace boost hash usage with TfHash in pxr/base/vt

nvmkuruc:
### Description of Change(s)
Requires #2173, #2304, #2298

To remove the dependency of pxr/base/vt on boost's hashing functions
* `VtDictionary` defers its hashing to `TfHash`'s default `std::map` hashing
* `VtArray` implements a `TfHashAppend` overload to take advantage of the hashability of its contiguous elements.  `hash_value` has been modified to use this via `TfHash`.
* Remove explicit `hash_value` support in `hash.h` as `TfHash` will fallback to `hash_value` if no explicit `TfHashAppend` is defined.  This may change the order of which hash implementation is chosen (now preferring `TfHashAppend` to `hash_value`) but the implementations should not vary for most types.
* `testVtCpp` now provides test coverage for `VtArray` hashing
* Since `TfHash` supports hashing via `std::type_index` and `std::pair`, an explicit implementation hashing the type index pair is no longer necessary

### Fixes Issue(s)

### DIFF
--- a/pxr/base/vt/array.h
+++ b/pxr/base/vt/array.h
@@ -38,7 +38,6 @@
 #include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/mallocTag.h"
 
-#include <boost/functional/hash.hpp>
 #include <boost/iterator_adaptors.hpp>
 #include <boost/iterator/reverse_iterator.hpp>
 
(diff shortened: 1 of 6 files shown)


sunyab:
Filed as internal issue #USD-7931

PR #2178, opened 2023-01-20
Export the Sync function in HdTask to allow for custom render task creation and use.

AdamFelt:
We would like to create new HdTasks outside of the hdx project which is not possible unless the HdTask::Sync function is exported.

Exporting this function allows for the creation and use of custom render tasks.

- [ X] I have submitted a signed Contributor License Agreement


sunyab:
Filed as internal issue #USD-7932

PR #2179, opened 2023-01-20
Fix the install syntax for testArchStackTrace.pdb to use TARGET_PDB_FILE 

AdamFelt:
Fix the install syntax for testArchStackTrace.pdb to use the TARGET_PDB_FILE directive which is the preferred way to install pdb files.
We have problems building and running unit tests locally without this fix.


sunyab:
Filed as internal issue #USD-7900

PR #2182, opened 2023-01-20
Replace boost hash usage with TfHash in pxr/usd/sdf/path.cpp

nvmkuruc:
### Description of Change(s)
To remove the dependency of pxr/usd/sdf/path.cpp on boost's hashing functions
* Replaces `boost::hash_combine` with `TfHash::Combine`

### Fixes Issue(s)


sunyab:
Filed as internal issue #USD-7902

PR #2186, opened 2023-01-20
Replace boost hash usage with TfHash in pxr/usd/pcp

nvmkuruc:
### Description of Change(s)
Requires #2173
To remove the dependency of pxr/usd/pcp on boost's hashing functions
* `PcpInstanceKey` now relies on `TfHash`'s hashing of containers instead of its own iteration.  To support this, the method `_Arc::GetHash` has been replaced with a `TfHashAppend` overload
* `PcpLayerStackIdentifier{,Str}`, `PcpMapExpression`, `PcpSite{,Str}`  use `TfHash::Combine` instead of `boost::hash_combine`
* Hashing test coverage has been added for `PcpInstanceKey` and `PcpMapFunction`
* `PcpMapFunction`'s hashing has been split off into an overload of `TfHashAppend` for its `_Data` which is `TfHash::Combine`d with the hash of its `_offset` field

### Fixes Issue(s)

### DIFF
--- a/pxr/usd/pcp/instanceKey.cpp
+++ b/pxr/usd/pcp/instanceKey.cpp
@@ -82,13 +82,7 @@ PcpInstanceKey::PcpInstanceKey(const PcpPrimIndex& primIndex)
         }
     }
     _variantSelection.assign(variantSelection.begin(), variantSelection.end());
-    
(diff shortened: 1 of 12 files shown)


sunyab:
Filed as internal issue #USD-7906

PR #2192, opened 2023-01-24
[build_usd] win: only add -A if generator supports it (ie, not ninja)

pmolodo:
### Description of Change(s)

build_usd.py would error on VS2019+ systems when building with `--generator Ninja`.  Fixes by only adding the `-A` flag to cmake if the generator supports it (ie, `Visual Studio *` generators do, `Ninja` doesn't.)

### Fixes Issue(s)
- error when building on windows + VS2019+ with Ninja generator


sunyab:
Filed as internal issue #USD-7912

PR #2194, opened 2023-01-24
[work] add missing include <algorithm> for std::max

pmolodo:
### Description of Change(s)

was getting errors when building without python on windows VS2022 - I think python and the older boost version were both bringing in the <algorithm> header; without either, it errored.

### Fixes Issue(s)
- error with no-python VS2022 build - max not a member of std, due to missing <algorithm> header


sunyab:
Filed as internal issue #USD-7914

PR #2199, opened 2023-01-25
Account for MaterialX namespaces when adding MaterialX nodes to a node graph.

AdamFelt:
Include the MaterialX namespace when adding MaterialX nodes to a node graph.  Namespace support was added last year, but somehow this piece was omitted from the namespace updates but is required when using MaterialX libraries that include namespaces.


sunyab:
Filed as internal issue #USD-7918

klucknav:
Thank you for raising this issue, we have a fix that should be in dev soon.

PR #2204, opened 2023-01-26
Remove use of the register keyword

dgovil:
### Description of Change(s)

The `register` keyword is no longer allowed as of C++17. It's also now considered to be a fairly weak hint that doesn't do anything in most scenarios.
As a result, I've removed all uses of it in the lexer files, which allows USD to build as a C++17 project. The builds seem to work just fine without it, and some brief testing shows no difference in performance.

- [X] I have submitted a signed Contributor License Agreement (Under the Apple CLA)

### DIFF
--- a/pxr/base/arch/stackTrace.cpp
+++ b/pxr/base/arch/stackTrace.cpp
@@ -609,10 +609,10 @@ nonLockingLinux__execve (const char *file,
 
 #if defined (ARCH_CPU_ARM)
     {
-        register long __file_result asm ("x0") = (long)file;
-        register char* const* __argv asm ("x1") = argv;
-        register char* const* __envp asm ("x2") = envp;
-        register long __num_execve asm ("x8") = 221;
+        long __file_result asm ("x0") = (long)file;
+        char* const* __argv asm ("x1") = argv;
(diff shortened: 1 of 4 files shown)


spiffmon:
Thanks, @dgovil - we're looking into upgrading our truly ancient version of flex to one that has the fix.  If we don't manage to pull that off for our VFX2022 toolset, we'll definitely take this PR as a stop-gap.

sunyab:
Filed as internal issue #USD-7923

stilllman:
We had to make the same fix for C++17 compatibility, I was about to push a PR when I saw this one. Note that the `register` keyword is used in two other files, `pxr/base/arch/stackTrace.cpp` and `third_party/renderman-24/plugin/hdPrman/virtualStructConditionalGrammar.lex.cpp`.

dgovil:
Ah good call out. I'll remove it there too and update this PR just in case it's useful. 

@spiffmon were there any updates on changing your flex version? Would it make sense to accept this PR in the interim, since they wouldn't conflict?

dgovil:
@stilllman updated the PR with the files you suggested. A quick ripgrep search didn't show other uses, so thanks for catching those.

PR #2206, opened 2023-01-26
Add [[maybe_unused]] attribute to `TfPyObjWrapperStub._stub` 

dgovil:
I'm trying to reduce a bunch of the warnings that take up the build logs (I build often...so any reduction really helps). TfPyObjWrapper is one of the heavy offenders in the logs, but `_stub` is necessary per https://github.com/PixarAnimationStudios/USD/issues/1716 , so this just says that `_stub` can be unused.

**NOTE:** This does require C++17, I'm just putting it up now so that it's logged for whenever that happens.

I also think this is something that should be applied to `_schemaTokens` in the API schemas too, as those are the other major offenders. But I can hit that up in a different PR if this one doesn't raise any concerns.

### DIFF
--- a/cmake/defaults/gccdefaults.cmake
+++ b/cmake/defaults/gccdefaults.cmake
@@ -40,4 +40,8 @@ endif()
 # See https://gcc.gnu.org/bugzilla/buglist.cgi?quicksearch=maybe%20uninitialized&list_id=394666
 _disable_warning("maybe-uninitialized")
 
+# MSVC and Clang support a super-set of GCC attributes.
+# Unfortunately, GCC warns when it doesn't use said attributes
+_disable_warning("ignored-attributes)
+
 set(_PXR_CXX_FLAGS "${_PXR_GCC_CLANG_SHARED_CXX_FLAGS}")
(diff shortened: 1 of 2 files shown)


sunyab:
Filed as internal issue #USD-7925

nvmkuruc:
Now that OpenUSD is on C++17, @dgovil do you think it's a good time to consider integrating this PR?

dgovil:
Yep that would be great if we could. I just rebased on dev to resolve the conflicts that had accumulated. 
I did also add `-Wno-ignored-attributes` for gcc, because unfortunately the situation is:
1. MSVC and Clang warn for unused attributes, so this attribute quells them.
2. GCC doesn't warn for it but then warns that the attribute does nothing as a result.

nvmkuruc:
That's a great catch for GCC.  If there's architecture dependent behavior, would it be better to create a define in `pxr/base/arch` (ie.  `ARCH_MAYBE_UNUSED_PRIVATE_MEMBER`) and set it to `[[maybe_unused]]` in the MSVC and Clang code paths and let it be empty for GCC?

dgovil:
Ah you know what, I missed after rebasing that Pixar have already added ARCH_PRAGMA_UNUSED_PRIVATE_FIELD above that line. So I'll just close out this PR as unnecessary.

nvmkuruc:
👍

Issue #2203, opened 2023-01-26
Prefer explicit deletion of copy constructor and assignment operator to `boost::noncopyable`


nvmkuruc:
### Description of Issue
There are several usages of `boost::noncopyable` which could be replaced with expicit `Object(const Object& o) = delete;` and `Object& operator=(const Object& o) = delete;` statements. Switching to this would reduce the number of dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

sunyab:
Filed as internal issue #USD-7922

nvmkuruc:
Closing, as this was addressed last year.

PR #2209, opened 2023-01-26
Replace boost::noncopyable with explicit deletion of copy constructor and assignment operator in pxr/usd/kind

nvmkuruc:
### Description of Change(s)
- `KindRegistry` has been updated to use explicit deletion of the copy constructor and assignment operator
### Fixes Issue(s)
- #2203


sunyab:
Filed as internal issue #USD-7928

PR #2215, opened 2023-01-29
Add missing #include for GCC13

musicinmybrain:
### Description of Change(s)

Adds `#include <cstdint>` for `uint64_t` and friends.

See “Header Dependency Changes” in https://www.gnu.org/software/gcc/gcc-13/porting_to.html.

### Fixes Issue(s)

(Fixes build of USD 22.05b in Fedora Linux Rawhide.)


tallytalwar:
Filed as internal issue #USD-7945

PR #2220, opened 2023-01-31
[build] Fix toolchain detection when using Microsoft Build Tools

ix-dcourtois:
### Description of Change(s)

The environment variable used by `build_usd.py` to detect the version of the MSVC compiler is in fact an environment variable storing the version of Visual Studio, which is the IDE and has a different versioning than the compiler. And when using Microsoft Build Tools, only the compiler and SDKs are installed, and that environment variable doesn't even exist.

This commit just uses the actual environment variable storing the MSVC compiler version (it's setup by both Visual Studio and the Build Tools) and fixes the `IsVisualStudio20XXOrGreater` helper functions to check against the compiler versions (see https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B the table `MSVC++ versions`)

**Edit**: It also skips adding the `-A x64` platform option to the CMake command if the generator used isn't a Visual Studio Variant, because only those support that option (it gives an error when using Ninja for instance)

### Fixes Issue(s)

- Fixes build using Microsoft Build Tools instead of Visual Studio.

- [ ] I have verified that all unit tests pass with the proposed changes (not applicable, build issue only)

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -127,11 +127,11 @@ def GetVisualStudioCompilerAndVersion():
 
     msvcCompiler = which('cl')
     if msvcCompiler:
-        # VisualStudioVersion environment variable should be set by the
+        # VCToolsVersion environment variable should be set by the
         # Visual Studio Command Prompt.
         match = re.search(
             r"(\d+)\.(\d+)",
-            os.environ.get("VisualStudioVersion", ""))
+            os.environ.get("VCToolsVersion", ""))
         if match:
             return (msvcCompiler, tuple(int(v) for v in match.groups()))
     return None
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-7950

ix-dcourtois:
Hi, sorry to bother. I'm not sure why the checks failed on Linux: the change only impacts Windows.
Is there anything I can do to fix things / help this get merged more easily?

spiffmon:
Hi @ix-dcourtois , we suspect this is because your PR conflicts with PR #2192 , which we pulled down last week, and which should be making its way out to github soon.  Once it does, you can rebase your PR upon it, and that should take care of it!

ix-dcourtois:
Ah yes, the fix for Ninja, good to know, thanks!

Issue #2222, opened 2023-01-31
`testVtCpp` can use `std::unique_ptr` instead of `boost::scoped_ptr`


nvmkuruc:
### Description of Issue
`testVtCpp` uses `boost::scoped_ptr` which cannot be moved or copied.  A `const std::unique_ptr` could be used to the same effect, reducing the number of dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

sunyab:
Hey @nvmkuruc, FYI we just checked in a fix for this internally last week and it'll be hitting the dev branch with the next push.

nvmkuruc:
Got it.  Thanks.

PR #2223, opened 2023-01-31
Replace `boost::scoped_ptr` with `std::unique_ptr` in `testVtCpp`

nvmkuruc:
### Description of Change(s)
- Replaces usage of `boost::scoped_ptr` with a `const` `std::unique_ptr` to prevent moving and copying
- `std::unique_ptr` is constructed with `std::make_unique` instead of `new` per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-make_unique

### Fixes Issue(s)
- #2222


nvmkuruc:
This is unnecessary per @sunyab's comment on #2222

Issue #2228, opened 2023-02-01
Replace `boost::iterator_adaptor` usage with fully specified iterator classes


nvmkuruc:
### Description of Issue

Most usages of `boost::iterator_adaptor` in pxr can be replaced with fully specified iterator classes.  Addressing this would reduce USD's dependence on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7956

nvmkuruc:
This was addressed by several PRs last year.

PR #2232, opened 2023-02-01
Replace `boost::iterator_adaptor` with explicitly specified iterator definition for `UsdPrimRange`

nvmkuruc:
### Description of Change(s)

- Introduces an `_UnderlyingIterator` type alias for `Usd_PrimDataConstPtr` in `UsdPrimRange::iterator`
- Introduces a `_PtrProxy` private class to be used as the `pointer` type for `UsdPrimRange::iterator`.  This special proxy is required because `UsdPrimRange::reference` is not a true reference type and it's not safe to take the address of a temporary.
- Explicitly implement the increment, dereference, and arrow operators for `UsdPrimRange::iterator`

### Fixes Issue(s)
- #2228 

### DIFF
--- a/pxr/usd/usd/primRange.cpp
+++ b/pxr/usd/usd/primRange.cpp
@@ -52,7 +52,7 @@ UsdPrimRange::Stage(const UsdStagePtr &stage,
 void
 UsdPrimRange::iterator::PruneChildren()
 {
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-7960

PR #2233, opened 2023-02-01
Replace `boost::iterator_adaptor` with explicitly specified iterator definition for `VtDictionary`

nvmkuruc:
### Description of Change(s)
- Explicitly implement the increment, dereference, equality, and arrow operators for `VtDictionary::iterator`
- Uses `nullptr` instead of `0` as the value to represent a null map per the C++ Core Guidelines (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nullptr)

### Fixes Issue(s)
- #2228 


tallytalwar:
Filed as internal issue #USD-7961

PR #2234, opened 2023-02-01
Replace `boost::iterator_adaptor` with explicitly specified iterator definition for `UsdPrim{Sibling,Subtree}Iterator`

nvmkuruc:
### Description of Change(s)
- Introduces an `_UnderlyingIterator` type alias for `const Usd_PrimData*` in `UsdPrimSiblingIterator`
- Introduces an `_UnderlyingIterator` type alias for `const UsdPrimDataConstPtr` in `UsdPrimSubtreeIterator`
- Introduces `_PtrProxy` private classes to be used as the pointer type for `UsdPrimSiblingIterator` and `UsdPrimSubtreeIterator`. This special proxy is required because their `reference` type aliases are not true reference types and it's not safe to take the address of a temporary.
- Explicitly implements the increment, dereference, equality and arrow operators for `UsdPrimSiblingIterator` and `UsdPrimSubtreeIterator`
- Fix doxygen string for pre/post increment operators

### Fixes Issue(s)
- #2228 

### DIFF
--- a/pxr/usd/usd/prim.h
+++ b/pxr/usd/usd/prim.h
@@ -43,7 +43,6 @@
 
 #include "pxr/usd/sdf/path.h"
 
-#include <boost/iterator/iterator_adaptor.hpp>
 #include <boost/range/iterator_range.hpp>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-7962

PR #2235, opened 2023-02-01
Replace `boost::iterator_adaptor` with explicitly specified iterator definition for `Usd_PrimData{Sibling,Subtree}Iterator`

nvmkuruc:
### Description of Change(s)
- Introduces an `_UnderlyingIterator` type alias for `Usd_PrimData*` in `Usd_PrimDataSiblingIterator` and  `Usd_PrimDataSubtreeIterator`
- Elides implementing `operator->` for these internal types as it is currently not used. As the `reference` type is not a true reference, a spurious proxy class would be required to properly support `pointer` semantics.
- Explicitly implements the increment, dereference, equality and arrow operators for `Usd_PrimDataSiblingIterator` and `Usd_PrimDataSubtreeIterator`

### Fixes Issue(s)
- #2228 

### DIFF
--- a/pxr/usd/usd/primData.h
+++ b/pxr/usd/usd/primData.h
@@ -41,7 +41,6 @@
 #include "pxr/usd/sdf/path.h"
 
 #include <boost/range/iterator_range.hpp>
-#include <boost/iterator/iterator_adaptor.hpp>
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-7963

Issue #2236, opened 2023-02-02
`boost::equality_comparable` can be replaced with explicit `operator!=` implementation


nvmkuruc:
### Description of Issue
`boost::equality_comparable` adds `operator!=` to any derived class.  Explicit implementation of `operator!=` would remove this dependency on `boost`.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7964

nvmkuruc:
This was addressed by #2701 and #2237

PR #2237, opened 2023-02-02
Replace `boost::equality_comparable` with explicit implementation of `operator!=` in `pxr/usd/sdf`

nvmkuruc:
### Description of Change(s)
- Adds explicit implementation of `operator!=` to `SdfAllowed`, `SdfChildrenProxy`, `SdfNamespaceEdit`, `SdfNamespaceEditDetail`, and `SdfUnregisteredValue`
- Equality operator test coverage has been extended to `Sdf.NamespaceEdit`, `Sdf.NamespaceEditDetail`, and `Sdf.UnregisteredValue`.

### Fixes Issue(s)
- #2236 

### DIFF
--- a/pxr/usd/sdf/namespaceEdit.h
+++ b/pxr/usd/sdf/namespaceEdit.h
@@ -30,8 +30,6 @@
 #include "pxr/usd/sdf/api.h"
 #include "pxr/usd/sdf/path.h"
 
-#include <boost/operators.hpp>
-
 #include <functional>
 #include <iosfwd>
 #include <string>
@@ -44,8 +42,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 /// A single namespace edit.  It supports renaming, reparenting, reparenting
 /// with a rename, reordering, and removal.
 ///
-struct SdfNamespaceEdit :
-    boost::equality_comparable<SdfNamespaceEdit> {
+struct SdfNamespaceEdit {
 public:
     typedef SdfNamespaceEdit This;
     typedef SdfPath Path;
@@ -114,6 +111,7 @@ public:
     }
 
     SDF_API bool operator==(const This& rhs) const;
+    SDF_API bool operator!=(const This& rhs) const;
 
 public:
     Path currentPath;   ///< Path of the object when this edit starts.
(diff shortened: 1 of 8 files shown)


tallytalwar:
Filed as internal issue #USD-7965

pxr/usd/sdf/namespaceEdit.h
```diff
@@ -146,6 +143,7 @@ struct SdfNamespaceEditDetail :
                            const std::string& reason);
 
     SDF_API bool operator==(const SdfNamespaceEditDetail& rhs) const;
```
crevilla, line 145: Just a nitpick. Since == uses "rhs", != should use "rhs" too

pxr/usd/sdf/testenv/testSdfBatchNamespaceEdit.py
```diff
@@ -31,6 +31,47 @@
...
+        self.assertTrue(Sdf.NamespaceEdit('/B', '/A') != Sdf.NamespaceEdit('/A', '/B'))
+
+        # Validate the equality operators by varying a single field from a prototype edit
+        # at a time
+        prototype = Sdf.NamespaceEditDetail(
+            Sdf.NamespaceEditDetail.Okay,
+            Sdf.NamespaceEdit('/A', '/B'),
+            "reason"
+        )
+        self.assertTrue(prototype == prototype)
```
crevilla: The prototype == prototype test case would be more robust if it were
self.assertTrue(prototype == Sdf.NamespaceEditDetail(
￼            Sdf.NamespaceEditDetail.Okay,
￼            Sdf.NamespaceEdit('/A', '/B'),
￼            "reason"
￼        )
Same for !=

nvmkuruc: Done, I also switched the tests to use `assertEqual` and `assertNotEqual` which is preferred per https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertFalse

pxr/usd/sdf/namespaceEdit.h
nvmkuruc: Done

PR #2239, opened 2023-02-02
Add support for long Windows paths

mistafunk:
### Description of Change(s)

Prefix all file system paths longer than 248 characters with "\\\\?\\" before passing into Win32 File I/O API functions.

### Fixes Issue(s)

Avoids errors when reading/writing USD files on long Windows paths.

### Notes
* ~~The CLA is pending on my side.~~
* This is my first stab at fixing this problem and also my first time contributing larger changes to USD.
* ~~This PR is not complete yet (Arch and Tf test coverage). I submit it early to get some basic feedback.~~
* ~~This PR contains temporary/helper commits not intended for merge.~~

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -24,6 +24,7 @@
 #include <filesystem>
 #include <memory>
 #include <utility>
+#include <vector>
 
 #include <fcntl.h>
 #include <sys/types.h>
@@ -49,10 +50,17 @@ using std::set;
 
 #if defined (ARCH_OS_WINDOWS)
 namespace {
+
+const std::string LONG_PATH_PREFIX = "\\\\?\\";
+const std::wstring LONG_PATH_PREFIX_W = L"\\\\?\\";
+const std::string UNC_LONG_PATH_PREFIX = LONG_PATH_PREFIX + "UNC\\";
+const std::wstring UNC_LONG_PATH_PREFIX_W = LONG_PATH_PREFIX_W + L"UNC\\";
+
 static inline HANDLE _FileToWinHANDLE(FILE *file)
 {
     return reinterpret_cast<HANDLE>(_get_osfhandle(_fileno(file)));
 }
+
 }
 #endif // ARCH_OS_WINDOWS
 
@@ -95,9 +103,11 @@ FILE* ArchOpenFile(char const* fileName, char const* mode)
         return nullptr;
     }
 
+    const std::wstring apiPath = ArchHandleLongWindowsPaths(ArchWindowsUtf8ToUtf16(fileName));
+
     // Call CreateFileW.
     HANDLE hfile = CreateFileW(
-        ArchWindowsUtf8ToUtf16(fileName).c_str(),
+        apiPath.c_str(),
         desiredAccess,
         shareMode,
         /* securityAttributes=*/nullptr,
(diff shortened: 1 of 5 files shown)


mistafunk:
Looking into the compile issues...

pxr/base/arch/fileSystem.cpp
```diff
@@ -1387,4 +1429,35 @@ void ArchFileAdvise(
 #endif
 }
 
+#if defined(ARCH_OS_WINDOWS)
+
+ARCH_API
+std::wstring ArchHandleLongWindowsPaths(const std::wstring& path)
```
meshula: minor note, the pattern 

`ArchHandleLongWindowsPaths(ArchWindowsUtf8ToUtf16(path));`

is very common in this patch. What about overloading this function instead, with a const char* path overload to match the callsites?

meshula:
Looks promising! Out of curiosity, I noticed you removed the Debug target, which I assume is a temporary development detail. Is there some aspect of Debug that was getting in the way, or is it just a patch to streamline a workflow you have?

tallytalwar:
Filed as internal issue #USD-7967

pxr/base/arch/fileSystem.cpp
mistafunk: Good idea, done.

mistafunk:
> Looks promising! Out of curiosity, I noticed you removed the Debug target, which I assume is a temporary development detail. Is there some aspect of Debug that was getting in the way, or is it just a patch to streamline a workflow you have?

Thanks. I saw CMake install errors regarding `testArchStackTrace.pdb` in non-Release modes. I don't grok these yet, but it seems to be related to the VS generator. Switching to Ninja makes these go away.

Anyways, I'm working on another PR iteration and will trigger a review request once I'm ready.

erikaharrison-adsk:
Re: testArchStackTrace.pdb, see https://github.com/PixarAnimationStudios/USD/pull/2179

mistafunk:
@meshula I've rebased the PR to latest dev and expanded the test coverage a bit.

Please note, I've yet to backport these changes to our product and test them there. But I consider the changes done.

mistafunk:
Before merging, the  `_ArchHandleLongWindowsPaths` also needs to remove dot and dotdot in the paths as well as two (or more) consecutive backslashes.

(Stumbled over this while adding long path support to another library.)

johnk-realtime:
Hey guys, just checking on the status of this as it appears the discussion has been inactive for a while. Besides the review is there anything else that's preventing this from moving forwards?

We are experiencing this issue at the moment and considering our options :)

mistafunk:
Hi, the ball is with Pixar :-) I'm happy to receive feedback and work some more on this...

spiffmon:
@mistafunk , it's taken awhile, but this has, coincidentally, recently finally risen to near top of queue, and we'll be having a look shortly.  Can't promise it will make it in to 24.08, but soon.

pxr/base/arch/fileSystem.cpp
```diff
@@ -648,21 +762,27 @@ MakeUnique(
 
 #endif
 
+namespace {
+
+const std::string TMPDIR_FMT = std::string("%s") + ARCH_PATH_SEP + "%s.XXXXXX";
```
gitamohr: Can this just be a literal?  `"%s" ARCH_PATH_SEP "%s.XXXXXX"` ?

mistafunk: good catch, will turn this into a `constexpr const char*`

mistafunk: done

mistafunk:
@spiffmon @meshula I've rebased the branch onto latest dev and addressed the one remaining comment as well as made sure the long paths are normalized (removal of dotdot and dot) as per MS specs.

mistafunk:
Oops, let me take care of the compile issues on Linux/macOS...

mistafunk:
Force pushed the last two commits and fixed Linux build by adding a missing `typename` (hopefully this takes care of macOS as well... otherwise I'll fire up my macbook in the next few days)

mistafunk:
@spiffmon @meshula resolved merge conflicts and rebased on `dev`... soon this PR turns 2 years old, do you think we manage to merge this before it turns into a teenager? :-)

demiaster:
Hello!
We are also interested in seeing this feature through.
Is there any outstanding issues that need resolving?

Thank you!

spiffmon:
Apologies, this has languished on our work list for awhile.  We've scheduled it for an upcoming release this year, and given the time that has elapsed since @mistafunk made the changes we requested, we will handle any further updating/changes we think necessary.

Issue #2243, opened 2023-02-03
Some `boost` headers are unused


nvmkuruc:
### Description of Issue
The following headers are included but unused in `pxr`.

`boost/preprocessor/arithmetic/add.hpp`
`boost/preprocessor/tuple/eat.hpp`
`boost/preprocessor/seq/push_back.hpp`
`boost/assign.hpp`
`boost/mpl/or.hpp`
`boost/ptr_container/ptr_map.hpp`

Removing them reduces the number of dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7971

nvmkuruc:
I think we can close this.  Any unused headers are now transitive through `tf/preprocessorUtils.h` and `vt/functions.h`

PR #2244, opened 2023-02-03
Remove include of unused `boost/preprocessor` headers

nvmkuruc:
### Description of Change(s)
- Removes unused `boost/preprocessor/tuple/eat.hpp` and `boost/preprocessor/arithmetic/add.hpp` from `pxr/base/tf/preprocessorUtils.h`
- Removes `boost/preprocessor/seq/push_back.hpp` from `pxr/base/tf/staticTokens.h`

### Fixes Issue(s)
- #2243 


tallytalwar:
Filed as internal issue #USD-7972

PR #2256, opened 2023-02-08
APPLE: Use unique name for debug View.

creijon:
### Description of Change(s)

Using `View` as a classname in the code can cause issues when integrating with other Cocoa applications.  To avoid this, rename it to `Garch_GLView`

### Fixes Issue(s)
- Potential name conflicts in Cocoa applications.


tallytalwar:
Filed as internal issue #USD-7990

Issue #2257, opened 2023-02-08
`boost::compressed_pair` can be replaced with explicit empty base optimization


nvmkuruc:
### Description of Issue

`boost::compressed_pair` is being leveraged in `TfDense{Map,Set}` to take advantage of the empty base optimization for the Hashing and Equality types.  This can be done explicitly with private inheritance on the containers.

Longer term, the `[[no_unique_address]]` attribute should be considered when `C++20` is an option.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7991

nvmkuruc:
This was addressed by #2522 and #2257

PR #2259, opened 2023-02-08
Replace usage of `boost::compressed_pair` with explicit empty base optimization

nvmkuruc:
### Description of Change(s)
- Update `TfDenseHash{Set,Map}` to explicitly use private inheritance to take advantage of the empty base optimization instead of indirectly through `boost::compressed_pair`
- Decouples initialization of vector storage, `HashFn`, and `Equality{Key,Element}` in constructors
- Updates constructors to use `std::make_unique` instead of `new` operator (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-make_unique)
- Decouples swapping of vector storage, `HashFn`, and `Equality{Key,Element}` in `swap` methods
- Remove references to `compressed_pair` in test for `TfDenseHashMap`
- Remove spurious include of `utility.hpp` in `pxr/base/tf/denseHash{Set,Map}.h`
- Remove spurious include of `boost/compressed_pair.hpp` in `pxr/usd/sdf/childrenView.h`

### Fixes Issue(s)
- #2257 

### DIFF
--- a/pxr/base/tf/denseHashMap.h
+++ b/pxr/base/tf/denseHashMap.h
@@ -32,9 +32,7 @@
 #include <memory>
 #include <vector>
 
-#include <boost/compressed_pair.hpp>
 #include <boost/iterator/iterator_facade.hpp>
-#include <boost/utility.hpp>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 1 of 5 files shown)


tallytalwar:
Filed as internal issue #USD-7993

pxr/base/tf/denseHashMap.h
```diff
@@ -56,7 +54,13 @@ template <
     class    EqualKey  = std::equal_to<Key>,
     unsigned Threshold = 128
 >
-class TfDenseHashMap
+
```
crevilla, line 58: Okay, we don't want to do this this way. The semantics of TfDenseHashMap now BEING a HashFn and an EqualKey is not desirable.
What I think you really want to do is reimplement the internal member class _VectorHashFnEqualFn using empty base optimization explicitly instead of the compressed_pair

pxr/base/tf/denseHashSet.h
```diff
@@ -55,7 +53,12 @@ template <
     class    EqualElement  = std::equal_to<Element>,
     unsigned Threshold = 128
 >
-class TfDenseHashSet
+class TfDenseHashSet :
```
crevilla: Same comment here; we don't want the DenseHashSet inheriting from these function types.

PR #2263, opened 2023-02-09
iOS rendering changes

slingthor:
Description of Change(s)

Allows building the CMAKE project
Allows rendering with iOS

Recreating this PR as it was quite risky and the history was hard to maintain the way it was set up.

In case of getting the work in, All the way up to step 9 is mandatory.

It would be our hope to get all the way to 7B in the possible future for all the expected features for iOS to be in.


tallytalwar:
Filed as internal issue #USD-7999

build_scripts/apple_utils.py
```diff
@@ -57,14 +60,69 @@ def GetLocale():
...
         pass
-    return None
+
+    # clean-up
+    for pipe in pipes:
+        pipe.wait()
+
+    return result
+
+def GetFrameworkRoot():
```
meshula: Can we omit GetFrameworkRoot?

It's not called anywhere in this PR, and it introduces a hard coded dependency to a Python.framework. Other python environments, such as those installed by conda, do not follow this pattern, and it would be unfortunate to introduce a new dependency on framework based Python distributions.

build_scripts/build_usd.py
```diff
@@ -248,7 +249,7 @@ def GetCPUCount():
     except NotImplementedError:
         return 1
 
-def Run(cmd, logCommandOutput = True):
+def Run(cmd, logCommandOutput = True, envOverride = None):
```
meshula: does this envOverride ever get set to anything but None? If no, could I request that that we don't introduce a new parameter. build_usd needs less features, not more, and adding an override seems to add a vector for new complexity down the line.

meshula: In this diff, it seems to be either set to None, or passed from function to function, but not uniquely set.

build_scripts/build_usd.py
```diff
@@ -682,6 +728,14 @@ def AnyPythonDependencies(deps):
 
 def InstallZlib(context, force, buildArgs):
     with CurrentWorkingDirectory(DownloadURL(ZLIB_URL, context, force)):
+        if context.targetIos:
```
meshula, line 714: is it possible to suppress the creation of the zlib tests instead? Also, is zlib part of the iOS SDK? If it is maybe we shouldn't be downloading and installing it at all?

build_scripts/build_usd.py
```diff
@@ -726,24 +780,31 @@ def InstallBoost_Helper(context, force, buildArgs):
                                              dontExtract=dontExtract)):
         bootstrap = "bootstrap.bat" if Windows() else "./bootstrap.sh"
 
+        if context.targetIos:
```
meshula: there's no link dependency on boost from iOS, because there's no support for python bindings to USD on iOS, right?

Can we simplify things by merely unarchiving boost, and pointing the iOS build at the headers, and calling it a day? In the unlikely event that one of the boost headers needs configuration for iOS, maybe we can just patch that header. But to the best of my recollection, even that shouldn't be necessary.

build_scripts/build_usd.py
```diff
@@ -910,6 +1017,17 @@ def InstallBoost(context, force, buildArgs):
                    "include/boost-1_70/boost/version.hpp",
                    "include/boost-1_78/boost/version.hpp")
 
+############################################################
+# IOS toolchain
+IOS_TOOLCHAIN_URL = "https://github.com/leetal/ios-cmake/archive/refs/tags/4.3.0.tar.gz"
```
meshula: I still prefer that we create our own toolchain file, or, that we intern a copy of Bogdan's cmake file here. The license on the file permits interning the file, and has no copyleft complications.

build_scripts/build_usd.py
```diff
@@ -1078,6 +1200,41 @@ def InstallJPEG_Turbo(context, force, buildArgs):
         extraJPEGArgs = buildArgs
         if MacOS():
             extraJPEGArgs.append("-DWITH_SIMD=FALSE")
+        if context.targetIos:
+            extraJPEGArgs.append('-DCMAKE_SYSTEM_PROCESSOR=aarch64')
+            extraJPEGArgs.append("-DENABLE_STATIC=TRUE")
+
+            # Replace test and utility executables with static libraries to avoid issues with code signing.
```
meshula: Instead of fixing tj for them, can we instead suppress the creation of test and utility executables for tj, as they are unrelated to building and deploying USD. It makes sense to me to upstream fixes to tj overall.

build_scripts/build_usd.py
```diff
@@ -1114,6 +1271,11 @@ def InstallTIFF(context, force, buildArgs):
                    [("add_subdirectory(tools)", "# add_subdirectory(tools)"),
                     ("add_subdirectory(test)", "# add_subdirectory(test)")])
 
+        if context.targetIos:
```
meshula: tbh, my opinion is that we should skip contrib on all platforms as people don't look to the USD project to build libtiff for them unless they are simply satisfying requirements of the oiio build... Not requesting an update though, as it's an existing behavior that contrib gets roped in.

build_scripts/build_usd.py
```diff
@@ -1135,6 +1297,24 @@ def InstallTIFF(context, force, buildArgs):
 def InstallPNG(context, force, buildArgs):
     with CurrentWorkingDirectory(DownloadURL(PNG_URL, context, force)):
         macArgs = []
+        if context.targetIos:
```
meshula: can we just skip all this on ios instead of fixing libtiff for them? I can't think of any reason we'd want to create these utilities.

build_scripts/build_usd.py
```diff
@@ -1161,13 +1341,160 @@ def InstallPNG(context, force, buildArgs):
     OPENEXR_URL = "https://github.com/AcademySoftwareFoundation/openexr/archive/refs/tags/v2.5.2.zip"
 else:
     OPENEXR_URL = "https://github.com/AcademySoftwareFoundation/openexr/archive/refs/tags/v2.4.3.zip"
+    
+def updateOpenEXRIOS(context, srcDir):
+    # IlmBase
```
meshula: let's fix OpenEXR upstream. It's okay to depend on a later version of OpenEXR specifically for iOS. I'm able to make commits in the OpenEXR repo, so I can stand by to help get these fixes landed as quickly as possible.

build_scripts/build_usd.py
```diff
@@ -1265,6 +1604,17 @@ def InstallOpenVDB(context, force, buildArgs):
             '-DOPENVDB_BUILD_BINARIES=OFF',
             '-DOPENVDB_BUILD_UNITTESTS=OFF'
         ]
+        if context.targetIos:
+            extraArgs.append('Boost_USE_STATIC_LIBS=ON')
```
meshula: I think the reason you are patching the boost generation code above is to support OpenVdb here. How important is the support of OpenVdb on iOS at this time? To expedite things and avoid having to support boost moving forward, would it be a reasonable approach to skip OpenVdb for now? I personally would rather not have boost on iOS more than I personal would want OpenVdb support on iOS.

build_scripts/build_usd.py
```diff
@@ -1306,6 +1659,24 @@ def InstallOpenImageIO(context, force, buildArgs):
                      '-DOIIO_BUILD_TESTS=OFF',
                      '-DUSE_PYTHON=OFF',
                      '-DSTOP_ON_WARNING=OFF']
+        if context.targetIos:
```
meshula: Once again, we could use a different version of OIIO on iOS. Can I request that we upstream fixes to OIIO, as I am sure @lgritz is able to assist integration.

build_scripts/build_usd.py
```diff
@@ -1465,14 +1840,64 @@ def InstallOpenSubdiv(context, force, buildArgs):
         # just 1 job for now. See:
         # https://github.com/PixarAnimationStudios/OpenSubdiv/issues/1194
         oldNumJobs = context.numJobs
+        extraEnv = os.environ.copy()
+        buildDirmacOS = ""
         if MacOS():
             context.numJobs = 1
 
+        if context.targetIos:
```
meshula, line 1551: @davidgyu Would it be possible to integrate these changes upstream at OpenSubdiv? Once again, we can point to a later version of OpenSubdiv for iOS in preference to patching OpenSubdiv.

pxr/imaging/hgiMetal/graphicsCmds.mm
```diff
@@ -818,6 +820,8 @@
     
     // Apple Silicon only support memory barriers between vertex stages after
     // macOS 12.3.
+    //For iOS we may want to introduve an alternative path
```
meshula: typo, "introduce".
Is there anything that can be said about what the alternative path is? That would help people reading the code in the future understand what was intended by this comment.

pxr/imaging/hgiMetal/hgi.mm
```diff
@@ -60,6 +60,7 @@
 , _workToFlush(false)
 {
     if (!_device) {
+#if defined(ARCH_OS_OSX)
         if( TfGetenvBool("HGIMETAL_USE_INTEGRATED_GPU", false)) {
```
meshula, line 64: I'm curious which macs need this override?

pxr/imaging/hgiMetal/hgi.mm
```diff
@@ -145,11 +146,10 @@
 bool
 HgiMetal::IsBackendSupported() const
 {
-    // Want Metal 2.0 and Metal Shading Language 2.2 or higher.
-    if (@available(macOS 10.15, ios 13.0, *)) {
+    if (@available(ios 14.0, *))
```
meshula: is no macOS check required any more? aka, will this conditional return true for macos due to the asterisk?

third_party/IlmBase/eLut.h
```diff
@@ -0,0 +1,71 @@
+//
```
meshula: please see my comments above about upstreaming OpenEXR fixes

meshula:
Thank you for refactoring this commit chain, it's looking much more tractable now :)

meshula:
Also, the comment says "dependent on 2035" but 2035 has been closed. I think the meaning is that this PR supersedes 2035?

build_scripts/build_usd.py
lgritz: I don't have a way to test on iOS myself, but I'm pretty sure that over the years, people have contributed patches to OIIO to fix the iOS build, so I'm presuming that at some point it worked (maybe still?). 

I'm not an iOS developer (nor a MacOS developer per se, except in the trivial sense that I use it all day long but almost entirely pretend it's just a generic Unix toolchain), so I don't know -- is it possible to build on MacOS but cross-compile for iOS, or even emulate iOS? If those tools are available on GitHub Actions runners, I'd be happy to add (that is, accept a PR from somebody who knows what they're doing) another entry to the test matrix for OIIO's CI, wherein we do an emulated iOS build just to ensure that it keeps working and we don't break it with future changes.

```python
if MacOS():
    # OIIO 2.3.15 adds fixes for Apple Silicon cross compilation.
    OIIO_URL = "https://github.com/OpenImageIO/oiio/archive/refs/tags/v2.3.15.0.zip"
else:
    OIIO_URL = "https://github.com/OpenImageIO/oiio/archive/Release-2.1.16.0.zip"
```

Oy, I am definitely available to advise anybody who would like to bring this up to a modern version. The current series is 2.4, and 2.1.16 dates from 2020.

build_scripts/build_usd.py
davidgyu: We've already fixed most (if not all) of these concerns upstream as of the current OpenSubdiv 3.5 release.
- We no longer need the numJobs override above.
- We no longer need a macOS dependent build for STRINGIFY.
- We should no longer need to extend RunCMake() with envOveride or hostPlatform optional args.

Volutionn:
Hi @slingthor, I’m interested to try this PR on iOS. Do you have indications on how to link this library to an Xcode project?

I tried to import the lib and while it's working overall, I'm facing an issue where the HdStorm renderer does not clear its buffer after each frame. I'm hoping that this might be because of the way I'm implementing it.

build_scripts/build_usd.py
meshula: Here's an example of github CI for iOS ~ I don't know if it's a useful reference for OIIO's automation ~ but it does work for me :) https://github.com/LabSound/LabSound/blob/main/.github/workflows/cmake-ios.yml

Since OpenImageIO isn't one of the vfxplatform specified versions, I can look into why it's set at 2.1.16.

build_scripts/build_usd.py
meshula: Oh, fantastic :) So pulling that version, at least for iOS, and identifying any remaining issues seems like it is the way forward.

build_scripts/build_usd.py
slingthor: In RunCmake it is passed into run in the CMake invocation, and then used to build OpenSubdiv - the parts are always moving and I immediately noticed it could be removed from InstallOpenEXR and possibly it can be removed from Opensubdiv by now, I will give that a try - there surely was an unneeded one and I see the interface complexity you are talking about

meshula: thanks :)

build_scripts/apple_utils.py
slingthor: I removed it and I could build all the core dependencies from commit 0 to 9 which is what I had hoped to get in to begin with, so removed -> will add in again if required for any other dependencies, but there's a good chance it's no longer needed and surely not needed for the core dependencies

build_scripts/build_usd.py
slingthor: Confirmed it can be removed by now! Good catch :D

build_scripts/build_usd.py
slingthor: For the time being, I agree it's more important to get the core functionality in and OpenVDB can join later - for users, it may a bit surprising to be able to attempt to build something not included - would you find it OK if I made a matrix for supported features by platform, and then we can explicitly tell the user it can't use that feature? We are seeing such suprises for users for OpenColorio for example, where nothing tells the user they can't use OCIO 1.x on Metal, but it however is not possible

build_scripts/build_usd.py
slingthor: Agree to do it upstream and we can start with the core components!

pxr/imaging/hgiMetal/hgi.mm
slingthor: This is an override for AMD based Macs that also have an Intel graphics for power saving - this has been helpful for testing.

pxr/imaging/hgiMetal/hgi.mm
slingthor: I discussed this further on our side - this was an opinion from me as 10.15 is quite old, but this can probably still stay for safety sakes

slingthor:
> Thank you for refactoring this commit chain, it's looking much more tractable now :)

A pleasure! For the time being I could remove the components from the whole chain that we won't include in the initial support for iOS.

build_scripts/build_usd.py
slingthor: After looking into it, I think Zlib's cmakelist is not very friendly for turning off tests.
Using -DBUILD_TESTING=OFF did not disable the test's generation as items.
The cmakelist simply always adds the executables.
image
From the zlib CmakeList

That then leads to an error in the build automation for Xcode projects that need bundle identifiers (iOS ones do)
image

Further down the line we could create automations that edit Xcode files, but adding an arbitrary bundle identifier meaningful for the building user is also something that is not very intuitive to automate as bundle identifiers should be unique for each user and probably we should not create identifiers for users in general.

We could however suppress them by writing a patch that modifies the zlib cmakelist file

build_scripts/build_usd.py
slingthor: Agree! I'll modify it to intern based on the license being suitable for that

build_scripts/build_usd.py
slingthor: In our initial submission we could skip supporting OpenImageIO, although it would be nice - are you aware if any of OpenImageIO's dependencies are optional?

build_scripts/build_usd.py
slingthor: Agree to that!

pxr/imaging/hgiMetal/graphicsCmds.mm
slingthor: Adjusted it now - 16.0 introduced the ability to perform the operation on iOS!

build_scripts/build_usd.py
meshula: That's a good suggestion. By matrix, I think you are meaning documentation, as opposed to a matrix of features in the code? Perhaps the best place for it is in the documentation here https://github.com/PixarAnimationStudios/USD/blob/dev/BUILDING.md#optional-components
- that would be a separate changelist and we could update it once these changes land.

We could also emit diagnostics, as part of this changelist, so that if someone specifies for example OCIO 1.x, it could explicitly inform them of the impossibility of it.

build_scripts/build_usd.py
meshula: errrmmm, 99% of them are? There are a lot of vendor package inclusions, documentation generators, Python dependencies, and OpenShadingLanguage. But I don't think we have to take particular action on that on behalf of iOS. On Mac, OCIO uses pybind11, not boost python thank goodness, so we are clean on the OCIO + boost front, as far as I can tell.

build_scripts/build_usd.py
meshula: If we must patch, then lets suppress them, rather than fix them, please.

meshula:
@slingthor is this change by @jonny-apple relevant to this change? https://github.com/PixarAnimationStudios/USD/pull/2274/files I assume it is independent.

Also I notice Jon updated a versions matrix there, so if that's the one you are referring to, another column for iOS makes sense...

build_scripts/build_usd.py
slingthor: Cool! Have modified it that way - the modifications I'm making, I might add as a final commit but later on possibly move into the nice commit chain

build_scripts/build_usd.py
slingthor: I'll give that a go

build_scripts/build_usd.py
meshula: I see the new commit :)

slingthor:
> @slingthor is this change by @jonny-apple relevant to this change? https://github.com/PixarAnimationStudios/USD/pull/2274/files I assume it is independent.
> 
> Also I notice Jon updated a versions matrix there, so if that's the one you are referring to, another column for iOS makes sense...

Back when this came about, we actually solved the same problem at a similar time in different places - I'll try to align mine with the version matrix

slingthor:
> Hi @slingthor, I’m interested to try this PR on iOS. Do you have indications on how to link this library to an Xcode project?
> 
> I tried to import the lib and while it's working overall, I'm facing an issue where the HdStorm renderer does not clear its buffer after each frame. I'm hoping that this might be because of the way I'm implementing it.

Hey! You would need to create an application that can display a surface Storm has rendered on your iOS screen - https://developer.apple.com/documentation/metal/metal_sample_code_library/creating_a_3d_application_with_hydra_rendering this example does it for MacOS, but you could modify it to run on iOS.

You might face challenges with:
- Code signing
- Sandboxing
- File access
- The ViewController class would need to be set for iOS
- You'd need an iOS app delegate

third_party/IlmBase/eLut.h
slingthor: Agree - In the "modifications" commit, I'll remove OpenEXR and OpenVDB for the time being

build_scripts/build_usd.py
slingthor: Hm, OpenImageIO will not respect -DUSE_OpenEXR=OFF

Volutionn:
> > Hi @slingthor, I’m interested to try this PR on iOS. Do you have indications on how to link this library to an Xcode project?
> > I tried to import the lib and while it's working overall, I'm facing an issue where the HdStorm renderer does not clear its buffer after each frame. I'm hoping that this might be because of the way I'm implementing it.
> 
> Hey! You would need to create an application that can display a surface Storm has rendered on your iOS screen - https://developer.apple.com/documentation/metal/metal_sample_code_library/creating_a_3d_application_with_hydra_rendering this example does it for MacOS, but you could modify it to run on iOS.
> 
> You might face challenges with:
> 
> * Code signing
> * Sandboxing
> * File access
> * The ViewController class would need to be set for iOS
> * You'd need an iOS app delegate

Thank you for your reply!  I actually did something really similar in my test, can you confirm that HdStorm is working properly on your side? Happy to help find the cause if it's not only an issue on my side. 

Here's a screenshot of what I see (the plane is animated)
Screen Shot 2023-02-18 at 10 19 32 am

meshula:
@Volutionn I think maybe you just need to clear the frame buffer at the start of the render? In my renderers I do something like this ~ 

```cpp
        renderPassDescriptor.colorAttachments[0].loadAction = mustClear ? MTLLoadAction.clear : MTLLoadAction.load
```

where `mustClear` is a flag I pass into my render routine indicating whether the pass needs to clear the framebuffer or not.

build_scripts/build_usd.py
slingthor: So for the time being I believe we won't include openimageio
https://github.com/OpenImageIO/oiio/blob/master/INSTALL.md
from my side, it's a reasonable step to only use jpgs and minimal textures to begin with and openimagiio can then be added later

build_scripts/build_usd.py
slingthor: Simplified all the things David mentioned!

build_scripts/build_usd.py
slingthor: That worked really well! Thank you so much!

build_scripts/build_usd.py
slingthor: Disabled openimagiio for the time being - OpenEXR needs it.

slingthor:
I've done all the simplifications mentioned - for the time being, disabled OpenImageIO as OpenEXR, ILM and it's related dependencies sounded quite tough to get in and happen to be required dependencies.
LibPNG and TurboJPG's created executables are also quite many and TJ and the tests are hard to disable with a patch, and probably it's best to submit those fixes and then enable OpenImageIO.

With what we ended up with, we can build iOS without extra dependencies as a first step.

build_scripts/build_usd.py
meshula: That sounds like a good intermediate step in order to get things up and running.

Volutionn:
> @Volutionn I think maybe you just need to clear the frame buffer at the start of the render? In my renderers I do something like this ~
> 
> ```c++
>         renderPassDescriptor.colorAttachments[0].loadAction = mustClear ? MTLLoadAction.clear : MTLLoadAction.load
> ```
> 
> where `mustClear` is a flag I pass into my render routine indicating whether the pass needs to clear the framebuffer or not.

Thank you @meshula for your reply. I tried adding this line but it has no impact on the output... I actually thought that the buffer clearing is managed by Hydra ?

What's weird also is that the exact same code when built for macOS, so also with USD built for macOS, does work perfectly without the need of adding this line. And, as you can see in my screenshot, the smooth shading does not work, but it does work on macOS with the same implementation. Can those issues be related to a bigger issue with one of the libraries ?

slingthor:
> @slingthor is this change by @jonny-apple relevant to this change? https://github.com/PixarAnimationStudios/USD/pull/2274/files I assume it is independent.
> 
> Also I notice Jon updated a versions matrix there, so if that's the one you are referring to, another column for iOS makes sense...

Oh apologies! I assumed it had to do with the texture filtering which also had sort of a matrix - these changes are independent

slingthor:
@meshula Hope the PR is looking good by now! I will be in the near future submitting directly to the dependencies to simplify the build scripts for the additional dependencies, but the build without them I hope is looking good now!

PR #2264, opened 2023-02-10
APPLE: Prevent mipmapping of 1x1 textures.

creijon:
### Description of Change(s)

In Metal, if a texture has any dimension being a single texel big the mipmapping call `generateMipmapsForTexture` will fail and throw an exception.  So we add a check to the existing one in `HgiMetalBlitCmds::GenerateMipMaps()` which catches this.

Thanks to @morteeza for this.

### Fixes Issue(s)
- Exception when mipmapping textures with any dimension being 1.


tallytalwar:
Filed as internal issue #USD-8000

sunyab:
This PR was merged into the `dev` branch in c6e09a4.

PR #2265, opened 2023-02-11
usdviewq: minor fixes when selecting and traversing collections in the property view

mattyjams:
Hello! :)

### Description of Change(s)

This PR addresses two small issues when working with `UsdCollectionAPI`s in the property view of `usdview`.

Collections can end up targeting other collections in their relationships using paths that identify the collection by name but otherwise do not correspond to a prim or a property. There were a few places in usdview that were getting tripped up by this.

The issues can both be seen when `usdview` is opened using the existing `UsdCollectionAPI` test scene:

```
usdview pxr/usd/usd/testenv/testUsdCollectionAPI/Test.usda
```

- Select the `/CollectionTest/Geom` prim in the prim view
- Expand the `collection:allGeom:includes` relationship in the property view
- Select the second target in the relationship that targets the path `/CollectionTest/Geom/Shapes.collection:allShapes`

You should see a warning similar to this output in the terminal:

```
Warning: in SdfPath at line 151 of C:\USD_src\pxr\usd\sdf\path.cpp -- Ill-formed SdfPath <>: syntax error
```

- Right-click that same target to open the context menu and click the "Select Target Path" action

You should see a warning similar to this output in the terminal:

```
      File "C:\USD_inst\lib\python\pxr\Usdviewq\attributeViewContextMenu.py", line 207, in RunCommand
        _selectPrimsAndProps(self._dataModel, paths)
      File "C:\USD_inst\lib\python\pxr\Usdviewq\attributeViewContextMenu.py", line 81, in _selectPrimsAndProps
        raise PropertyNotFoundException(path)
    pxr.Usdviewq.common.PropertyNotFoundException: Property not found at path in stage: /CollectionTest/Geom/Shapes.collection:allShapes
```

And the selection will not change off of the `/CollectionTest/Geom` prim.

---

With the changes here, neither of these warnings should appear, and using the "Select Target Path" action on a UsdCollectionAPI path will cause that collection's prim to be selected.

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/usdImaging/usdviewq/attributeViewContextMenu.py
+++ b/pxr/usdImaging/usdviewq/attributeViewContextMenu.py
@@ -24,6 +24,7 @@
 
 from .qt import QtGui, QtWidgets, QtCore
 from pxr import Sdf
+from pxr import Usd
 from .usdviewContextMenuItem import UsdviewContextMenuItem
 from .common import (PropertyViewIndex, PropertyViewDataRoles,
                      PrimNotFoundException, PropertyNotFoundException)
@@ -68,6 +69,11 @@ def _selectPrimsAndProps(dataModel, paths):
         if not prim:
             raise PrimNotFoundException(primPath)
         prims.append(prim)
+
+        # No property will exist for a CollectionAPI path, so skip trying
+        # to get one and just use the prim on which the collection is authored.
+        if Usd.CollectionAPI.IsCollectionAPIPath(path):
+            continue
 
         if path.IsPropertyPath():
             prop = prim.GetProperty(path.name)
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8001

crevilla:
So we have a change in progress that will make these changes unnecessary.
The first is that we are about to update the UsdCollectionAPI schema so that the collection:{name} property path  will actually exist as an opaque valued type attribute in the CollectionAPI schema. With this change, the "Select Target Path" will actually be able select the collection attribute.
The second is that even with the collection:{name} attribute existing, there's a bug in the _updatePropertiesFromPropertyView code that assumes all relationship targets must be prim paths when in actuality they can be prim or property paths. With that fixed _ensureValidTargetPath should no longer receive an empty path when the collection:{name} target is selected.

mattyjams:
Cool! Both of those changes sound great to me.

Thanks for taking a look @crevilla!

tallytalwar:
@crevilla 's change discussed above is available in dev branch now: 6c826c5408bdc4d9a4b3f138bf59ae8d6337ea46

Closing this PR.

mattyjams:
I just gave this a try in the dev branch and it's working great for me.

Thanks again, @crevilla and @tallytalwar!

PR #2269, opened 2023-02-13
Replace `boost::iterator_adaptor` with explicitly specified iterator definition for `UsdNotice::ObjectsChanged::PathRange`

nvmkuruc:
### Description of Change(s)
- Introduces an `_UnderlyingIterator` type alias for `_PathsToChangesMap::const_iterator` in `UsdNotice::ObjectsChanged::PathRange::iterator`
- Explicitly implement the equality, increment, dereference, and arrow operators for `UsdNotice::ObjectsChanged::PathRange::iterator`
- Add `GetBase` method to the iterator for direct query of the underlying iterator

### Fixes Issue(s)
- #2228 

### DIFF
--- a/pxr/usd/usd/notice.h
+++ b/pxr/usd/usd/notice.h
@@ -149,15 +149,38 @@ public:
         {
         public:
             /// \class iterator
-            class iterator : public boost::iterator_adaptor<
-                iterator,                           // crtp base,
-                _PathsToChangesMap::const_iterator, // base iterator
-                const SdfPath&                      // value type
-            >
-            {
+            class iterator {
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-8007

pxr/usd/usd/notice.h
```diff
@@ -171,15 +194,22 @@ class UsdNotice {
                 /// details.
                 USD_API bool HasChangedFields() const;
```
crevilla, line 196: Per our discussion, it'll be better  to include a deprecated `base()` function that returns the underlyingIterator to maintain downstream compatibility in the meantime.

Issue #2272, opened 2023-02-13
`boost::transform_iterator` can be replaced


nvmkuruc:
### Description of Issue
There are currently two usages of `boost::transform_iterator` in USD.

In `UsdProperty`, it's used as the input to `SdfPathFindLongestPrefix`.  However, `SdfPathFindLongestPrefix` also takes a function argument that is applied to the iterator. This argument can be used in `transform_iterator`'s stead.

In `stage.cpp`, it's used twice.  In `_Stringify`, an intermediate `SdfPathVector` is currently already created and so `std::transform` can be used without additional overhead. The other usage transforms the inputs for usage in a private internal helper function.  However, the helper function is only used once in the file so the contract can be easily modified and the transformation moved inside of the helper.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-8010

nvmkuruc:
This was addressed last year and can be closed.

PR #2273, opened 2023-02-14
Replace usage of `boost::transform_iterator` in `pxr/usd/usd`

nvmkuruc:
### Description of Change(s)
- Use the function argument to `SdfPathFindLongestPrefix` to apply the transform instead of `boost::transform_iterator` in `UsdProperty`
- Use `std::transform` to apply the transformation to an explicitly named temporary in `_Stringify` in `pxr/usd/usd/stage.h`
- Change private helper `_ComputeSubtreesToRecompose` in `UsdStage` to accept map iterators and apply the transformation inside of the function instead of via `boost::transform_iterator`

### Fixes Issue(s)
- #2272 

### DIFF
--- a/pxr/usd/usd/property.cpp
+++ b/pxr/usd/usd/property.cpp
@@ -27,8 +27,6 @@
 #include "pxr/usd/usd/stage.h"
 #include "pxr/usd/pcp/targetIndex.h"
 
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-8011

PR #2275, opened 2023-02-14
CMake: append should be used for `CMAKE_MODULE_PATH`

CraigHutchinson:
### Description of Change(s)
`CMAKE_MODULE_PATH` is a directory list and shall be appended via `list(APPEND`

### Note
No major change but ensures when `CMAKE_MODULE_PATH` has been defined with alternative defaults on the command-line or via use of `CMAKE_PROJECT_TOP_LEVEL_INCLUDES`

PR Note:
- First commit to confirm CLA
- This change is small enough and deemed self-explanatory to not affect tests, I will however look into executing tests as a sanity check soon but do not see this PR being blocked on this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ if (NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8013

Issue #2276, opened 2023-02-15
`boost::iterator_range` can be replaced with explicit range types


nvmkuruc:
### Description of Issue
`boost::iterator_range` is used in `prim.h` and `primData.h`.

The typedefs, functions, and methods which use in `primData.h` are all private, unused, and be removed.

`UsdPrim{SiblingRange,SubtreeRange}` already have a skeleton interface for doxygen. Fleshing this out will allow for the removal of `iterator_range` as a core dependency.

This issue addresses usage in the core only and does not address the usage in `examples`.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-8014

nvmkuruc:
Closing as this was addressed last year.

PR #2280, opened 2023-02-15
Replace `boost::iterator_range` usage for ranges of `UsdPrim`s

nvmkuruc:
### Description of Change(s)
Depends on #2235
- Implement `UsdPrim{Sibling,Subtree}Range` using the existing doxygen stubs
- Relocate class declarations to after their corresponding iterator declarations
- Make templated constructor and operators explicit
- Add `cbegin` and `cend` methods
- Remove unused references to `boost::iterator_range` in `pxr/usd/usd/primData.h`
- Add test coverage for equality operators of `UsdPrimSubtreeRange` to `testUsdPrimGetDescendants`

### Fixes Issue(s)
- #2276 

### DIFF
--- a/pxr/usd/usd/prim.h
+++ b/pxr/usd/usd/prim.h
@@ -43,8 +43,7 @@
 
 #include "pxr/usd/sdf/path.h"
 
-#include <boost/range/iterator_range.hpp>
-
+#include <iterator>
 #include <string>
 #include <type_traits>
 #include <vector>
@@ -72,10 +71,10 @@ class UsdVariantSet;
 class SdfPayload;
 
 class UsdPrimSiblingIterator;
-typedef boost::iterator_range<UsdPrimSiblingIterator> UsdPrimSiblingRange;
+class UsdPrimSiblingRange;
 
 class UsdPrimSubtreeIterator;
-typedef boost::iterator_range<UsdPrimSubtreeIterator> UsdPrimSubtreeRange;
+class UsdPrimSubtreeRange;
 
 /// \class UsdPrim
 ///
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-8018

pxr/usd/usd/prim.h
```diff
@@ -2379,8 +2312,138 @@ class UsdPrimSiblingIterator {
...
+    UsdPrimSiblingRange(const UsdPrimSiblingRange&) = default;
+    UsdPrimSiblingRange(UsdPrimSiblingRange&&) = default;
+    UsdPrimSiblingRange& operator=(const UsdPrimSiblingRange&) = default;
+    UsdPrimSiblingRange& operator=(UsdPrimSiblingRange&&) = default;
+
+    /// Construct with a pair of iterators.
+    UsdPrimSiblingRange(UsdPrimSiblingIterator begin,
+                        UsdPrimSiblingIterator end) : _begin(begin),
+                                                      _end(end) {}
+
```
crevilla, line 2338: I'm assuming these templated constructors and assignment operators are to perfectly match what boost::iterator_range provides?
I feel like these are overkill given that I would never expect anyone to construct a sibling range without calling the Get*Children functions that return them (and it would be honestly dangerous to do so since sibling iteration is a proxy). How do you feel about just leaving these out?

pxr/usd/usd/prim.h
```diff
@@ -2379,8 +2312,138 @@ class UsdPrimSiblingIterator {
...
+    /// Const iterator type.
+    typedef UsdPrimSiblingIterator const_iterator;
+    /// Iterator difference type.
+    typedef std::ptrdiff_t difference_type;
+    /// Iterator value_type.
+    typedef iterator::value_type value_type;
+    /// Iterator reference_type.
+    typedef iterator::reference reference;
+
+    UsdPrimSiblingRange() = default;
```
crevilla, line 2332: I'm curious about having to explicit default the implicit copy/move constructors/assignment. Is this necessary because the templated constructors below create a user-defined copy constuctor when expanded against UsdPrimSiblingRange?
If we get rid of the templated constructor/assignment, can we get rid of these declarations too?

pxr/usd/usd/prim.h
```diff
@@ -2379,8 +2312,138 @@ class UsdPrimSiblingIterator {
...
+    bool empty() const { return begin() == end(); }
+
+private:
+    /// Equality comparison.
+    friend bool operator==(const UsdPrimSiblingRange &lhs,
+                           const UsdPrimSiblingRange &rhs) {
+        return lhs.equal(rhs);
+    }
+
+    /// Equality comparison.
```
crevilla, line 2387: I'm also wondering if the template comparision operators are really needed (though they seem slightly less unneeded than the templated constructors.

pxr/usd/usd/prim.h
```diff
@@ -2591,8 +2585,141 @@ class UsdPrimSubtreeIterator {
...
+    typedef UsdPrimSubtreeIterator iterator;
+    /// Const iterator type.
+    typedef UsdPrimSubtreeIterator const_iterator;
+    /// Iterator difference type.
+    typedef std::ptrdiff_t difference_type;
+    /// Iterator value_type.
+    typedef iterator::value_type value_type;
+    /// Iterator reference_type.
+    typedef iterator::reference reference;
+
```
crevilla, line 2588: Exact same comments for this range as the sibling range.

PR #2288, opened 2023-02-17
[usdviewq] Hide render frame when --norender is set

asluk:
Strawman PR to hide render frame with --norender is set; happy to move this logic somewhere earlier in the function if desired and also avoid populating hidden UI contents...


sunyab:
Filed as internal issue #USD-8025

spiffmon:
Oh wow, thanks for catching that, @asluk !  That's a regression from the work to make the cool new render view toolbar.  I found a better change that also corrects errors that spew when you drag the splitter in `--norender` mode, but will still do the checkin against this PR!

Issue #2290, opened 2023-02-17
`boost::in_place` can be replaced by `emplace` methods


nvmkuruc:
### Description of Issue

`TfType` and `UsdStage` use `boost::in_place` to initialize `boost::optional` instances.  The STL has introduced `emplace` as the standard way to do `in_place` initialization. As recent versions of `boost::optional` support `emplace`, `boost::in_place` usage can be removed. Doing so will make it easier to swap out `boost::optional` for `std::optional` in the future and removes another boost dependency.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

sunyab:
Filed as internal issue #USD-8027

PR #2291, opened 2023-02-17
Replace `boost::in_place` with `emplace` method

nvmkuruc:
### Description of Change(s)
- Replaces `boost::in_place` with `emplace` method in `TfType` and `UsdStage` usage

### Fixes Issue(s)
- #2290 


sunyab:
Filed as internal issue #USD-8028

Issue #2250, opened 2023-02-08
`boost::totally_ordered` can be replaced with explicit operator overloads


nvmkuruc:
### Description of Issue

`boost::totally_ordered` adds `operator!=`, `operator>`, `operator<=`, and `operator>=` to any derived class.  Explicit implementation of these operators would remove this dependency on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7984

nvmkuruc:
This was addressed by several PRs last year.

PR #2293, opened 2023-02-17
Replace `boost::totally_ordered` with explicit operator overloads for `SdfAssetPath`

nvmkuruc:
### Description of Change(s)
- Add explicit implementations of `operator{<=,>,>=,!=}` to `SdfAssetPath`
- Expose copy constructor to python
- Expose comparison operators to python
- Add unit test coverage for `Sdf.AssetPath` operators

### Fixes Issue(s)
- #2250 


sunyab:
Filed as internal issue #USD-8030

pxr/usd/sdf/testenv/testSdfReference.py
```diff
@@ -27,6 +27,84 @@
 from pxr import Sdf, Tf
 import unittest
```
crevilla: I'm glad to see test cases added for SdfAssetPath, but I don't feel they belong in testSdfReference (it's not obvious to look for them here.) Can we instead add a new testSdfAssetPath to hold these tests? I'm a little surprised we don't have a testSdfAssetPath already.
Additionally, we typically make one unittest.TestCase class and have multiple test case methods for each case. That would make more sense here too for the trio of SdfAssetPath tests.

Issue #2296, opened 2023-02-17
`pxr/usd/usd/crateFile.{h,cpp}` can use `robin_map` instead of `boost::container::flat_map`


nvmkuruc:
### Description of Issue

`CrateFile` uses `boost::flat_map` to record a registry of type specific functions and a time code ordered list of values.

The first usage does not rely on the order of `boost::flat_map` and be replaced directly with `robin_map`.

The second usage requires an addition sort of the keys / iterators after the map has been populated.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-8037

nvmkuruc:
This was addressed by #2297 and can be closed.

PR #2297, opened 2023-02-17
Replace `boost::container::flat_map` with `robin_map`

nvmkuruc:
### Description of Change(s)
- Change `_packValueFunctions` to be a `robin_map` as it does not rely on the order provided by `boost::container::flat_map`
- Use `robin_map` to store `allValuesAtAllTimes` and create a sorted view of the map iterators for the subsequent order dependent operation

### Fixes Issue(s)
- #2296 


tallytalwar:
Filed as internal issue #USD-8038

PR #2298, opened 2023-02-18
Specialize `TfHashAppend` for `std::vector<bool>`

nvmkuruc:
### Description of Change(s)
The `TfHashAppend` specialization for `std::vector` uses the `.data()` method to hash contiguous elements. However, `std::vector<bool>` is not guaranteed to provide a `.data()` method and does not guarantee the underlying data's length is the container's `size() * sizeof(bool)`.

This change modifies the `TfHashAppend` specialization to not match against `std::vector<bool>` and provides a more explicit specialization that uses `std::hash` instead.

This issue was detected during testing of #2176 which changes `VtHash` to prefer `TfHash` over `hash_value` for hashing.

### Fixes Issue(s)
- #2172 

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -89,12 +89,20 @@ TfHashAppend(HashState &h, std::pair<T, U> const &p)
     h.Append(p.second);
 }
 
-// Support std::vector.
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8039

PR #2304, opened 2023-02-21
Specialize `TfHashAppend` for `std::shared_ptr` and `std::unique_ptr`

nvmkuruc:
### Description of Change(s)
`TfHash` and `VtHash` may utilize ADL of `hash_value` for their hashing implementations. As usage of `boost::hash_value` is being phased out, it's possible for other `hash_value` implementations to be picked up. Specifically, while testing the removal of `boost::hash_value` from `pxr/base/vt/hash.h` (#2176), `stdext::hash_value` was getting discovered on certain Windows builds when hashing `std::shared_ptr`.  `boost::hash_value` was previously providing the `std::shared_ptr` specialization for `VtHash` but with #2176 `VtHash` prefers `TfHash` and `TfHash` may resolve to the `stdext::hash_value` implementation.

This PR provides an implementation of `TfHashAppend` for `std::shared_ptr` and `std::unique_ptr` which hash their underlying pointers and takes precedence over `hash_value`.

While providing a `TfHashAppend` implementation for `std::shared_ptr` and `std::unique_ptr` precludes `hash_value` ADL, it's worth noting that for user code using `VtHash` or `TfHash`, other types may be affected by this and additional specializations may be required.

This should not affect the specialization of `TfHashAppend` for `std::shared_ptr<_Untyped>` used in `pxr/usd/ar` which hashes the pointed to value instead of the pointer.

### Fixes Issue(s)
- #2172 

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -34,6 +34,7 @@
 #include <cstring>
 #include <string>
 #include <map>
+#include <memory>
 #include <set>
 #include <typeindex>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8044

PR #2175, opened 2023-01-18
Replace boost hash usage with TfHash in pxr/base/gf

nvmkuruc:
### Description of Change(s)
Requires: #2173
To remove the dependency of pxr/base/gf on boost's hashing functions
* Copy constructor for python bindings added when missing to support test coverage (`Gf.Rotation`, `Gf.Quaternion`, `Gf.Interval`)
* `__hash__` operator overloads are added when missing to support test coverage (`Gf.BBox3d`, `Gf.Frustum`, `Gf.Quat{f,h,d}`, `Gf.Rect2i`, `Gf.Rotation`)  
* `__hash__` test coverage added to `testGfRotation`, `testGfVec`, `testGfRect2i`, `testGfRange`, `testGfQuarternion`, `testGfMatrix`, `testGfInterval`, `testGfFrustum`, `testGfDualQuaternion`, `testGfBBox3d`, and `testGfFrustum`.
* Replaced `boost::hash_combine` with `TfHash::Combine` in `hash_value` implementations.  Preserving `hash_value` allows `boost::hash` to still work with these types without the explicit dependency.  `TfHash` will use these `hash_value` implementations when no `TfHashAppend` overload has been provided.

### Fixes Issue(s)
-#2172 (more PRs forthcoming)

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2310, opened 2023-02-23
Explicitly delete hashing of `float` via implicit conversion to `GfHalf`

nvmkuruc:
### Description of Change(s)
During testing of #2175 (which removes boost's hashing from multiple code sites), it was observed that ADL was not resolving an implementation of `hash_value` for some qualifier variations of `GfHalf` under certain compiler conditions (specifically, on Windows when the `PXR_NS` has been disabled). SFINAE was being used to suppress hashing via implicit conversion of `float` to `GfHalf`. We can avoid SFINAE and suppress the implicit conversion by explicitly deleting `pxr_half::hash_value(float)`. With this change, ADL now works as expected in concert with #2175.

This change also makes the `half` input argument to `hash_value` passed by value under the principle of passing cheap to copy types by value.

Test coverage has been added for `hash_value`.

### Fixes Issue(s)
- #2172 

### DIFF
--- a/pxr/base/gf/half.h
+++ b/pxr/base/gf/half.h
@@ -35,8 +35,6 @@
 #include "pxr/base/gf/ilmbase_halfLimits.h"
 #include "pxr/base/gf/traits.h"
 
-#include <type_traits>
-
 PXR_NAMESPACE_OPEN_SCOPE
 
 /// A 16-bit floating point data type.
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8052

tallytalwar:
Available in dev as a3a66af11a81c56584f51a9da191a0cd43fccbae. Thanks @nvmkuruc

Issue #2311, opened 2023-02-23
Replace `boost::filter_iterator` usage with fully specified iterator class


nvmkuruc:
### Description of Issue

`pxr/usd/sdf/childrenView.h` uses `boost::filter_iterator` to apply a filtering predicate to another iterator. This can be explicitly specified to remove another boost dependency.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-8055

nvmkuruc:
Closing, as #2312 addressed this.

PR #2312, opened 2023-02-23
Replace `boost::filter_iterator` with explicit iterator implementation

nvmkuruc:
### Description of Change(s)
Converts the `_Predicate` private class into an explicit `_FilterIterator` bidirectional iterator implementation.

### Fixes Issue(s)
- #2311 

### DIFF
--- a/pxr/usd/sdf/childrenView.h
+++ b/pxr/usd/sdf/childrenView.h
@@ -31,7 +31,6 @@
 #include "pxr/usd/sdf/children.h"
 #include "pxr/base/tf/iterator.h"
 
-#include <boost/iterator/filter_iterator.hpp>
 #include <boost/iterator/iterator_facade.hpp>
 #include <boost/iterator/reverse_iterator.hpp>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8056

pxr/usd/sdf/childrenView.h
```diff
@@ -76,27 +75,75 @@ template <typename _Owner, typename _InnerIterator, typename _DummyPredicate>
...
         {
             return _owner->GetPredicate()(
                 _Owner::Adapter::Convert(x));
         }
 
-    private:
-        const _Owner* _owner;
+        void filter()
+        {
+            while (_underlyingIterator != _end &&
```
crevilla, line 141: I'm pretty sure you have this inverted. According to the boost filter_iterator documentation, the iterator retains the element if the predicate returns true and skips the element if the predicate is false. Here, you're skipping when the predicate is true.

pxr/usd/sdf/childrenView.h
```diff
@@ -76,27 +75,75 @@ template <typename _Owner, typename _InnerIterator, typename _DummyPredicate>
...
+            return _underlyingIterator != other._underlyingIterator;
+        }
 
-        _Predicate() : _owner(NULL) { }
-        _Predicate(const _Owner* owner) : _owner(owner) { }
+        const _InnerIterator& GetBase() const { return _underlyingIterator; }
 
-        bool operator()(const value_type& x) const
+    private:
+        // Skip any iterators that don't satisfy the predicate
```
crevilla, line 132: These `shouldFilter` and `filter` functions aren't here for API conformance right? If I'm right, they should follow our coding conventions for private methods, i.e. capitalized and starting with an underscore

pxr/usd/sdf/childrenView.h
```diff
@@ -76,27 +75,75 @@ template <typename _Owner, typename _InnerIterator, typename _DummyPredicate>
...
+    class _FilterIterator {
     public:
-        typedef typename _Owner::value_type value_type;
+        using iterator_category = std::forward_iterator_tag;
+        using value_type = typename _InnerIterator::value_type;
+        using reference = typename _InnerIterator::reference;
+        using pointer = typename _InnerIterator::pointer;
+        using difference_type = typename _InnerIterator::difference_type;
+
+        _FilterIterator() = default;
```
crevilla, line 88: The body of all three of these should be on a new line like you did for all the rest of the functions.

pxr/usd/sdf/childrenView.h
nvmkuruc: That was it.  All tests are now passing.

PR #2317, opened 2023-03-01
usd: align Python UsdProperty property stack getter API with C++

mattyjams:
### Description of Change(s)

On the C++ side, the `GetPropertyStack()` and `GetPropertyStackWithLayerOffsets()` member functions of `UsdProperty` take a `UsdTimeCode` as an optional argument. This change makes that argument optional through the Python API as well.


tallytalwar:
Filed as internal issue #USD-8065

PR #2319, opened 2023-03-02
Add missing HD_API in hdGp

thibaud-lambert:
### Description of Change(s)
Missing HD_API prevents from inheriting from `HdGpGenerativeProcedural` and `HdGpGenerativeProceduralPlugin` on windows because symbols are not exported (for example in the case of an external plugin) . This PR simply adds the missing HD_API in generativeProcedural.h and generativeProceduralPlugin.h.

### DIFF
--- a/pxr/imaging/hdGp/CMakeLists.txt
+++ b/pxr/imaging/hdGp/CMakeLists.txt
@@ -18,6 +18,9 @@ pxr_library(hdGp
         generativeProceduralPluginRegistry
         generativeProceduralResolvingSceneIndex
         sceneIndexPlugin
+        
+    PUBLIC_HEADERS
+        api.h
 
     RESOURCE_FILES
         plugInfo.json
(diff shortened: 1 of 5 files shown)


sunyab:
Filed as internal issue #USD-8080

sunyab (changes requested):
Hi @thibaud-lambert, since HdGp is its own library we need to add a new HDGP_API macro instead of reusing HD_API. This should be pretty straightforward:

- Add a new api.h to hdGp. This can be done by copying hd/api.h into this directory and then adjusting all of the HD prefixes to HDGP instead.
- Add api.h as a PUBLIC_HEADER to CMakeLists.txt -- see hd/CMakeLists.txt as an example
- Include pxr/imaging/hdGp/api.h in these two headers and use HDGP_API instead of HD_API

Thanks!

thibaud-lambert:
Hi, I made the appropriate changes. `HdGP` now use its own macro `HDGP_API` rather than reusing `HD_API`.

Issue #2318, opened 2023-03-01
Preprocessor usage can be reduced in pxr/base/tf


nvmkuruc:
### Description of Issue

Both `pxr/base/tf/enum.h` and `pxr/base/tf/scopeDescription.h` provide macros which use the boost preprocessor.

The `enum.h` usage can be removed in favor of always passing a string since the default argument is an empty string.
The `scopeDescription.h` usage can be removed in favor of variadic templates.  Specific forwarding overloads can be provided to avoid unnecessarily creating new strings if no formatting is needed.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-8066

nvmkuruc:
This was addressed by #2320.  Closing.

PR #2320, opened 2023-03-02
Remove preprocessor usage in macros in `pxr/base/tf`

nvmkuruc:
### Description of Change(s)
- The boost preprocessor usage in `enum.h` are replaced by simply passing the variadic arguments as `std::string` arguments. If there are no arguments, the macro expands to `std::string{}` which is the empty string (and equivalent to the default argument). Otherwise, the variadic arguments will be passed to the `std::string` constructor.
- The boost preprocessor usage in `scopeDescription.h` are replaced with variadic templates. Forwarding overloads are provided for common string types to avoid needless construction of a new string when no formatting arguments are provided.

### Fixes Issue(s)
- #2318 

### DIFF
--- a/pxr/base/tf/enum.h
+++ b/pxr/base/tf/enum.h
@@ -37,7 +37,6 @@
 #include "pxr/base/tf/api.h"
 
 #include <boost/operators.hpp>
-#include <boost/preprocessor/punctuation/comma_if.hpp>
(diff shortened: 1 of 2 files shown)


sunyab:
Filed as internal issue #USD-8081

Issue #1471, opened 2021-03-03
[tbb] Please support tbb version 2021.1.1 


JackBoosY:
Hi guys, I'm vcpkg maintainer.
I tried to update tbb to the latest version 2021.1.1 in vcpkg (PR microsoft/vcpkg#16321), but seems that usd doesn't support this version. And some other libraries has been adapted for that.
Can you consider updating the code to adapt to this version?

Thanks.

jilliene:
Filed as internal issue #USD-6600

manuelk:
Confirmed - after much trial & error, was able to get a working build using the 'tbb_2019' branch (which unfortunately did not have a cmake build or exported config yet).

The biggest hurdle seems to be the obsoleting of tbb::atomic (replaced with std::atomic sometime in late 2020). 

IMHO, there is little reason to not submodule tbb & just let tbb's cmake config take care of everything, but just in case, I am attaching the snippet i had to modify in the FindTBB.cmake module (because the version file moved too)

```
  find_file(TBB_VERSION_FILE 
      NAMES oneapi/tbb/version.h tbb/version.h tbb/tbb_stddef.h       
      HINTS ${TBB_INCLUDE_DIRS} ${TBB_SEARCH_DIR}
      PATHS ${TBB_DEFAULT_SEARCH_DIR})

  if(TBB_VERSION_FILE)
    file(READ "${TBB_VERSION_FILE}" _tbb_version_file)

    string(REGEX REPLACE ".*#define TBB_VERSION_MAJOR ([0-9]+).*" "\\1"
        TBB_VERSION_MAJOR "${_tbb_version_file}")

    string(REGEX REPLACE ".*#define TBB_VERSION_MINOR ([0-9]+).*" "\\1"
        TBB_VERSION_MINOR "${_tbb_version_file}")

    string(REGEX REPLACE ".*#define TBB_INTERFACE_VERSION ([0-9]+).*" "\\1"
        TBB_INTERFACE_VERSION "${_tbb_version_file}")

    set(TBB_VERSION "${TBB_VERSION_MAJOR}.${TBB_VERSION_MINOR}")
  endif()
```

meshula:
Maybe the biggest hurdle can be cleared by moving USD to std::atomic :) You know what gets my vote! @manuelk , is that the approach you took?

manuelk:
[...]

PR #2321, opened 2023-03-02
Use non-deprecated TBB headers in `pxr/base/work`

nvmkuruc:
### Description of Change(s)
`tbb/task.h` is deprecated / removed in newer versions of TBB, but `tbb/task_group.h` is not. Using `tbb/task_group.h` when possible reduces the warning count when building with TBB 2020.3 and is a small step in making `pxr/base/work` compatible with oneAPI.

### Fixes Issue(s)
- #1600
- #1471 
This PR is related to the issues above, but does not fully address.


sunyab:
Filed as internal issue #USD-8082

Issue #2211, opened 2023-01-27
Use STL type_traits instead of boost traits


nvmkuruc:
### Description of Issue

C++{11,14} offers type traits that avoid the need to depend on boost.

### Steps to Reproduce

N/A

### System Information (OS, Hardware)

N/A

### Package Versions

N/A

### Build Flags

N/A

nvmkuruc:
I have a fix for most of these (except for `has_left_shift`).

tallytalwar:
Filed as internal issue #USD-7941

nvmkuruc:
Closing, as this was addressed in changes made last year.

PR #2323, opened 2023-03-03
Remove unused boost headers from `pch.h` files

nvmkuruc:
### Description of Change(s)
`boost/mpl/or.hpp`, `boost/mpl/if.hpp`, `boost/mpl/and.hpp`, `boost/type_traits/is_signed.hpp`, and `boost/utility/in_place_factory.hpp` are no longer used in the `pxr` code base and can be safely removed from all `pch.h` files.

### Fixes Issue(s)
- #2211 
- #2291 


sunyab:
Filed as internal issue #USD-8084

spiffmon:
@nvmkuruc , we're going to close this in favor of rerunning our internal pch-generator script so we can be confident we'll continue to get the right results going forward.  Thanks, though!

PR #2324, opened 2023-03-06
Apple: MaterialX on Metal

slingthor:
### Description of Change(s)
- Enables MaterialX support on Metal. Makes the MaterialX shader generator polymorphic and implements the Metal based functionality.
- Points to Apple's branched hosted version of MaterialX until official release of MaterialX with Metal
- Makes sure that on Metal we don't attempt to mipmap textures containing only one pixel.
- Metal versions 2.2 and earlier don't return upon usage of the discard keyword. Improve discard handling to pass MaterialX unit tests on Metal.

Co-authored by:
Morteza Mostajabodaveh
Jon Creighton

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1574,11 +1574,12 @@ DRACO = Dependency("Draco", InstallDraco, "include/draco/compression/decode.h")
 ############################################################
 # MaterialX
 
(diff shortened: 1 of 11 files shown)


sunyab:
Filed as internal issue #USD-8085

PR #2325, opened 2023-03-06
APPLE: Explicitly return after discarding in shader programs where appropriate

slingthor:
### Description of Change(s)
Metal 2.2 and older requires explicit returns for discards in order to stop shader execution.
Along with an improvement made in the MaterialX pull request https://github.com/PixarAnimationStudios/USD/pull/2324 , discards in the system on the used version of Metal are more correct and explicit, should increase performance and reduce risk of errors and should fix some errors as well.

### Fixes Issue(s)
- Not explicitly returning after discards on Metal 2.2


sunyab:
Filed as internal issue #USD-8086

Issue #2305, opened 2023-02-21
Replace `boost::iterator_facade` usage with fully specified iterator classes


nvmkuruc:
### Description of Issue

Most usages of `boost::iterator_facade` in pxr can be replaced with fully specified iterator classes. Addressing this would reduce USD's dependence on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

tallytalwar:
Filed as internal issue #USD-8045

nvmkuruc:
This was addressed by several PRs last year.

PR #2328, opened 2023-03-13
Replace `boost::iterator_facade` with explicit iterator definition for `pcp/iterator.h`

nvmkuruc:
### Description of Change(s)
- Replace `iterator_facade` with explicit implementation for `PcpNodeIterator`, `PcpPrimIterator`, and `PcpPropertyIterator`.
- Provide `_PtrProxy` type for `PcpNodeIterator` and `PcpPrimIterator` (whose reference type is a proxy value and not a true reference) to provide a safe `operator->` implementation.
- Use `nullptr` instead of `0` to default initialize pointer types (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nullptr)
- Use in-class initialization for defaults (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer)
- While testing removal of `boost::reverse_iterator` in some parallel work (#2333), it was observed that the current implementation of `PcpNodeIterator` hangs when `std::prev` was used. The source of the hang was not determined but not observed in this implementation. `testPcpIterator` has been augmented to include tests to validate that the increment / decrement operators and `std::prev` / `std::next` can be used and produce symmetrical end states.

### Fixes Issue(s)
- #2305 

### DIFF
--- a/pxr/usd/pcp/iterator.cpp
+++ b/pxr/usd/pcp/iterator.cpp
@@ -34,12 +34,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 ////////////////////////////////////////////////////////////
 
-PcpPrimIterator::PcpPrimIterator()
-    : _primIndex(NULL)
-    , _pos(PCP_INVALID_INDEX)
-{
-    // Do nothing
-}
+PcpPrimIterator::PcpPrimIterator() = default;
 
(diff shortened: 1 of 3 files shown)


sunyab:
Filed as internal issue #USD-8089

PR #2205, opened 2023-01-26
Replace `boost::reverse_iterator` with `std::reverse_iterator` for proxy-less iterators

nvmkuruc:
### Description of Change(s)
`std::reverse_iterator` can be used whenever the underlying iterator's `operator*` returns true references (and not proxy temporary values).

The C++ specification states that until C++20, `std::addressof(operator*())` should be used as `std::reverse_iterator`'s implementation of `operator->`. This does not work when the result of `operator*` is a temporary.

For the remaining cases, explicit reverse iterator implementations will be required compilers consistently support the C++20 specification (which is compatible with proxy reference types). This is addressed by #2333.

### Fixes Issue(s)
- #2202

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

Issue #2202, opened 2023-01-26
Prefer `std::reverse_iterator` instead of `boost::reverse_iterator`


nvmkuruc:
### Description of Issue

`std::reverse_iterator` is already used in several places in the USD code base.  Switching `boost::reverse_iterator` usage to this reduces the number of dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

sunyab:
Filed as internal issue #USD-7921

nvmkuruc:
Closing as this was fixed in #2333

PR #2333, opened 2023-03-14
Provide adapter for `std::reverse_iterator` to replace `boost::reverse_iterator` for proxy reference iterators

nvmkuruc:
### Description of Change(s)
Depends on #2205, #2328

#2205 replaces `std::reverse_iterator` in several but not all instances. The C++ standard states `std::reverse_iterator::operator->` should return the equivalent of `std::addressof(operator*())`. For iterators that return proxy types as references (say `PcpNodeRef`), taking the address of a temporary is a problematic operation. The C++20 standard modifies this specification to `std::prev(/*underlying iterator*/).operator->()`. Some compilers are already using the newer specification.

This change introduces `Tf_ProxyReferenceReverseIterator` as an adapter for `std::reverse_iterator` that specializes `operator->` to use the C++20 specification, deferring all other operations to `std::reverse_iterator`. When possible, this is explicitly done with `using` directives. However, in many cases, the return and argument types require wrapper methods and functions.

This change depends on #2328 due to a bug with the current implementation of `PcpNodeIterator` and `std::prev`.

### Fixes Issue(s)
- #2202 

### DIFF
--- a/pxr/base/tf/iterator.h
+++ b/pxr/base/tf/iterator.h
@@ -409,6 +409,150 @@ constexpr size_t TfArraySize(const T (&array)[N]) noexcept
     return N;
 }
 
+/// A reverse iterator adapter for `std::reverse_iterator` that provides
+/// an `operator->` compatible with proxy reference types.
+/// This should only be used when the underlying iterator's reference
+/// is a value type and should become unnecessary in newer compilers and C++20.
+/// This implementation was written for use with random access iterators but
+/// could be extended to bidirectional iterators if necessary.
+template <typename UnderlyingIterator>
+class Tf_ProxyReferenceReverseIterator :
+    private std::reverse_iterator<UnderlyingIterator> {
+    // private API for interacting with an STL reverse_iterator of the
+    // UnderlyingIterator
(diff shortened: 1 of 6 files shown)


tallytalwar:
Filed as internal issue #USD-8122

pxr/base/tf/iterator.h
```diff
@@ -409,6 +409,148 @@ constexpr size_t TfArraySize(const T (&array)[N]) noexcept
...
+                 "Tf_ProxyReferenceReverseIterator must wrap a random "
+                 "access iterator.");
+
+    Tf_ProxyReferenceReverseIterator() = default;
+    explicit Tf_ProxyReferenceReverseIterator(UnderlyingIterator it) :
+        ReverseIterator(it) {
+    }
+
+    // Operators and functions which can just use the underlying STL
+    // implementation
```
crevilla, line 454: Since these three functions are the only ones that work as is from std::reverse_iterator, I'm wondering if it would be better to use a private inherit of std::reverse_iterator and just explicitly implement the pass throughs to the parent for these functions also. The extra protection against converting to or returning as the underlying STL implementation seems beneficial.

PR #2334, opened 2023-03-14
Avoid implicit conversion to `TfWeakPtr` when iterating in `dynamicFileFormatContext.cpp`

nvmkuruc:
### Description of Change(s)
The return type of `PcpLayerStack::GetLayers` is a container of `SdfLayerRefPtr`. However, in `pxr/usd/pcp/dynamicFileFormatContext.cpp`, a range based for loop was iterating over this container using `const SdfLayerHandle&` (an alias of `TfWeakPtr`). This produced the following warning in GCC11.

```
dynamicFileFormatContext.cpp:81:36: warning: loop variable ‘layer’ of type ‘const SdfLayerHandle&’ {aka ‘const pxrInternal_v0_23__pxrReserved__::TfWeakPtr<pxrInternal_v0_23__pxrReserved__::SdfLayer>&’} binds to a temporary constructed from type ‘const pxrInternal_v0_23__pxrReserved__::TfRefPtr<pxrInternal_v0_23__pxrReserved__::SdfLayer>’ [-Wrange-loop-construct]
```

To avoid taking a reference to a copy via implicit conversion from `TfRefPtr` to `TfWeakPtr`, this changes the type of `layer` in this loop to `const auto&`.

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/usd/pcp/dynamicFileFormatContext.cpp
+++ b/pxr/usd/pcp/dynamicFileFormatContext.cpp
@@ -78,7 +78,7 @@ private:
     {
         // Search the node's layer stack in strength order for the field on
         // the spec.
-        for (const SdfLayerHandle &layer : node.GetLayerStack()->GetLayers()) {
+        for (const auto &layer : node.GetLayerStack()->GetLayers()) {
             VtValue value;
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8123

Issue #2327, opened 2023-03-10
Setting `VtArray` elements via index in python does not work with `[-1]` but does work correctly with other negative indices


nvmkuruc:
### Description of Issue
The python bindings for `VtArray` implements setting by index in terms of setting by slice, incrementing the index by 1 to create the slice.  However, for `-1`, that creates an empty slice and the operation silently fails.

### Steps to Reproduce
1. Append the following test case to `testVtArray.py`
```
    def test_SetNegativeIndices(self):
        d = Vt.DoubleArray([1.0, 3.0, 5.0])
        d[-1] = 2.0
        d[-2] = 4.0
        self.assertEqual(d, [1.0, 4.0, 2.0])
```
2. Build
3. `ctest -C Release -R testVtArray`
You should see the following error message.
```
    self.assertEqual(d, [1.0, 4.0, 2.0])
AssertionError: Vt.DoubleArray(3, (1, 4, 5)) != [1.0, 4.0, 2.0]
```
You can see that the second element from back was correctly set to `4.0` but the last element was not set to `2.0`.

### System Information (OS, Hardware)
Linux

### Package Versions
Python 3.10
TBB 2020.3

### Build Flags
`--tests --no-imaging`

sunyab:
Filed as internal issue #USD-8088

FlorianZ:
@nvmkuruc good find! Are you interested in tackling this? We'd consider a contribution, for sure. Otherwise, we can pursue this internally, but it doesn't currently rise to the top of our internal priority list. Thank you!!

nvmkuruc:
Happy to submit one, but didn't want to start in case there was a strong opinion about the best fix.  My instinct is to just special case `-1` and convert it to `[-1 + size(), size())`.

gitamohr:
I have a fix for this, thanks for reporting, @nvmkuruc !

PR #2335, opened 2023-03-15
Fix setting `VtArray` elements by index for `[-1]`

nvmkuruc:
### Description of Change(s)
The code path for setting `VtArray` elements by index converts the index into a slice by adding 1 to the index.  However, for index -1, this produces the slice `[-1:0]` which is empty.

This change fixes the behavior of setting elements with index equal to -1. `slice{-1, slice_nil}`  is now used to create a slice equivalent to to `[-1:]`. This is both non-empty and only contains the last element.

A test case has been added to exercise setting by negative indices.

### Fixes Issue(s)
- #2327 

### DIFF
--- a/pxr/base/vt/testenv/testVtArray.py
+++ b/pxr/base/vt/testenv/testVtArray.py
@@ -60,6 +60,12 @@ class TestVtArray(unittest.TestCase):
         with self.assertRaises(TypeError):
             a = d['foo']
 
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8124

FlorianZ:
This should be fixed with the next dev push. Thank you!

PR #2337, opened 2023-03-15
Allow customizing the C++ standard used for building

stilllman:
### Description of Change(s)

Instead of hardcoding `CMAKE_CXX_STANDARD` to `14`, set it only if it isn't set already so that consuming projects can customize it, to build with a consistent standard.


stilllman:
Recreated with different account: #2339

PR #2338, opened 2023-03-15
Apple: Add clip plane handling on Metal in Storm

slingthor:
### Description of Change(s)
- Adds clip plane outputs in vertex stages
- Adds ability to write arrays to Metal shader stage output
- Is selective about when to write out arrays as current assumptions about the array out are around the post tessellation stage, and we don't want those attributes to create array style outputs

### Fixes Issue(s)
- Clip plane distance having no effect on Metal

### DIFF
--- a/pxr/base/plug/initConfig.cpp
+++ b/pxr/base/plug/initConfig.cpp
@@ -38,9 +38,10 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


tallytalwar:
Filed as internal issue #USD-8126

slingthor:
@tallytalwar This one has been modified for correctness

slingthor:
Replaced by PR 2444

PR #2339, opened 2023-03-15
Allow customizing the C++ standard used for building

touraill-adobe:
### Description of Change(s)

Instead of hardcoding CMAKE_CXX_STANDARD to 14, set it only if it isn't set already so that consuming projects can customize it, to build with a consistent standard.

### DIFF
--- a/cmake/defaults/CXXDefaults.cmake
+++ b/cmake/defaults/CXXDefaults.cmake
@@ -25,10 +25,20 @@ include(CXXHelpers)
 include(Version)
 include(Options)
 
-# Require C++14
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+# Default to C++14
+if (NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
+if (NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+endif()
+if (NOT DEFINED CMAKE_CXX_EXTENSIONS)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
+
+if (CMAKE_CXX_STANDARD EQUAL 98 OR CMAKE_CXX_STANDARD LESS 14)
+    message(FATAL_ERROR "USD requires C++14 at least, CMAKE_CXX_STANDARD must be >= 14")
+endif()
 
 if (CMAKE_COMPILER_IS_GNUCXX)
     include(gccdefaults)


meshula:
Hi Luc! I wonder if we might test, if the `CMAKE_CXX_STANDARD` has been externally supplied, that the `CMAKE_CXX_STANDARD` is at least 14, to provide users with guidance if they are attempting to build under say c++11?

thomthom:
I had a similar question in the PR: https://github.com/PixarAnimationStudios/OpenSubdiv/pull/1297#discussion_r1132215396

One option is to set this only if this project is the root project (not consumed as a sub project). That's what I've done for my own CMake library project.

https://github.com/PixarAnimationStudios/OpenSubdiv/blob/8ffa2b6566be10209529d7a0d1db02a0796b160c/CMakeLists.txt#L62-L65

cmake/defaults/CXXDefaults.cmake
```diff
@@ -25,8 +25,10 @@ include(CXXHelpers)
 include(Version)
 include(Options)
 
-# Require C++14
-set(CMAKE_CXX_STANDARD 14)
+# Default to C++14
+if (NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
thomthom: Probably worth giving these other CXX values the same treatment.

touraill-adobe:
> Hi Luc! I wonder if we might test, if the `CMAKE_CXX_STANDARD` has been externally supplied, that the `CMAKE_CXX_STANDARD` is at least 14, to provide users with guidance if they are attempting to build under say c++11?

Definitely, let me update the PR.

cmake/defaults/CXXDefaults.cmake
touraill-adobe: I was hesitant to do so but I think you are right for `CMAKE_CXX_STANDARD_REQUIRED`: if a consuming project wants to build with C++20 but fallback to C++17 if 20 is not available, it should probably be allowed to do so. On the other hand I don't know if it makes sense to let `CMAKE_CXX_EXTENSIONS` be customizable considering USD doesn't use extensions. But maybe it can cause issues if some headers change depending on extensions, as they would be different in different translation units :thinking: . Let's leave it configurable as well.

touraill-adobe: Done

tallytalwar:
Filed as internal issue #USD-8127

erslavin:
Perhaps consider using `target_compile_features` instead to set the minimum c++ standard required (bonus - it can be set as an `INTERFACE` property).  You can still override what you use by setting `CMAKE_CXX_STANDARD` so maybe you get the best of both worlds?

touraill-adobe:
> Perhaps consider using `target_compile_features` instead to set the minimum c++ standard required (bonus - it can be set as an `INTERFACE` property). You can still override what you use by setting `CMAKE_CXX_STANDARD` so maybe you get the best of both worlds?

Unfortunately, `target_compile_features` will not simply _check_ the requirement, it will _enforce_ it, possibly propagating it to other targets. So you might end up building different targets of your project with different C++ standard, which is precisely what this change is trying to avoid. We definitely not want to have half of our project building with C++20 instead of C++17 because one dependency down the graph decided to build with C++20 behind our back for instance. I honestly don't understand why cmake implemented this in such a way, considering how brittle it is to mix build flags in a project; I wish an error could be raised if the requirements are not met, so that we can make an educated decision about them.

erslavin:
It enforces a minimum, which is a good thing when you link against a target that built with a particular standard.  However, you can still override that minimum for your project by setting `CMAKE_CXX_STANDARD` - but yes, if you don't rebuild the target for that same standard you are mixing.  Why would it be a problem for OpenUSD targets to declare their minimum target version required in the interface requirements?

touraill-adobe:
> It enforces a minimum, which is a good thing when you link against a target that built with a particular standard. However, you can still override that minimum for your project by setting CMAKE_CXX_STANDARD - but yes, if you don't rebuild the target for that same standard you are mixing. Why would it be a problem for OpenUSD targets to declare their minimum target version required in the interface requirements?

Because `cxx_std_xxx` supersedes `CXX_STANDARD`, thus it becomes harder to control what is used for building. Consider the following example:

```
cmake_minimum_required(VERSION 3.30.0)

project(test)

set(CMAKE_CXX_STANDARD 11)

# Deep in OpenUSD
add_library(usd INTERFACE)
target_compile_features(usd INTERFACE cxx_std_17)

add_library(foo STATIC)
target_sources(foo PRIVATE foo.cpp)

add_library(exe)
target_sources(exe PRIVATE main.cpp)
target_link_libraries(exe PRIVATE foo usd)

# foo is built with C++11, usd and exe with C++17
```

By setting `CMAKE_CXX_STANDARD` to `11`, I would expect my whole project to build with C++11, but the usd target would silently bump the standard for itself and for my own targets. This can very easily lead to ABI issues or ODR violations (think about what happens if some headers included by both `foo` and `exe` have a content that depends on the standard used).

There is actually a [feature request in cmake](https://gitlab.kitware.com/cmake/cmake/-/issues/22592) to improve this behavior.

PR #2341, opened 2023-03-15
[cmake] Replace MATCHES with IN_LIST

touraill-adobe:
### Description of Change(s)

In several places in the cmake code, the following pattern is used to check if a list contains a value:

```
if (";${list};" MATCHES ";${var};")
```

This pattern doesn't work when `var` contains special characters such as `[` or `$`, which can happen when dealing with paths or when using generator expressions. This PR replaces this pattern with the more robust and readable `IN_LIST` operator.

### Fixes Issue(s)

I stumbled upon this issue when using my own `FindXXX.cmake` which "returned" variables with generator expressions. The `if MATCHES` were not matching and some lists were growing too much, to the point that cmake hung.


tallytalwar:
Filed as internal issue #USD-8129

PR #2342, opened 2023-03-15
Fix memory leak during framebuffer resize

schevrel:
When the size of the framebuffer is updated, Hgi textures leak when Hydra delegates use a HdRenderBuffer.

### Description of Change(s)
Destroy the previous texture if applicable before creating a new one.

### Fixes Issue(s)
Memory leak leading to crash when resizing framebuffer.

### Note
I observed two tests failing when testing **both** the public USD release branch and my fix's branch:
348 - testUsdBugs (Failed)
658 - testUsdResolverExample (Failed)


tallytalwar:
Filed as internal issue #USD-8130

PR #2345, opened 2023-03-16
Remove `boost::recursive_wrapper` usage from `JsValue`

nvmkuruc:
### Description of Change(s)
It's expected that upon adoption of C++17, USD will replace `boost::variant` with `std::variant`.  However, `std::variant` does not support `boost::recursive_wrapper` and does not have an equivalent concept.

This PR introduces `Js_Wrapper`, internally leveraging `std::unique_ptr`, but adds the invariant that it is always dereferenceable (ie. always non-NULL) and implements equality operators via the dereferenced value not the pointer value.

`Js_Wrapper` can then be used to add a level of indirection to `JsValue::_Holder` for types (`JsObject` and `JsArray`) that may contain `JsValue`s.

This PR adds test coverage for an array of objects nested in an object to validate both conversion and equality testing for more complicated structures.

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/base/js/testenv/testJsConverter.cpp
+++ b/pxr/base/js/testenv/testJsConverter.cpp
@@ -335,6 +335,20 @@ int main(int argc, char const *argv[])
             indent << "checking null conversion" << std::endl;
             TF_AXIOM(object[p.first].IsNull());
             TF_AXIOM(IsEmpty(p.second));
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-8132

PR #2348, opened 2023-03-21
Add missing include directive

touraill-adobe:
### Description of Change(s)

Adds `#include <algorithm>` for `std::find`.

### Fixes Issue(s)

This was causing a compilation error on some toolchains.


tallytalwar:
Filed as internal issue #USD-8135

PR #2349, opened 2023-03-21
Replace `boost::multi_index_container` with custom container in `stageCache.cpp`

nvmkuruc:
### Description of Change(s)
To reduce the dependency on `boost`, this converts the `StageContainer` into a class containing three synchronized unordered maps for each lookup type.  For many operations that are read only, the underlying maps can simply be exposed.  For insertion / erasure operations, explicit methods are provided on `StageContainer` to ensure the underlying containers don't violate the synchronization invariant.

### Fixes Issue(s)

### DIFF
--- a/pxr/usd/usd/stageCache.cpp
+++ b/pxr/usd/usd/stageCache.cpp
@@ -25,23 +25,17 @@
 #include "pxr/usd/usd/stageCache.h"
 
 #include "pxr/usd/sdf/layer.h"
-#include "pxr/base/tf/hash.h"
+#include "pxr/usd/usd/common.h"
 #include "pxr/usd/usd/debugCodes.h"
 #include "pxr/usd/usd/stage.h"
 
 #include "pxr/usd/ar/resolverContext.h"
-
-#include <boost/multi_index_container.hpp>
-#include <boost/multi_index/hashed_index.hpp>
-#include <boost/multi_index/identity.hpp>
-#include <boost/multi_index/member.hpp>
-#include <boost/multi_index/global_fun.hpp>
-#include <boost/range/begin.hpp>
-#include <boost/range/end.hpp>
+#include "pxr/base/tf/hash.h"
 
 #include <atomic>
 #include <vector>
 #include <thread>
+#include <unordered_map>
 #include <utility>
 
 PXR_NAMESPACE_OPEN_SCOPE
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8136

pxr/usd/usd/stageCache.cpp
```diff
@@ -65,69 +58,153 @@ std::atomic_long idCounter(9223000);
...
+
+    // If the stage is already in the cache, return the pair {Id, false}
+    // If the stage is not in the cache, generate a new id, populate
+    // the synchronized maps, and return {NewId, true}
+    std::pair<Id, bool> Insert(const UsdStageRefPtr& stage) {
+        const auto it = _byStage.find(stage);
+        if (  it != _byStage.cend()) {
+            return std::make_pair(it->second, false);
+        }
+        const Id id = GetNextId();
```
crevilla, line 101: Notes about all the uses of TF_AXIOM:
1. We definitely do not want TF_AXIOM wrapping any statements with a side effect like the emplace statements here. The docs for TF_AXIOM state that these statements are allowed to be converted to no-ops in an optimized build (even though we don't do that in any builds currently)
2. We prefer to use TF_VERIFY around an expression that is required to return true or even a normal if conditional that produces a TF_CODING_ERROR it the condition is not true.
3. Do we need axioms/verifies here at all?

pxr/usd/usd/stageCache.cpp
```diff
@@ -65,69 +58,153 @@ std::atomic_long idCounter(9223000);
...
+        TF_AXIOM(_byStage.emplace(stage, id).second);
+        TF_AXIOM(_byId.emplace(id, stage).second);
+        _byRootLayer.emplace(stage->GetRootLayer(), stage);
+        return std::make_pair(id, true);
+    }
 
-struct ByStage {};
+    // Return true if the stage was successfully erased
+    bool Erase(const UsdStageRefPtr& stage) {
+        const auto it = _byStage.find(stage);
```
crevilla, line 110: It's preferred that the `return false;` is on a separate line and wrapped in curly braces.

pxr/usd/usd/stageCache.cpp
```diff
@@ -65,69 +58,153 @@ std::atomic_long idCounter(9223000);
...
+        _byStage.erase(it);
+        return true;
+    }
 
-struct ByRootLayer {
-    static SdfLayerHandle Get(const Entry &entry) {
-        return entry.stage->GetRootLayer();
+    // Return true if the stage was successfully erased
+    bool Erase(const Id id) {
+        const auto it = _byId.find(id);
```
crevilla, line 122: Same comment as above.

pxr/usd/usd/stageCache.cpp
```diff
@@ -143,15 +220,11 @@ struct DebugHelper
 
     bool IsEnabled() const { return _enabled; }
```
crevilla, line 248: Is the variadic template necessary here?
It looks like all callers of AddEntry pass the same UsdStageRefPtr, Id parameters.

pxr/usd/usd/stageCache.cpp
```diff
@@ -553,11 +627,10 @@ UsdStageCache::Insert(const UsdStageRefPtr &stage)
     Id ret;
 
     { LockGuard lock(_mutex);
-        StagesByStage &byStage = _impl->stages.get<ByStage>();
-        auto iresult = byStage.insert(Entry(stage, GetNextId()));
+        auto iresult = _impl->stages.Insert(stage);
         if (iresult.second && debug.IsEnabled())
```
crevilla: Can we add the braces around this debug.AddEntry too

pxr/usd/usd/stageCache.cpp
```diff
@@ -284,10 +356,12 @@ UsdStageCache::RequestStage(UsdStageCacheRequest &&request)
 
     { LockGuard lock(_mutex);
         // Does the cache currently have a match?  If so, done.
-        StagesByStage &byStage = _impl->stages.get<ByStage>();
-        for (auto const &entry: byStage)
-            if (request.IsSatisfiedBy(entry.stage))
-                return std::make_pair(entry.stage, false);
+        {
+            const IdsByStage &byStage = _impl->stages.ByStage();
```
crevilla, line 385: While we're here, can we add braces around these nested statements.

pxr/usd/usd/stageCache.cpp
nvmkuruc: `emplace` returning false would be a red flag that the class's invariants have been violated. I think some sort of unlikely failure assertion would be helpful. I can switch this to a `TF_VERIFY`.

pxr/usd/usd/stageCache.cpp
nvmkuruc: It's not necessary. It may have been useful in an earlier version of the code, but `AddEntry` should now just take the Ptr / Id parameters.

nvmkuruc:
Thanks for taking a look.  I've updated the PR to address the notes.

pxr/usd/usd/stageCache.cpp
```diff
@@ -65,69 +58,159 @@ std::atomic_long idCounter(9223000);
...
+    size_t _EraseAllIf(const SdfLayerHandle& rootLayer,
+                       ConditionFn&& conditionFn,
+                       vector<Entry>* erased) {
+        const auto range = _byRootLayer.equal_range(rootLayer);
+        size_t erasedCount = 0;
+        // In C++20, consider using a filtered range with the condition
+        // function
+        for (auto it = range.first; it != range.second;) {
+            if (conditionFn(it->second)) {
+                const auto byStageIt = _byStage.find(it->second);
```
crevilla, line 200: Sorry for nitpicking, but the only issue here is that it this verify does happen to fail, this will loop infinitely since the iterator doesn't get updated. Maybe starting at 200, the code should be
```
                    it = _byRootLayer.erase(it);
                    ++erasedCount;
                    continue;
                }
            } 
            ++it;
```

nvmkuruc: I restructured the loop to avoid the need for a `continue` in the spirit of avoiding in non-trivial loops (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-continue).

PR #2350, opened 2023-03-22
Fix ranged loop type in `SdfLayer`

nvmkuruc:
### Description of Change(s)
When building with GCC11, the following warning was observed.
```
 warning: loop variable ‘tokenPath’ of type ‘const std::pair<pxrInternal_v0_23__pxrReserved__::TfToken, pxrInternal_v0_23__pxrReserved__::SdfPath>&’ binds to a temporary constructed from type ‘std::pair<const pxrInternal_v0_23__pxrReserved__::TfToken, pxrInternal_v0_23__pxrReserved__::SdfPath>’ [-Wrange-loop-construct]
```
The value type of a `std::map<TfToken, SdfPath>` is `std::pair<const TfToken, SdfPath>` not `std::pair<TfToken, SdfPath>`. This change uses `auto` in the range loop to avoid the unnecessary temporary copy.

### Fixes Issue(s)
N/A


tallytalwar:
Filed as internal issue #USD-8137

PR #2352, opened 2023-03-22
APPLE: Add an option to turn off dome light background in usdrecord

creijon:
### Description of Change(s)
Add an option to turn off dome light background in usdrecord

### Fixes Issue(s)
- For creating cards, it is helpful to remove the dome light background.


tallytalwar:
Filed as internal issue #USD-8139

PR #2353, opened 2023-03-22
Replace `boost::multi_index_container` with custom container in `Sdf_LayerRegistry`

nvmkuruc:
### Description of Change(s)
To reduce the number of dependencies on boost, this PR introduces a custom container that synchronizes a set of unordered maps.  Internally, it leverages layer `Sdf_AssetInfo` as the generator of keys for `Insert`, `Update`, and `Erase` operations.

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/usd/sdf/layerRegistry.h
+++ b/pxr/usd/sdf/layerRegistry.h
@@ -27,14 +27,12 @@
 /// \file sdf/layerRegistry.h
 
 #include "pxr/pxr.h"
+#include "pxr/usd/sdf/assetPathResolver.h"
 #include "pxr/usd/sdf/declareHandles.h"
 #include "pxr/base/tf/hash.h"
 
-#include <boost/multi_index_container.hpp>
-#include <boost/multi_index/hashed_index.hpp>
-#include <boost/multi_index/identity.hpp>
-
 #include <string>
+#include <unordered_map>
 #include <iosfwd>
 
 PXR_NAMESPACE_OPEN_SCOPE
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-8140

pxr/usd/sdf/layerRegistry.h
```diff
@@ -91,74 +88,58 @@ class Sdf_LayerRegistry
...
-        typedef std::string result_type;
-        result_type operator()(const SdfLayerHandle& layer) const;
+    // A wrapper around a set of unordered_maps that maps layers
+    // bidirectionally to their various string representations (realPath,
+    // identifier, and repositoryPath)
+    class _Layers final {
+    public:
+        _Layers() = default;
+        using LayersByRealPath =
+            std::unordered_map<std::string, SdfLayerHandle, TfHash>;
```
crevilla, line 103: I don't believe that LayersByIdentifier and LayersByRepositoryPath are going to be equivalent to the original behavior if implemented as std::unordered_map as the multi index container had those two keys set as non-unique.
I'm trying to find out from internal conversations as to if/why this is necessary.

nvmkuruc: This is a good catch!  I think "one => many" can be similarly modeled with a `unordered_multimap`, but it'd would be good to understand the practical situations where the `unordered_map` implementation fails so we can get test coverage for it.

crevilla: I've learned at least one of the reasons is that you can have different identifiers resolve to the same real layer path under different ArResolverContexts. And the good news is we do have a test case for this in testSdfLayer so we should be able to catch if there's an issue pretty easily.

nvmkuruc: `testSdfLayer` passes. It seems like `SdfLayer` tries to avoid look ups using identifiers when they are context dependent (deferring to the real path instead) which probably explains why the code is working.  It seems like the case that might fail is that if you mixed context dependent paths and non-context dependent paths and the context-dependent path boots the non-context dependent path lookup from the registry.  I'll see if I can fashion that test case and move things over to use `unordered_multimap`.

nvmkuruc: `testSdfLayer` was passing because the test which exercises multiple contexts was being skipped due to 

```
    @unittest.skipIf(preferredResolver != "ArDefaultResolver",
                     "Test uses search-path functionality specific to "
                     "ArDefaultResolver")
```

By default, `testSdfLayer` is configured to run only against a test resolver.

```
preferredResolver = os.environ.get(
    "TEST_SDF_LAYER_RESOLVER", "Sdf_TestResolver")
```

PR #2355, opened 2023-03-23
Apple: Update dynamic state from render pass for pipelinedrawbatch

slingthor:
### Description of Change(s)
- Lets the render pass state set the dynamic state on classes that use the pipelinedrawbatch and HGI resource generation
### Fixes Issue(s)
- Dynamic state not being updated from the renderpassstate for later frames / timecodes during a scene's playback


tallytalwar:
Filed as internal issue #USD-8141

slingthor:
We've moved that forward as a part of another fix

PR #2356, opened 2023-03-23
[usdz, ar] Support ArAsset / ArWritableAsset for USDZ

nvidia-jomiller:
### Description of Change(s)

[usdz] USDZ zip files would always force the use of file system paths.
For asset management systems that don't directly rely on the file system this could be inconvenient. It would require downloading content to temp files, writing to a temp file, then uploading the temp zip file. With Ar 2.0, we can now use OpenAsset / OpenAssetForWrite so the underlying ArResolver can manage this directly. This change adds support so we can write directly to an ArWritableAsset to abstract the file system.

[ar] Required a change to ArFilesystemWritableAsset to discard the TfSafeOutput file if Close() was not explicitly called.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/2374

### DIFF
--- a/pxr/usd/ar/filesystemWritableAsset.cpp
+++ b/pxr/usd/ar/filesystemWritableAsset.cpp
@@ -76,7 +76,17 @@ ArFilesystemWritableAsset::ArFilesystemWritableAsset(TfSafeOutputFile&& file)
     }
 }
 
-ArFilesystemWritableAsset::~ArFilesystemWritableAsset() = default;
+ArFilesystemWritableAsset::~ArFilesystemWritableAsset()
+{
+    // XXX: If Close() has not been explicitly called for replacing an asset
+    // we will discard it, so we don't overwrite that asset with invalid data.
+    // When an asset has been opened for update we will allow the destructor
+    // of TfSafeOutput to close it properly
+    if (!_file.IsOpenForUpdate())
+    {
+        _file.Discard();
+    }
+}
 
 bool
 ArFilesystemWritableAsset::Close()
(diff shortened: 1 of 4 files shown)


tallytalwar:
Filed as internal issue #USD-8142

pxr/usd/ar/filesystemWritableAsset.cpp
```diff
@@ -76,7 +76,17 @@ ArFilesystemWritableAsset::ArFilesystemWritableAsset(TfSafeOutputFile&& file)
...
+ArFilesystemWritableAsset::~ArFilesystemWritableAsset()
+{
+    // XXX: If Close() has not been explicitly called for replacing an asset
+    // we will discard it, so we don't overwrite that asset with invalid data.
+    // When an asset has been opened for update we will allow the destructor
+    // of TfSafeOutput to close it properly
+    if (!_file.IsOpenForUpdate())
+    {
+        _file.Discard();
+    }
```
nvidia-jomiller, line 88: I found the behavior of `ArFilesystemWritableAsset` a little odd when using WriteMode::Replace.

From the [WriteMode::Replace](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/ar/resolver.h#L306-L309) documentation it says that if the file exists it's original content will be discarded when the `ArWritableAsset` is destroyed. But the documentation for [ArWritableAsset](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/ar/writableAsset.h#L52-L55) says that finalization of writing the asset will be done when `Close()` is called. This is fine and makes sense, but the behavior of `ArFilesystemWritableAsset` kind of ignores that and will finalize / commit the changes to the underlying `TfSafeOutput` file regardless if `Close()` is called or the `ArFilesystemWritableAsset` instance is destroyed.

The change above will discard any writes to the `TfSafeOutput` file if it's been opened for replacement and `Close()` has not been called

PR #2357, opened 2023-03-23
Apple: iOS - garch changes

slingthor:
### Description of Change(s)
- Adds the ability to build with iOS in garch
- Modifies the platform contexts to support iOS


tallytalwar:
Filed as internal issue #USD-8143

slingthor:
This builds locally - can we nudge it?

PR #2360, opened 2023-03-24
Apple: iOS - arch changes 

slingthor:
### Description of Change(s)
- Correctly finds system includes on iOS in the arch module


tallytalwar:
Filed as internal issue #USD-8146

slingthor:
This builds locally, can we nudge it?

PR #2363, opened 2023-03-24
Apple: iOS - hdst, hdx, hgigl changes 

slingthor:
### Description of Change(s)
- Modifies the cmake files to allow for building on iOS


sunyab:
Filed as internal issue #USD-8160

PR #2364, opened 2023-03-24
Apple: iOS - Add Cmake Toolchain

slingthor:
### Description of Change(s)
- Adds a CMake toolchain to build on iOS
- Has several build options and is based on a tried and tested toolchain whose copyright reference is included
- That toolchain has been simplified


sunyab:
Filed as internal issue #USD-8161

PR #2366, opened 2023-03-26
Apple: iOS - hgimetal changes

slingthor:
### Description of Change(s)
- Removes unneeded library references
- Does not perform didmodifyrange as not needed and not supported on iOS
- Disables certain texture formats on iOS


slingthor:
This won't build until
https://github.com/PixarAnimationStudios/USD/pull/2365
has been pulled in

sunyab:
Filed as internal issue #USD-8163

PR #2368, opened 2023-03-29
Load plugins using LoadLibraryEx with LOAD_WITH_ALTERED_SEARCH_PATH on Windows

SimonBoorer:
### Description of Change(s)

Search for USD plugin dependencies in the folder that the plugin is loaded from as well as the application's directory on Windows. This is to support situations where USD itself is a plugin for an application. See https://github.com/PixarAnimationStudios/USD/issues/2367 for more information.

I found when testing LoadLibraryEx that the LOAD_WITH_ALTERED_SEARCH_PATH flag only works when using back slashes instead of forward slashes, so I've added a ArchWindowsPreferredPath method to convert between them based off of [std::filesystem::path::make_preferred](https://en.cppreference.com/w/cpp/filesystem/path/make_preferred) in C++17.

This limitation is mentioned [here](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa):
> If the string specifies a fully qualified path, the function searches only that path for the module. When specifying a path, be sure to use backslashes (\\), not forward slashes (/).

I originally added ArchPreferredPath that simply returned the input path for MacOS/Linux but decided against the extra copy.

I also made converting to back slashes the responsibility of the caller (PlugPlugin::_Load) rather than ArchLibraryOpen in keeping with LoadLibraryEx.

Happy to revisit either of these decisions.

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/USD/issues/2367

### DIFF
--- a/pxr/base/arch/library.cpp
+++ b/pxr/base/arch/library.cpp
@@ -44,7 +44,7 @@ void* ArchLibraryOpen(const std::string &filename, int flag)
 {
 #if defined(ARCH_OS_WINDOWS)
     arch_lastLibraryError = 0;
-    if (void* result = LoadLibrary(filename.c_str())) {
+    if (void* result = LoadLibraryEx(filename.c_str(), NULL, flag)) {
         return result;
     }
     else {
--- a/pxr/base/plug/plugin.cpp
+++ b/pxr/base/plug/plugin.cpp
@@ -253,8 +253,12 @@ PlugPlugin::_Load()
         else {
             string dsoError;
             {
-                TRACE_FUNCTION_SCOPE("dlopen");
+                TRACE_FUNCTION_SCOPE("dlopen");   
+#if defined(ARCH_OS_WINDOWS)
+                _handle = TfDlopen(ArchWindowsPreferredPath(_path).c_str(), ARCH_LIBRARY_ALTERED_SEARCH_PATH, &dsoError);
+#else
                 _handle = TfDlopen(_path.c_str(), ARCH_LIBRARY_NOW, &dsoError);
+#endif
             }
             if (!_handle ) {
                 TF_CODING_ERROR("Failed to load plugin '%s': %s in '%s'",  
--- a/pxr/base/arch/fileSystem.h
+++ b/pxr/base/arch/fileSystem.h
@@ -32,6 +32,7 @@
 #include "pxr/base/arch/api.h"
 #include "pxr/base/arch/defines.h"
 #include "pxr/base/arch/inttypes.h"
+#include <algorithm>
 #include <memory>
 #include <cstdio>
 #include <string>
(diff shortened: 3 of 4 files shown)


sunyab:
Filed as internal issue #USD-8175

pxr/base/plug/plugin.cpp
```diff
@@ -253,8 +253,12 @@ PlugPlugin::_Load()
         else {
             string dsoError;
             {
-                TRACE_FUNCTION_SCOPE("dlopen");
+                TRACE_FUNCTION_SCOPE("dlopen");   
+#if defined(ARCH_OS_WINDOWS)
+                _handle = TfDlopen(ArchWindowsPreferredPath(_path).c_str(), ARCH_LIBRARY_ALTERED_SEARCH_PATH, &dsoError);
+#else
                 _handle = TfDlopen(_path.c_str(), ARCH_LIBRARY_NOW, &dsoError);
+#endif
```
sunyab, line 261: Instead of having an #ifdef here I think it'd be simpler to have the one function call and just combine `ARCH_LIBRARY_NOW | ARCH_LIBRARY_ALTERED_SEARCH_PATH` since these are already set to platform-specific values in `arch/library.h`.

pxr/base/arch/library.cpp
```diff
@@ -44,7 +44,7 @@ void* ArchLibraryOpen(const std::string &filename, int flag)
 {
 #if defined(ARCH_OS_WINDOWS)
     arch_lastLibraryError = 0;
-    if (void* result = LoadLibrary(filename.c_str())) {
+    if (void* result = LoadLibraryEx(filename.c_str(), NULL, flag)) {
```
sunyab, line 47: Interestingly enough the docs for the old `LoadLibrary` mention the same backslash requirement -- I guess it just didn't make an actual difference until this change. 

In any case, if we follow the letter of the docs then the filename we pass to LoadLibraryEx must always use backslashes, so I'd advocate for doing the conversion to preferred path here. That way, the logic is centralized and downstream code wouldn't have to be platform-aware. Does that sound OK?

pxr/base/plug/plugin.cpp
sunyab: Could you please add a comment about how `ARCH_LIBRARY_ALTERED_SEARCH_PATH` on Windows ensures that .dll dependencies in the same directory as the specified library will be picked up without requiring modifications to the environment?

sunyab: Also, I just noticed `TfDlopen` takes a `std::string` and not a `const char*`, so if you could drop the `.c_str()` while you're here that'd be great

sunyab (changes requested):
Hi @SimonBoorer, I left some notes but it's just some reorganization of what you've already got. Let me know what you think, thanks!

nvmkuruc:
Now that OpenUSD uses C++17, I wonder if this could use `std::filesystem`'s `make_preferred`.

PR #2369, opened 2023-03-30
Fix thread safety issues in UsdSkel_SkelDefinition.

cameronwhite:
### Description of Change(s)

- In methods like `_ComputeJointWorldInverseBindTransforms()`, check the compute flag again after acquiring the lock to avoid potentially recomputing the result again if multiple threads were waiting on the mutex. Although the computed values would not change, it is not safe to call mutable member functions of the VtArray (which can cause a copy-on-write detach) while other threads may be in the middle of making a copy of it. An example sequence that could lead to a crash:
  - Thread A finishes the critical section, and continues on to make a copy of the array in `_GetJointWorldInverseBindTransforms()` (refcount == 2)
  - Thread B was waiting on the mutex and starts to redo the computation. This calls non-const member functions and enters the body of `_DetachIfNotUnique()` since `_IsUnique()` is false
  - Thread C enters `_GetJointWorldInverseBindTransforms()` and observes that the compute flag is set, and starts to make its own copy of the array, but hasn't [bumped the refcount yet](https://github.com/PixarAnimationStudios/USD/blob/5c5ebddff35012461a2b0ba773c47f05730cbab4/pxr/base/vt/array.h#L300).
  - Thread A finishes with its copy and decrements the refcount (refcount == 1)
  - Thread B [decrements the refcount](https://github.com/PixarAnimationStudios/USD/blob/5c5ebddff35012461a2b0ba773c47f05730cbab4/pxr/base/vt/array.h#L832), taking it to zero and destroying the data, before switching to its new copy of the data
  - Thread C now has an array with a reference to the deleted data block

- Prefer using `operator|=` to atomically set the flag rather than doing a read -> bitwise OR -> atomic store sequence which could cause flags to be lost if there are concurrent writes. Currently the writes are all guarded by the same mutex so the previous approach was not actually problematic, but this is safer if e.g. in the future there are separate locks for each cached array.

This isn't the easiest to reproduce (e.g. Storm will not encounter this, since it computes the skinning transforms during the single-threaded sprim sync), but I had a customer file that reproduced this very reliably in the Houdini GL delegate with a 32-core machine, and verified that this fix resolved the issue.

### Fixes Issue(s)
- #1742

### DIFF
--- a/pxr/usd/usdSkel/skelDefinition.cpp
+++ b/pxr/usd/usdSkel/skelDefinition.cpp
@@ -244,19 +244,21 @@ UsdSkel_SkelDefinition::_ComputeJointSkelRestTransforms()
     if (TF_VERIFY(GetJointLocalRestTransforms(&jointLocalRestXforms))) {
 
         std::lock_guard<std::mutex> lock(_mutex);
-        
-        VtArray<Matrix4>& skelXforms = _jointSkelRestXforms.Get<Matrix4>();
-        skelXforms.resize(_topology.size());
 
-        const bool success =
-            UsdSkelConcatJointTransforms(_topology, jointLocalRestXforms,
-                                         skelXforms);
+        if (!(_flags & ComputeFlag)) {
+            VtArray<Matrix4>& skelXforms = _jointSkelRestXforms.Get<Matrix4>();
+            skelXforms.resize(_topology.size());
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-8176

PR #2372, opened 2023-03-31
Add a helper function to add mtlx support to a scene delegate

hshakula:
### Description of Change(s)
This helper function provides an easy way of adding support of MaterialX to a custom HdSceneDelegate:
```
class CustomSceneDelegate : public HdSceneDelegate {
    // ...
    VtValue GetMaterialResource(SdfPath const& id) override {
        std::string mtlxPath = GetMaterialXForPrim(id);

        HdRenderDelegate* renderDelegate = GetRenderIndex().GetRenderDelegate();

        HdMaterialNetworkMap materialNetworkMap;
        UsdImagingMtlxConvertMtlxToHdMaterialNetworkMap(
            mtlxPath,
            renderDelegate->GetShaderSourceTypes(),
            renderDelegate->GetMaterialRenderContexts(),
            &materialNetworkMap);
        return VtValue(materialNetworkMap);
    }
};

```

### Fixes Issue(s)
- No easy way to pass MaterialX from a custom HdSceneDelegate in the standard way (MaterialX translated to HdMaterialNetwork as a UsdShade graph).

### DIFF
--- a/pxr/usdImaging/usdImagingMtlx/adapter.cpp
+++ b/pxr/usdImaging/usdImagingMtlx/adapter.cpp
@@ -1,0 +1,82 @@
+#include "pxr/usdImaging/usdImagingMtlx/adapter.h"
+
+#include "pxr/usdImaging/usdImaging/materialParamUtils.h"
+
+#include "pxr/imaging/hd/material.h"
+#include "pxr/imaging/hd/tokens.h"
+
+#include "pxr/usd/usdMtlx/reader.h"
+#include "pxr/usd/usdMtlx/utils.h"
+
+#include "pxr/usd/ar/resolver.h"
+#include "pxr/usd/ar/resolverContextBinder.h"
+#include "pxr/usd/ar/resolverScopedCache.h"
+
+#include "pxr/usd/usdShade/material.h"
+#include "pxr/usd/usdShade/shader.h"
+
+#include "pxr/base/arch/fileSystem.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+void UsdImagingMtlxConvertMtlxToHdMaterialNetworkMap(
+    const std::string& mtlxPath,
+    const TfTokenVector& shaderSourceTypes,
+    const TfTokenVector& renderContexts,
+    HdMaterialNetworkMap* out)
+{
+    if (mtlxPath.empty()) {
+        return;
+    }
+
+    std::string basePath = TfGetPathName(mtlxPath);
+    if (basePath.empty()) {
+        basePath = ".";
+    }
+
+    ArResolver& resolver = ArGetResolver();
+    const ArResolverContext context = resolver.CreateDefaultContextForAsset(mtlxPath);
+    ArResolverContextBinder binder(context);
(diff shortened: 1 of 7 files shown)


sunyab:
Filed as internal issue #USD-8179

spiffmon:
We'll send this to the Hydra team for their take, @hshakula !  I see the attraction of this approach, as it introduces very little code.  Though by leveraging the same indirection through a UsdStage, it incurs all the same cost (actually, it pushes some cost from "asset load" time to "render-time" that may will get paid repeatedly).  But it also means it needs to live in usdImaging, and so any unrelated-to-USD scene adapter then needs to pull in a usdImaging dependency.

If instead we went directly from .mtlx document to HdMaterialNetwork2 (which admittedly means alot more logic to implement):
* the code could just be added to hdMtlx in pxr/imaging
* it would be considerably faster

I don't know if that last point is of general interest to the "MaterialX and Hydra" community, but maybe worth discussing.

hshakula:
You are absolutely correct. This approach is feasible only as a quick stub implementation. For production, direct conversion .mtlx-> HdMaterialNetwork2 must be implemented. We in our scene delegate did not come to this yet, and I don’t think we will in near future. This PR was opened in the hope to help others to get their scene delegates to MVP state faster. Obviously, implementing the direct conversion requires a lot of effort and not everyone is ready to take it upon themselves.

pxr/usdImaging/usdImagingMtlx/adapter.cpp
```diff
@@ -0,0 +1,82 @@
...
+                           "MaterialX cycle found: %s\n",
+                           x.what());
+        return;
+    } catch (MaterialX::Exception &x) {
+        Tf_PostErrorHelper(TF_CALL_CONTEXT,
+                           TF_DIAGNOSTIC_RUNTIME_ERROR_TYPE,
+                           "MaterialX error: %s\n",
+                           x.what());
+        return;
+    }
```
spiffmon, line 62: It was pointed out to me that all of this should be replaceable with simply:
`UsdStageRefPtr stage = UsdStage::Open(mtlxPath);`

... even if someone else has opened the same file and is keeping it open, because UsdMtlxFileFOrmat is read-only, they won't have been able to mutate the result, so you should be guaranteed to get the same results, and it also gives you the potential for "caching" this translation step at a higher level for reuse.

ashwinbhat:
Hi @hshakula @spiffmon 
I'm trying to track down this PR in dev. It says it's merged, but can't seem to locate usdImagingMtlx in the repo. Did it get removed?

spiffmon:
Apologies, @ashwinbhat , there must have been gremlins in our automation... the reference commit here addresses PR #2373 , **not** this PR, 2372

Our internal trackers all seem consistent, but something happened on the way to gitHub.  I'm not sure if we can clean up the "merged" state here, but we'll see if we can.  This PR has not yet been committed.

ashwinbhat:
Thanks @spiffmon for checking into this. I wanted to evaluate this change and will grab it from the PR source repo.

PR #2373, opened 2023-03-31
Disable shadows on the "simple" headlight

marktucker:
When using a "simple" headlight, turn off shadows generated by this light.

Because it is a distant light, and not an actual "headlight" co-located with the camera, interior scenes will be entirely in shadow. Submitting this on behalf of @crydalch spawned from this discussion on usd-interest: https://groups.google.com/g/usd-interest/c/ShnbqGtpi_0

### Description of Change(s)

Changed the code in the TaskController that translates a GlfSimpleLight into explicit values stored in the task controller's custom scene delegate so that simple dome lights have shadows enabled, but simple distant light have shadows disabled.

### DIFF
--- a/pxr/imaging/hdx/taskController.cpp
+++ b/pxr/imaging/hdx/taskController.cpp
@@ -857,6 +857,8 @@ HdxTaskController::_SetParameters(SdfPath const& pathName,
     if (light.IsDomeLight()) {
         _delegate.SetParameter(pathName, HdLightTokens->textureFile,
                                _GetDomeLightTexture(light));
+        _delegate.SetParameter(pathName, HdLightTokens->shadowEnable, 
+            VtValue(true));
     }
     // When not using storm, initialize the camera light transform based on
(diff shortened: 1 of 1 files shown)


marktucker:
We gave some consideration to extending GlfSimpleLight to have a flag indicating whether shadows should be turned on or off (for either dome or headlight mode), but opted against it since we felt this simpler change would satisfy most use cases without adding user/UI complexity.

sunyab:
Filed as internal issue #USD-8198

marktucker:
Should this PR now be closed? There is some confusion in the commit referenced above (https://github.com/PixarAnimationStudios/OpenUSD/commit/28cefab6164838ae2ac6c017477fc9030fd0ba48), where the commit contents are from this PR, but the comment is from a different PR. But if the change here has been merged (and it looks like it has), this should be closed?

spiffmon:
Yes, thanks for adding the commit, for posterity, @marktucker

PR #2383, opened 2023-04-07
APPLE: Protect IBL convolution from NaNs in textures

creijon:
### Description of Change(s)

On Apple platforms with the Metal port of Storm, any NaNs in the input dome light textures will propagate through the IBL convolution calculation and cause completely black lighting. on these channels.

This doesn't seem to be a problem with openGL, but it's a small change that shouldn't have a performance hit.

### Fixes Issue(s)
- Incorrect lighting when NaNs are present in input texture to dome lights.


sunyab:
Filed as internal issue #USD-8218

PR #2388, opened 2023-04-18
APPLE: Fix for breaking API change in PySide 6.5.0

creijon:
### Description of Change(s)

The most recent version of PySide 6.5.0 appears to have introduced a breaking change in the initialiser of `QObject`.  Check for the Qt version number and pass in a null `parent` and `name` to handle it.

Small change: split the qVersion string only once and store it as a function attribute, to avoid perf hit.

### Fixes Issue(s)
- usdview crash on startup with PySide 6.5.0


creijon:
The issue actually appears to be related to multiple inheritance.  I'm closing this PR until I have a clean fix.

PR #2389, opened 2023-04-18
Replace `boost::iterator_facade` with explicit implementation for `pcp/node.h`

nvmkuruc:
### Description of Change(s)
- Replace `iterator_facade` with explicit operators for `PcpNodeRef_ChildrenIterator` and `PcpNodeRef_ChildrenReverseIterator`

### Fixes Issue(s)
- #2305 

### DIFF
--- a/pxr/usd/pcp/node.h
+++ b/pxr/usd/pcp/node.h
@@ -31,9 +31,6 @@
 #include "pxr/base/tf/iterator.h"
 #include "pxr/base/tf/hashset.h"
 
-#include <boost/iterator/iterator_facade.hpp>
-#include <boost/iterator/reverse_iterator.hpp>
-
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-8252

pxr/usd/pcp/node.h
```diff
@@ -372,8 +369,28 @@ class PcpNodeRef_ChildrenIterator
     PCP_API
     PcpNodeRef_ChildrenIterator(const PcpNodeRef& node, bool end = false);
```
crevilla, line 381: I'm pretty confident that you will need operator-> for both of these iterator classes. They're exposed publicly as child_const_iterator and child_const_reverse_iterator (via typedef) and I can confirm that at least the maya open source code uses these iterators and dereferences via ->

PR #2390, opened 2023-04-19
Replace `boost::iterator_facade` with explicit implementation for `SdfChildrenView`'s internal iterator

nvmkuruc:
### Description of Change(s)
* Replace `iterator_facade` with explicit operators for `SdfChildrenView::_InnerIterator`
* Use `nullptr` instead of `NULL` to default initialize pointer types (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nullptr)
* Use in-class initialization for defaults (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer)
* Use `using` instead of `typedef` for defining iterator traits (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using)

### Fixes Issue(s)
- #2305 

### DIFF
--- a/pxr/usd/sdf/childrenView.h
+++ b/pxr/usd/sdf/childrenView.h
@@ -32,7 +32,6 @@
 #include "pxr/base/tf/iterator.h"
 
 #include <boost/iterator/filter_iterator.hpp>
-#include <boost/iterator/iterator_facade.hpp>
 #include <boost/iterator/reverse_iterator.hpp>
 #include <algorithm>
 #include <vector>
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-8253

pxr/usd/sdf/childrenView.h
```diff
@@ -188,23 +187,102 @@ class SdfChildrenView {
...
+        using value_type = SdfChildrenView::value_type;
+        using reference = value_type;
+        using pointer = void;
+        using difference_type = std::ptrdiff_t;
 
-        _InnerIterator() :
-            _owner(NULL), _pos(0) { }
+        _InnerIterator() = default;
         _InnerIterator(const This* owner, const size_t& pos) :
             _owner(owner), _pos(pos) { }
```
crevilla, line 209: I'm pretty sure you're going to need operator-> here too. This iterator is the type for SdfChildrenView::const_iterator (after all the template expansion) so it's highly likely that we have Presto code that dereferences via ->

PR #2391, opened 2023-04-19
APPLE: Disable presentation in UsdAppUtilsFrameRecorder

creijon:
### Description of Change(s)

The `UsdImagingGLEngine` in `UsdAppUtilsFrameRecorder` has `HdxPresentTask` presentation enabled by default.  When this is used in Metal this means that the openGL context is still required.  Since the frame recorder doesn't present to the viewport we should disable this.

Addendum: the presentation flag is also reset to enabled if the renderer plugin is changed after the frame recorder is created, so it is now set to disabled then too.

### Fixes Issue(s)
- Requirement for openGL context when using the UsdAppUtilsFrameRecorder on Metal.


sunyab:
Filed as internal issue #USD-8254

Issue #1331, opened 2020-09-17
non-ascii utf-8 string encoding error


takahito-tejima:
### Description of Issue

When I store non-ascii utf-8 string (such as '日本語' or 'ピクサー') as a string attribute in usd, they always get corrupted when I export into usda.

https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/sdf/fileIO_Common.cpp#L679
    static const char* hexdigit = "0123456789abcedf";
This table looks somehow shuffled ('e' and 'd'). Is this intentional? (I hope not...)

### Steps to Reproduce
prim = stage.DefinePrim('/prim')
prim.CreateAttribute('str', Sdf.ValueTypeNames.String).Set('ピクサー')
stage.ExportToString()
  (snip)
> custom string str = "\\xd3\\x83\\x94\\xd3\\x82\\xaf\\xd3\\x82\\xb5\\xd3\\x83\\xbc"
this must be
> custom string str = "\\xe3\\x83\\x94\\xe3\\x82\\xaf\\xe3\\x82\\xb5\\xe3\\x83\\xbc"
or ideally
> custom string str = 'ピクサー'

### Package Versions
20.08

takahito-tejima:
correction: exporting usdc is fine (sorry!). just usda.

takahito-tejima:
another example. this may need a different fix.

t1.usda
```
#usda 1.0
def "prim"
{
    string s = "É00"
}
```

usdcat t1.usda > t2.usda
```
#usda 1.0
def "prim"
{
    string s = "\xc3\x8900"     # this makes sense.  É = 0xc389
}
```
usdcat t2.usda
```
#usda 1.0
def "prim"
{
    string s = "\xc3\x00"       # I guess char(\x8900) becomes \00 ?
}
```

jtran56:
Filed as internal issue #USD-6353

c64kernal:
Hey @takahito-tejima -- thanks so much! Great catch, we'll try to get this fixed up as soon as we can!

takahito-tejima:
thank you for the fix!
but I don't think that change addresses the second case ("É00" becomes "\xc3\x00" instead of "\xc3\x89\x00").
are there any other fix for that one in 20.11?

gitamohr:
Oh!  I missed that second case Takahito.  Sorry about that. :-/  I will take a look and see what's going on there.

PR #2395, opened 2023-04-20
Replace documentation references to "ASCII" with "text"

nvmkuruc:
### Description of Change(s)
Utilities using `usda` historically reference it as the "ASCII" file format. However, for some time, strings can be UTF-8 encoded (#2055, #1331). Prim and property identifiers will be getting similar support soon as well (#2120).

For clarity, this PR changes references to ASCII at several points in documentation to the more general and accurate descriptor of "text".

This PR does not update test case descriptions or file names.

### Fixes Issue(s)
- #2120 

### DIFF
--- a/pxr/usd/sdf/overview.dox
+++ b/pxr/usd/sdf/overview.dox
@@ -4,13 +4,13 @@
 \mainpage Sdf : Scene Description Foundations
 \endif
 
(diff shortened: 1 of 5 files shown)


sunyab:
Filed as internal issue #USD-8257

Issue #2396, opened 2023-04-21
Missing usage requirements for `usd_ms` in `pxrTargets.cmake`


nicolaspopravka:
### Description of Issue
When building USD as a monolithic shared library, the generated and installed CMake file to import USD into another CMake project is unfortunately not quite usable as is. For example, the `INTERFACE_INCLUDE_DIRECTORIES` property is not set on the  `usd_ms` target.

### Steps to Reproduce
1. Build USD as a monolithic shared library
2. In another CMake project, find the `pxr` package, and try a build against it. Here's an example of a minimal `CMakeLists.txt` for this step:
```
find_package(pxr CONFIG REQUIRED)

file(WRITE main.cc 
"#include <iostream>
#include \"pxr/pxr.h\"
int main() { 
  std::cout << \"Hello, USD v\" << PXR_MAJOR_VERSION << \".\" << PXR_MINOR_VERSION << \".\" << PXR_PATCH_VERSION << std::endl;
  return 0;
}"
)

add_executable(hello_usd main.cc)
if (TARGET usd_ms)
  target_include_directories(hello_usd PRIVATE $<TARGET_PROPERTY:usd_ms,INTERFACE_INCLUDE_DIRECTORIES>)
else()
  target_include_directories(hello_usd PRIVATE $<TARGET_PROPERTY:usd,INTERFACE_INCLUDE_DIRECTORIES>)
endif()
```
3. Compiler reports `fatal  error C1083: Cannot open include file: 'pxr/pxr.h': No such file or directory`

Same build succeeds if USD is **not** built as a monolithic shared library.

### System Information (OS, Hardware)
Windows 10

### Package Versions
USD v23.05

### Build Flags
`--build-monolithic`

nicolaspopravka:
I have a candidate patch for this issue:
```
diff --git a/cmake/macros/Private.cmake b/cmake/macros/Private.cmake
index 23172745e..c44b3f776 100644
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -1424,7 +1424,10 @@ function(_pxr_library NAME)
     endif()

     if(isObject)
-        # Nothing
+        install(
+            TARGETS ${NAME}
+            EXPORT pxrTargets
+        )
     else()
         # Do not include plugins libs in externally linkable targets
         if(isPlugin)
diff --git a/cmake/macros/Public.cmake b/cmake/macros/Public.cmake
[...]

PR #2400, opened 2023-04-21
Fix export of USD as a monolithic shared library.

nicolaspopravka:
### Description of Change(s)
Outside projects could not easily be built against USD as a monolithic shared library because of some usage requirements not set on the exported `usd_ms` target. For example, the `INTERFACE_INCLUDE_DIRECTORIES` target property was missing.

This is fixed by reverting the use of the `$<BUILD_INTERFACE:...>` export generator expression.

However, outside builds would still fail because the `usd_m` and other internal-to-monolithic targets were not exported. This is fixed too.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/2396

### DIFF
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -1424,7 +1424,10 @@ function(_pxr_library NAME)
     endif()
 
     if(isObject)
-        # Nothing
+        install(
+            TARGETS ${NAME}
+            EXPORT pxrTargets
+        )
     else()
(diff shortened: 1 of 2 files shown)


sunyab:
Filed as internal issue #USD-8264

sunyab:
#2396 should be fixed with commit f5da4e2 in the `dev` branch, so this PR is no longer needed. Closing this out, thanks!

PR #2401, opened 2023-04-22
APPLE: Buffer binding null pointer checks

creijon:
### Description of Change(s)

If an empty USD file was loaded, a null buffer would be passed into the step function and indirect command buffer bindings, causing a segfault.  This protects against it.

### Fixes Issue(s)
- Segfault on binding null buffers


sunyab:
Filed as internal issue #USD-8265

PR #2408, opened 2023-04-26
Allow instancing of bounding boxes drawn on prims with unloaded payloads

marktucker:
Allow instancing of bounding boxes drawn on prims with unloaded payloads.

### Description of Change(s)
Reordered code in UsdImagingDelegate::_AdapterLookup to check for IsInstanceable _before_ checking for IsLoaded. This allows the instancer adapter to instance the bounding box rprims created for instanceable prims with unloaded payloads. Previously even with the instanceable flag set, every unloaded payload prim would create a separate rprim for its bounding box.

This change also required changing the draw mode adapter because that code was explicitly compensating for the fact that unloaded payload prims were not being instanced properly.


tallytalwar:
Filed as internal issue #USD-8273

PR #2409, opened 2023-04-26
Make sure add_dll_directory is always called with an absolute path.

marktucker:
Make sure add_dll_directory is always called with an absolute path.
Calling it with a relative path on Windows raises an exception.
This would happen if the PATH variable contains a relative path.

### Description of Change(s)
In the Tf __init__.py module, get the absolute path of each component of the PATH variable before calling add_dll_directory.


tallytalwar:
Filed as internal issue #USD-8274

PR #2414, opened 2023-04-28
Fix spelling: indicies -> indices

Vertexwahn:
### Fixes Issue(s)
-

- [ ] I have submitted a signed Contributor License Agreement

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,20 @@
+{
+    "files.associations": {
+        "array": "cpp",
+        "deque": "cpp",
+        "filesystem": "cpp",
+        "forward_list": "cpp",
+        "list": "cpp",
+        "queue": "cpp",
+        "ranges": "cpp",
+        "regex": "cpp",
+        "span": "cpp",
+        "stack": "cpp",
+        "vector": "cpp",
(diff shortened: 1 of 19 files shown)


tallytalwar:
Filed as internal issue #USD-8285

FlorianZ:
Hi @Vertexwahn, would you mind please providing a description for your PR and supplying information on which version of boost python introduced the get_indices method?
Also, if you wouldn't mind please reverting the .vscode configuration file that inadvertently slipped into this PR?
Thank you!!

Vertexwahn:
Sorry - didn't get that this is a boost Python binding thing. Was thinking this is a typo - just checked Boost 1.82.0 - it is still called `get_indicies` - sorry for the confusion

FlorianZ:
@Vertexwahn, no you're certainly on to something. I believe this is actually a typo in boost. They later corrected it by adding the `get_indices` API alongside the misspelled `get_indicies`. Both methods do the same thing - one just calls into the other. We can certainly fix this by accepting a PR that calls the correctly spelled method, but we need to be careful to make sure we're not breaking backwards compatibility with any older versions of boost that we still support. I don't know which version of boost introduced the corrected spelling.

PR #2415, opened 2023-04-28
Fix VtArray declaraions of extern template instantiations so that the…

marktucker:
Fix VtArray declaraions of extern template instantiations so that the code in array.cpp will actually instantiate and export the templates as expected. Currently the templates are marked as extern even while compiling array.cpp so the compiler isn't actually instantiating anything. This problem seems to only apply to Windows builds.

### Description of Change(s)
The existing code in vt/array.cpp doesn't actually generate exported symbols for some compilers as demonstrated here: https://godbolt.org/z/Gz1vc5eEe

The array.h file should be using the VT_API_TEMPLATE_CLASS macro so that when compiling array.cpp there is no confusion about whether code should be generated to instantiate the templates. In array.cpp VT_API must be added explicitly to the instantiation macro to ensure the generated symbols are exported.

Note that for many compilers there are many configurations here that all seem to work okay, either because of inlining of VtArry functions, or different interpretations of classes defined first as extern in the header, then non-extern in the cpp.

Thanks go to @e4lam for pointing this out.

### DIFF
--- a/pxr/base/vt/array.cpp
+++ b/pxr/base/vt/array.cpp
@@ -47,7 +47,7 @@ Vt_ArrayBase::_DetachCopyHook(char const *funcName) const
 
 // Instantiate basic array templates.
 #define VT_ARRAY_EXPLICIT_INST(r, unused, elem) \
-    template class VtArray< VT_TYPE(elem) >;
+    template class VT_API VtArray< VT_TYPE(elem) >;
 BOOST_PP_SEQ_FOR_EACH(VT_ARRAY_EXPLICIT_INST, ~, VT_SCALAR_VALUE_TYPES)
 
 
--- a/pxr/base/vt/array.h
+++ b/pxr/base/vt/array.h
@@ -919,7 +919,7 @@ class VtArray : public Vt_ArrayBase {
 // Declare basic array instantiations as extern templates.  They are explicitly
 // instantiated in array.cpp.
 #define VT_ARRAY_EXTERN_TMPL(r, unused, elem) \
-    extern template class VtArray< VT_TYPE(elem) >;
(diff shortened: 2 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-8286

wi1k1n:
Had the exact same issue: clang on windows complaining about ```lld-link : error : undefined symbol: public: __cdecl pxrInternal_v0_23__pxrReserved__::VtArray<float>::VtArray<float>(void)``` in debug variant (strangelly was fine in release variant). applying changes from the commit solved the issue!

Many many thanks!

moshev:
I suspect this went unnoticed because MSVC does not (and will not) implement `extern template`, see [https://developercommunity.visualstudio.com/t/c11-extern-templates-doesnt-work-for-class-templat/157868](https://developercommunity.visualstudio.com/t/c11-extern-templates-doesnt-work-for-class-templat/157868). So when compiling with cl.exe the template is simply instantiated each time, but when using a different compiler that follows the standard, it actually looks for the symbols and you get errors like:
```error LNK2019: unresolved external symbol "public: float const & __cdecl pxrInternal_v0_22__pxrReserved__::VtArray<float>::operator[](unsigned __int64)const " (??A?$VtArray@M@pxrInternal_v0_22__pxrReserved__@@QEBAAEBM_K@Z) referenced in function```
(posting it here so it's easier to google for others). Note that even though MSVC doesn't follow the standard and ignores the `extern template` declaration, the explicit instantiation is still exported in the DLL, so it's all fine - you can for example compile USD with MSVC and then compile and link a project that uses that USD with clang-cl.

Edit: @wi1k1n that is also why you're only seeing the error in a debug build. In a release build, the compiler (both MSVC and Clang) inlines most of VtArray.

PR #2420, opened 2023-05-03
APPLE: Fix for primID rendering in AOVs

creijon:
### Description of Change(s)

When rendering primID as an AOV, the `IntToVec4` function originally packed the 32-bit ID into all 4 channels, which will correspond to RGBA. However, using A for the top 8 bits causes problems.  First, somewhere downstream RGB will be premultiplied by A and information will be lost. Second, when these end up in an output image, there is no way to know what pixels have valid IDs, unless we pick a specific RGBA like (0,0,0,0) to mean "no object".

While this means we can only handle 16 million IDs, this seems to be the safest option right now. See similar code in frameRecorder.cpp which interprets these values.

Thanks to the wider team at Apple for this.

### Fixes Issue(s)
- Incorrect output in primID AOV

### DIFF
--- a/pxr/imaging/hdSt/shaders/renderPass.glslfx
+++ b/pxr/imaging/hdSt/shaders/renderPass.glslfx
@@ -276,12 +276,22 @@ void RenderOutput(vec4 Peye, vec3 Neye, vec4 color, vec4 patchCoord)
 // with AoVs.  We generate "primId" and "instanceId", and depending on how
 // the client called glDrawBuffers, they will get one or both.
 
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8291

slingthor:
Re-created here
https://github.com/PixarAnimationStudios/OpenUSD/pull/2622/files

davidgyu:
This has been addressed via 524ad40 and e6e01c7 and related commits.

PR #2421, opened 2023-05-03
WIP: EnhancingUSDPreviewSurface via alignment with Khronos glTF PBR + Autodesk Standard Surface

bhouston:
NOTE: "WIP" in the title means "Work in Progress" and it should not be considered a merge candidate.

### Description of Change(s)

In response to [this discussion back in December](https://github.com/PixarAnimationStudios/USD/issues/2119), I have started on the work to align USDPreviewSurface with the common real-time oriented subset of Adobe Standard Material, Autodesk Standard Surface and the Khronos glTF PBR specification.

I am working off of [this spreadsheet of mine that compares/contrasts here](https://docs.google.com/spreadsheets/d/1Af5Oevg-ES4aEH3BrH6tpzrUtyoPxN2O6wm5YRDgnZE/edit#gid=0) and specifically I am adding the lines that are marked as [GAP] in the USDPreviewSurface column:

### Scope of Work

Specifically this PR aims the following parameters to the USDPreviewSurface:

- specularFactor, common definition
- specularColorFactor, common definition
- specularAnisotropy (float), using Autodesk Standard Surface definition, per @jstone-lucasfilm's recommendation
- specularRotation (float), using Autodesk Standard Surface definition, per @jstone-lucasfilm's recommendation
- clearcoatNormal (normal), common definition
- clearcoatColor (color), common definition
- sheen (color), common definition
- sheenRoughness (float), common definition
- transmission (float), common definition
- attenuationColor (color), common definition
- attenuationDistance (float), common definition

I've purposely left out the following for the first pass, just to try to make progress in bite-sized chunks:
- iridescenceFactor
- scattering
- diffuse transmission

### Changes

To start with I've added them to shaderDef.usda here:

https://github.com/PixarAnimationStudios/USD/blob/1e4c1dfd300e38d7f0a9cfe49e5e7212a77571d1/pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda

I've also added additional documentation linking existing parameters to both the Khronos glTF PBR and the Autodesk Standard Surface specifications.  This makes it clear that USD Preview Surface is a subset of existing standards thus furthering the goal of interoperability.

### Open for Feedback

I've just started on adding support for these new parameters.  I am just learning the USD library so it may take a bit of time figuring out how the USD library works, but at least this long proposed alignment is now becoming concrete.  I welcome any contributions or suggestions on how best to do this alignment.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/2119

### DIFF
--- a/pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda
+++ b/pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda
@@ -50,7 +50,7 @@ def Shader "UsdPreviewSurface" (
              Transition between the two colors according to Schlick fresnel 
              approximation."""
     )
- 
+
     float inputs:metallic = 0.0 (
         doc = """Used only in the metalness workflow.
             1 for metallic surfaces and 0 for non-metallic.
@@ -69,6 +69,14 @@ def Shader "UsdPreviewSurface" (
         GGX or Beckmann lobe."""
     )
  
+    float inputs:anisotropyStrength = 0.0 (
+        doc = """Anisotropy strength of the specular lobe. 0 means isotropic, 1 means fully anisotropic."""
+    )
+ 
+    float inputs:anisotropyRotation = 0.0 (
+        doc = """Anisotropy rotation of the specular lobe. Orientation of anisotropy; range [0,1] (1 means 360°)"""
+    )
+ 
     float inputs:clearcoat = 0.0 (
         doc = """Second specular lobe amount. The color is white."""
     )
@@ -76,7 +84,16 @@ def Shader "UsdPreviewSurface" (
     float inputs:clearcoatRoughness = 0.01 (
         doc = """Roughness for the second specular lobe."""
     )
- 
+  
+    color3f inputs:clearcoatTint = (1.0, 1.0, 1.0) (
+        doc = """The color to tint the clearcoat layer."""
+    ) 
+    
+    normal3f inputs:clearcoatNormal = (0.0, 0.0, 1.0) (
+        doc = """The normal of the clearcoat layer.  The same format and restrictions
+        that apply to *normal* also apply to this clearcoat normal."""
+    )
+
     float inputs:opacity = 1.0 (
         doc = """Opacity of the material."""
     )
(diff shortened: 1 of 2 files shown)


pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda
```diff
@@ -124,6 +159,59 @@ def Shader "UsdPreviewSurface" (
...
+
+    float inputs:attenuationDistance = +Inf (
+        doc = """Attenuation distance.  This is the distance that light travels through the
+        surface before it is attenuated.  This is useful for glass-like materials.  Equivalent to 
+        attenuationDistance from KHR_materials_volume and transmission_depth from Autodesk
+        Standard Surfaced."""
+    )
+
+    float inputs:iridescenceFactor = 0.0 (
+        doc = """The iridescence intensity factor."""
```
emackey: Small nitpick, I'd avoid the terms `Factor` (and `Texture` if that's anywhere).  "Factor" is a glTF-specific term that is roughly the equivalent of an untextured shader uniform, and its presence often coincides with a lack of texturemapping and/or is being used to enable an extension.

Here for example, one might call this simply `iridescence`, not `iridescenceFactor`.

emackey: Likewise for `specularFactor` -> `specular`, and `specularColorFactor` -> `specularColor`, etc.

bhouston: BTW I just can not use the term "specularColor" as it is already in USD Preview Surface with another meaning:

```
    color3f inputs:specularColor = (0.0, 0.0, 0.0) (
        doc = """Used only in the specular workflow.
             Specular color to be used.
             This is the color at 0 incidence. Edge color is assumed white.
             Transition between the two colors according to Schlick fresnel 
             approximation."""
    )
```

Given it defaults to (0,0,0) and it has a different meaning and may already be used in a lot of files, I feel I should introduce another term in seeking alignment with Autodesk and Khronos to avoid any potential confusing.

bhouston: Also I removed "iridescence" for both in part because it introduced the min/max thickness variables which no one else uses except glTF.  I figured it would be a distracting conversation which we can revisit in the future.  I figured if I stay within the low hanging fruit, it would be more productive.

tallytalwar:
Filed as internal issue #USD-8292

bhouston:
Closing per @jstone-lucasfilm recommendation.

PR #2433, opened 2023-05-11
APPLE: Move ResolveCullMode to geometry shader and remove accessors.

creijon:
### Description of Change(s)
The `_ResolveCullMode` method in `hdSt/renderPassState.cpp` requires a number of accessors on the `HdSt_GeometricShader`, specifically:

- `GetCullStyle`
- `GetUseHardwareFaceCulling`
- `GetHasMirroredTransform`
- `GetDoubleSided`
- 
It seems to make more sense to move this function into the geometric shader and remove these accessors entirely.

### Fixes Issue(s)
- General cleanup


tallytalwar:
Filed as internal issue #USD-8317

PR #2435, opened 2023-05-12
Using the correct integer compression to get decompression working space size

NickWu:
### Description of Change(s)
There is an issue in `_DecompressIntegers` function regarding dealing with integers bigger than 4 bytes. When dealing with integers bigger than 4 bytes, it is still using `Usd_IntegerCompression::GetDecompressionWorkingSpaceSize` to calculate the output buffer size where it should have used `Usd_IntegerCompression64::GetDecompressionWorkingSpaceSize` instead. 

This is the fix for this issue https://github.com/PixarAnimationStudios/USD/issues/2434

### Fixes Issue(s)
- Fixed an error at https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/fastCompression.cpp#L116 

### DIFF
--- a/pxr/usd/usd/integerCoding.cpp
+++ b/pxr/usd/usd/integerCoding.cpp
@@ -389,9 +389,14 @@ template <class Int>
 size_t _DecompressIntegers(char const *compressed, size_t compressedSize,
                            Int *ints, size_t numInts, char *workingSpace)
 {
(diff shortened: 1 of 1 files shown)


tallytalwar:
Filed as internal issue #USD-8319

PR #2436, opened 2023-05-16
APPLE: Replace bitwise with logical operators

creijon:
### Description of Change(s)

The predicates of the if statements should use logical rather than bitwise operators.

### Fixes Issue(s)
- Warnings in the build.

### DIFF
--- a/pxr/usd/usdSkel/animation.cpp
+++ b/pxr/usd/usdSkel/animation.cpp
@@ -282,8 +282,8 @@ UsdSkelAnimation::SetTransforms(const VtMatrix4dArray& xforms,
     VtVec3hArray scales;
     if (UsdSkelDecomposeTransforms(xforms, &translations,
                                    &rotations, &scales)) {
-        return GetTranslationsAttr().Set(translations, time) &
-               GetRotationsAttr().Set(rotations, time) &
+        return GetTranslationsAttr().Set(translations, time) &&
+               GetRotationsAttr().Set(rotations, time) &&
                GetScalesAttr().Set(scales, time);
     }
     return false;


tallytalwar:
Filed as internal issue #USD-8341

pxr/base/tf/stringUtils.cpp
```diff
@@ -806,8 +806,8 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
             // Add 5 mod 32 makes '_' sort before all letters.
             return ((l + 5) & 31) < ((r + 5) & 31);
         }
-        else if (IsDigit(l) | IsDigit(r)) {
-            if (IsDigit(l) & IsDigit(r)) {
+        else if (IsDigit(l) || IsDigit(r)) {
+            if (IsDigit(l) && IsDigit(r)) {
                 // We backtrack to find the start of each digit string, then we
```
gitamohr: The use of bitwise operators here are intentional, to make the generated code branchless.

pxr/base/tf/stringUtils.cpp
```diff
@@ -869,7 +869,7 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
                 curEnd = lcur + std::min(std::distance(lcur, lend),
                                          std::distance(rcur, rend));
             }
-            else if (IsDigit(l) | IsDigit(r)) {
```
gitamohr: Same as above.

pxr/base/tf/stringUtils.cpp
creijon: Ah, sorry.  I'll back that out.

creijon: But then, further down you've got:

```
else if (!IsAlpha(l) || !IsAlpha(r)) {
```

pxr/base/tf/stringUtils.cpp
creijon: I've backed out those changes, just leaving the ones in usdSkel

pxr/base/tf/stringUtils.cpp
creijon: I was checking this with a simple test function in compiler explorer and didn't see any logical difference in the generated output with `-O3`.  But I might have missed something.

PR #2439, opened 2023-05-17
[usdAbc] Reenable USD_ABC_READ_ARCHIVE_USE_MMAP

shadeops:
With this change a user may set USD_ABC_READ_ARCHIVE_USE_MMAP=0 to disable memory map read support.

### Description of Change(s)

The USD_ABC_READ_ARCHIVE_USE_MMAP TfGetEnvSetting was defined but had no effect when set. Originally intended to allow for the user to choose between pread and mmap, the functionality was lost in commit 5a795b2cb.

We restore that functionality and change USD_ABC_READ_ARCHIVE_USE_MMAP's default from 'false' to 'true' to match the current behavior of Alembic 1.7.9+ and the usdAbc plugin.

### Fixes Issue(s)
- n/a


sunyab:
Filed as internal issue #USD-8353

PR #2440, opened 2023-05-18
Report the path of the invalid spec (not the inducing spec) in `SdfCopySpec`

nvmkuruc:
### Description of Change(s)
`SdfCopySpec` iterates over child specs when copying.  If any of those specs is invalid (`GetSpecType() == SdfSpecTypeUnknown`), a coding error is reported.  However, the error message reports the path to the inducing spec (the source path argument to `SdfCopySpec`) as invalid when a child is invalid.

This PR fixes the error message so that the right spec is reported as being invalid.

### Fixes Issue(s)


sunyab:
Filed as internal issue #USD-8354

PR #2442, opened 2023-05-18
Set correct aspect for depth buffers with HgiVulkan

schevrel:
### Description of Change(s)
Set the correct region aspect when dealing with depth buffers so that Vulkan copies the data correctly

### Fixes Issue(s)
aovInputTask does not correctly copy depth buffers from hydra delegate when running Hgi Vulkan.


sunyab:
Filed as internal issue #USD-8356

PR #2444, opened 2023-05-23
APPLE: Add clip distance handling on Metal

slingthor:
Replacement for an older PR. This one includes more robust support for out handling and initialisation

### Description of Change(s)
- Add clip distance support on Meta
- We always put a stub in for clip plane handling of size 1.
- This is done as having a size of 0 clip planes is hard to handle in the generated code that does not otherwise know if it should be there or not
### Fixes Issue(s)
- Metal respects explicit clip planes


sunyab:
Filed as internal issue #USD-8358

PR #2448, opened 2023-05-26
Support MaterialX 1.38.7

adro79:
### Description of Change(s)
Support TextureBaker.h changes in newer MaterialX versions.

### Fixes Issue(s)
- Building with MaterialX 1.38.7


sunyab:
Filed as internal issue #USD-8364

klucknav:
Thank you for the PR. We recently added support for MaterialX v1.38.7 and as a result we already have a similar change in the dev branch. Thanks again!

PR #2453, opened 2023-05-31
Add strict validation mode for AR resource identifier schemes

nvmkuruc:
### Description of Change(s)
The current implementation of AR resource identifiers (URI/IRI) allows any string to be registered as the scheme component.  The URI/IRI specification restricts the character set of the scheme component to ASCII alphanumeric characters plus `+`, `.`, and `-`.  Consider handling of a scheme which contains `:` as an example of where an unrestricted character set for scheme can go awry.  This change adds validation, warning the user when a scheme is not valid under the specification.  By default, the scheme is ignored, but users can set `PXR_AR_DISABLE_STRICT_SCHEME_VALIDATION` to be more permissive.

Validation could be implemented as a regular expression match, but was done with explicit checks to produce clearer error messages for users.  A note is left to output the invalid characters once #2120 has been accepted.

The `test_output` scheme currently used in some of the testing is an example of an invalid scheme per the URI/IRI specifications.  This scheme has been added to a list of invalid test cases and been replaced with `test-output`.

### Fixes Issue(s)

### DIFF
--- a/pxr/usd/ar/resolver.cpp
+++ b/pxr/usd/ar/resolver.cpp
@@ -87,6 +87,12 @@ TF_DEFINE_ENV_SETTING(
     PXR_AR_DISABLE_PLUGIN_URI_RESOLVERS, false,
     "Disables plugin URI resolver implementations.");
 
+TF_DEFINE_ENV_SETTING(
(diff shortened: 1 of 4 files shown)


sunyab:
Filed as internal issue #USD-8372

PR #2456, opened 2023-06-01
geomUtil: use Tf.PreparePythonModule() in __init__.py

mattyjams:
### Description of Change(s)

This follows the pattern used by other library Python modules and ensures that the `pxr.GeomUtil` Python module can be imported correctly on all platforms.

In particular, this fixes an issue on Windows where the `pxr.GeomUtil` Python module fails to import:

```
C:\Users\MattJohnson\Developer>python
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from pxr import GeomUtil
  File "D:\USD_dev\lib\python\pxr\GeomUtil\__init__.py", line 28, in <module>
    from . import _geomUtil
ImportError: DLL load failed while importing _geomUtil: The specified module could not be found.
>>>
```

In my case, I wasn't actually trying to use anything out of `GeomUtil` in Python, but because the module is included in the `__all__` list in the top-level `pxr` module's `__init__.py`, you'll see the same failure-to-import when doing `from pxr import *`.

### DIFF
--- a/pxr/imaging/geomUtil/__init__.py
+++ b/pxr/imaging/geomUtil/__init__.py
@@ -25,15 +25,6 @@
 Utilities to help image common geometry.
 """
 
-from . import _geomUtil
 from pxr import Tf
-Tf.PrepareModule(_geomUtil, locals())
-del _geomUtil, Tf
-
-try:
-    import __DOC
-    __DOC.Execute(locals())
-    del __DOC
-except Exception:
-    pass
-
+Tf.PreparePythonModule()
+del Tf


sunyab:
Filed as internal issue #USD-8374

mattyjams:
Just rebased on the current dev branch to kick the CI into running again.

PR #1908, opened 2022-06-16
From tbb::atomic to std::atomic

boberfly:
### Description of Change(s)

Swaps out tbb::atomic with std::atomic, with the help of the people in #1471

_PrototypeTask and _Entry structs need defined copy/move constructors to be able to compile (std::atomic has copy deleted and move undefined). The adding into a TfHashMap with a make_pair is considered a move.

### Fixes Issue(s)
- In the process of supporting a later TBB version https://github.com/PixarAnimationStudios/USD/issues/1471 tbb::atomic is considered deprecated in favour of std::atomic.

- [x] I have submitted a signed Contributor License Agreement

PR #2460, opened 2023-06-01
oneTBB: tbb::atomic and tbb::concurrent_unordered_map changes in usdImagining

brechtvl:
### Description of Change(s)
Swaps out tbb::atomic with std::atomic, and correct tbb::concurrent_unordered_map iterators to be non-const where older TBB was too lenient.

Split off from #1908

### Fixes Issue(s)


sunyab:
Filed as internal issue #USD-8378

PR #2463, opened 2023-06-01
oneTBB: change tbb::tbb_thread to std::thread

brechtvl:
### Description of Change(s)
The only place where tbb::thread was still used.

### Fixes Issue(s)


sunyab:
Filed as internal issue #USD-8381

PR #2467, opened 2023-06-01
Nominally reference the IRI specification in documentation

nvmkuruc:
### Description of Change(s)
The URI specification nominally referenced throughout the `Ar` documentation restricts the character set to ASCII characters. This update to documentation clarifies that--
* Resolvers registered via `uriSchemes` may also accept UTF-8 encoded IRIs (a unicode supporting resource identifier specification)
* Resolvers working with protocols that only accept URIs may (but are not required to) support IRIs encoded in layers and handle percent encoding idempotently at the resolver level
* USD does not validate URI or IRI paths beyond what's needed to dispatch via a resource identifier's scheme component. Resolvers handle their own parsing, validation, and normalization of asset paths.

This change also converts the invalid `my_scheme` URI/IRI scheme to `my-scheme`.

### Fixes Issue(s)
- #2120 

### DIFF
--- a/pxr/usd/ar/overview.dox
+++ b/pxr/usd/ar/overview.dox
@@ -94,15 +94,20 @@ that will search for an asset in a specified list of directories.
 See documentation on ArDefaultResolver for more information on configuring
 this behavior.
 
-\subsection ar_uri_resolvers URI Resolvers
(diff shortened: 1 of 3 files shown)


sunyab:
Filed as internal issue #USD-8385

PR #2468, opened 2023-06-02
usdBakeMtlx, usdrecord: skip building usdImaging tools if GPU and/or OpenGL are unavailable

mattyjams:
This fixes two issues when building USD with `PXR_BUILD_USD_IMAGING` **enabled** but `PXR_ENABLE_GPU_SUPPORT` and/or `PXR_ENABLE_GL_SUPPORT` **disabled**.

The build of `usdAppUtils` is gated on `PXR_BUILD_GPU_SUPPORT` being enabled, but there was no such check for `usdrecord`, so the build would be aborted when CMake goes to build `usdrecord` but discovers the dependency is missing.

The `usdBakeMtlx` change might be slightly more contentious, but in environments where OpenGL is not supported, it's likely that building MaterialX will yield an installation that is missing the `MaterialXRenderGlsl` library. Attempting to then build USD in that environment against such an installation of MaterialX with `PXR_ENABLE_GL_SUPPORT` disabled would be aborted when CMake goes to build `usdBakeMtlx` but discovers the dependency is missing.

### DIFF
--- a/pxr/usdImaging/bin/usdBakeMtlx/CMakeLists.txt
+++ b/pxr/usdImaging/bin/usdBakeMtlx/CMakeLists.txt
@@ -1,6 +1,14 @@
 set(PXR_PREFIX pxr/usdImaging)
 set(PXR_PACKAGE usdBakeMtlx)
 
+# MaterialX's MaterialXRenderGlsl library, which is used by usdBakeMtlx,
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-8399

PR #2471, opened 2023-06-06
Apple: Mesh shader stage - high level

slingthor:
### Description of Change(s)
- Adds a mesh shader stage to HGI and Storm
- Minimal changes that should allow us getting the stage in, with shader generator, binding, codegen, geometric shader and mesh shader itself to follow.


slingthor:
Would it be possible to run this through tests again?

jesschimein:
Filed as internal issue #USD-8402

PR #2473, opened 2023-06-06
Autodesk: Fix bug where ResolveParameter doesn't support type conversion

erikaharrison-adsk:
Currently it is required that the type of the value must match the type of the parameter. However sometimes we still want to convert the value. For example, the Wrap parameter of the sampler requires a TfToken type, which can be TfToken(“repeat”), TfToken(“clamp”) and so on. But in the USD file, it is common that we just write a string as the value. When we resolve the parameter, we need to cast from a string to token. This fix enables us to get the value from the usd file for the parameters, by casting to the type of parameter.

### Description of Change(s)

When the type of the node is different from the required type, we try to cast the value of the node to the required type.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/materialNetwork.cpp
+++ b/pxr/imaging/hdSt/materialNetwork.cpp
@@ -551,6 +551,12 @@ _ResolveParameter(
         if (value.IsHolding<T>()) {
             return value.UncheckedGet<T>();
         }
+        else if (value.CanCast<T>())
+        {
+            VtValue castResult = value;
+            castResult.Cast<T>();
+            return castResult.UncheckedGet<T>();
+        }
     }
 
     // Then fallback to SdrNode.
@@ -560,6 +566,12 @@ _ResolveParameter(
             const VtValue &value = input->GetDefaultValue();
             if (value.IsHolding<T>()) {
                 return value.UncheckedGet<T>();
+            }
+            else if (value.CanCast<T>())
+            {
+                VtValue castResult = value;
+                castResult.Cast<T>();
+                return castResult.UncheckedGet<T>();
             }
         }
     }


jesschimein:
Filed as internal issue #USD-8403

spiffmon:
Hi @erikaharrison-adsk !  While it's in the Sdr code, it looks like we have left out one of the important SdrPropertyType -> SdfValueTypeName rules in the [Sdr overview](https://openusd.org/release/api/sdr_page_front.html).  Specifically, most shading systems have an annotation/metadata to indicate a parameter has "options", i.e. is a string (or string with integer index selection) enumeration.  The `token` SdfValueTypeName is USD's way of indicating a property represents an enumeration, so it is semantically meaningful that `inputs:wrapMode` is `token` and not `string`, and we **want** that to be validated and validatable.

So I don't think we would want this change, and hope that your authoring tools can use Sdr node definitions to guide GUI into writing the proper types for shader parameters?

erikaharrison-adsk:
Thanks for this. We're going to take a look at your suggestion and see whether we need to iterate on this discussion or not. If its straightforward (ie. no other tweaks to propose), we'll close this PR in a week or so if that works.

spiffmon:
Thanks so much, @erikaharrison-adsk - that sounds great!

PierreWang:
Hi @spiffmon . thank you for your comments. This change was created by me and @erikaharrison-adsk helped me to push it to pixar repo. I think the problem is, in the usda file, if a parameter is a string enumeration, the value it is assigned can only be a string. So here when we resolve the parameter, the value of the SdrProperty will hold a string, but the value we want to get is TfToken. We need a cast from string to TfToken.

If we only need the conversion from string to TfToken, not a general cast as I wrote, I can add check for the type before the cast.

spiffmon:
Hi @PierreWang - apologies I'm not quite grasping where the problem is occurring in your toolchain.  Are you aware of  `SdrShaderProperty::GetTypeAsSdfType()` ?  For an enumeration-type property, it will tell you the SdfValueTypeName is `Sdf.ValueTypeNames.Token`, which you can use to construct a properly-type UsdAttribute.  Does that help?

PierreWang:
Hi @spiffmon . Indeed it is a different case. The SdfValueTypeName is string, and we want a token. But I understand what is the right behaviour. If valuetype is a string, we should get the string from the value, and then construct the token from the string. We don't need to modify ResolveParameter. Thank you. I think this pull request can be closed.

sunyab:
Thanks @PierreWang, closing per your last comment.

spiffmon:
Also, @PierreWang , I'm curious why you can't change the "shader definition" (i.e. whatever is parsed by a NdrParserPlugin) to get that `Wrap` parameter to be a `token` ?  Is this an Arnold node, or MaterialX ?

PR #2474, opened 2023-06-06
Autodesk: Fix for shader compile error with mx::heighttonormal nodes

erikaharrison-adsk:
When we use MaterialX heighttonormal node to convert a height map to normal map, Storm reports a shader error. error C7522: OpenGL requires constants to be initialized

The root cause is that the filter kernels are uninitialized

const float c_box_filter_weights[84];
const float c_gaussian_filter_weights[84];

Also logged as USD github issue https://github.com/PixarAnimationStudios/USD/issues/2281

### Description of Change(s)
Ensure blur filters are initialized correctly by MaterialX glslfx codegen

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/2281

### DIFF
--- a/pxr/imaging/hdSt/materialXShaderGen.cpp
+++ b/pxr/imaging/hdSt/materialXShaderGen.cpp
@@ -689,12 +689,12 @@ HdStMaterialXShaderGen<Base>::_EmitConstantsUniformsAndTypeDefs(
     Base::emitLineBreak(mxStage);
     Base::emitTypeDefinitions(mxContext, mxStage);
 
-    // Add all constants
+    // Add all constants and ensure that values are initialized
     const mx::VariableBlock& constants = mxStage.getConstantBlock();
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8404

ashwinbhat:
Tagging @klucknav for visibility. Thanks

ashwinbhat:
Hi 
Upon further analysis this fix does not work for macOS Metal.

ashwinbhat:
**Update:**
This PR requires MaterialX [PR 1576] (https://github.com/AcademySoftwareFoundation/MaterialX/pull/1576)

Without the MaterialX PR, MacOS metal results will fail, but Windows/Linux are ok. 
With the MaterialX PR, Test should pass on all platforms. 



FYI: @jesschimein

PR #2476, opened 2023-06-06
Autodesk: Use TfNormPath to fix TfPathUtils test failures on Windows

erikaharrison-adsk:
### Description of Change(s)

Unit tests were failing on our Windows setup due to path issues. Wrap with TfNormPath.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8406

pxr/base/tf/testenv/pathUtils.cpp
```diff
@@ -104,11 +104,11 @@ TestTfRealPath()
 
     if (testSymlinks) {
         // Leaf dir is symlink
-        TF_AXIOM(TfRealPath("d", true) == TfAbsPath("subdir"));
+        TF_AXIOM(TfNormPath(TfRealPath("d", true)) == TfNormPath(TfAbsPath("subdir")));
```
tallytalwar, line 111: Since the updates here are primarily to fix the test on windows and the failure is not noticed on linux or macOS, it was mentioned in an internal review if we can split and add some ARCH_OS_WINDOWS guards for these updates?

tallytalwar (changes requested):
Hi @erikaharrison-adsk 

Thanks for submitting the PR.

Added a note about using "ARCH_OS_WINDOWS" for the test updates as the test fixups are required for windows only.

Also in an internal review, we were wondering if you can provide an example of a path which fails, just in case there are any code improvements we can do to mitigate this specific case?

Thanks

pxr/base/tf/testenv/pathUtils.cpp
zhangha182: Thanks a lot for your comments. We'll wrap these changes under ARCH_OS_WINDOWS and add some comments as well.

The reason is that TfRealPath on windows explicitly lower-cases drive letters if the given path contains symlinks, otherwise it returns the same as os.path.abspath. The similar workaround is already used in [pxr/base/tf/testenv/testTfPathUtils.py](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/base/tf/testenv/testTfPathUtils.py#L58-L68)

erikaharrison-adsk: Branch updated. @tallytalwar - let us know if this works or if you need any more follow-up.

PR #2482, opened 2023-06-08
APPLE: Keep the color target as an FP16 format in hgiInterop

creijon:
### Description of Change(s)

Keep the colour target as an FP16 in HgiInteropMetal, which allows openGL clients to use EDR.  This doesn't change the behaviour in usdview, we'll have to add some changes into the Python code to create the right final target.

### Fixes Issue(s)
- [Allow for 16bit colour targets in Metal->openGL applications](https://github.com/PixarAnimationStudios/USD/issues/2481)


jesschimein:
Filed as internal issue #USD-8412

sunyab:
This PR was merged into the `dev` branch in 075fc26.

PR #2484, opened 2023-06-08
[build] add a --cmake-build-args flag to build_usd.py

pmolodo:
### Description of Change(s)

add a --cmake-build-args flag to build_usd.py

useful if you need to make a change to all dependencies that use cmake (including USD itself)


jesschimein:
Filed as internal issue #USD-8414

PR #2492, opened 2023-06-16
Autodesk: Fix bug where Creation of HgiGLSampler will always set GL_TEXTURE_MAX_ANISOTROPY_EXT to 16

erikaharrison-adsk:
### Description of Change(s)

HgiGLSampler will always enable ANISOTROPY. But when the filter is nearest, we should disable anisotropy. In this branch we fix this issue. In the constructor of HgiGLSampler, we will check if the minFilter or magFilter is nearest. If yes, we will not set GL_TEXTURE_MAX_ANISOTROPY_EXT. When GL_TEXTURE_MAX_ANISOTROPY_EXT is set, it will ignore the minFilter and magFilter on some platform. This means that we will never create a nearest sampler.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8424

PR #2493, opened 2023-06-16
Autodesk: Add support for input/output qualifiers

erikaharrison-adsk:
### Description of Change(s)

Summary: Add support for input/output qualifiers for input/output layout. "flat", "noperspective", "centroid" and "sample" are supported.

Currently USD doesn’t support interpolation qualifiers in the shader input/output layout. USD uses the “layout” structure as the shader input/output, which can be parsed into platform-dependent code. One example is:
```
-- layout Curves.Vertex.Patch

[
    ["out block", "CurveVertexData", "outData",
        ["vec4", "Peye"],                                                          
        ["vec3", "Neye"]
    ]
]
```
But you can not add “flat” qualifier to the member. The HdSt_ResourceLayout class can not identify it. This will block any feature that requires flat interpolation. 
This code change adds interpolation support to the layout. It can be correctly translated to GLSL or MSL. The layout will be like this:
```
-- layout Curves.Vertex.Patch

[
    ["out block", "CurveVertexData", "outData",
        ["vec4", "Peye"],                                                          
        ["vec3", "Neye"]
        ["float", "flatData", "", "flat"] // The third string is already defined by pixar as array size.
    ]
]
```

Currently we support “flat”, “noperspective”, “centroid”, and “sample” interpolation, as these are the interpolations that are supported in global value.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/CMakeLists.txt
+++ b/pxr/imaging/hdSt/CMakeLists.txt
@@ -207,6 +207,7 @@ pxr_library(hdSt
         shaders/visibility.glslfx
         shaders/volume.glslfx
         textures/fallbackBlackDomeLight.png
+        testenv/testHdstQualifiers/testQualifiers.glslfx
 
     DOXYGEN_FILES
         overview.dox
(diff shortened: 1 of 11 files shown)


jesschimein:
Filed as internal issue #USD-8425

PR #2496, opened 2023-06-21
Replace `boost::lexical_cast` with `pxr/base/tf/stringUtils.h` in `UsdStageCache`

nvmkuruc:
### Description of Change(s)
To reduce the dependency on boost, replace usage of `boost::lexical_cast` with `TfStringify` and `TfStringToLong`.

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/usd/usd/stageCache.h
+++ b/pxr/usd/usd/stageCache.h
@@ -28,8 +28,7 @@
 #include "pxr/usd/usd/api.h"
 #include "pxr/usd/sdf/declareHandles.h"
 #include "pxr/base/tf/declarePtrs.h"
-
-#include <boost/lexical_cast.hpp>
+#include "pxr/base/tf/stringUtils.h"
 
 #include <string>
 #include <memory>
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8430

pxr/usd/usd/stageCache.h
```diff
@@ -107,15 +106,23 @@ class UsdStageCache
         /// Create an Id from a string value.  The supplied \p val must have
         /// been obtained by calling ToString() previously.
         static Id FromString(const std::string &s) {
-            return FromLongInt(boost::lexical_cast<long int>(s));
+            bool overflow = false;
+            int64_t result = TfStringToInt64(s, &overflow);
```
crevilla: Using `long result = TfStringToLong(s, &overflow);` would be more technically correct here.

nvmkuruc: Agreed!

PR #2497, opened 2023-06-22
Improve UsdLux->Sdr translation

johnhaddon:
It may be that this would be better as a query on `usd-interest`, but I thought it might be worth having a stab at coding it so that there's something concrete to talk about. My goal is to be able to use `Sdr` as a source of shader and light definitions that can be loaded onto nodes in Gaffer, for presentation to the user via Gaffer's UI. The UsdLux `schema.usda` defines light inputs in a user-friendly order and with useful metadata for `displayGroup` and `displayName`, but that is currently lost when the schema is parsed into the `SdrRegistry`. This results in things like `RectLight.width` and `RectLight.height` being separated, and `enableColorTemperature` no longer being alongside `colorTemperature`. In this PR I've attempted to preserve this information during parsing to Sdr so that the user is presented with a well organised UI in Gaffer.

Questions in my mind : Was this a reasonable expectation to have from Sdr in the first place? Does whitelisting `propertyOrder` like this in `usdGenSchema` have unwanted side-effects I don't know about?

### Description of Change(s)

- Preserve property order in UsdLux_LightDefParserPlugin.
- Translate Usd displayGroup/displayName to Sdr page/label.

- [ ] I have verified that all unit tests pass with the proposed changes (`ctest` reports `No tests were found!!!`, so clearly I did something wrong)
- [x] I have submitted a signed Contributor License Agreement (via Cinesite)

### DIFF
--- a/pxr/usd/usd/usdGenSchema.py
+++ b/pxr/usd/usd/usdGenSchema.py
@@ -1643,7 +1643,7 @@ def GenerateRegistry(codeGenPath, filePath, classes, validate, env):
             #   'specifier' - This is a required field and will always exist.
             # Any other metadata is an error.
             allowedAPIMetadata = [
-                'specifier', 'customData', 'documentation']
+                'specifier', 'customData', 'documentation', 'propertyOrder']
             # Single apply API schemas are also allowed to specify 'apiSchemas'
             # metadata to include other API schemas.
             if apiSchemaType == SINGLE_APPLY or apiSchemaType == MULTIPLE_APPLY:
--- a/pxr/usd/usdLux/generatedSchema.usda
+++ b/pxr/usd/usdLux/generatedSchema.usda
@@ -34,6 +34,7 @@ class "LightAPI" (
     """
 )
 {
+    reorder properties = ["inputs:intensity", "inputs:exposure", "inputs:diffuse", "inputs:specular", "inputs:normalize", "inputs:color", "inputs:enableColorTemperature", "inputs:colorTemperature"]
     uniform bool collection:lightLink:includeRoot = 1
     uniform bool collection:shadowLink:includeRoot = 1
     color3f inputs:color = (1, 1, 1) (
@@ -360,6 +361,7 @@ class "ShapingAPI" (
     doc = "Controls for shaping a light's emission."
 )
 {
+    reorder properties = ["inputs:shaping:focus", "inputs:shaping:focusTint", "inputs:shaping:cone:angle", "inputs:shaping:cone:softness", "inputs:shaping:ies:file", "inputs:shaping:ies:angleScale", "inputs:shaping:ies:normalize"]
     float inputs:shaping:cone:angle = 90 (
         displayGroup = "Shaping"
         displayName = "Cone Angle"
@@ -412,6 +414,7 @@ class "ShadowAPI" (
     non-physical controls that are valuable for visual lighting work."""
 )
 {
+    reorder properties = ["inputs:shadow:enable", "inputs:shadow:color", "inputs:shadow:distance", "inputs:shadow:falloff", "inputs:shadow:falloffGamma"]
     color3f inputs:shadow:color = (0, 0, 0) (
         displayGroup = "Shadows"
         displayName = "Shadow Color"
@@ -691,6 +694,7 @@ class DistantLight "DistantLight" (
     Also known as a directional light."""
 )
 {
+    reorder properties = ["inputs:angle", "inputs:intensity"]
     float inputs:angle = 0.53 (
         displayGroup = "Basic"
         displayName = "Angle Extent"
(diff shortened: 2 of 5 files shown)


jesschimein:
Filed as internal issue #USD-8432

spiffmon:
Hi @johnhaddon ,
    Definitely a worthy problem, and the shaderDefUtils change is a good catch!  Our concern about propagating `propertyOrder` into the schema definition is about what happens when typed schemas are composed with applied schemas in the UsdPrimDefinition (which is the thing that's supplying the `propertyOrder`).  What happens with the contributions from API schemas?  Should _they_ be allowed  to contribute ordering?  And if so, how do the different orderings combine?  We'd want to spend some time thinking about it before opening that door.

   We also wonder if, given your fix for preserving displayGroup information, you could instead add/leverage displayGroups in the Lux schema definitions such that, even though you'd still get lexicographic ordering of the properties as a starting point, you'd be able to (e.g.) group `width` and `height` in a `shape` or other group?  Which your UI could then pay attention to?

johnhaddon:
Thanks for the feedback @spiffmon. I suspected the `propertyOrder` stuff might have implications beyond this narrow topic and my limited USD knowledge. In practice, I will need to roll something out in Gaffer before even the `displayGroup` stuff lands in a release, so I'll just enforce an ordering separately on the Gaffer side for now. But I do think preserving the natural order _somehow_ is a reasonable ask if Sdr is intended as a means of loading definitions into user interfaces.

Here's a comparison using displayGroups with and without the ordering being maintained. Lexicographical on the left, and natural (as presented in `schema.usd`) on the right :

image

Lexicographical isn't _completely_ horrible, but it's not great either. Having `enableShadows` in the middle of its group rather than the top breaks the natural top-to-bottom flow an artist would normally visit parameters in. Having height before width and colour before exposure is also unusual, and it's sheer luck that Shadows and Shaping come late in the alphabet. I don't think it's something I could roll out without some push back.

> What happens with the contributions from API schemas? Should they be allowed to contribute ordering? And if so, how do the different orderings combine? 

With my UsdLux->Sdr tunnel vision, I'd say yes - they are concatenated, concrete schema first and applied schemas after, in the order they were applied. But I have no idea of the wider implications.

Anyway, I think I've discovered a bigger hitch in my plans. Taking the Sdr definition at face value, every newly created light has a `shaping:cone:angle` input defaulting to 90 degrees, which limits the illumination to the hemisphere in front of the light. The individual APIs are something the user should opt in to explicitly, but Sdr presents each light as a single bag of inputs, with no (correct me if I'm wrong) way of determining that some of them came from an optional API. Maybe I shouldn't be using Sdr as the source of truth for light definitions, and should be using the schemas directly instead? That would be a pity - all our existing integrations have managed to avoid a separate code path for loading lights.

All of this is will be dealt with easily enough by hardcoding some special cases about ordering and activation of parameters/schemas on our side, but I had hoped that Sdr would provide enough information to do everything generically. Looking at the two USD-native environments I have access to, there is quite a difference in the way lights are presented in each though, so perhaps I'm hoping for too much in wanting a single source of truth and a standard UI presentation?

spiffmon:
Hi @johnhaddon , you've made the case for (optionally, since it does add cost) allowing schemas to dictate their property order quite compellingly.  We also realized that in our internal goal of replacing most of Presto's scenegraph with USD, it will be a feature-parity concern.  We still need to discuss whether we want to accept your PR now, as-is, or wait till we can make the time to address the API-combinatoric aspects, as well.

As to the issue illuminated by the `shaping` properties, API schemas provide an elegant solution that (I think) maps well to the physicality of lights, where the shaping equipment is (conceptually, at least) "added on" to the basic light, and we've had some internal discussions about making it easier/more obvious to configure prims with API schemas, though no work has been scheduled yet.

There is no official mechanism for property "conditional visibility/activation" in Sdr.  Personally, I'd like to really explore a future where Sdr is primarily a rendering-focused subsystem (needed because Hydra doesn't want to assume the existence of a UsdStage or schemas as input), but enable USD clients/tools to remain unaware of it.  In other words, push forward on what we started in our integration of RenderMan into USD, where `usdgenschemafromsdr` can be leveraged to produce schemas not just for "shader like" RenderMan objects/API's, but for shaders in general.  This wouldn't have been practical without both codeless schemas *and* schema versioning, and there are still dev workflow issues to address, but I think tools in general would benefit from there being *one way* to interact with UsdPrims.

johnhaddon:
> We still need to discuss whether we want to accept your PR now, as-is, or wait till we can make the time to address the API-combinatoric aspects, as well.

It won't impact my work in the short term either way, so certainly no pressure from me to accept as-is if you can do a better job in the longer term. One thing that felt odd in my approach was having to manually add `propertyOrder` when the properties had clearly already been described in the required order, so perhaps some order-preserving option in the `.usda` format reader might be an improvement?

> Personally, I'd like to really explore a future where Sdr is primarily a rendering-focused subsystem...but enable USD clients/tools to remain unaware of it. 

Do you think that future is likely enough that I should prefer the UsdLux schemas over Sdr _now_, knowing that I'll be able to use schemas for _everything_ in future? The only other thing I was hoping to load in the short term was UsdPreviewSurface stuff, and that has schemas already, so it should work fine for me I think.

PR #2501, opened 2023-06-26
Validating UsdCollectionAPI before accessing in UsdShadeMaterialBindingAPI::ComputeBoundMaterial

### LINKED ISSUE #2500: UsdShadeMaterialBindingAPI crashes if collection points to a non-existing property
### Description of Issue
The `UsdShadeMaterialBindingAPI::ComputeBoundMaterial` crashes when a collection on the primitive points to a non-existing property.

While the attached scene is invalid, `ComputeBoundMaterial` should not crash.

### Steps to Reproduce
1. Open the scene from 
[shirt_02_simplified.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/11867960/shirt_02_simplified.zip)
in usdview

Windows 10

### Package Versions
HEAD of dev

sirpalee:
### Description of Change(s)
Validating UsdCollectionAPI before accessing in UsdShadeMaterialBindingAPI::ComputeBoundMaterial.

This is a quick proposal to fix #2500 , but I'm unsure if this is the right place to do so. I.e. should we validate collections earlier, like in `UsdShadeMaterialBindingAPI::_GetCollectionBindings` or a similar place?

### Fixes Issue(s)
- #2500

### DIFF
--- a/pxr/usd/usdShade/materialBindingAPI.cpp
+++ b/pxr/usd/usdShade/materialBindingAPI.cpp
@@ -791,6 +791,9 @@ UsdShadeMaterialBindingAPI::ComputeBoundMaterial(
 
                 const UsdCollectionAPI &collection = 
                     collBinding.GetCollection();
+                if (!collection) {
+                    continue;
+                }
                 const SdfPath &collectionPath = collBinding.GetCollectionPath();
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8440

PR #2504, opened 2023-06-26
Fix Hgi Vulkan shader generation flip UV y coordinate on texture access.

schevrel:
### Description of Change(s)

Modifying shader generation on HgiVulkan to flip the UV y coordinate when looking up textures so Vulkan shaders behave like on other backends.

### Fixes Issue(s)
Aovs processed by UsdImagingGl tasks (Color correction, selection, etc.) appear flipped vertically from their initial state after processing when using Hgi Vulkan.


jesschimein:
Filed as internal issue #USD-8444

Issue #2502, opened 2023-06-26
`SdfMapEditProxy` has inconsistent comparison operator overloads


nvmkuruc:
### Description of Issue
While removing the dependency on `boost::totally_ordered`, it was observed many of the comparison operator overloads (`==`, `<`, `>`, etc.) return `false` if a `_Validate` method fails.  This could lead to situations where (for example) `==` and `!=` both return `false` for the same values.

However, the operators that are inherited from `boost::totally_ordered` cannot know about this `_Validate` method, leading to inconsistent behavior depending on the value types being compared and which operators are being exercised. The boost removal effort intends to preserve the existing behavior. It was requested that this be filed as a reminder to examine after `boost::totally_ordered` has been removed.

### Steps to Reproduce
N/A

jesschimein:
Filed as internal issue #USD-8441

PR #2506, opened 2023-06-26
Replace `boost::totally_ordered` with explicit operator overloads for `SdfMapEditProxy`

nvmkuruc:
### Description of Change(s)
- Add explicit implementations of `operator{<=,>=,!=}` to `SdfMapEditProxy` when compared against its `Type` template parameter
- #2502 documents some inconsistencies between operator overloads that should be looked at in a future change. This PR attempts to preserve the existing behavior.

### Fixes Issue(s)
- #2250 


jesschimein:
Filed as internal issue #USD-8446

PR #2507, opened 2023-06-27
Apple: Fix floating point error caused by FMA conversion

slingthor:
### Description of Change(s)
Adds a compile flag that prevents FMA conversion for floating points due to additions and multiplications which changed behaviour of unit tests.
-ffp-contract specifically controls fused multiply-add operations
https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-contract

### Fixes Issue(s)
The following tests FAILED:
111 - testGfCamera (Failed)
433 - testUsdGeomPointInstancer (Failed)
472 - testUsdSkelBakeSkinning (Failed)

When building on Xcode 14.3 and newer.
With the change, they now pass.

- [] I have verified that all unit tests pass with the proposed changes

### DIFF
--- a/cmake/defaults/clangdefaults.cmake
+++ b/cmake/defaults/clangdefaults.cmake
@@ -26,6 +26,12 @@ include(gccclangshareddefaults)
 
 set(_PXR_CXX_FLAGS "${_PXR_GCC_CLANG_SHARED_CXX_FLAGS}")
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8447

slingthor:
Moved the change to clang defaults

PR #2514, opened 2023-06-28
Apple: Mesh shaders glslfx files

slingthor:
### Description of Change(s)
Includes changes to the glslfx files involved for Mesh shading. Implements an inline index optimisation for drawing with mesh shaders.


jesschimein:
Filed as internal issue #USD-8455

PR #2522, opened 2023-07-03
Revise removal of `boost::compressed_pair` to better preserve object layout

nvmkuruc:
### Description of Change(s)
Some minor performance regressions were observed in the initial removal of `boost::compressed_pair` (#2259) . This PR restores the order of inheritance of the equality and hash function objects which were inadvertently switched as well as some of the original initialization logic.

### Fixes Issue(s)
- #2257 


jesschimein:
Filed as internal issue #USD-8477

PR #2524, opened 2023-07-05
Remove unneeded boost dependencies from python bindings

nvmkuruc:
### Description of Change(s)
Several usages of boost in the python bindings can be easily removed.

* Remove unused header references to `boost/functional/hash.hpp`
* Introduce a simple pair and range-like struct for iterators to remove `boost::begin` and `boost::end` in favor of `std::begin` and `std::end`
* Replace `boost::lexical_cast` with `TfStringify`

### Fixes Issue(s)
N/A

### DIFF
--- a/pxr/base/plug/wrapRegistry.cpp
+++ b/pxr/base/plug/wrapRegistry.cpp
@@ -31,7 +31,6 @@
 #include "pxr/base/tf/pySingleton.h"
 #include "pxr/base/tf/stringUtils.h"
 
-#include <boost/range.hpp>
 #include <boost/noncopyable.hpp>
 #include <boost/python.hpp>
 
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-8480

pxr/base/plug/wrapRegistry.cpp
```diff
@@ -54,6 +50,23 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace {
```
crevilla: I feel like this would be simpler if we just change the PluginNames function to no longer be templated and just take two `const PlugPluginPtrVector::iterator &` parameters instead for begin and end. This especially given it's only called in two places.

nvmkuruc: Done!

PR #2525, opened 2023-07-07
Fix inconsistent units used in documentation string

eoineoineoin:
### Description of Change(s)

Discussed this with @eberled - just noticed a minor issue with the units in the documentation for UsdPhysicsDriveAPI's damping.

### Fixes Issue(s)

The motor force equation is given by:

> Force or acceleration = stiffness * (targetPosition - position) + damping * (targetVelocity - velocity)

The stiffness term is correct. However, for angular motors, the documentation for the damping term uses incorrect units:

```
        doc = """Damping of the drive. Units: 
		if linear drive: mass/second
		If angular drive: mass*DIST_UNITS*DIST_UNITS/second/second/degrees."""
```

i.e. kg m^2 s^-2 deg^-1
So the term "damping * (targetVelocity - velocity)" has units of "kg m^2 s^-3" which is not the units of torque (kg m^2 s^-2)

While in the area, I changed the stiffness term to use the word "degrees" rather than "degree," which is more consistent with the surrounding documentation. I'll also note that the file "wp_rigid_body_physics.rst" has a copy of the physics schema from when it was a proposal, but it's out of date (and has a warning to that effect) and I have not updated it.

### DIFF
--- a/pxr/usd/usdPhysics/driveAPI.h
+++ b/pxr/usd/usdPhysics/driveAPI.h
@@ -328,7 +328,7 @@ public:
     // --------------------------------------------------------------------- //
     /// Damping of the drive. Units: 
     /// if linear drive: mass/second
-    /// If angular drive: mass*DIST_UNITS*DIST_UNITS/second/second/degrees.
+    /// If angular drive: mass*DIST_UNITS*DIST_UNITS/second/degrees.
     ///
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-8481

Issue #196, opened 2017-04-08
Add stubs and docstrings for python bindings


chadrik:
There have been some big strides made recently with static type analysis in python, and I have a perhaps unhealthy obsession with it (hopefully to everyone's benefit).  I'm wrapping up work on a script that creates stub files which can be used by an IDE to provide code hints and warnings for the `pxr` python package.  I find to be indispensable, especially when working with a new API.  I'd like to discuss the possibility of adding the execution of this script as a build option.

Python stub files are normal python files with whose class and function definitions contain `pass` statements instead of functionality.  There are a handful of extensions used for stub files, depending on the IDE.  The ones I'm aware of are .py, .pi, .pypredef.  The IDE inspects the stub files instead of the *real* modules, which is particularly useful when the real modules are compiled extensions and therefore cannot be statically analyzed, such as with USD.

Recently there's been a big push in the python community to add static type hints to python, described in [PEP484](https://www.python.org/dev/peps/pep-0484/), recently released in the form of the [typing](https://docs.python.org/3/library/typing.html) module for python 2&3 and actual syntax support in versions 3.5+.  Python itself doesn't *do* anything with the type hints:  the point of them is to be available to third-party static type checking tools.  The gold standard of these is [mypy](http://mypy.readthedocs.io/en/latest/), but PyCharm also has very good support as of version 2017.1.

PEP484 describes its own type of stub file, saved with a pyi extension.  The pyi file sits next to its .py equivalent and in addition to providing membership hints an docstrings like normal stubs, it also adds type hints. 

Long story short, my stub generator can create either kind of stub file.

The script works like this:
- generate xml data using Doxygen
- parse the xml files, extract the arguments and typing info, and run it through a set of rules to convert it to our best prediction of the corresponding python types and object names. 
- import the built `pxr` module, inspect the contents, looking up argument and type data in the structure from the previous step, and write out .pyi stub files for each `__init__.py` file.

Note: we rely on inspection to drive stub generation rather than creating stubs directly from Doxygen, because we don't have perfect knowledge of how C++ classes are wrapped in python, and it's better to have missing type info than missing objects/members.

[...]

PR #2526, opened 2023-07-07
convertDoxygen: Add docstrings to more python objects

chadrik:
### Description of Change(s)

With these changes, the `convertDoxgen` build script adds docstrings for many additional python objects.  

For example, from Sdf these objects are now documented:

```
SpecType
Specifier
Permission
Variability
AuthoringError
GetNameForUnit
GetUnitFromName
ValueHasValidType
GetTypeForValueTypeName
GetValueTypeNameForValue
ConvertToValidMetadataDictionary
JustCreatePrimAttributeInLayer
CopySpec
ComputeAssetPathRelativeToLayer
ListOpType
CreatePrimInLayer
JustCreatePrimInLayer
CreateVariantInLayer
```

The primary change is that we no longer filter "file" xml type: these are necessary to find info on non-class functions.

In addition, this PR also provides a change necessary to support creation of [python type stubs](https://github.com/PixarAnimationStudios/OpenUSD/issues/196):  the parser now captures function default arguments. 

### Performance impact

Casting a wider net during parsing increased parse time from 25s to 1m45s.  The addition of slots shaved off ~10s and reduced memory usage. 

I've run this through a profiler, and the majority of the time is spent in the xml parsing, with about half of the parsing-time spent in pxr code and the other half in the parser base class.  I think there's room to make this more efficient if we can avoid instantiating `XMLNode` unnecessarily.  I noticed that for every class/function there is a useless "doxygen" root node, so I tried collapsing that down to a single root, but it only saved 10s and it creates some other problems.  

I also tried using `multiprocessing` to speed up parsing, but even distributing the parsing of files across 16 processes did not speed it up -- I suspect this was due to the amount of `DocElement` instances that needed to be serialized and marshaled across to the primary process after parsing, due to the way that the `multiprocessing` module works in python.   If that inference is correct, this approach could be improved by filtering the resulting `DocElement` classes while parsing, as there are many extraneous functions being captured. 

tl;dr  with this change, the doc generation process takes longer, but there's not a simple solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .p4*
 .DS_Store
 .AppleDouble
-
+*.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .p4*
 .DS_Store
 .AppleDouble
-*.pyc
--- a/docs/python/convertDoxygen.py
+++ b/docs/python/convertDoxygen.py
@@ -106,11 +106,13 @@ if ',' in modules:
     # Loop through module list and create a Writer for each module to 
     # load and generate the doc strings for the specific module
     for moduleName in moduleList:
+        if not moduleName:
+            continue
         writer = Writer(packageName, moduleName)
         # Parser.traverse builds the docElement tree for all the 
         # doxygen XML files, so we only need to call it once if we're
         # processing multiple modules
-        if (docList == None):
+        if (docList is None):
             docList = parser.traverse(writer)
             Debug("Processed %d DocElements from doxygen XML" % len(docList))
         Debug("Processing module %s" % moduleName)
@@ -120,8 +122,12 @@ if ',' in modules:
         module_output_file = os.path.join(module_output_dir, "__DOC.py")
         writer.generate(module_output_file, docList)
 else:
+    moduleName = modules
     # Processing a single module. Writer's constructor will sanity
     # check module and verify it can be loaded
-    writer = Writer()
+    if not output_file.endswith(".py"):
+        module_output_dir = os.path.join(output_file, moduleName)
+        output_file = os.path.join(module_output_dir, "__DOC.py")
+    writer = Writer(packageName, moduleName)
     docList = parser.traverse(writer)
     writer.generate(output_file, docList)
--- a/docs/python/doxygenlib/cdDocElement.py
+++ b/docs/python/doxygenlib/cdDocElement.py
@@ -29,8 +29,13 @@
 #
 # The Parser class is responsible for building up a list of these objects.
 #
+from collections import namedtuple
 
 from .cdUtils import Warn
+
+
+Param = namedtuple("Param", "type name default")
+
 
 class DocElement:
     """
@@ -44,6 +49,9 @@ class DocElement:
     in that list represents an alternative calling signature for the
     same function, i.e., in C++ parlance, the overloaded methods.
     """
+
+    __slots__ = ("name", "kind", "prot", "doc", "location", "children", "const", "virt", "explicit",
+                 "static", "inline", "returnType", "argsString", "definition", "params")
 
     def __init__(self, name, kind, prot, doc, location):
         self.name = name                     # the name of this class/method
@@ -60,7 +68,10 @@ class DocElement:
         self.returnType = None               # return type of a method/function
         self.argsString = None               # arguments for this method/func
         self.definition = None               # full C++ definition for method
-        self.params = None                   # name and type of each parameter
+        self.params = None                   # type, name, and default of each parameter
+
+    def __repr__(self):
+        return "%s(%r, %r, %r, ...)" % (self.__class__.__name__, self.name, self.kind, self.location)
 
     def isFunction(self):
         """Is this doc element a function?"""
@@ -86,6 +97,10 @@ class DocElement:
         """Is this doc element the root of the doxygen XML tree?"""
         return self.kind == 'root'
 
+    def isStatic(self):
+        """Is this doc element static?"""
+        return self.static is not None and self.static == 'yes'
+    
     def addChildren(self, children):
         """Adds the list of nodes as children of this node."""
         for child in children:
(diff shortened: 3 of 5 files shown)


docs/python/doxygenlib/cdWriterDocstring.py
```diff
@@ -52,29 +52,6 @@ class Writer:
     # we can combine property docstrings for getters/setters.
     propertyTable = {}
 
-    # Default constructor that assumes we're processing a single module
-    def __init__(self):
-        #
```
chadrik, line 57: python doesn't support overloads like C++:  you can't simply define a function twice.  This `__init__` was being replaced by the one below it, so calling `Writer()` without args never actually worked.  I've fixed the use of this class in `convertDoxygen`

docs/python/doxygenlib/cdWriterDocstring.py
```diff
@@ -295,11 +272,14 @@ def generate(self, output_file, docElements):
...
+        if not os.path.exists(outputDir):
+            os.makedirs(outputDir)
+
         # output the lines to disk...
         try:
-            logfile = open(output_file, 'w')
-            logfile.write('\n'.join(lines))
-            logfile.close()
+            with open(output_file, 'w') as logfile:
+                logfile.write('\n'.join(lines))
```
chadrik, line 282: use context manager to ensure that the file is closed in case of an exception (also, more idiomatic)

docs/python/doxygenlib/cdParser.py
```diff
@@ -42,13 +42,18 @@ class XMLNode:
     """
     Rrepresent a single node in the XML tree.
     """
+    __slots__ = ("parent", "name", "attrs", "text", "childNodes")
+
```
chadrik, line 46: Added slots for performance:  we make a _lot_ of these classes.

docs/python/doxygenlib/cdDocElement.py
```diff
@@ -86,6 +97,10 @@ def isRoot(self):
         """Is this doc element the root of the doxygen XML tree?"""
         return self.kind == 'root'
 
+    def isStatic(self):
+        """Is this doc element static?"""
+        return self.static is not None and self.static == 'yes'
+
```
chadrik, line 103: Added `isStatic` and `__repr__` methods for convenience

docs/python/convertDoxygen.py
```diff
@@ -106,11 +106,13 @@
...
     # load and generate the doc strings for the specific module
     for moduleName in moduleList:
+        if not moduleName:
+            continue
         writer = Writer(packageName, moduleName)
         # Parser.traverse builds the docElement tree for all the 
         # doxygen XML files, so we only need to call it once if we're
         # processing multiple modules
-        if (docList == None):
+        if (docList is None):
```
chadrik, line 115: using `is` instead of `==` is considered idiomatic because `None` is a singleton.

jesschimein:
Filed as internal issue #USD-8482

.gitignore
```diff
@@ -1,4 +1,4 @@
 .p4*
 .DS_Store
 .AppleDouble
-
+*.pyc
```
pmolodo: Since you're adding this - don't suppose you could add `__pycache__` to the list as well?
Had that queued up in another PR, but if you just add that to your MR, would avoid a merge conflict...

docs/python/doxygenlib/cdParser.py
```diff
@@ -221,8 +226,7 @@ def parseDoxygenIndexFile(self, doxygen_index_file):
                     continue
                 if (kind == "dir"):
                     continue
-                if (kind == "file"):
-                    continue
```
pmolodo, line 225: So I think this change is the heart of this PR - unfortunately, when we tested it, it had some negative side effects.
Specifically, many of our classes lost their docstrings after running `convertDoxygen.py` with this change.

One example was `pxr.Usd.ZipFileWriter` - in `Usd/__DOC.py`, we now get:

```python
   result["ZipFileWriter"].__doc__ = """"""
```

instead of:

```python
   result["ZipFileWriter"].__doc__ = """
Class for writing a zip file. This class is primarily intended to
support the .usdz file format. It is not a general-purpose zip writer,
as it does not implement the full zip file specification. However, all
files written by this class should be valid zip files and readable by
external zip modules and utilities.

"""
```

We're using Doxygen 1.8.18... which is old, but also newer than the 1.8.14 listed in [VERSIONS.md](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/VERSIONS.md).

@chadrik, can you confirm whether or not you're seeing the same, and what version of Doxygen you're using?

spiffmon: Hey @pmolodo , you made us realize we neglected to update VERSIONS.md for the addition of doxygen-awesome, which requires 1.9.x (we are using 1.9.6 internally).  We will be pushing out updates to VERSIONS.md shortly, but can you retry your experiments with a 1.9.x version?

chadrik:
I did test the results before and after my change to ensure proper results
but it’s quite possible that I missed something. I was doing a synthetic
test of just the docstring build rather than a full cmake build so that
could also have led to differences. I will double check this on my end.

docs/python/doxygenlib/cdParser.py
dsyu-pixar: I tried a test with only removing these lines and comparing build output (using doxygen 1.8.x). The new __DOC.py appears to have duplicate (empty) values for some things, like ZipFileWriter as @pmolodo notes, which end up overwriting the earlier defined values.

chadrik:
I pushed a fix for this.

.gitignore
pixar-oss: Hi @chadrik, is the .gitignore change critical for the main goal of this PR? For tracking changes for USD releases, it helps if all the changes for a PR are related. 

If the .gitignore change isn't specifically needed for this PR, would it be possible to move the .gitignore change (maybe combined with @pmolodo's change) to a separate PR?

dsyu-pixar: Hi @chadrik, is the .gitignore change critical for the main goal of this PR? For tracking changes for USD releases, it helps if all the changes for a PR are related.

If the .gitignore change isn't specifically needed for this PR, would it be possible to move the .gitignore change (maybe combined with @pmolodo's change) to a separate PR?

chadrik: Removed!

docs/python/doxygenlib/cdParser.py
chadrik: This is fixed.

PR #2527, opened 2023-07-07
Replace `boost::call_traits` in `pxr/usd/sdf/accessorHelpers.h`

nvmkuruc:
### Description of Change(s)

The macros in `pxr/usd/sdf/accessorHelpers.h` use `boost::call_traits` to convert the parameter `T` to `const T` if `T` is a primitive type and the cost of the reference outweighs the cost of the copy (`const T&` otherwise).

This replaces that macro with a templated `using` statement and standard STL traits.

### Fixes Issue(s)

### DIFF
--- a/pxr/usd/sdf/accessorHelpers.h
+++ b/pxr/usd/sdf/accessorHelpers.h
@@ -173,6 +173,13 @@ SDF_ACCESSOR_CLASS::name_(                                                     \
 
 // Convenience macros to provide common combinations of value accessors
 
+// Convert non-trivial types like `std::string` to `const std::string&` while
+// preserving the type for `int`, `bool`, `char`, etc.
+template <typename T>
+using Sdf_SetParameter = std::conditional<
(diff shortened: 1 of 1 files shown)


pxr/usd/sdf/accessorHelpers.h
```diff
@@ -173,6 +173,13 @@ SDF_ACCESSOR_CLASS::name_(                                                     \
 
 // Convenience macros to provide common combinations of value accessors
```
crevilla, line 175: Okay, so I think we could make this more complete by adding to the conditional by making it 
std::Is_arithmetic<T>::value || std::is_enum<T>::value
All the places I see that use the SDF_DEFINE_TYPED_GET_SET macro are specifically to for enum values which leads me to believe enums were a hole in the original boost call_traits.
However, I'm okay with leaving this as is in your change and following a followup bug to add is_enum and replace the SDF_DEFINED_TYPED_GET_SET usages

jesschimein:
Filed as internal issue #USD-8484

PR #2530, opened 2023-07-10
Replace `TF_NUM_ARGS` with `TF_PP_VARIADIC_SIZE` in `staticTokens.h`

nvmkuruc:
### Description of Change(s)
For non-empty argument lists, the boost independent `TF_PP_VARIADIC_SIZE` returns the same result as the boost dependent `TF_NUM_ARGS`. With this change, the USD code base is no longer dependent on `TF_NUM_ARGS`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8489

PR #2214, opened 2023-01-27
Replace boost traits and mpl with STL traits in pxr/usd/sdf, pxr/usd/pcp, and pxr/usd/usd

nvmkuruc:
### Description of Change(s)
- Replaces `boost::remove_cv` and `boost::remove_reference` with `std::remove_cv_t` and `std::remove_reference_t`
- Replaces `boost::true_type` with `std::true_type`
- Replaces `boost::is_base_of` with `std::is_base_of`
- Replaces `boost::has_trivial_constructor`, `boost::has_trivial_copy`, `boost::has_trivial_assign`, and `has::trivial_destructor` with `std::is_trivially_constructible`, `std::is_trivially_copyable`, `std::is_trivially_assignable`, and `std::is_trivially_desctructible`.

### Fixes Issue(s)
- #2211

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2533, opened 2023-07-12
Replace use of boost traits with `std::conditional`

nvmkuruc:
### Description of Change(s)
- Replace `boost::mpl::if_` with `std::conditional`

### Fixes Issue(s)
- #2214 


jesschimein:
Filed as internal issue #USD-8494

PR #2532, opened 2023-07-12
Move `TF_EAT_PARENS` to `tf/preprocessorUtilsLite.h`

nvmkuruc:
### Description of Change(s)
Depends on #2241 

- Move `TF_EAT_PARENS` to `pxr/base/tf/preprocessUtilsLite.h` now that it's not dependent on boost
- Update `pxr/base/vt/traits.h` and `pxr/base/tf/testenv/meta.cpp` dependencies

### Fixes Issue(s)
- #2240 

- [ ] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2241, opened 2023-02-02
Make `TF_PP_EAT_PARENS` not dependent on boost

nvmkuruc:
### Description of Change(s)
- Replaces usage of `BOOST_PP_CAT` with `TF_PP_CAT` 

### Fixes Issue(s)
- #2240 

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

Issue #2240, opened 2023-02-02
`TF_PP_EAT_PARENS` can be made boost independent and moved to `preprocessorUtilsLite.h`


nvmkuruc:
### Description of Issue

`TF_PP_EAT_PARENS` only depends on `BOOST_PP_CAT` for which an equivalent `TF_PP_CAT` already exists.  Making this change can reduce dependencies on boost.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

tallytalwar:
Filed as internal issue #USD-7968

nvmkuruc:
This was addressed by #2241 and #2532

PR #2535, opened 2023-07-12
Remove unused `tf/preprocessorUtils.h` dependency

nvmkuruc:
### Description of Change(s)
Depends on #2241 and #2532.

Several files depend on `tf/preprocessorUtils.h` unnecessarily.  As `tf/preprocessorUtils.h` brings in a lot of boost dependencies, it's helpful to minimize its usage.

After this PR lands, its only usage will be in `tf/staticTokens.h`.

### Fixes Issue(s)
#2240 


jesschimein:
Filed as internal issue #USD-8496

PR #2543, opened 2023-07-17
Remove spurious include of boost header in `sdf/types.h`

nvmkuruc:
### Description of Change(s)

`boost/preprocessor/punctuation/comma.hpp` is not being used by `sdf/types.h` and can be removed

### Fixes Issue(s)
- #2243 


jesschimein:
Filed as internal issue #USD-8508

PR #2544, opened 2023-07-18
Remove support for `TfToken` arrays in `tf/staticTokens.h` macros

nvmkuruc:
### Description of Change(s)
The macros in `tf/staticTokens.h` currently allow the following syntax.
```
((myArray, ((value1) (value2) (value3))))
```
which will add `value1`, `value2`, and `value3` as tokens as well as `TfToken myArray[3]` to token static data structure with references to `value1`, `value2`, and `value3`. This feature is unused in the OpenUSD code base and brings in a lot of additional boost dependencies to support.

Future refinements will be required to completely remove the boost dependency from this file, but removing the array support is a first step.

This feature can be replaced by simply declaring the values of the array as tokens.
```
(value1)
(value2)
(value3)
```
The `allTokens` vector would have the same contents (though the order may change).  Usage which benefits from having a static array of tokens can be replaced in the scopes they're needed.
```
void function() {
    static const std::array<TfToken, 3> myArray = {{Tokens->value1, Tokens->value2, Tokens->value3}};
    ...
}
```
### Fixes Issue(s)

### DIFF
--- a/pxr/base/tf/pyStaticTokens.h
+++ b/pxr/base/tf/pyStaticTokens.h
@@ -39,12 +39,9 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-// TODO: Should wrap token arrays to Python.
-
 /// Macro to wrap static tokens defined with \c TF_DEFINE_PUBLIC_TOKENS to
 /// Python.  It creates a class of name \p name in the current scope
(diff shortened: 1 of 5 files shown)


jesschimein:
Filed as internal issue #USD-8510

nvmkuruc:
Closing-- this was done in 62f96936efcfc8d3de535351cb29ec9830089489

PR #2548, opened 2023-07-24
Replace `boost::iterator_facade` with explicit iterator implementation for `SdfListProxy`

nvmkuruc:
### Description of Change(s)
* Replace `iterator_facade` with explicit implementation for `SdfListProxy`'s iterator

### Fixes Issue(s)
- #2305 

### DIFF
--- a/pxr/usd/sdf/listProxy.h
+++ b/pxr/usd/sdf/listProxy.h
@@ -35,7 +35,6 @@
 #include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/errorMark.h"
 #include "pxr/base/tf/iterator.h"
-#include <boost/iterator/iterator_facade.hpp>
 #include <boost/iterator/reverse_iterator.hpp>
 #include <boost/optional.hpp>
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8521

pxr/usd/sdf/listProxy.h
```diff
@@ -136,43 +135,129 @@ class SdfListProxy {
...
+                      "not be a lvalue reference type");
+
+        _Iterator() = default;
 
         _Iterator(Owner owner, size_t index) : _owner(owner), _index(index)
         {
             // Do nothing
         }
 
+        reference operator*() const { return dereference(); }
```
crevilla, line 176: I'm confused as to why this -> returns the address directly when you define pointer = _PtrProxy and static_assert that value type of reference is not an lvalue. My expectation is that your assert would mean we always return the _PtrProxy or that we want a conditional type here like you did SdfMapEditProxy

nvmkuruc: You're right!  That was an error.  Thanks.

PR #2549, opened 2023-07-24
Replace `boost::iterator_facade` with explicit implementation for `SdfMapProxy`

nvmkuruc:
### Description of Change(s)
- Replace `iterator_facade` with explicit implementation for `SdfMapProxy`
- Uses type traits / SFINAE to provide two implementations of `operator->` depending on whether the reference type is a true reference and can be safely converted to a pointer or if it's a proxy type and needs a pointer proxy to manage lifetime. Added a comment to recommend deprecating usage of `operator->` for C++20 to avoid this complexity.

### Fixes Issue(s)
- #2305 


jesschimein:
Filed as internal issue #USD-8522

PR #2550, opened 2023-07-24
Fix issues using Hydra and Storm with OpenGL core profile 

brechtvl:
### Description of Change(s)

These changes together make Hydra and Storm work without warnings in Blender, which uses the OpenGL core profile rather than the compatibility profile.

* Don't use point size or line width for core profile, since these are not available and generate many warnings. This only silences the warnings, ideally these features would be implemented in shaders.
* Fix missing VAO binding that is required with the core profile, causing Storm renders to fail.
* Also fix VAO double free and leak in GL graphics pipeline that I noticed looking into this.

---

The solution for VAOs in `HdSt_IndirectDrawBatch` does not seem ideal, as OpenGL usage should be abstracted away in `hgiGL`. However OpenGL vertex buffer binding still seems to be part of `hdSt` and it's not clear to me where it will move exactly.

Also note that while it may seem that in `HdSt_IndirectDrawBatch::_ExecuteFrustumCull`, `cullGfxCmds->BindPipeline` will bind the VAO, this doesn't actually work because those commands are queued and only executed after functions like `state.BindResourcesForViewTransformation` that bind the vertex buffer objects.

Advice for a better solution would be welcome.


jesschimein:
Filed as internal issue #USD-8536

DarkDefender:
@unhyperbolic Is there any update on this? Sorry if I'm poking the wrong person.
We are applying multiple patches to the USD version that we officially ship with Blender, this is one of them.
As we are up to quite a few patches at this point (7) and more are starting to pile up, we would like to have these fixes upstream.

jesschimein:
Hi @DarkDefender we're hoping to take a look at this PR soon!

PR #2551, opened 2023-07-25
Fix crash when reading corrupted USD files

seando-adsk:
### Description of Change(s)
- Throw an exception when detecting an out-of-bound read instead of returning random data.
- CrateFile helpers re-throw the exceptions they caught.
- Catch exception in the parallel read dispatcher, otherwise the dispatcher destructors that run afterwards cause double-exception termination or read invalid memory.
- CreateData remembers that exceptions were thrown during I/O and returns an error. Cratedata cannot throw an exception because it is invoked in parallel code that crash when exceptions are thrown. Error return are used instead.
- With this, the error propagates to the caller that tried to load the layer or stage.

### Fixes Issue(s)
- Crash when reading corrupted USD files.

### DIFF
--- a/pxr/usd/usd/crateFile.cpp
+++ b/pxr/usd/usd/crateFile.cpp
@@ -4008,7 +4008,7 @@ CrateFile::_UnpackValue(ValueRep rep, T *out) const
                          "%s, returning a value-initialized object",
                          GetAssetPath().c_str(),
                          ArchGetDemangled<T>().c_str());
-        *out = T();
+        throw;
     }
 }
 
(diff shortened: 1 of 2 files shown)


pxr/usd/usd/crateFile.cpp
```diff
@@ -595,8 +596,7 @@ struct _MmapStream {
                     "Read out-of-bounds: %zd bytes at offset %td in "
                     "a mapping of length %zd",
                     nBytes, offset, mapLen);
-                memset(dest, 0x99, nBytes);
-                return;
+                throw std::runtime_error("Read out of bounds");
```
seando-adsk, line 599: This is the actual fix.

seando-adsk:
I accidentally created this PR to the wrong branch.

PR #2561, opened 2023-07-31
Autodesk: Fix for MSAA configuration on macOS

erikaharrison-adsk:
### Description of Change(s)

Here's the issue:

Most of our apps use an MSAA configuration, and on macOS we are getting the error:

* [MTLDebugRenderCommandEncoder setRenderPipelineState]:1598: failed assertion 'Set Render Pipeline State Validation
the sample count (4) does not match the renderPipelineState's sample count (1).

The issue is that the PSO does not have the correct (MSAA) sample count for the target.

I tracked this down to some code in pxr/imaging/hdSt/pipelineDrawBatch.cpp, in _GetDrawPipeline() where the correct state configured by a call to InitGraphicsPipelineDesc() is later overwritten. The fix here simply moves the call to InitGraphicsPipelineDesc() a few rows down, after the hardcoded initialization.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/pipelineDrawBatch.cpp
+++ b/pxr/imaging/hdSt/pipelineDrawBatch.cpp
@@ -1315,9 +1315,6 @@ _GetPTCSPipeline(
     if (pipelineInstance.IsFirstInstance()) {
         HgiGraphicsPipelineDesc pipeDesc;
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8550

PR #2562, opened 2023-07-31
Autodesk: Add normal bias support to shadow mapping

erikaharrison-adsk:
### Description of Change(s)

Add normal bias support to Shadow Mapping. There is only depth bias in the built-in USD-Hydra shadow mapping. We need to support normal bias otherwise we will have, for example, acnes on faces parallel to the light direction. Some reference: https://doc.babylonjs.com/features/featuresDeepDive/lights/shadows#normal-bias-since-32

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8551

erikaharrison-adsk:
Hi @jesschimein or is there another person I should tag? Any update on this PR? We've a shipping product that's looking to incorporate this work. Let us know if there's anything on our end we can do to help this along. Maybe we can get some feedback before addressing the new merge conflicts?

PR #2563, opened 2023-07-31
Use spirv_reflect from vulkansdk

dlarsson-adobe:
### Description of Change(s)
Uses spirv_reflect from the vulkan sdk instead of a version bundled with USD
Adds the vulkan files in hgiInterop to allow building with PXR_ENABLE_VULKAN_SUPPORT=ON
### Fixes Issue(s)
In environments where there are multiple clients of the Vulkan, having different versions of
spirv_reflect can cause issues at runtime. Using the version in the sdk will make it more likely
USD is in sync with the rest of the application. 

(Note, these tests failed with or without the patch applied)
```
    Start 430: testUsdZipFile
1/3 Test #430: testUsdZipFile ...................***Failed    0.93 sec
    Start 683: testUsdviewNavigationKeys
2/3 Test #683: testUsdviewNavigationKeys ........***Failed   10.16 sec
    Start 699: testUsdResolverExample
3/3 Test #699: testUsdResolverExample ...........***Failed    0.71 sec
```


jesschimein:
Filed as internal issue #USD-8552

davidgyu:
We think this has been addressed by ab47438

Could you verify and let us know if it looks like we need anything more to address this concern?

Thanks!

PR #2564, opened 2023-08-01
Support both OpenGL and Vulkan Hgi from single build.

marsupial:
### Description of Change(s)
Makes it possible to use hgiOpenGL or hgiVulkan from a single USD distribution, with the choice to use which happening at runtime.

### Fixes Issue(s)
Currently only a single OpenGL, Vulkan, or Metal backend is built with USD.  While the exclusionary nature of OpenGL/Metal makes sense for macOS, this is less true for Window or Linux and means an app cannot choose which to use at runtime in addition to making updating a render-delegate to be aware of hgiVulkan a bit more difficult than it needs to be.

Additionally, with [proposals such as Autodesk DirectX backend](https://github.com/autodesk-forks/USD-proposals/tree/adsk/feature/DirectX12Hgi), it makes sense to think of these interop plug-ins as a matrix where multiple backends can transfer to one-another as needed.

### DIFF
--- a/pxr/imaging/hgiInterop/CMakeLists.txt
+++ b/pxr/imaging/hgiInterop/CMakeLists.txt
@@ -26,6 +26,12 @@ else()
     list(APPEND optionalPrivateHeaders opengl.h)
 endif()
 
+if (PXR_ENABLE_VULKAN_SUPPORT)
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-8554

PR #2567, opened 2023-08-04
Remove unused `boost/call_traits.h` header

nvmkuruc:
### Description of Change(s)
`boost::call_traits` was removed from `pxr/usd/sdf/accessorHelpers.h` in #2527 but the header was not removed. This change removes the now unused header.

### Fixes Issue(s)
N/A


jesschimein:
Filed as internal issue #USD-8557

Issue #849, opened 2019-05-14
USD running as different user on Windows?


bsilvaLS:
### Description of Issue

Hello! We're having issues with permissions on Windows when working on network drives. It appears that USD is running as a different user, probably a system user.

### Steps to Reproduce
1. Using command prompt, change directory to a network drive location, where I have full-control permissions but Everyone does not.
2. Open a Python shell and create a new stage:
```
>>> from pxr import Usd, UsdGeom
>>> stage = Usd.Stage.CreateNew('HelloWorld.usda')
```

A traceback occurs:

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> pxr.Tf.ErrorException:
>         Error in 'pxrInternal_v0_19__pxrReserved__::SdfTextFileFormat::WriteToFile' at line 305 in file <...>\pxr\usd\lib\sdf\textFileFormat.cpp : 'Insufficient permissions to write to destination directory '<...>''

This will also happen with safeOutputFile (HelloWorld.usdc).

If I change the directory permissions so Everyone has full permissions, it will write the file without complaint.

This occurs when running from within Maya as well, and trying to use the AL_USDMaya export command, since it uses `stage->GetRootLayer()->Save()`

Notably, it does NOT occur when using the built-in usdExport command in Maya -- that works fine and will write the file.

Any help appreciated!

### System Information (OS, Hardware)
Windows 10.0.16299 Build 16299, x64
Intel(R) Core(TM) i9-7940X CPU

### Package Versions
USD 19.05
AL_USDMaya 0.31.1

c64kernal:
Hi @bsilvaLS, USD doens't attempt to run as a different user... though what our text file format does upon writing is that it first tries to create a temporary file, write to that temporary file completely, and if successful, moves it to the file you said to write to.  The reason we do this is so that if the save fails for any reason, it wouldn't corrupt the existing file.

My suspicion is that your user doesn't have permission to create and write to this temporary file.

Hope that helps!

bsilvaLS:
Thanks @c64kernal ! Is that temporary file written in place in the current directory, or in a temp folder elsewhere? This happens when I'm running within a folder I have full control permissions (which is the same as the destination directory listed in the error message).

And this also happens when writing a binary USD, does it write to a temporary file in the same way?

jtran56:
Filed as internal issue #USD-5287.

c64kernal:
[...]

PR #2570, opened 2023-08-07
[tf] AtomicRenameUtil - add short retry period on windows

pmolodo:
### Description of Change(s)

On windows, things like virus checkers and indexing service can often grab a handle to newly-created files; they usually release them fairly quickly, though, so just need a short retry period

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/USD/issues/2141

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -1094,14 +1094,9 @@ static int Arch_FileAccessError()
     }
 }
 
-int ArchFileAccess(const char* path, int mode)
+int ArchWindowsFileAccess(const char* path, DWORD accessMask)
 {
-    // Simple existence check is handled specially.
     std::wstring wpath{ ArchWindowsUtf8ToUtf16(path) };
-    if (mode == F_OK) {
-        return (GetFileAttributesW(wpath.c_str()) != INVALID_FILE_ATTRIBUTES)
-                ? 0 : Arch_FileAccessError();
-    }
 
     const SECURITY_INFORMATION securityInfo = OWNER_SECURITY_INFORMATION |
                                               GROUP_SECURITY_INFORMATION |
@@ -1152,7 +1147,6 @@ int ArchFileAccess(const char* path, int mode)
         mapping.GenericExecute = FILE_GENERIC_EXECUTE;
         mapping.GenericAll = FILE_ALL_ACCESS;
 
-        DWORD accessMask = ArchModeToAccess(mode);
         MapGenericMask(&accessMask, &mapping);
 
         if (AccessCheck(security,
@@ -1176,6 +1170,17 @@ int ArchFileAccess(const char* path, int mode)
     CloseHandle(token);
 
     return result ? 0 : -1;
+}
+
+int ArchFileAccess(const char* path, int mode)
+{
+    // Simple existence check is handled specially.
+    if (mode == F_OK) {
+        std::wstring wpath{ ArchWindowsUtf8ToUtf16(path) };
+        return (GetFileAttributesW(wpath.c_str()) != INVALID_FILE_ATTRIBUTES)
+                ? 0 : Arch_FileAccessError();
+    }
+    return ArchWindowsFileAccess(path, ArchModeToAccess(mode));
 }
 
 // https://msdn.microsoft.com/en-us/library/windows/hardware/ff552012.aspx
--- a/pxr/base/arch/fileSystem.h
+++ b/pxr/base/arch/fileSystem.h
@@ -135,6 +135,7 @@ ArchOpenFile(char const* fileName, char const* mode);
 #endif
 
 #if defined(ARCH_OS_WINDOWS)
+    ARCH_API int ArchWindowsFileAccess(const char* path, DWORD accessMask);
     ARCH_API int ArchFileAccess(const char* path, int mode);
 #else
 #   define ArchFileAccess(path, mode)   access(path, mode)
--- a/pxr/base/tf/CMakeLists.txt
+++ b/pxr/base/tf/CMakeLists.txt
@@ -417,6 +417,15 @@ pxr_build_test_shared_lib(TestTfRegistryFunctionPlugin
         testenv/TestTfRegistryFunctionPlugin.cpp
 )
 
+if(WIN32)
+    pxr_build_test(testAtomicRenameUtil
+        LIBRARIES
+            tf
+        CPPFILES
+            testenv/testAtomicRenameUtil.cpp
+    )
+endif()
+
 pxr_build_test(testTfCast
     LIBRARIES
         tf
@@ -489,6 +498,7 @@ pxr_build_test(testTf
 )
 
 pxr_test_scripts(
+    testenv/testTfAtomicRenameUtil.py
     testenv/testTfCrashHandler.py
     testenv/testTfFileUtils.py
     testenv/testTfMallocTagReport.py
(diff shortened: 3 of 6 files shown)


jesschimein:
Filed as internal issue #USD-8565

pmolodo:
Would love to get this merged... continues to cause problems when testing stock windows builds in our CI.

gitamohr:
Hey Paul, looking at this one today.  Oooof... this Windows behavior is horrid!  Is the lock grabbed on the temporary file that we've written to after we close it but before we rename it over?  Is there some way we can prevent that?  I guess I'm just trying to see if there's some better way to do this b/c having to sit and retry for 300 msec (!) feels icky and unreliable.

gitamohr:
Oh I see from reading the test that it's not the temp file, it's the actual output file.  Does this mean we should do this kind of retrying *always* everytime we try to create a file on Windows?

pmolodo:
> Oh I see from reading the test that it's not the temp file, it's the actual output file. Does this mean we should do this kind of retrying _always_ everytime we try to create a file on Windows?

So - I think this isn't a problem so much on initial file creation, as it is when you try to then to replace or remove a file shortly after creation.

ie, i think what happens is this:

- Some unittest creates `test.usda`
- This generally goes through ok, even without my changes
- The virus checker or indexer or whatever notices a new file was created, starts inspecting it
- The unittest then does some more stuff, saves over `test.usda` with new content
- This write fails because the virus checker still has a lock on the original file

As I've implemented, though, I'm not doing any sort of checks to see if the file already exists before applying retrying logic... so it's doing the retry loop on both the initial creation and later attempt to overwrite.

Also - when I was writing the test, I was thinking in terms of the "final" file being the problem... but I suppose the temp file could also be causing issues: ie, we create the temp file, virus checker grabs a lock on it, then when we try to do a move operation on it, and it fails.  The retry logic is on a move operation, so we don't really know if the root of the problem was the deletion of the old (temp) half of it, or the rewriting of the new (final) file half.

But to the larger point... yes, this would be a potential problem on any windows write or delete style operation, I think.  I tackled it in `atomicRenameUtil` because that's what was popping up in my tracebacks when I was doing spot checks, so practically speaking, I think it gets "most" of the use cases?

gitamohr:
Thanks for the info Paul -- are you saying that just running the shipped USD tests on Windows are failing for you like this?  We have Windows CIs and don't see these kinds of failures.  I also ran this by Mark T. at SideFX b/c I know he has a lot of Windows experience, and he says he hasn't seen behavior like this before either.  Do you have a sense of which virus checker is causing this for you?

pmolodo:
Yeah, the standard tests fail... not at a high rate, but high enough to be noticeable and annoying.  Unfortunately never dedicated time to get exact figures, but at a rough guess I'd say about 10% of the complete test runs on machines that have the virus checker enabled would have at least one test fail due to something that looked like this.

As I understand it, the only virus checker we have enabled on those test machines is Windows Defender.  Not sure if the windows indexing service is enabled or not.

pmolodo:
Also - a bit of a chagrined mea culpa - while looking at the PR again I noticed that my modified `ArchFileAccess` wasn't always returning a value!  I [pushed a fix](https://github.com/PixarAnimationStudios/OpenUSD/compare/ebdd58f506c07701eb02a29bbe6ade26a1418c51..f0ebdea77dbe755ffefab07f8ab572d0f87e9b76)... but I'm surprised that this didn't show up earlier, because we've been using this code ourselves for a while now.
My best guess is that the undefined behavior in this particular case was resulting in it always returning 0, which meant that the [file-access](https://github.com/PixarAnimationStudios/OpenUSD/blob/v23.11/pxr/base/tf/atomicRenameUtil.cpp#L138) [pre-checks](https://github.com/PixarAnimationStudios/OpenUSD/blob/v23.11/pxr/base/tf/atomicRenameUtil.cpp#L150) always passed.  Which, given [some of the issues here](https://github.com/PixarAnimationStudios/OpenUSD/issues/849#issuecomment-989098984), may not have been the worst side effect...
Still, apologies that that slipped through!

gitamohr:
Cool -- thanks for all the info!  And yeah I think we have a task to remove the file access checks on Windows too -- I know the SideFX folks delete those too.  I will check to see if we have Windows Defender on in our CI setup.  I'd love to repro this ourselves if possible.

marktucker:
The thing I keep coming back to here is the question of why this would be a "USD problem", if the explanation behind the source of the problem is correct? Why don't the unit tests of any other library have this problem? I'm fairly certain (having spent a little time ten years ago writing Windows kernel device drivers) that a properly written device driver of the sort anti-virus software uses will _not_ cause this kind of problem. They either delay the operation until they are done with the file, or they instantly release their lock on the file so the requested operation can complete. If this wasn't the case, every library and application on Windows would face this problem, but I'm not aware of any other software that implements this kind of "retry loop" when writing files...

One complicating factor might be if this only happens on network file systems, which are generally problematic on Windows. But especially so if there are multiple machines with this drive/directory mounted? Maybe another machine's antivirus is scanning the new/temp file and getting in the way?

pmolodo:
@marktucker - some very good questions!  I'm relatively new to windows development, so I thought perhaps this sort of thing was common.  My other thought was that perhaps most other companies simply don't have virus checking enabled on CI test runners, as perhaps it was just "known" to create problems.

If neither of those are true, then it does indeed make our situation a bit more singular.

I should also note that I haven't definitively placed the blame for this behavior on Virus Checking / Windows Defender - I wanted to do some A/B tests with some runners that had Defender disabled (but were otherwise identical to machines that typically display this problem), but I couldn't get support from the IT teams that handle our CI.

Here's what I know:
- these write errors seem to happen happen on most of our locally-managed test runner groups
- these test runner groups have Windows Defender enabled
- they don't happen on any our AWS runner groups
- which don't have Windows Defender
- adding these retries reduced incidence of these issues

So "Windows Defender is the culprit" is my working hypothesis, but it's certainly possible there's some other pertinent difference in the runners which is causing this difference in behavior.  If people are reluctant to implement these sort of changes without some more definitive information, I can try leaning on our IT team again to see if I can get some help for testing.

pmolodo:
@marktucker - Haven't had a chance to do more digging on our end, but while doing an unrelated task, I stumbled across this bit of code in CMake, which at least indicates we're not the only people having to retry when using `MoveFileEx`:

https://github.com/Kitware/CMake/blob/master/Source/cmSystemTools.cxx#L1234:
>      Try multiple times since we may be racing against another process
>      creating/opening the destination file just before our MoveFileEx.  */

https://github.com/Kitware/CMake/blob/master/Source/cmSystemTools.cxx#L1252

>     // Try again only if failure was due to access/sharing permissions.
>     // Most often ERROR_ACCESS_DENIED (a.k.a. I/O error) for a directory, and
>     // ERROR_SHARING_VIOLATION for a file, are caused by one of the following:
>     // 1) Anti-Virus Software
>     // 2) Windows Search Indexer
>     // 3) Windows Explorer has an associated directory already opened.

pmolodo:
That made me curious, and I did some quick searching on github - and found several reputable software packages that are doing similar retry loops.

ie, here's one in golang:

https://github.com/golang/go/blob/master/src/testing/testing.go#L1254

> // removeAll is like os.RemoveAll, but retries Windows "Access is denied."
> // errors up to an arbitrary timeout.

...and another in elastic:

[Increase the time between RemovePath retries by AndersonQ · Pull Request #3349 · elastic/elastic-agent](https://github.com/elastic/elastic-agent/pull/3349)

marktucker:
Thanks @pmolodo. Sorry for the delayed response, but all of that is pretty compelling. Sometimes Microsoft does not make it easy to defend (pun intended) them :)

PR #2573, opened 2023-08-08
[docs] raise cmake error if converting python docstrings fails

pmolodo:
### Description of Change(s)

Currently, if there is an error during python docstring generation, CMake will continue without erroring, and it is very easy to miss that there was an error.

Since python docstring generation is off by default, if it's on, the user likely wants to be know if there was a problem, so this change makes CMake error if docstring generation fails.

**NOTE:**


jesschimein:
Filed as internal issue #USD-8568

PR #2577, opened 2023-08-08
[docs] py docstrings: fix docstring injection for pxr.GeomUtil

pmolodo:
### Description of Change(s)

For some reason, pxr.GeomUtil had non-standard code for handling it's python module setup; this code wasn't properly ported to python 3, as it does `import __DOC` instead of `from . import __DOC`, as relative imports are no longer default

However, replacing the non-standard code with the normal Tf.PreparePythonModule() works fine, so I just went with that approach

**NOTE:**

### DIFF
--- a/pxr/imaging/geomUtil/__init__.py
+++ b/pxr/imaging/geomUtil/__init__.py
@@ -25,15 +25,7 @@
 Utilities to help image common geometry.
 """
 
-from . import _geomUtil
 from pxr import Tf
-Tf.PrepareModule(_geomUtil, locals())
-del _geomUtil, Tf
-
-try:
-    import __DOC
-    __DOC.Execute(locals())
-    del __DOC
-except Exception:
-    pass
+Tf.PreparePythonModule()
+del Tf
 


mattyjams:
Hey @pmolodo! :)

I hadn't noticed how this factored into docs generation, but I happened to make the same change in #2456 since I found it actually made a functional difference in some builds where the module might fail to import. Not sure if that was something you were seeing as well?

pmolodo:
> Hey @pmolodo! :)
> 
> I hadn't noticed how this factored into docs generation, but I happened to make the same change in #2456 since I found it actually made a functional difference in some builds where the module might fail to import. Not sure if that was something you were seeing as well?

Hey Matt! Interesting... no, we didn't see a functional difference.  Was just an attempt to get the docstrings working on our part.
I'll go ahead an close this PR in favor of yours, though...

pmolodo:
Closing in favor of https://github.com/PixarAnimationStudios/OpenUSD/pull/2456

PR #2580, opened 2023-08-08
[docs] py docstrings: get rid of, ie "USD_API" type name prefixes

pmolodo:
### Description of Change(s)

The python function signatures still had the X_API modifiers in them.  Since this is a C++ only concept, strip them when going from C++ to python.

**NOTE:**


jesschimein:
Filed as internal issue #USD-8574

PR #2581, opened 2023-08-08
[docs] py docstrings: better conversion of C++ types to python

pmolodo:
### Description of Change(s)

Numerous fixes to translation of C++ types to python types

**NOTE:**

### DIFF
--- a/docs/python/doxygenlib/cdWriterDocstring.py
+++ b/docs/python/doxygenlib/cdWriterDocstring.py
@@ -103,14 +103,14 @@ def getAllPxrModules() -> Dict[str, types.ModuleType]:
                         pxrModuleName = basename
             if pxrModuleName:
                 modules[pxrModuleName] = importPxrModule(pxrModuleName)
-    # return a sorted dict
-    return {k: modules[k] for k in sorted(modules)}
+    return modules
 
 pxrModules = getAllPxrModules()
+pxrModuleNames = sorted(pxrModules)
 
 # we use the reversed sorted order of the module names, because this ensures that longer names (ie, UsdGeom) come
 # before shorter ones (Usd), so the more exact matches are preferred
-PXR_MODULES_OR_JOINED = "|".join(re.escape(m) for m in reversed(pxrModules))
+PXR_MODULES_OR_JOINED = "|".join(re.escape(m) for m in reversed(pxrModuleNames))
 PXR_MODULE_PREFIX = re.compile(rf"^\s*(?P<module>{PXR_MODULES_OR_JOINED})(?P<suffix>[A-Za-z0-9_]+)")
 
 
--- a/docs/python/doxygenlib/cdWriterDocstring.py
+++ b/docs/python/doxygenlib/cdWriterDocstring.py
@@ -50,7 +50,7 @@ from .cdUtils import *
 API_RE_FIND = re.compile(r"""\b[A-Z]+_API(?:\s+|$)""")
 API_RE_REPLACE = ""
 
-WORDBREAK_RE = re.compile(r"\b")
+NONWORD_RE = re.compile(r"(\W+)")
 
 MULTIPLE_WHITESPACE_RE = re.compile(r'\s+')
 MULTIPLE_WHITESPACE_REPLACE = ' '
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8575

docs/python/doxygenlib/cdWriterDocstring.py
```diff
@@ -35,13 +35,82 @@
...
+            if pxrModuleName:
+                modules[pxrModuleName] = importPxrModule(pxrModuleName)
+    # return a sorted dict
+    return {k: modules[k] for k in sorted(modules)}
+
+pxrModules = getAllPxrModules()
+
+# we use the reversed sorted order of the module names, because this ensures that longer names (ie, UsdGeom) come
+# before shorter ones (Usd), so the more exact matches are preferred
+PXR_MODULES_OR_JOINED = "|".join(re.escape(m) for m in reversed(pxrModules))
```
dsyu-pixar: @pmolodo in verifying this change, I'm running into an issue where reversed() isn't supported for dictionaries in Python 3.6 & 3.7 (but is supported in 3.8+).  Additionally, it sounds like dictionaries didn't really support strict ordering at all until Python 3.6 (see https://groups.google.com/g/dev-python/c/76kIoAEKlew). 

Would there be a way to make this work for earlier versions of Python 3?

dsyu-pixar: Alternatively, it'd be nice to have some sort of "requires Python 3.8 or better" check or error message. The current error you get on 3.6 is something along the lines of "Error: 'dict' object is not reversible"

pmolodo:
rebased to fix merge conflict...

pmolodo:
Pushed changes to make it not rely on dicts being ordered or reversable:

https://github.com/PixarAnimationStudios/OpenUSD/pull/2581/commits/f39826a2915b49829d56985692685d25f05a59b2

dsyu-pixar:
Hi @pmolodo thanks for the update! Unfortunately, I now get the following issue (when running with Python 3.6 -- this does not occur when testing with Python 3.9):

```
  File "/home/dsyu/trees/dev-usd/docs/python/doxygenlib/cdWriterDocstring.py", line 552, in __convertTypeName
    tokens = [self.__convertTypeNameToken(x) for x in WORDBREAK_RE.split(ret) if x]
ValueError: split() requires a non-empty pattern match.
```

I think this might be due to an re.split() issue in 3.6, possibly fixed in 3.7 (see https://bugs.python.org/issue43222). 

I realize trying to craft things to work in older versions of Python 3 (that have known issues) can turn into an exercise in futility, so I'm fine with making a min requirement of Python 3.8 or 3.9 -- I believe the VFX 2022 reference platform recommends Python 3.9, and we're trying to test/build with that in mind anyway. What do you think? Maybe just adding a python_version() check somewhere early in convertDoxygen.py would be enough to be transparent enough to anyone using older versions?

pmolodo:
I pushed a version which fixes that `re.split` issue on python-3.6.  I tried to test with python-3.6 myself, but ran into another python-3.6 bug related to `locale.getpreferredencoding()` on the docker image we use for builds (that occurs earlier in the build process), so I can't confirm this is the last python-3.6 related issue.

If another one pops up, I say we just state that you need python-3.7+ to run this script...

pmolodo:
Also, fyi - I confirmed that this regexp should give the same results once split with this script:

```python
import inspect
import os
import re

WORDBREAK_RE = re.compile(r"\b")
NONWORD_RE = re.compile(r"(\W+)")

THIS_FILE = os.path.abspath(inspect.getsourcefile(lambda: None) or __file__)
THIS_DIR = os.path.dirname(THIS_FILE)

prim_cpp_path = os.path.join(THIS_DIR, "pxr", "usd", "usd", "prim.cpp")

with open(prim_cpp_path, "r", encoding="utf8") as reader:
    test_cpp_text = reader.read()

def tokenize_1(text: str):
    return [x for x in WORDBREAK_RE.split(text) if x]

def tokenize_2(text: str):
    return [x for x in NONWORD_RE.split(text) if x]

split1 = tokenize_1(test_cpp_text)
split2 = tokenize_2(test_cpp_text)

if split1 == split2:
    print("yay!")
else:
    raise Exception("oh noes")
```

PR #2585, opened 2023-08-12
Replace `BOOST_PP_CAT` with `TF_PP_CAT`

nvmkuruc:
### Description of Change(s)
`BOOST_PP_CAT` can be replaced with `TF_PP_CAT` in many places, reducing the dependency of OpenUSD on boost.

### Fixes Issue(s)
- 


jesschimein:
Filed as internal issue #USD-8582

PR #2590, opened 2023-08-14
Implement `sdf/types.h` internal unit macros with variadics instead of `BOOST_PP_LIST`

nvmkuruc:
### Description of Change(s)
By changing `_SDF_UNITS` to a variadic tuple instead of a `BOOST_PP_LIST`, several internal macros in `pxr/usd/sdf/types.h` can be implemented in terms of existing `tf/preprocessorUtilsLite.h` macros.  Removing these macros reduces the project's dependency on several boost preprocessor headers (like `boost/preprocessor/list`).

This change additionally converts some size macros to `constexpr` variables and C-style arrays to `std::array`.  The `constexpr` changes avoids expanding the macro in multiple places and localizes error messages.  `std::array` is preferred per the C++ Core Guidelines (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon1-prefer-using-stl-array-or-vector-instead-of-a-c-array) and its usage simplified the migration.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8587

PR #2592, opened 2023-08-15
Make `Sdf_LayerRegistry::FindByIdentifier` private

nvmkuruc:
### Description of Change(s)
Layer identifiers when multiple contexts are bound are not guaranteed to be unique. As such, `FindByIdentifier` may not return the intended result.  It's only used once externally (in a `TF_VERIFY` statement).  This updates that statement to use `Find` with an explicitly resolved path and makes `FindByIdentifier` private.

Simplifying the public API for `Sdf_LayerRegistry` will help the boost removal effort by scoping what kind of queries the replacement data structure needs to support.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8591

PR #2593, opened 2023-08-15
Fix Windows dll-interface warnings (C4251) for clients of `TfSingleton`

andrewkaufman:
### Description of Change(s)

The `std::atomic` is a private implementation detail and is inaccessible by clients of `TfSingleton`.

This pragma was originally included in `TfSingleton` on a private `std::mutex`, but was removed when it was refactored in 7899a94edb41a7b4d08d065a6f971f048bf1597c. That commit also removed `ARCH_PRAGMA_UNDEFINED_VAR_TEMPLATE` from `GetInstance()`. However, that one was added back in 4cec3635bc4b11e6ba5ce3d372ff6e4703c06391, while `ARCH_PRAGMA_NEEDS_EXPORT_INTERFACE` was never re-introduced.

### DIFF
--- a/pxr/base/tf/singleton.h
+++ b/pxr/base/tf/singleton.h
@@ -195,7 +195,12 @@ public:
 private:
     static T *_CreateInstance(std::atomic<T *> &instance);
     
+    // Suppress dll-interface warnings from MSVC; _instance
+    // is an inaccessible implementation detail of TfSingleton.
+    ARCH_PRAGMA_PUSH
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8592

FlorianZ:
As @nvmkuruc pointed out, I *think* this would be subsumed by #2678. Please, correct me if I am wrong!

PR #2595, opened 2023-08-16
[tf] fix for python dll loading order

pmolodo:
### Description of Change(s)

Tf uses `WindowsImportWrapper` to add directories on the `PATH` to the list of allowable directories to find .dll files, so python modules relying on finding .dlls on the `PATH` will continue to work.

However, it seems that `os.add_dll_directory` works in a "most recently added wins" manner, so we need to reverse the order that we call it for entries on the `PATH`.

### Fixes Issue(s)
- Errors loading incorrect dlls if order of entries on PATH matters

### DIFF
--- a/pxr/base/tf/CMakeLists.txt
+++ b/pxr/base/tf/CMakeLists.txt
@@ -8,6 +8,101 @@ if(WIN32)
     set(WINLIBS Shlwapi.lib)
 endif()
 
+# Add a test that WindowsImportWrapper adds entries on the PATH in the correct order, to ensure
+# that the correct dlls needed by python modules are loaded
+function(add_py_dll_link_test)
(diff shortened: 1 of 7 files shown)


jesschimein:
Filed as internal issue #USD-8593

pmolodo:
Hi - if this PR seems a bit bulky, the vast majority is an added test for correct PATH ordering, which was a bit tricky to get right.

The functional change is a single line (really, a single additional call to `reversed()`):

https://github.com/PixarAnimationStudios/OpenUSD/pull/2595/files#diff-78e4c3d1f9dc10a8e04395293a5f6fb8d46d3ac9d7123a0e1b3ee7867e7a003eR57

I'm ok with dropping the test if the test seems too bulky...

PR #2596, opened 2023-08-16
Cylinders and Capsules with two radii

eoineoineoin:
### Description of Change(s)

Adds a new schema version of Cylinder and Capsule, which accept different radii for each end. (Some discussion at https://groups.google.com/g/usd-interest/c/zC7wYK7wN-Y/m/PZoEiEkXAgAJ) This allows construction of tapered capsules and conic sections.

Uses schema versioning to add a Cylinder_1 as well as a Capsule_1; as such, most of the code in this PR is auto-generated. However, in addition to adding these types, adds some additional fixes:

1. Adds an "includeSchemaFamily" option to the usdImaging pluginInfo.json, to allow one adapter to be used to render multiple different schema versions (25889c439cec28709a4a19c0fe3d8c7ebbe095ca)
2. Default values used by implicit surface generator were inconsistent with the values specified in the schema (62acf3867437a89709fadeb8f4379e67a3d0c396)
3. Capsule mesh generator now generates a continuous surface between the end caps and the conical section. (1370773fd065ae53790710df6d321211403ba722)

Along the way, I split the HdCapsuleSchema and HdCylinderSchema into two classes (HdCapsule_1Schema and HdCylinder_1Schema) - though the two schema versions are very similar, it felt like there was limited overlap between the "version 0" and "version 1" to enable effective code re-use, particularly in the `Builder` interface. Am aware the schema versioning is quite new though, so expect some teething issues like this - open to ideas on how this could be organized better.

### Fixes Issue(s)
-

Am seeing some failures on my machine; comparing reference images, they appear unrelated to these changes; some line-drawing aliasing appears different as well as some textures have failed to load.


jesschimein:
Filed as internal issue #USD-8594

PR #2604, opened 2023-08-17
[sdrOsl] Support building plugin against OSL builds that use Imath

nrusch:
### Description of Change(s)

Simple change to fix the build of the `sdrOsl` plugin when using an OSL build that depends on `Imath` rather than `IlmBase`.

The changes use the same pattern as the `hioOiio` plugin build.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8600

PR #2607, opened 2023-08-20
VisualStudio finding failure bugfix

AndrewChan2022:
### Description of Change(s)

Bug fix for VisualStudio finding failure of build script.

### Fixes Issue(s)
- VisualStudio finding fail if user not set environment variable, and most users will not set it by default.
- When VisualStudio finding fail but clang or g++ installed, CMake use MSBuild but build option is -j cpu_count, and correct option should be /M:cpu_count

431 - testUsdZipFile (Failed)
677 - testUsdviewNavigationKeys (Failed)
693 - testUsdResolverExample (Failed)

I only change build script, not change the source code.

the 3 cases also failed before I change the script, with manually specified visual studio generator.

So the failure should unrelated to these changes.


meshula:
Hi Andrew, this PR is against release, rather than dev, that's why it's picking up so many other changes. Maybe try opening a new PR against dev?

AndrewChan2022:
Ok，thanks

PR #2608, opened 2023-08-21
fix VisualStudio find fail

AndrewChan2022:
### Description of Change(s)

Visual Studio finding will fail if user not set environment variable, which happened to most users.

And if clang or g++ was installed, build will continue with MSBuild but with -j cpu_count, not /M:cpu_cout. This makes users confused and don't known how to build.

This fix use CMake to find visual studio,  cmake use sophisticated approach to find visual studio.

A python implement of this sophisticated finding algorithm also provided on my github [PyVisualStudioSetupConfiguration](https://github.com/AndrewChan2022/PyVisualStudioSetupConfiguration), but I think it is unnecessary, cmake should be good enough.

### Fixes Issue(s)
- Visual Studio finding fails

431 - testUsdZipFile (Failed)
677 - testUsdviewNavigationKeys (Failed)
693 - testUsdResolverExample (Failed)

the 3 cases also failed before my change, should unrelated to this change.

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -116,7 +116,8 @@ def GetXcodeDeveloperDirectory():
 def GetVisualStudioCompilerAndVersion():
     """Returns a tuple containing the path to the Visual Studio compiler
     and a tuple for its version, e.g. (14, 0). If the compiler is not found
-    or version number cannot be determined, returns None."""
+    or version number cannot be determined, returns None.
+    NOTE: path possibly to be None, please don't depend on it"""
     if not Windows():
         return None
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8603

sunyab:
@AndrewChan2022 Thank you for this PR. However, I believe the `VisualStudioVersion` env var is set automatically when using the Visual Studio developer command prompt as noted in our [README.md](https://github.com/PixarAnimationStudios/OpenUSD#windows). Alternatively, Visual Studio provides a vcvarsall.bat file for setting this environment up manually -- see [here](https://learn.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=msvc-170).

Are you using the VS command prompt? Or are you building in a different environment?

AndrewChan2022:
Sorry, I didn't read the document carefully so I missed the command.

And it is fine after running the command.

sunyab:
Good to hear! I'm going to close this PR out then, thanks!

PR #2609, opened 2023-08-21
Autodesk: Fix C4273 warning from Sdf_PathExpressionEvalBase

erikaharrison-adsk:
### Description of Change(s)

The declaration of the function Sdf_MakePathExpressionEvalImpl has SDF_API prefix, but when it is used as a friend function in Sdf_PathExpressionEvalBase, the SDF_APL prefix is missing. This will cause C4273 warning if this header is included in an external project. This small change fixes the warning.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8606

sunyab:
I'm going to close this PR out since the underlying issue was addressed by 72452dc. Thanks!

Issue #2601, opened 2023-08-17
[tests] `testUsdImagingGLCards` EXR texture incompatible with OpenImageIO 2.2.4+


nrusch:
### Description of Issue

OpenImageIO 2.2.4+ no longer automatically renames the `worldToNDC` metadatum to `worldtoscreen` on file read, or vice versa on file write.

One of the textures used in `testUsdImagingGLCards` for testing the `fromTexture` mode for `model:cardGeometry` is an EXR that only contains the `worldToNDC` metadata key, and the metadata indicates that it was generated using OIIO 2.1:

**exrheader output:**
```
file format version: 2, flags 0x0
Exif:ImageHistory (type string): "oiiotool WoodPlanksDonut.png -a --attrib:type=matrix worldtoscreen 1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 1,1,0,0,-1,1,0,0,0,0,1,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 0,1,0,0,0,0,1,0,1,0,0,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 0,1,0,0,1,0,1,0,1,0,-1,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 1,1,0,0,-1,1,0,0,0,0,1,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 0,1,1,0,0,-1,1,0,1,0,0,0,0,0,0,1 -o WoodPlanksDonut.exr
oiiotool WoodPlanksDonut.exr -a --attrib:type=matrix worldtoscreen 0,1,-1,0,0,1,1,0,1,0,0,0,0,0,0,1 -o WoodPlanksDonut.exr
[...]

PR #2610, opened 2023-08-21
Autodesk: Upgrade OpenImageIO to 2.3.15 + update unit test

erikaharrison-adsk:
### Description of Change(s)

Upgrade OpenImageIO to 2.3.15 and need update one exr image to make sure test case testUsdImgingGLCards passed.

Background: After OIIO upgrade start from 2.2.4, the attribute "worldtoNDC" is not mapped with "worldtoscreen" attribute in OIIO. The test case testUsdImgingGLCards needs revised exr file with legacy attribute name "worldtoscreen" that is supported in HioOiio plugin.

Note: The unit test fix is also part of https://github.com/PixarAnimationStudios/OpenUSD/pull/2603 but we need it for the OIIO here too.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2601 


jesschimein:
Filed as internal issue #USD-8607

build_scripts/build_usd.py
```diff
@@ -1264,11 +1264,7 @@ def InstallOpenVDB(context, force, buildArgs):
 ############################################################
 # OpenImageIO
 
-if MacOS():
```
tallytalwar, line 1267: Hi @erikaharrison-adsk 

Noting that in the current dev branch and in upcoming 23.11 release, we have already unified this and moved OIIO to 2.3.21. https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/build_scripts/build_usd.py#L1247

tallytalwar (commented):
Hi @erikaharrison-adsk 

Since we are already on page to update our OIIO dependency to 2.3.21, I will go and close this PR. Thanks a bunch.

erikaharrison-adsk:
Thanks so much! Looks like your changes have also addressed the unit test impact. Thank you for that too!

PR #2611, opened 2023-08-21
Autodesk: Fix C3861 error for std::sort in base/tf

erikaharrison-adsk:
### Description of Change(s)

Adding include to address error C3861: 'sort': identifier not found.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8608

PR #2614, opened 2023-08-22
APPLE: OIT issue fix for MaterialX

slingthor:
### Description of Change(s)
- Explicit returns for discards for Metal shading language version less than 2_4
- Parts of our earlier MaterialX PR which couldn't be included in last release are explicitly put out in PR, for discussion and hopeful landing in next release

### Fixes Issue(s)
- OIT buffer invalidly written into when the intention was to stop shader execution due to discard
- When one such asset is included, especially with MaterialX, the whole frame buffer becomes noisy and glitchy

### DIFF
--- a/pxr/imaging/hdSt/shaders/mesh.glslfx
+++ b/pxr/imaging/hdSt/shaders/mesh.glslfx
@@ -1510,6 +1510,7 @@ void main(void)
 #ifdef HD_MATERIAL_TAG_MASKED   
     if (ShouldDiscardByAlpha(color)) {
         discard;
+        return;
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-8614

dgovil:
Just leaving a note here for others who hit it, that this PR fixes a lot of the corruption issues that people will see using MaterialX in usdview, where you'll see tiles of images just completely disappearing and flickering inter-frame.

Without this PR, MaterialX on usdview becomes very unusable on a Mac.

ashwinbhat:
Thanks @slingthor it works well!!

PR #2616, opened 2023-08-23
Autodesk: Remove include of "dispatcher.h" which introduces tbb headers

erikaharrison-adsk:
### Description of Change(s)

Currently, TBB header is directly included in dispatcher.h. And when someone includes the header file colorCorrectionTask.h or taskController.h in their application, dispatcher.h is included in colorCorrectionTask.h and the TBB headers will be leaked to end users causing some compatibility issues. 

In this commit, the unnecessary dispatcher.h included in colorCorrectionTask.h is removed to reduce the risk of such problems.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8616

PR #2620, opened 2023-08-24
Apple: Make C++ CLI apps linkable with python plugins

slingthor:
### Description of Change(s)
- Expose the ability to allow external CLIs to link with Python


meshula:
Hi Thor, this is great to see. I have been wrestling with this issue. Is it not true that not only do the applications, such as `usdcat` need the extra linkage you are introducing, but also, we must do the same for all the test programs? I was able to get all the USD test programs running inside Xcode, but I had to make a similar patch to what you have done, to each test.

jesschimein:
Filed as internal issue #USD-8623

PR #2623, opened 2023-08-24
Apple: Allow toggling visualization

slingthor:
### Description of Change(s)
- Storm applies a "visualization" to some passes so they appear nicer in usdview, some users prefer not to have that option on. This is a contribution that allows us to toggle that easily on or off
### Fixes Issue(s)
- Can't toggle visualization on or off


jesschimein:
Filed as internal issue #USD-8626

PR #2625, opened 2023-08-25
Replace deprecated QFontMetrics.width with horizontalAdvance 

dgovil:
### Description of Change(s)

usdview was using a deprecated `width` method on `QFontMetrics` that was finally dropped in Qt6.
This PR (thanks to John Hood for the fix) replaces it with  `horizontalAdvance`

This method has been available since Qt 5.11 and is the recommended replacement per https://doc.qt.io/qt-5/qfontmetrics-obsolete.html#width

Since this encompasses every version of the VFX Platform setup, I haven't tried to maintain compatibility behind 5.11

### Fixes Issue(s)
- Allows python auto-complete again with Qt 6


jesschimein:
Filed as internal issue #USD-8628

PR #2630, opened 2023-08-28
extras: make the build of extras/imaging/examples conditional on PXR_BUILD_EXAMPLES

mattyjams:
The `extras/imaging/examples` directory currently contains `hdTiny` and `hdui`, and since `hdui` (intentionally) does not build as part of the open source distribution, this makes it so that specifying `PXR_BUILD_EXAMPLES=OFF` prevents `hdTiny` from being built.

This mimics the pattern already in use in `extras/usd/CMakeLists.txt`.


jesschimein:
Filed as internal issue #USD-8633

PR #2631, opened 2023-08-29
fix: fixed potential overflow when compressing integer

sssooonnnggg:
### Description of Change(s)
When compressing integer, capacity passed to LZ4 is incorrect, because we already passed `compressed + 1` as output ptr, the capacity parameter should be `LZ4_compressBound` instead of `LZ4_compressBound + 1`.

image

FYI, the caller in `crateFile.cpp` only allocate a buffer with `compressBound + 1` size :

image

Then inside the callee, we passed LZ4 an incorrect capacity larger than actual, which causes undefined behaviour when serializing usdc files

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8634

gitamohr:
Nice catch, thanks!

sssooonnnggg:
@gitamohr Hi, can this PR be merged? We are currently using a custom repo with the fix, and cannot use the upstream code.😄

spiffmon:
It has been committed internally to Pixar, and should be flowing out to
GitHub dev branch shortly!

PR #2517, opened 2023-06-30
Autodesk: Fix for Full debug USD with OpenImageIO fails

erikaharrison-adsk:
### Description of Change(s)

Fixes the following build error on Debug config:

-- Could NOT find Imath (missing: Imath_DIR)
-- Imath not found. Looking for OpenEXR instead.
CMake Error at C:/Program Files/CMake/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenEXR (missing: OPENEXR_LIBRARIES) (found version "2.5.2")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindOpenEXR.cmake:94 (find_package_handle_standard_args)
  cmake/defaults/Packages.cmake:342 (find_package)
  CMakeLists.txt:23 (include)

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2516

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

Issue #2516, opened 2023-06-29
Full debug USD with OpenImageIO fails


ashwinbhat:
### Description of Issue
Building a full debug USD with OpenImageIO is broken due to dll/lib missing `_d` qualifier.
### Steps to Reproduce
1. run usd build script with options
  Building                      Shared libraries
    Variant                     Debug
    Target
    Imaging                     On
      Ptex support:             Off
      OpenVDB support:          Off
      OpenImageIO support:      On
      OpenColorIO support:      Off
      PRMan support:            Off
    UsdImaging                  On
      usdview:                  On
    Python support              On
      Python Debug:             Off
      Python 3:                 On
      Python docs:              Off
    Documentation               Off
    Tests                       On
    Examples                    On
    Tutorials                   On
    Tools                       On
    Alembic Plugin              Off
      HDF5 support:             Off
    Draco Plugin                Off
    MaterialX Plugin            On

### System Information (OS, Hardware)
Windows 11
### Build Flags
--tests

ashwinbhat:
Will post a PR for this soon

jesschimein:
Filed as internal issue #USD-8467

PR #2633, opened 2023-08-30
Autodesk: Use OIIO and OpenEXR imported targets

erikaharrison-adsk:
### Description of Change(s)

Use imported targets instead of FindXXX.cmake.

This replaces the approach previously attempted with https://github.com/PixarAnimationStudios/OpenUSD/pull/2517 and related to https://github.com/PixarAnimationStudios/OpenUSD/pull/2079

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2516 


jesschimein:
Filed as internal issue #USD-8637

cmake/defaults/Packages.cmake
```diff
@@ -211,8 +211,16 @@ if (PXR_BUILD_IMAGING)
         find_package(OpenImageIO REQUIRED)
         set(OIIO_VERSION "${OpenImageIO_VERSION_MAJOR}.${OpenImageIO_VERSION_MINOR}.${OpenImageIO_VERSION_PATCH}")
         add_definitions(-DPXR_OIIO_PLUGIN_ENABLED)
+        if (CMAKE_BUILD_TYPE)
+            string(TOUPPER "${CMAKE_BUILD_TYPE}" UPPER_BUILD_TYPE)
```
meshula: This is a global setting, so wouldn't it make sense to do it more globally? I don't know if anything else references the CMAKE_BUILD_TYPE, but it might be weird for it to conditionally switch in the middle of the build configuration.

meshula: I'm looking at issues at cmake ~ e.g. https://gitlab.kitware.com/cmake/cmake/-/issues/22591 ~ and it leaves me a bit confused. It seems like things are case sensitive, but should match e.g. Release or Debug, etc. Is the problem that OIIO is expecting RELEASE and DEBUG, etc? If so, perhaps there's an issue to file with OIIO?

zhangha182: We're not changing `CMAKE_BUILD_TYPE` but use its value with upper case to find the proper `IMPORTED_LOCATION_XXX ` property of OIIO.

[cmake-build-type](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html#cmake-build-type) says that its typical values include Debug, Release, RelWithDebInfo and MinSizeRel, but custom build types can also be defined, and maybe treated case-sensitively or case-insensitively depending on the situation. But property IMPORTED_LOCATION_XXX is always in upper case, so we need to use upper case to find the property defined when creating OIIO targets.

cmake/defaults/Packages.cmake
```diff
@@ -209,9 +209,20 @@ if (PXR_BUILD_IMAGING)
...
+                string(TOUPPER "${CMAKE_BUILD_TYPE}" UPPER_BUILD_TYPE)
+                get_target_property(OIIO_idiff_BINARY OpenImageIO::idiff IMPORTED_LOCATION_${UPPER_BUILD_TYPE})
+            else()
+                get_target_property(OIIO_idiff_BINARY OpenImageIO::idiff IMPORTED_LOCATION)
+            endif()
+        endif()
         if (OIIO_idiff_BINARY)
             set(IMAGE_DIFF_TOOL ${OIIO_idiff_BINARY} CACHE STRING "Uses idiff for image diffing")
+        elseif (PXR_BUILD_TESTS)
+            message(FATAL_ERROR "Unable to find idiff for image diffing tests")
```
meshula: Sorry for a slow response on this issue. If one builds imaging, builds the OIIO plugin, does wish to build tests, and hasn't got idiff (perhaps one did not build the binary tools, just the library), then cmake will fatal ~ wouldn't it make more sense to warn, not fatal that idiff based tests won't succeed? I think it's not sufficient that the idiff binary is found here, it also has to be true that the idiff binary is in the executable environment when the tests are run, and the cmake check isn't sufficient for that.

zhangha182: Thanks for the comment and I'll make necessary changes.

Issue #2636, opened 2023-09-01
UsdUtils.ComputeAllDependencies does not catch missing thumbnails in USDZ files


dgovil:
### Description of Issue

UsdUtils.ComputeAllDependencies appears to work differently for USDZ and USD files in that the checker seems to miss that the assetInfo dictionary is pointing to an invalid asset path when the file is a USDZ, but It works correctly when you unzip the same file

This is with USD 23.8.

The attached script will run over the two files and print when it misses the check

### Steps to Reproduce
1. Download [Test.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/12501259/Test.zip)
2. Run the included Python script

dgovil:
Ah this seems to be because `UsdUtils_FileAnalyzer` has this bit to mention. 

```
// If the newly opened layer is a package, it we do not need to traverse
// into it as the entire package will be included as a dependency.
if (_layer->GetFileFormat()->IsPackage()) {
        return;
}
```

So at least the cause is clear, though I think maybe the assumption in the code isn't accurate, because a USDZ may have external references that are unresolvable (per our previous discussions)

I'll submit a PR with that removed. I don't believe there are any side effects to the removal in my testing

jesschimein:
Filed as internal issue #USD-8640

PR #2637, opened 2023-09-01
Remove assumption in UsdUtils_FileAnalyzer that USDZ files are fully self contained

dgovil:
### Description of Change(s)

`UsdUtils_FileAnalyzer` assumed that USDZ files are fully self contained.
I'm removing this assumption because:

1. According to prior discussion, USDZ files *may perhaps* refer to external files,  even if that's not ideal
2. This was skipping validation in files that did have missing references, thereby not catching incorrect files

### Fixes Issue(s)
[2636](https://github.com/PixarAnimationStudios/OpenUSD/issues/2636)

### DIFF
--- a/pxr/usd/usdUtils/assetLocalization.cpp
+++ b/pxr/usd/usdUtils/assetLocalization.cpp
@@ -80,12 +80,6 @@ UsdUtils_FileAnalyzer::UsdUtils_FileAnalyzer(const std::string &referencePath,
             return;
         }
 
-        // If the newly opened layer is a package, it we do not need to traverse
-        // into it as the entire package will be included as a dependency.
-        if (_layer->GetFileFormat()->IsPackage()) {
-            return;
-        }
-
         _AnalyzeDependencies();
     }
 


meshula:
Irrespective of point 1, 

> USDZ files may perhaps refer to external files

the test you removed is making a pretty big assumption that the referenced package itself is valid. USDZ files do have the requirement that they are reproducibly resolvable, so removing the "early out" from the test should be both safe to do, and increase confidence about the overall analysis. My follow up question would be whether the analyzer checks for the "anchored" restriction, or if there might be further work required?

> For Reproducible Results, Encapsulate Using Anchored Asset Paths 
> ----------------------------------------------------------------
> To insure uniform consumption of assets shipped to clients via usdz, we feel it
> is prudent to curtail the power of USD's asset resolution system, so that the
> asset references within a package resolve uniformly on any consuming system,
> regardless of how that system's USD ArResolver is configured. Rather than make
> restrictions within the FileFormat itself, we propose that such "encapsulation"
> of packages be achieved by restricting the *content* to only use **anchored**
> paths (paths that begin with "./" or "../", both of which are interpreted in the
> virtual filesystem described by the package's internal layout.

dgovil:
I'm not sure on that front. I'd have to go code spelunking, but at a cursory glance I don't believe it does.
However there is `ARKitPackageEncapsulationChecker` at the UseChecker level that does check for encapsulation, but only for assets that actually resolve which was why nothing was flagging it.

jesschimein:
Filed as internal issue #USD-8641

meshula:
Reopened, as the noted commit only partially addressed the PR.

dgovil:
@meshula It looks like this was fixed somewhere along the way in the last few USD versions, even though I can't find a commit that intentionally did so. I'll (re)close this one out.

PR #2641, opened 2023-09-04
Fix typo in hgiInterop CMakeLists.txt

pablode:
### Description of Change(s)
The line was copied over from the metal if-branch, but the file extension was not adjusted.


jesschimein:
Filed as internal issue #USD-8645

PR #2644, opened 2023-09-05
Apple: Mesh shaders - stubs and tokens

slingthor:
### Description of Change(s)
Add stubs and tokens that will later allow for mesh shading in Storm


jesschimein:
Filed as internal issue #USD-8650

Issue #2247, opened 2023-02-07
`TF_PP_CAT` can be used to make some files not depend on `boost/preprocessor`


nvmkuruc:
### Description of Issue

`pxr/usd/pcp/diagnostic.h` and `pxr/base/tf/pyModule.h` bring in the `boost/preprocessor` library solely for `BOOST_PP_CAT`.

Using `TF_PP_CAT` in these cases reduces the dependencies on `boost/preprocessor`.

### Steps to Reproduce
N/A

### System Information (OS, Hardware)
N/A

### Package Versions
N/A

### Build Flags
N/A

nvmkuruc:
I have a fix.

tallytalwar:
Filed as internal issue #USD-7981

nvmkuruc:
This was addressed by #2645 and #2249

PR #2645, opened 2023-09-05
Replace `BOOST_PP_CAT` with `TF_PP_CAT` in `tf/staticTokens.h`

nvmkuruc:
### Description of Change(s)

#2585 removed the header include of `boost/preprocessor/cat.hpp` but didn't remove the usage of the `BOOST_PP_CAT` macro from `tf/staticTokens.h`.  This PR completes that change.

### Fixes Issue(s)
- #2247 


jesschimein:
Filed as internal issue #USD-8651

PR #2646, opened 2023-09-05
Avoid `TfToken` copy while iterating in `usd/primDefinition.cpp`

nvmkuruc:
### Description of Change(s)

The following warning was observed in GCC 11

```
warning: loop variable ‘srcField’ creates a copy from type ‘const pxrInternal_v0_23__pxrReserved__::TfToken’ [-Wrange-loop-construct]
  523 |     for (const TfToken srcField : defProp.ListMetadataFields()) {
      |                        ^~~~~~~~
note: use reference type to prevent copying
  523 |     for (const TfToken srcField : defProp.ListMetadataFields()) {
      |                        ^~~~~~~~
      |                        &
```

Making `srcField` a reference type avoids the copy of the `TfToken`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8652

PR #2647, opened 2023-09-05
Apple: Mesh shaders - shader modifications

slingthor:
### Description of Change(s)
- Add shader that takes advantage of going wide on the GPU and uses triangle backface and frustum culling.
- Introduce a technique for packing and unpacking normals in device space. This method is slightly lossy but may be good enough. Another solution exists which performs the whole calculation.
- Stubs for renderPass


jesschimein:
Filed as internal issue #USD-8653

PR #2648, opened 2023-09-05
Apple: Mesh shaders - shader generator modifications

slingthor:
### Description of Change(s)
- Extend shader generator to support
    - Mesh shader payload
    - Mesh shader per primitive in out
    - Mesh shader style handling of in outs
    - Handling of threadgroup locals that need to be declared in the entry point function
- Add new metadata to handle shader descriptions for mesh shaders


jesschimein:
Filed as internal issue #USD-8654

PR #2649, opened 2023-09-05
Apple: Mesh shaders - codegen and decisions

slingthor:
### Description of Change(s)
- Extend codegen to gracefully handle mesh shaders
- Extend mesh shader key to build up a mesh shading method for triangles
- Extend geometric shader to support mesh shading decisions


jesschimein:
Filed as internal issue #USD-8655

PR #2213, opened 2023-01-27
Replace boost traits and mpl with STL traits in pxr/base/tf

nvmkuruc:
### Description of Change(s)
- Remove spurious `boost/utility` include from `errorMark.cc`
- Removes spurious reference to `boost/mpl/vector/vector10.hpp` in `pySingleton.h`
- Replaces `boost::mpl::if_` with `std::conditional_t`
- Replaces `boost::is_same` with `std::is_same`
- Replaces `boost::remove_reference` with `std::remove_reference_t`
- Replaces `boost::enable_if` with `std::enable_if_t`
- Replaces `boost::disable_if` with `std::enable_if_t` and the negation of the condition
- Replaces `boost::is_base_of` with `std::is_base_of`

### Fixes Issue(s)
- #2211

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2656, opened 2023-09-06
Replace boost type traits from python bindings

nvmkuruc:
### Description of Change(s)

#2213 replaced boost's type traits with the STL equivalents but excluded the python bindings.  This PR extends the replacement to the usage in python bindings.

### Fixes Issue(s)
- #2211 


jesschimein:
Filed as internal issue #USD-8662

PR #2657, opened 2023-09-06
Replace `boost::lexical_cast` with `TfStringify` in python bindings

nvmkuruc:
### Description of Change(s)
#2496 replaced `boost::lexical_cast` with `TfStringify` for its non-python usage.  This PR extends that replacement to all usage in python bindings.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8663

Issue #974, opened 2019-09-19
usdMtlx ignores an active file prefix


willmuto-lucasfilm:
### Description of Issue
usdMtlx ignores an active file prefix defined by the fileprefix attribute in a parent tag.

### Steps to Reproduce
1. Given the following MaterialX file (ExpandFilePrefix.mtlx):

```xml
<?xml version="1.0"?>
<materialx version="1.36">
    <nodedef name="ND_layerShader" type="surfaceshader" node="layerShader">
        <input name="in" type="filename"/>
    </nodedef>
    <material name="noPrefix">
        <shaderref name="layered_sr" node="layerShader">
            <bindinput name="in" type="filename" value="img1.jpg"/>
        </shaderref>
    </material>
    <material name="prefixInValue">
        <shaderref name="layered_sr" node="layerShader">
            <bindinput name="in" type="filename" value="images/img2.jpg"/>
        </shaderref>
    </material>
    <material name="prefixInAttr"  fileprefix="images">
        <shaderref name="layered_sr" node="layerShader">
            <bindinput name="in" type="filename" value="img2.jpg"/>
        </shaderref>
    </material>
</materialx>
```
And running the following python code:

```python
from pxr import Usd, UsdShade, Sdf
stage = Usd.Stage.CreateInMemory()
stage = stage.Open('ExpandFilePrefix.mtlx')
material_noprefix = UsdShade.Material.Get(stage, Sdf.Path('/MaterialX/Materials/noPrefix'))
inputs = material_noprefix.GetInputs()
for i in inputs:
     print i.GetFullName(), i.Get().resolvedPath

material_prefixed1 = UsdShade.Material.Get(stage, Sdf.Path('/MaterialX/Materials/prefixInValue))
inputs = material_prefixed1.GetInputs()
for i in inputs:
     print i.GetFullName(), i.Get().resolvedPath

material_prefixed2 = UsdShade.Material.Get(stage, Sdf.Path('/MaterialX/Materials/prefixInAttr))
inputs = material_prefixed2.GetInputs()
for i in inputs:
     print i.GetFullName(), i.Get().resolvedPath
```

Returns the following results:
```
inputs:in /home/wmuto/usd_fileprefix/img1.jpg
inputs:in /home/wmuto/usd_fileprefix/images/img2.jpg
inputs:in 
```
This fix ensures that the fileprefix attribute is respected:
```
inputs:in /home/wmuto/usd_fileprefix/img1.jpg
[...]

PR #2660, opened 2023-09-07
Material-X: flatten filenames upon read  #974

Derek-Nvidia:
### Description of Change(s)
1. Call mx::flattenFilenames() anytime a mx::read* function is called.  This results in a document in which all of the 'fileprefix' attributes have been fully resolved and prepended to any inputs of type 'filename'
2. Add `test_ExpandFilePrefix` to `TestFileFormat` to test `fileprefix` resolution at various scopes in the MaterialX document.

### Fixes Issue(s)
- #974 


jesschimein:
Filed as internal issue #USD-8666

Issue #1544, opened 2021-06-23
MaterialX: Referencing a MaterialX fails


JGamache-autodesk:
### Description of Issue

Trying to reference a MaterialX document in a USD file to leverage the SdfFileFormat capabilities of usdMtlx:
```
def "MaterialX"
(
	references = @./standard_surface_jade.mtlx@
)
{
}
```
Fails with the following error:
```
Warning: in _ReportErrors at line 2825 of ...\USD\pxr\usd\usd\stage.cpp -- Unresolved reference path <defaultPrim> on prim @./UsdImportMaterialXSidecar.usda@,@anon:000002829ECFAAE0@</MaterialX>. (instantiating stage on stage @./UsdImportMaterialXSidecar.usda@ <000002829C48B470>)
```

Converting the file to USD using Python's `UsdMtlx._TestFile("standard_surface_jade.mtlx")` shows that the resulting stage is missing a `defaultPrim="MaterialX"` statement at the top level, which prevents referencing it.

Is this the expected behavior? Am I using the feature incorrectly?

JGamache-autodesk:
Forgot the most important step. Try with the latest tip of usdMtlx. Closing since I see commits more recent that most probably fix this...

JGamache-autodesk:
OK. Confirmed it still does not work on tip of dev branch, at commit 3fd536ae4 of June 14th.

JGamache-autodesk:
It works if I use subLayers though. Here is a test scene.
[ReferencingTest.zip](https://github.com/PixarAnimationStudios/USD/files/6703947/ReferencingTest.zip)
The Jade material uses an edited `.usda` stage dump of the `.mtlx` where a default prim was added. The Gold material uses a subLayer.
It would be nice if I could reference standard_surface_gold.mtlx instead.

spiffmon:
This is expected behavior, since we don't know which prim in the converted Mtlx document should be considered "the" canonical prim for the file.  You need to reference specific materials within the file.  If folks can make a reasonable argument that there's a knowable canonical prim to reference, we'd be happy to consider a PR that adds a setting of _defaultPrim_ .

JGamache-autodesk:
Thank you.

For reference, the fix is:
```
def "MaterialX"
(
	references = @./standard_surface_jade.mtlx@</MaterialX>
)
{
}
```
Which works perfectly fine.

spiffmon:
[...]

PR #2661, opened 2023-09-07
usdMtlx: set a default prim

expenses:
### Description of Change(s)

Having to specify `</MaterialX>` when referencing MaterialX files was annoying me and it has tripped other people up in the past. There's no reason not to set a default prim for these files so I thought I'd just do that.

I haven't ran the tests yet simply because I haven't found the right directory to run `ctest` in.

Before:
`usdcat test.mtlx`
```
#usda 1.0
(
    customLayerData = {
        string colorSpace = "lin_rec709"
    }
)

def "MaterialX"
{
    def "Materials"
    {
...
```
After:
`usdcat test.mtlx`
```
#usda 1.0
(
    customLayerData = {
        string colorSpace = "lin_rec709"
    }
    defaultPrim = "MaterialX"
)

def "MaterialX"
{
    def "Materials"
    {
...
```

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/1544 (https://github.com/PixarAnimationStudios/OpenUSD/issues/1544#issuecomment-867162857 specifically)

### DIFF
--- a/pxr/usd/usdMtlx/reader.cpp
+++ b/pxr/usd/usdMtlx/reader.cpp
@@ -2612,6 +2612,9 @@ UsdMtlxRead(
     // Translate all materials.
     ReadMaterials(mtlx, context);
 
+    // Set the default prim.
+    stage->SetDefaultPrim(stage->GetPrimAtPath(internalPath));
+
     // If there are no looks then we're done.
     if (mtlx->getLooks().empty()) {
         return;
--- a/pxr/usd/usdMtlx/reader.cpp
+++ b/pxr/usd/usdMtlx/reader.cpp
@@ -2613,7 +2613,9 @@ UsdMtlxRead(
     ReadMaterials(mtlx, context);
 
     // Set the default prim.
-    stage->SetDefaultPrim(stage->GetPrimAtPath(internalPath));
+    if (auto internalPrim = stage->GetPrimAtPath(internalPath)) {
+        stage->SetDefaultPrim(internalPrim);
+    }
 
     // If there are no looks then we're done.
     if (mtlx->getLooks().empty()) {
--- a/pxr/usd/usdMtlx/testenv/testUsdMtlxFileFormat.testenv/baseline/CustomNodeDef.usda
+++ b/pxr/usd/usdMtlx/testenv/testUsdMtlxFileFormat.testenv/baseline/CustomNodeDef.usda
@@ -1,4 +1,7 @@
 #usda 1.0
+(
+    defaultPrim = "MaterialX"
+)
 
 def "MaterialX"
 {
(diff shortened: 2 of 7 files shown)


jesschimein:
Filed as internal issue #USD-8667

spiffmon:
Great enhancement, @expenses !  I think you _are_ going to need to update some baselines, namely the ones [here](https://github.com/PixarAnimationStudios/OpenUSD/tree/dev/pxr/usd/usdMtlx/testenv/testUsdMtlxFileFormat.testenv/baseline).  Looking into what directory to run ctest from for these tests...

spiffmon:
... and I'm told that you can run ctest from the directory where build was done.  If you do this:
```
ctest -C Release -R testUsdMtlx
```
it should run all the test having "testUsdMtlx" in the name.

expenses:
I've run and fixed the tests now. I just forgot to build them earlier :)

dlarsson-adobe:
I like the idea of a default prim in a MaterialX document but I don't like it not being an actual Material Prim.
We have recently gone through a similar struggle with our material related usd documents and decided to go away from the MaterialX/Materials like layout (which is super nice when sublayering in a bunch of materials and it all ends up in a nice library) to make all the materials root level prims so we can have a default prim being a directly referrable Material.

One way of eating the cake and having it too would be to reference out the first material to the root level namespace through a reference and make that one the default prim but it would also litter the root level namespace with material prims.

Also both materialx viewer and graph editor will pick the first surfacematerial node in the document as its "default material" so there is some level of precedence on what is the default material in the materialx ecosystem

spiffmon:
We'd defer to the usd-materialx community on what _should_ be the best `defaultPrim`, but you do point out a problem that  we can and likely should address, @dlarsson-adobe ... `defaultPrim` came about long before we allowed sub-root references.  Now that we do, we should remove the "root prim only" restriction - would you care to file an Issue for that?

dlarsson-adobe:
I'll file an issue straight away
https://github.com/PixarAnimationStudios/OpenUSD/issues/2951

fhechtAdobe:
An alternative approach might be to choose the behavior based on the specific .mtlx file. If the file contains a single material it would be nice to have a default prim that points at the USD version of the material, so that one can reference the material directly when referencing the file.

If the .mtlx file contains multiple materials, then it is more like a material library and having a top level scope that is also the default prim might be a better setup for this case. Referencing the file will bring in all the materials.

The consuming application would have to check the type of the referenced default prim to understand the contents of the file. So the argument @dlarsson-adobe made to make sure the default prim always points at a material would avoid having to do this extra check.

Comes down to how many materials are usually in a .mtlx file? If it is common to have single material in there, then it would be nice to be able to just reference the file.

spiffmon:
Do note that UsdAbc shows how [we can use SdfFIleFormatArgs in the reference asset path to control the "unrolling" behavior](https://openusd.org/dev/api/usdabc_page_front.html).  If there's no clear favoring of library-over-single-material, this would provide toolmakers another way to control whether we're referencing "allMaterials" or "singleMaterial".  I'm not promoting that as the _best_ option, but seems like this issue should get some discussion among the MaterialX/USD toolmakers?

Issue #2663, opened 2023-09-07
UsdMtlx: reader ignores unit attribute


Derek-Nvidia:
### Description of Issue
Certain Material X parameters, the user can supply `unit` and `unittype` attributes.

_MaterialX allows floating-point and vector values to be defined in terms of a specific unit and unit type, and can automatically convert values from their specified unit into a scene unit of the same type specified by the application. This allows images and the quantities they represent such as displacement amount to be specified at an absolute real-world size, and then be converted automatically to the expected scene units of the application._

Presently `unittype` can be one of two values.

distance: - with unit values of:  nanometer, inch, yard, cm, meter, etc.
angle - with unit values of radian or degree

The reader should be updated so that if the unit attribute is present it's value is stored as custom data.

### System Information (OS, Hardware)
Windows

### Package Versions
USD 22.11
MaterialX - 1.38.4

jesschimein:
Filed as internal issue #USD-8669

spiffmon:
Hi @Derek-Nvidia ,
   The `distance` metric is not currently tractable to operate on in full generality, though we are hoping to propose changes to how USD handles such metrics, later this year, that should allow us to make these units more meaningful.

   For `angle`, [UsdGeom dictates that all user-encoded angles are in degrees](https://openusd.org/release/api/usd_geom_page_front.html#UsdGeom_LinAlgBasics), so I would counter-propose that UsdMtlx convert all angles-tagged-as-radians to degrees.  However, I do not know if there are assumptions in MaterialX shaders about angles, and/or whether the `angle` `unittype` is meant to be a declaration of what the shader expects - tagging @jstone-lucasfilm for comment.  If that **is** the case, then we may need to combine our proposals: UsdMtlx converts all angles in radians to degrees, but also adds some tag that makes it through to HdMtlx that causes conversion from degrees back to radians.  It would be **better** (simpler and faster) if this angular expectation were part of the SdrShaderNode definition for affected shaders, however this will only work if it is sufficient to encode on Shaders and not NodeGraphs, because Sdr does not yet allow NodeGraphs to be registered.

PR #2664, opened 2023-09-07
UsdMtlx: add support for unit attribute #2663

Derek-Nvidia:
### Description of Change(s)
1. Update `UsdMtlx` reader to store unit attribute value as custom data.
2. `TestFileFormat` - add `test_unit_attribute` test and relevant test data.

### Fixes Issue(s)
- #2663 


jesschimein:
Filed as internal issue #USD-8670

FlorianZ:
@Derek-Nvidia , closing this since as per discussion in the associated issue, we think we need to handle this a different way, which we hope to tackle before too long! Thanks again for your contribution and for raising this issue, though. We really appreciate it!

PR #2667, opened 2023-09-08
Remove unused units table in `pxr/usd/sdf/types.cpp`

nvmkuruc:
### Description of Change(s)
While testing #2590, it was observed that one of the tables is unused. This followup PR removes the unused table.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8678

PR #2669, opened 2023-09-10

aaye:
### Description of Change(s)
Small fixes to address warnings while working to get clang/ninja build working on Windows. Some warnings from MSVC most recent update.

### Fixes Issue(s)
- Multiple cases of printf semantics referring to size_t variables using a "l" semantic instead of "z". For example "lu" instead of "zu".
- A few cases of a boolean bit operator instead of the logical operator.
- A few cases of for_each resulting in a copy instead of a const reference in the loop control.
- A few cases where a reference was created on a temporary value (return in the loop control).

### DIFF
--- a/pxr/base/arch/symbols.cpp
+++ b/pxr/base/arch/symbols.cpp
@@ -128,7 +128,7 @@ ArchGetAddressInfo(
         }
 
         if (symbolName) {
-            *symbolName = symbol->Name ? symbol->Name : "";
+            *symbolName = symbol->Name[0] ? symbol->Name : "";
         }
 
         if (symbolAddress) {
--- a/pxr/base/tf/stringUtils.cpp
+++ b/pxr/base/tf/stringUtils.cpp
@@ -806,8 +806,8 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
             // Add 5 mod 32 makes '_' sort before all letters.
             return ((l + 5) & 31) < ((r + 5) & 31);
         }
-        else if (IsDigit(l) | IsDigit(r)) {
-            if (IsDigit(l) & IsDigit(r)) {
+        else if (IsDigit(l) || IsDigit(r)) {
+            if (IsDigit(l) && IsDigit(r)) {
                 // We backtrack to find the start of each digit string, then we
                 // scan each digit string, ignoring leading zeros to put the two
                 // strings into alignment with their most significant digits.
@@ -869,7 +869,7 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
                 curEnd = lcur + std::min(std::distance(lcur, lend),
                                          std::distance(rcur, rend));
             }
-            else if (IsDigit(l) | IsDigit(r)) {
+            else if (IsDigit(l) || IsDigit(r)) {
                 if (lcur == lstr.c_str()) {
                     return l < r;
                 }
--- a/pxr/imaging/hio/OpenEXR/deflate/lib/deflate_decompress.c
+++ b/pxr/imaging/hio/OpenEXR/deflate/lib/deflate_decompress.c
@@ -1081,9 +1081,9 @@ typedef enum libdeflate_result (*decompress_func_t)
 
 #ifdef arch_select_decompress_func
 static enum libdeflate_result
-dispatch_decomp(struct libdeflate_decompressor *d,
-		const void *in, size_t in_nbytes,
-		void *out, size_t out_nbytes_avail,
+dispatch_decomp(struct libdeflate_decompressor * restrict d,
+		const void * restrict in, size_t in_nbytes,
+		void * restrict out, size_t out_nbytes_avail,
 		size_t *actual_in_nbytes_ret, size_t *actual_out_nbytes_ret);
 
 static volatile decompress_func_t decompress_impl = dispatch_decomp;
--- a/pxr/imaging/hio/fieldTextureData.h
+++ b/pxr/imaging/hio/fieldTextureData.h
@@ -103,7 +103,7 @@ private:
 ///
 /// A base class to make HioFieldTextureData objects, implemented by plugins.
 ///
-class HIO_API HioFieldTextureDataFactoryBase : public TfType::FactoryBase
+class HioFieldTextureDataFactoryBase : public TfType::FactoryBase
 {
 protected:
     friend class HioFieldTextureData;
--- a/pxr/imaging/hio/image.h
+++ b/pxr/imaging/hio/image.h
@@ -199,7 +199,7 @@ HioImage::GetMetadata(TfToken const & key, T * value) const
     return true;
 }
 
-class HIO_API HioImageFactoryBase : public TfType::FactoryBase {
+class HioImageFactoryBase : public TfType::FactoryBase {
 public:
     virtual HioImageSharedPtr New() const = 0;
 };
--- a/pxr/usd/sdf/textFileFormat.tab.cpp
+++ b/pxr/usd/sdf/textFileFormat.tab.cpp
@@ -3197,13 +3197,6 @@ yydestruct (yymsg, yytype, yyvaluep, context)
   if (!yymsg)
     yymsg = "Deleting";
   YY_SYMBOL_PRINT (yymsg, yytype, yyvaluep, yylocationp);
-
-  switch (yytype)
-    {
-
-      default:
-	break;
-    }
 }
 
 /* Prevent warnings from -Wmissing-prototypes.  */
--- a/pxr/base/arch/pragmas.h
+++ b/pxr/base/arch/pragmas.h
@@ -102,6 +102,12 @@
     #define ARCH_PRAGMA_INSTANCE_METHOD_NOT_FOUND \
         _Pragma("clang diagnostic ignored \"-Wobjc-method-access\"")
 
+    #define ARCH_PRAGMA_INT_FLOAT_CONVERSION
+        _Pragma("clang diagnostic ignored \"-Wimplicit-const-int-float-conversion\"")
+
+    #define ARCH_PRAGMA_POTENTIALLY_EVALUATED_EXPRESSION
+        _Pragma("clang diagnostic ignored \"-Wpotentially-evaluated-expression\"")
+
 #elif defined(ARCH_COMPILER_MSVC)
 
     #define ARCH_PRAGMA_PUSH \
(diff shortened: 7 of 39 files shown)


nvmkuruc:
It's exciting to see all the warning fixes.  If you're open to the suggestion, when iterating over a dictionary you can preserve the reference and avoid conversions by using `const auto&` or `const std::pair<const KEYTYPE, VALUETYPE>&`.  #2350 has an example of this.

jesschimein:
Filed as internal issue #USD-8680

pxr/imaging/hio/OpenEXR/openexr-c.c
```diff
@@ -31,6 +31,12 @@
 // defined but not referenced or exported.
 ARCH_PRAGMA_UNUSED_FUNCTION
 
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wmacro-redefined"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic ignored "-Wmacro-redefined"
+#endif
```
meshula: Which macro is reported as redefined? I didn't see this on linux/mac.

pxr/base/arch/symbols.cpp
```diff
@@ -128,7 +128,7 @@ ArchGetAddressInfo(
         }
 
         if (symbolName) {
-            *symbolName = symbol->Name ? symbol->Name : "";
+            *symbolName = symbol->Name[0] ? symbol->Name : "";
```
spiffmon, line 131: Perhaps a safer check here would be `symbol->NameLen > 0` ?

pxr/imaging/hio/fieldTextureData.h
```diff
@@ -103,6 +103,8 @@ class HioFieldTextureData
 ///
 /// A base class to make HioFieldTextureData objects, implemented by plugins.
 ///
+ARCH_PRAGMA_PUSH
+ARCH_PRAGMA_NON_DLL_INTERFACE_CLASS_USED_AS_BASE_FOR_DLL_INTERFACE_CLASS
```
meshula: FactoryBase is properly adorned. Is the real issue that HioFieldTextureDataFactoryBase doesn't export any symbols of its own? Would removing HIO_API fix it?

pxr/base/tf/stringUtils.cpp
```diff
@@ -806,8 +806,8 @@ TfDictionaryLessThan::_LessImpl(const string& lstr, const string& rstr) const
             // Add 5 mod 32 makes '_' sort before all letters.
             return ((l + 5) & 31) < ((r + 5) & 31);
         }
-        else if (IsDigit(l) | IsDigit(r)) {
-            if (IsDigit(l) & IsDigit(r)) {
+        else if (IsDigit(l) || IsDigit(r)) {
+            if (IsDigit(l) && IsDigit(r)) {
```
spiffmon, line 810: Although we neglected to document it here, this choice (and below) were done intentionally because they produced more efficient code, and this is a performance-sensitive function.  We're OK with these changes if it can be documented (e.g. via godbolt) that gcc/clang, at least, produce equivalent instructions.

pxr/base/arch/symbols.cpp
meshula, line 131: Double checking ~ this fix matches documentation. https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfromaddr

pxr/imaging/hio/OpenEXR/deflate/lib/deflate_decompress.c
```diff
@@ -1081,9 +1081,9 @@ typedef enum libdeflate_result (*decompress_func_t)
 
 #ifdef arch_select_decompress_func
 static enum libdeflate_result
-dispatch_decomp(struct libdeflate_decompressor *d,
-		const void *in, size_t in_nbytes,
-		void *out, size_t out_nbytes_avail,
+dispatch_decomp(struct libdeflate_decompressor * restrict d,
```
meshula, line 1084: corresponds to upstream warning (not functional) fix at https://github.com/ebiggers/libdeflate/commit/343ec0a6d67767c8ce10595555e8dae0ceb973ef

pxr/imaging/hio/image.h
```diff
@@ -199,10 +199,13 @@ HioImage::GetMetadata(TfToken const & key, T * value) const
     return true;
 }
 
+ARCH_PRAGMA_PUSH
+ARCH_PRAGMA_NON_DLL_INTERFACE_CLASS_USED_AS_BASE_FOR_DLL_INTERFACE_CLASS
```
meshula: same issue, no symbols exported., is the proper fix to remove HIO_API here?

spitzak:
I've also seen `auto &&` used for the interator to preserve const references

pxr/usd/sdf/textFileFormat.tab.cpp
```diff
@@ -3197,13 +3197,6 @@ yydestruct (yymsg, yytype, yyvaluep, context)
   if (!yymsg)
     yymsg = "Deleting";
   YY_SYMBOL_PRINT (yymsg, yytype, yyvaluep, yylocationp);
-
-  switch (yytype)
-    {
-
-      default:
-	break;
-    }
```
spiffmon, line 3206: Note this is auto-generated code!

spiffmon (changes requested):
Thanks for taking this on!  In addition to a few specific notes, we have an overall request on factoring.  As we note in our [contributing guidelines](https://openusd.org/release/contributing_to_usd.html#pull-request-guidelines), it is easier to review and safer to deploy when changes are smaller.  Could we ask that you break this up into several changes:
1. pxr/base/arch
2. pxr/base/{tf,vt}
3. pxr/usd
4. pxr/{imaging,usdImaging}

We think that should be granular enough to digest.

pxr/imaging/hio/fieldTextureData.h
aaye: If the structure is not meant to be DLL Exported, then removing the API decoration would also fix it. The core issue is that you have a DLL data structure inheriting from a non-DLL data structure.

pxr/base/arch/symbols.cpp
aaye: The fix is because symbol->Name is a "single" array, and so would always have a valid address which was the original code. The modified code I think was the intent of checking for an empty string.

pxr/imaging/hio/image.h
aaye: Potentially - I do not have sufficient understanding of the code base to make that call.

pxr/base/arch/symbols.cpp
aaye: Can do that, missed that the field existed.

pxr/base/tf/stringUtils.cpp
aaye: I can understand that - the change would be to disable the warning then and comment the reason for the it.

pxr/usd/sdf/textFileFormat.tab.cpp
aaye: Annoying :)

pxr/imaging/hio/OpenEXR/openexr-c.c
aaye: I hit most of these with elevated warnings attempting to get a ninja/clang build working on Windows. I can run it again to get an answer as to what triggered it.

aaye:
Once I see the changes pass through the gates, will break it up into four PRs as requested. Including getting back to the open questions.

pxr/base/arch
pxr/base/{tf,vt}
pxr/usd
pxr/{imaging,usdImaging}

pxr/base/tf/stringUtils.cpp
aaye: Diving into it a bit more to explain, I was quick to go to a documentation approach. Runtime performance would be dependent on the probability that the quick outs are hit or not. There are more instructions since the logical operators must include the test and jump when compared to the bitwise. This is a case where a pure instruction equivalency would not provide the right data, as it is more of an empirical test with real-world data that would provide insights into performance in normal situations. I'm going to assume something like that was done, so documenting it sounds like the right course of action. The other factor would be non-x86 hardware and how the CPU handles branch prediction and out-of-order processing.

pxr/imaging/hio/image.h
meshula: Could you test removing the decoration here to see if it resolves the warning?

aaye: Will do.

pxr/imaging/hio/OpenEXR/openexr-c.c
aaye: MSVC 17.7.2, PXR_ENABLE_PYTHON_SUPPORT=OFF (Probably the reason it has not been seen previously)
pxr\imaging\hio\OpenEXR\OpenEXRCore/openexr_conf.h(23,1): warning C4005: 'ILMTHREAD_THREADING_ENABLED': macro redefinition
pxr\imaging\hio\OpenEXR\OpenEXRCore/openexr_conf.h(25,1): warning C4005: 'OPENEXR_C_STANDALONE': macro redefinition

pxr/imaging/hio/fieldTextureData.h
aaye: Original warning:
/pxr/imaging/hio/image.h(202,52): warning C4275: non dll-interface class 'pxrInternal_v0_23__pxrReserved__::TfType::FactoryBase' used as base for dll-interface class 'pxrInternal_v0_23__
pxrReserved__::HioImageFactoryBase'

Removal of the HIO_API does fix it and is the preferred approach?

pxr/imaging/hio/image.h
aaye: Same as above.

pxr/imaging/hio/OpenEXR/openexr-c.c
aaye: Not sure if it helps, but the only other occurrence that I saw in my environment was a config file in OpenEXR where they were defining ILMTHREAD_THREADING_ENABLED 1. Also, may no longer be valid - not seeing it in dev.

pxr/base/tf/stringUtils.cpp
aaye: For comparison and FYI:

The boolean operator (2 compare and jumps - hard to say what that looks like on x86 out-of-order performance)
        cmp     dword ptr [rbp - 8], 0
        jne     .LBB1_2
        cmp     dword ptr [rbp - 12], 0
        je      .LBB1_3

The bitwise operator (1 compare and jump, and a mov and or - same number of memory references)
        mov     eax, dword ptr [rbp - 8]
        or      eax, dword ptr [rbp - 12]
        cmp     eax, 0
        je      .LBB1_2

code:
    int varA=0, varB=1;
    if (varA | varB)

spiffmon:
Hi @aaye - checkin in on this, whether a) newer releases have addressed anything, b) if you're still planning on splitting this into smaller PR's, as discussed?

Thanks!

aaye:
(1) Have not tried new releases. (2) I did break it down into smaller PRs (4-6 if I remember correctly). Can close this one down in favor of the other ones.

PR #2671, opened 2023-09-11
Fix for git bash on windows using cp 1252 instead of Unicode for stdout

erslavin:
### Description of Change(s)

git bash on Windows uses cp 1252 for terminal encoding by default.  In newer OpenUSD builds, there are file name that contain Unicode characters that will cause functions in the build script like e.g. `PrintCommandOutput` to raise a `UnicodeEncodeError` when attempting to write the names of these files to `stdout`.  This fix reconfigures `stdout` in the build script to use `utf-8` explicitly.

### Fixes Issue(s)

[X] I have submitted a signed Contributor License Agreement

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2270,6 +2270,13 @@ except Exception as e:
 
 verbosity = args.verbosity
 
+if os.name == "nt" and sys.version_info > (3, 6):
+    # fix for git-bash on Windows using code page 1252 as default instead
+    # of unicode - functions like e.g. PrintCommandOutput fail writing
+    # unicode characters for filenames containing them
+    # `reconfigure` only exists in versions greater than Python 3.6
+    sys.stdout.reconfigure(encoding='utf-8')
+
 # Augment PATH on Windows so that 3rd-party dependencies can find libraries
 # they depend on. In particular, this is needed for building IlmBase/OpenEXR.
 extraPaths = []
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2270,7 +2270,7 @@ except Exception as e:
 
 verbosity = args.verbosity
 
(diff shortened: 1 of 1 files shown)


build_scripts/build_usd.py
```diff
@@ -2270,6 +2270,13 @@ def ForceBuildDependency(self, dep):
 
 verbosity = args.verbosity
 
+if (sys.version_info[0] == 3 and sys.version_info[1] > 6) or sys.version_info[0] > 3:
+    # fix for git-bash on Windows using code page 1252 as default instead
+    # of unicode - functions like e.g. PrintCommandOutput fail writing
+    # unicode characters for filenames containing them
+    # `reconfigure` only exists in versions greater than Python 3.6
```
asluk, line 2277: What role does "reconfigure" play? Is there a spec we can link? Thanks!

erslavin: It's because the default encoding of `TextIOWrapper` is locale-specific (which is the type of object `sys.stdin` and `sys.stdout` are):

https://docs.python.org/3/library/io.html?highlight=reconfigure#io-text-encoding

jesschimein:
Filed as internal issue #USD-8682

build_scripts/build_usd.py
```diff
@@ -2270,6 +2270,13 @@ def ForceBuildDependency(self, dep):
 
 verbosity = args.verbosity
 
+if os.name == "nt" and sys.version_info > (3, 6):
```
nvmkuruc: I think you're going to want to do `sys.version_info >= (3, 7)` as `(3, 6, PATCH_NUMBER)` will evaluate to being greater than `(3, 6)`.

nvmkuruc:
@erslavin Since OpenUSD's minimum version is 3.9, do you mind updating this PR to not have the version guard?

PR #2675, opened 2023-09-12
Enable MSVC conformant preprocessor

aaye:
### Description of Change(s)
 Add the compiler option in msvc defaults to include the conformant preprocessor (Zc:preprocessor)
 Modify two of the preprocessor util include files to test for whether the conformant or traditional processor is being used.

### Fixes Issue(s)
- Increases the conformance between MSVC and other frontends used in the project.


aaye:
Compiling against windows SDK 17763 - not seeing the compile issues under 22621. Microsoft must have fixed the issues with their include chain between the two versions. I am assuming that changing Windows SDK is not trivial and not worth doing for a conformance preprocessor update?

pxr/base/tf/preprocessorUtils.h
```diff
@@ -63,7 +63,7 @@ ARCH_PRAGMA_MACRO_TOO_FEW_ARGUMENTS
 ///
 /// \ingroup group_tf_Preprocessor
 /// \hideinitializer
-#if defined(ARCH_COMPILER_MSVC)
+#if defined(ARCH_COMPILER_MSVC) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
```
nvmkuruc: If you're open to the suggestion, you might want to create a `ARCH_MSVC_TRADITIONAL_PREPROCESSOR` macro equivalent to 
`defined(ARCH_COMPILER_MSVC) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)` that can be referenced in `preprocessorUtils.h` and `preprocessorUtilsLite.h`.

Support for the conformant preprocessor is exciting to see.

aaye: Can do that - will still be blocked by the warnings emitted from the MSVC include chain. As noted above, I do not see them locally, as I am compiling them against a newer version. With that update, it compiles cleanly "on my machine.com"

jesschimein:
Filed as internal issue #USD-8689

pxr/base/tf/preprocessorUtils.h
aaye: > Microsoft Blog: Windows 10 SDK version 2104 has been released, which includes the changes needed for C11 and C17 as well as the conformant preprocessor
This seems to match with Visual Studio 2019.
I will update the PR with the asked-for change (since I keep them as patch local), but then this should either be closed or left unresolved until an SDK update.

nvmkuruc:
@jesschimein This might be addressed by #2990 which just got merged.  Perhaps this can be closed?

pixar-oss:
Thank you!  I think you're right Matt. Closing as a duplicate of #2990

PR #2678, opened 2023-09-12

aaye:
### Description of Change(s)
 - Fix for using "long" type in printf format strings when paired with a size_t argument. Change from "l" to "z".
 - Fix for not using the "long" type when referring to a wchat_t in a print format string. Change from "s" to "ls".
 - Added a comment for using a bitwise operator in performance-critical code.
 - Template that will SFINAE throws a warning during the first pass compilation. Pragma ignored the warning not to break strict builds.
 - Wrapping the static atomic in singleton.h with an ignore compilation warning due to lack of export symbols (DLL export). This is theoretically safe as long as care is taken about access of singletons across DLL boundaries.

### Fixes Issue(s)
- Compile warnings

### DIFF
--- a/pxr/base/arch/pragmas.h
+++ b/pxr/base/arch/pragmas.h
@@ -102,6 +102,24 @@
     #define ARCH_PRAGMA_INSTANCE_METHOD_NOT_FOUND \
         _Pragma("clang diagnostic ignored \"-Wobjc-method-access\"")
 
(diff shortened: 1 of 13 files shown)


nvmkuruc:
Just noting that #2593 also tries to address the `static std::atomic` issue.

jesschimein:
Filed as internal issue #USD-8692

PR #2643, opened 2023-09-05
Avoid conversion of `SdfLayerRefPtr` to `SdfLayerHandle` in `pcp/cache.cpp`

nvmkuruc:
### Description of Change(s)

The following warning was observed in GCC 11.

```
warning: loop variable ‘layer’ of type ‘const SdfLayerHandle&’ {aka ‘const pxrInternal_v0_23__pxrReserved__::TfWeakPtr<pxrInternal_v0_23__pxrReserved__::SdfLayer>&’} binds to a temporary constructed from type ‘const pxrInternal_v0_23__pxrReserved__::TfRefPtr<pxrInternal_v0_23__pxrReserved__::SdfLayer>’ [-Wrange-loop-construct]
 1232 |         for (const SdfLayerHandle& layer: layerStack->GetLayers()) {
```

`GetLayers()` returns a vector of `SdfLayerRefPtr`s.  Iterating using a handle will trigger an implicit conversion to a temporary `SdfLayerHandle`.  This PR uses `auto` to preserve the type of the vector contents and avoid any implicit conversions.

### Fixes Issue(s)
-

- [ ] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2679, opened 2023-09-12

aaye:
### Description of Change(s)
 - Fix for the use of "long" type in printf format strings when paired with a size_t argument. Change from "l" to "z".
 - Fix for for loop iterators using non-reference types, causing a full object copy to occur. Change to use const reference.
 - Fix for an int to float conversion warning. Change to disable the warning.
 - Fix for using the result of an assignment as an if condition. Change to wrap assignment in parentheses.
 - Removed an unused variable from a _UsdImagingDataSourceShadingNodeInputs and UsdImagingDataSourceVolumeFieldBindings
 - Fix for an issue when using typeid. A warning is thrown when the parameter expression may have side effects. Disabled the compilation warning.

### Fixes Issue(s)
- Compile warnings

### DIFF
--- a/pxr/usd/pcp/cache.cpp
+++ b/pxr/usd/pcp/cache.cpp
@@ -1229,7 +1229,7 @@ PcpCache::ReloadReferences(PcpChanges* changes, const SdfPath& primPath)
     // local layers.
     SdfLayerHandleSet layersToReload;
     for (const PcpLayerStackPtr& layerStack: layerStacksAtOrUnderPrim) {
-        for (const SdfLayerHandle& layer: layerStack->GetLayers()) {
+        for (const SdfLayerHandle layer: layerStack->GetLayers()) {
             if (!_layerStack->HasLayer(layer)) {
                 layersToReload.insert(layer);
             }
--- a/pxr/usd/usd/primDefinition.cpp
+++ b/pxr/usd/usd/primDefinition.cpp
@@ -397,8 +397,8 @@ UsdPrimDefinition::_FindOrCreatePropertySpecForComposition(
     // we create a new layer for this prim definition to write its composed
     // properties.
     if (_composedPropertyLayer) {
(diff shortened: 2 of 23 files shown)


pxr/usd/usd/primDefinition.cpp
```diff
@@ -520,7 +520,7 @@ UsdPrimDefinition::_ComposeOverAndReplaceExistingProperty(
     // get the property spec with the overrides applied. Any fields that
     // are defined in the override spec are stronger so we copy the defined
     // spec fields that aren't already in the override spec.
-    for (const TfToken srcField : defProp.ListMetadataFields()) {
+    for (const TfToken &srcField : defProp.ListMetadataFields()) {
```
nvmkuruc, line 523: Just a heads up this has been fixed internally already with https://github.com/PixarAnimationStudios/OpenUSD/pull/2646.  You might need to rebase after the next push.

pxr/usd/pcp/cache.cpp
```diff
@@ -1229,7 +1229,7 @@ PcpCache::ReloadReferences(PcpChanges* changes, const SdfPath& primPath)
     // local layers.
     SdfLayerHandleSet layersToReload;
     for (const PcpLayerStackPtr& layerStack: layerStacksAtOrUnderPrim) {
-        for (const SdfLayerHandle& layer: layerStack->GetLayers()) {
+        for (const SdfLayerHandle layer: layerStack->GetLayers()) {
```
nvmkuruc, line 1232: Just a heads up this has been fixed internally with https://github.com/PixarAnimationStudios/OpenUSD/pull/2643.  You might need to rebase after the next push.

pxr/usd/usd/primDefinition.cpp
aaye: Thanks, and appreciated the heads up :)

jesschimein:
Filed as internal issue #USD-8693

PR #2680, opened 2023-09-12

aaye:
### Description of Change(s)
- Fix for the use of "long" type in printf format strings when paired with a size_t argument. Change from "l" to "z".
- Fix for loop iterators using non-reference types, causing a full object copy to occur. Change to use const reference.
- Fix for loop iterators using reference types on temporary values (function return values). Change to use a full type.
- Fix a base class that did not have a virtual destructor.
- Removed DLL Export tag on a class inheriting from a non-DLL export class.
- Unnamed structures are not allowed to initialize values. Name the structure in image to allow for defaulting gamma to 0.0.

### Fixes Issue(s)
- Compiler Warnings


meshula:
Could you try this PR again? It's made against release, rather than dev. That's why it accidentally picked up so many unrelated changes.

jesschimein:
Filed as internal issue #USD-8694

aaye:
Changed the base. Sorry for the delay; I finally finished getting the repo to compile against Ninja/Clang on Windows. Expect a few more PRs from that effort :)

pxr/imaging/hdsi/coordSysPrimSceneIndex.cpp
```diff
@@ -302,7 +302,7 @@ HdsiCoordSysPrimSceneIndex::GetChildPrimPaths(const SdfPath &primPath) const
     if (it == _targetedPrimToNameToRefCount.end()) {
         return result;
     }
-    for (const std::pair<TfToken, size_t> &nameAndRefCount : it->second) {
+    for (const std::pair<TfToken, size_t> nameAndRefCount : it->second) {
```
nvmkuruc: You should be able to avoid the copy by making `nameAndRefCount` `const auto&` or `const std::pair<const TfToken, size_t>&`.

pxr/imaging/hio/glslfxConfig.cpp
```diff
@@ -310,7 +310,7 @@ HioGlslfxConfig::_GetSourceKeyMap(VtDictionary const & dict,
 
     const VtDictionary& specDict = techniqueSpec.UncheckedGet<VtDictionary>();
     // get all of the shader stages specified in the spec
-    for (const std::pair<std::string, VtValue>& p : specDict) {
+    for (const std::pair<std::string, VtValue> p : specDict) {
```
nvmkuruc: You should be able to avoid the copy by making `p` `const auto&` or `const std::pair<const std::string, VtValue>&`.

pxr/imaging/hio/glslfxConfig.cpp
```diff
@@ -438,7 +438,7 @@ HioGlslfxConfig::_GetParameters(VtDictionary const & dict,
 
     const VtDictionary& paramsDict = params.UncheckedGet<VtDictionary>();
     // pre-process the paramsDict in order to get the merged ordering
-    for (const std::pair<std::string, VtValue>& p : paramsDict) {
+    for (const std::pair<std::string, VtValue> p : paramsDict) {
```
nvmkuruc: You should be able to avoid the copy by making `p` `const auto&` or `const std::pair<const std::string, VtValue>&`.

pxr/imaging/hio/glslfxConfig.cpp
```diff
@@ -546,7 +546,7 @@ HioGlslfxConfig::_GetTextures(VtDictionary const & dict,
     }
 
     const VtDictionary& texturesDict = textures.UncheckedGet<VtDictionary>();
-    for (const std::pair<std::string, VtValue>& p : texturesDict) {
+    for (const std::pair<std::string, VtValue> p : texturesDict) {
```
nvmkuruc: You should be able to avoid the copy by making `p` `const auto&` or `const std::pair<const std::string, VtValue>&`.

pxr/imaging/hio/glslfxConfig.cpp
```diff
@@ -617,7 +617,7 @@ HioGlslfxConfig::_GetAttributes(VtDictionary const & dict,
 
     const VtDictionary& attributesDict =
         attributes.UncheckedGet<VtDictionary>();
-    for (const std::pair<std::string, VtValue>& p : attributesDict) {
+    for (const std::pair<std::string, VtValue> p : attributesDict) {
```
nvmkuruc: You should be able to avoid the copy by making `p` `const auto&` or `const std::pair<const std::string, VtValue>&`.

pxr/imaging/hdsi/coordSysPrimSceneIndex.cpp
aaye: Both worked, will update. I had tried your other suggestion of auto&& and that failed, should have tried the others as the time.

PR #2684, opened 2023-09-13
Replace `boost::any` with `std::any` in `pxr/base/tf`

nvmkuruc:
### Description of Change(s)

C++17 provides `std::any` and `std::any_cast` which can replace `boost::any` and `boost::any_cast`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8699

PR #2685, opened 2023-09-13
Replace `boost::shared_array` with `std::shared_ptr` of C-style array

nvmkuruc:
### Description of Change(s)
In C++17, `std::shared_ptr` can manage C-style arrays without a custom deleter. This uses that feature to replace the `UsdGeomBBoxCache`'s usage of `boost::shared_array`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8700

PR #2686, opened 2023-09-13
fix: add long path support when calling ArchAbsPath on Windows 10

sssooonnnggg:
### Description of Change(s)

According to MSDN, starting with Windows 10, Version 1607, `GetFullPathNameW` now support get paths which has no limitation with MAX_PATH 

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew

image

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8701

PR #2462, opened 2023-06-01
oneTBB: replace TBB utility functions for placement new

brechtvl:
### Description of Change(s)
No need to rely on TBB for this.

### Fixes Issue(s)

Part of #1471, adding oneTBB support.

- [x] I have verified that all unit tests pass with the proposed changes
- [x] I have submitted a signed Contributor License Agreement

PR #2694, opened 2023-09-14
OneAPI TBB selector for 2019/2021 adapted from brechtvl

aaye:
### Description of Change(s)
- Originally filed under USD-6600, Issue #1471
- Looked like brechtvl did not continue to see their change through the pipeline. Took and adapted the change to fix compilation, pathing, and build associated issues.
- Successfully compiled with both 2019 and 2021

### Fixes Issue(s)
- Facilitates updating to more recent releases for OneAPI TBB

### DIFF
--- a/cmake/defaults/Packages.cmake
+++ b/cmake/defaults/Packages.cmake
@@ -40,6 +40,10 @@ set(PXR_THREAD_LIBS "${CMAKE_THREAD_LIBS_INIT}")
 # Find Boost package before getting any boost specific components as we need to
 # disable boost-provided cmake config, based on the boost version found.
 find_package(Boost REQUIRED)
+if(Boost_FOUND)
+    include_directories(${Boost_INCLUDE_DIRS})
+    link_directories(${Boost_LIBRARY_DIRS})
+endif()
 
 # Boost provided cmake files (introduced in boost version 1.70) result in 
 # inconsistent build failures on different platforms, when trying to find boost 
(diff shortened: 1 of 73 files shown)


jesschimein:
Filed as internal issue #USD-8708

loqs:
Is this intended to replace #2457 #2458 #2459 #2460 #2461 #2462 #2463 #2464 #2465 and #2466?

spiffmon:
Thanks for asking that, @loqs , as:

1. We have @brechtvl 's PR's queued up to process shortly
2. We specifically asked him to break the changes up as he did for fine-grained testing and regression analysis, and we would still not want to accept a monolithic change.

@aaye , can you please confirm, and let us know if you can follow the pattern of the previous PR's, and also let us know the scale of the fixes you made on top of the earlier work?

brechtvl:
From what I can tell the main addition here is build system fixes for Windows? I did not test my changes on Windows, so getting that to work would be great. I can add additional changes to my PRs if needed (might be just #2466).

cmake/defaults/Packages.cmake
```diff
@@ -152,8 +156,23 @@ endif()
 
 
 # --TBB
-find_package(TBB REQUIRED COMPONENTS tbb)
-add_definitions(${TBB_DEFINITIONS})
+if (PXR_ENABLE_ONEAPI_TBB)
+    find_package(TBB REQUIRED COMPONENTS tbb12) # OneAPI TBB encodes the major version on the library name.
```
brechtvl, line 160: I think it would be preferable to put as much logic as possible in `FindTBB.cmake`, and to avoid having a `PXR_ENABLE_ONEAPI_TBB` option.

From what I can tell the library name can be `tbb12` or `tbb` on Windows, and is always `tbb` on other platforms.

In Embree it looks for both on Windows:
https://github.com/embree/embree/blob/master/common/cmake/FindTBB.cmake#L319

While in OpenImageIO it only looks for `tbb` (but maybe can find `tbb12` by using cmake config files instead):
https://github.com/OpenImageIO/oiio/blob/master/src/cmake/modules/FindTBB.cmake

brechtvl: I updated #2466 now with a fix for `tbb12` detection on Windows. Other build system changes in this PR may be needed as well, but their purpose is not clear to me.

spiffmon:
Closing this as we are finishing up work to incorporate @brechtvl 's PR's.

PR #2695, opened 2023-09-15
Improve the documentation for UsdSkelRoot

cameronwhite:
### Description of Change(s)

These are a couple fixes and improvements to the SkelRoot docs, from https://forum.aousd.org/t/getting-different-results-with-computeworldbound-for-when-root-is-xform-vs-skelroot/409

- Document that SkelRoot has an extent computation which is used when the extents are not authored, and add a recommendation that the extents should be pre-authored for better performance.

- Fix incorrect code snippet, which used `UsdGeomTransform` rather than `UsdGeomXform`. For C++ code, `UsdSkel.Root` was also incorrect and should be just `UsdSkelRoot`

### Fixes Issue(s)
N/A


jesschimein:
Filed as internal issue #USD-8709

Issue #2425, opened 2023-05-04
UsdView renders Mesh with Blendshapes in wrong place when using GPU vs CPU implementation


talislfb:
### Description of Issue
Having a prim Mesh with BlendShapes, but no UsdSkel joints associated with it causes UsdView to render the mesh wrongly at the origin UsdView with CPU vs GPU. 

Expected result achieved by running the `UsdSkel.BakeSkinning(usdviewApi.stage.Traverse())`
![image](https://user-images.githubusercontent.com/77503455/236298465-d569f427-7b2c-4d91-9c35-5b5dad97f679.png)

Current default results
![image](https://user-images.githubusercontent.com/77503455/236298596-7341e61e-073d-4518-bb94-6ae3c677371a.png)

As you can see, from the images above, the sphere is rendered at the origin (even though the actual prim location hasn't changed).

cameronwhite:
Attaching the example from the mailing list: [shapes.usd.gz](https://github.com/PixarAnimationStudios/USD/files/11407384/shapes.usd.gz)

The different behaviour can be observed by comparing against `env USDSKELIMAGING_FORCE_CPU_COMPUTE=1 usdview shapes.usd`, or applying the BakeSkinning operation mentioned above

My comments from mailing list:
```
From a quick glance, if there are only blendshapes (no skinning) then the deformed point positions are already in the prim's local space.
These transforms shouldn't be necessary in this scenario since they're expecting the points to be in the skeleton's space after skinning:
https://github.com/PixarAnimationStudios/USD/blob/b53573ea2a6b29bc4a6b129f604bbb342c35df5c/pxr/usdImaging/usdSkelImaging/shaders/skinning.glslfx#L227
(and the geomBindXform earlier on also looks unnecessary when no skinning is being applied?)
```

tallytalwar:
Filed as internal issue #USD-8311

FlorianZ:
Great find, @cameronwhite. Would you be up for contributing a fix for this? Thank you!!

cameronwhite:
> Great find, @cameronwhite. Would you be up for contributing a fix for this? Thank you!!

Yep! I've submitted PR #2696

PR #2696, opened 2023-09-15
Fix incorrectly transformed points for GPU blendshapes with no skinning.

cameronwhite:
### Description of Change(s)
When the mesh only has blendshapes (no skinning), the deformed point positions are already in the prim's local space so it's simplest to just skip applying any additional transforms.

The previous behaviour was incorrect because the final skel space -> world space -> model space transform expected the points to be in skel space, which was not the case when only blendshapes were present. Similarly, the bind transform is only required when skinning is enabled.

The results are now consistent with the CPU skinning computation.
### Fixes Issue(s)
- #2425

### DIFF
--- a/pxr/usdImaging/usdSkelImaging/shaders/skinning.glslfx
+++ b/pxr/usdImaging/usdSkelImaging/shaders/skinning.glslfx
@@ -23,9 +23,6 @@ const float EPS = 1e-5;
 
 void compute(int index)
 {
-    // model space -> bind space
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8710

tcauchois:
Hey all, this will go in momentarily, but do you happen to have a test asset demonstrating the problem that we could add to our unit test suite?  Thanks!

Issue #2621, opened 2023-08-24
Basic UsdSkel doesn't animate in usdview-23.08


mati-nvidia:
### Description of Issue
usdview-23.08 viewport does not show the animation from this basic UsdSkel example: https://openusd.org/release/api/_usd_skel__schema_overview.html

### Steps to Reproduce
1. Save the USDA from *Skinning an Arm* example: https://openusd.org/release/api/_usd_skel__schema_overview.html
2. Open the USDA file with usdview (23.08)
3. Press play or scrub through the timeline
4. RESULT: The interpolated values for `/Model/Skel/Anim.rotations` are displayed in the Properties window, but it does not animate in the viewport.

### System Information (OS, Hardware)
Windows 10

### Package Versions
USD-23.08, Python 3.10

### Build Flags
Default

jesschimein:
Filed as internal issue #USD-8624

spiffmon:
Oh Shazam, @mati-nvidia - good find, and reinforcement that we need to find time to turn all of our doc examples into unit tests.  It looks like when we made the changes to "strictify" the application of SkelBindingAPI, the bindings discovery process lost the ability to find animation bindings inside/under Skeleton prims.  If I simply move the `skel:animation` binding up onto `</Model>` then all works as expected.  We'll get this fixed for 23.11... give a shout if this is a pattern that's being used commonly - looks like _we_ do not use it anywhere except in this example!

mati-nvidia:
I checked around and, for example, currently our Blender exporter sets the `skel:animationSource` on the Skeleton prim, but also correctly applies the `SkelBindingAPI` on the Skeleton prim so it plays fine in usdview.

@spiffmon is there some code change you were planning on making for this issue? If it's just a matter of fixing the example by moving `skel:animationSource` to the SkelRoot, I can include that in some other doc improvements I'm working on.

spiffmon:
Oh hah!  I *totally* missed that the Skeleton was lacking the schema
application.. but sure enough:
a) usdchecker complains about it - yay!
b) adding it fixes the example.

Sure, @mati-nvidia, if you were already making a PR, this would be an
appreciated fix.

--spiff

tallytalwar:
usdfixbrokenpixarschemas should also fix the application of UsdSkelBindingAPI: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/bin/usdfixbrokenpixarschemas/usdfixbrokenpixarschemas.py#L51

PR #2697, opened 2023-09-15
Improve UsdSkel docs

mati-nvidia:
### Description of Change(s)
- Fix UsdSkel docs code/usda samples.
- Add more info to UsdSkel docs for clarity.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2621

### DIFF
--- a/pxr/usd/usdSkel/doxygen/apiIntro.dox
+++ b/pxr/usd/usdSkel/doxygen/apiIntro.dox
@@ -744,10 +744,12 @@ down into its component parts.
 
 - C++:
 \code{.cpp}
+using namespace pxr;
+
 bool
 WriteAnimatedSkel(
     const UsdStagePtr& stage,
-    const SdfPath& skelPath,
+    const SdfPath& skelRootPath,
     const SdfPathVector& jointPaths,
     const std::vector<GfMatrix4d>& rootTransformsPerFrame,
     const std::vector<VtMatrix4dArray>& jointWorldSpaceTransformsPerFrame,
@@ -762,11 +764,14 @@ WriteAnimatedSkel(
     if (bindTransforms.size() != jointPaths.size())
         return false;
 
-    UsdSkelSkeleton skel = UsdSkelSkeleton::Define(stage, skelPath);
-    if (!skel) {
-        TF_WARN("Failed creating a Skeleton prim at <%s>.", skelPath.GetText());
+    UsdSkelRoot skelRoot = UsdSkelRoot::Define(stage, skelRootPath);
+    if (!skelRoot) {
+        TF_WARN("Failed creating a Skeleton prim at <%s>.", skelRootPath.GetText());
         return false;
     }
+
+    UsdSkelSkeleton skel = UsdSkelSkeleton::Define(stage, skelRootPath.AppendChild(TfToken("Skel")));
+    
 
     const size_t numJoints = jointPaths.size();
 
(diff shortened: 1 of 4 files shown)


pxr/usd/usdSkel/doxygen/apiIntro.dox
```diff
@@ -795,10 +800,10 @@ WriteAnimatedSkel(
     }
 
     UsdSkelAnimation anim = UsdSkelAnimation::Define(
-        stage, skelPath.AppendChild(TfToken("Anim")));
+        stage, skel.GetPath().AppendChild(TfToken("Anim")));
 
     UsdSkelBindingAPI binding = UsdSkelBindingAPI::Apply(skel.GetPrim());
-    binding.CreateSkeletonRel().SetTargets(
+    binding.CreateAnimationSourceRel().SetTargets(
```
mati-nvidia, line 806: Bug fix.

pxr/usd/usdSkel/doxygen/schemas.dox
```diff
@@ -76,10 +76,10 @@ name and order joints in vectorized data. For example:
 
 \code
 def SkelAnimation "Anim" {
-    uniform token[] = ["A", "A/B"]
+    uniform token[] = ["A/B", "A"]
 }
 def Skeleton "Skel" {
-    uniform token[] = ["A/B", "A"]
+    uniform token[] = ["A", "A/B"]
```
mati-nvidia: Updated to correctly list parent joints before child joints in the Skeleton. Switched the order in SkelAnimation too just to still show that the joint order can be different between the two prims.

pxr/usd/usdSkel/doxygen/schemas.dox
```diff
@@ -530,12 +536,51 @@ using the UsdSkelBindingAPI. The _jointIndices_ primvar provides an array giving
 the joint index of an influence, while the _jointWeights_ primvar provides a
 weight value corresponding to each of those indices.
 
+For example:
```
mati-nvidia, line 546: This was a missing example. I made it up based on the explanation paragraph below.

pxr/usd/usdSkel/doxygen/skinnedArm.usda
```diff
@@ -12,7 +12,10 @@ def SkelRoot "Model" (
     prepend apiSchemas = ["SkelBindingAPI"]
 )
 {
-    def Skeleton "Skel" {
+    def Skeleton "Skel" (
```
mati-nvidia, line 15: Bug fix for new strict applied schemas.

pxr/usd/usdSkel/doxygen/apiIntro.dox
```diff
@@ -744,10 +744,12 @@ down into its component parts.
 
 - C++:
 \code{.cpp}
+using namespace pxr;
+
 bool
 WriteAnimatedSkel(
     const UsdStagePtr& stage,
-    const SdfPath& skelPath,
+    const SdfPath& skelRootPath,
```
mati-nvidia, line 752: I think it'd be good to include SkelRoot in this example.

pxr/usd/usdSkel/doxygen/schemas.dox
```diff
@@ -218,19 +218,25 @@ as the authored _joints_ array. The effect of a skel animation prim may
 also be directly nullified by either deactivating the primitive, or by blocking
 the component attributes.
 
+\subsection UsdSkel_SkelAnimation_Blendshapes Skel Animation Schema: Blend Shape Animation
```
mati-nvidia, line 228: New subsection. Just better for scannability and linking.

jesschimein:
Filed as internal issue #USD-8711

pxr/usd/usdSkel/doxygen/apiIntro.dox
dsyu-pixar: I could be wrong, but I believe the repeated section of code down near line 1081 needs to be updated to use CreateAnimationSourceRel() instead of CreateSkeletonRel() as you did here (and similarly for the Python snippet). I'm working on merging this change in, so I'll update accordingly on my end.

mati-nvidia: Good catch. I missed the fix in the breakdown section. I went ahead and updated it in a new commit along with another suggestion from Spiff on the AOUSD forum.

PR #2699, opened 2023-09-18
Use unqualified calls to `get_pointer` for `Tf{Ref,Weak}Ptr`

nvmkuruc:
### Description of Change(s)

Rely on ADL to pick up on the `get_pointer` implementations for `Tf` pointer types in the `PXR_NS` without using `boost::get_pointer` as an intermediary.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8715

PR #2251, opened 2023-02-08
Replace `boost::totally_ordered` with explicit operator overloads for `TfEnum`

nvmkuruc:
### Description of Change(s)
- Add explicit implementations of `operator<=`, `operator>`, `operator>=`, and `operator!=` to `TfEnum`
- Update doxygen comments so `operator!=`'s references `operator==` and  `operator<=`, `operator>`, and `operator>=` reference `operator<`
- Add operator test coverage for `TfEnum` 

### Fixes Issue(s)
- #2250 

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2701, opened 2023-09-18
Replace `boost::totally_ordered` and `boost::equality_comparable` in python bindings

nvmkuruc:
### Description of Change(s)
`boost::totally_ordered` and `boost::equality_comparable` have been removed from non-python binding related usage in several previous changes (such as #2251). This extends their replacement with explicit operator overloads for the python bindings.

### Fixes Issue(s)
- #2250
- #2236 


jesschimein:
Filed as internal issue #USD-8717

PR #2702, opened 2023-09-18
Replace `boost::optional` with `std::optional` for `SdfCopySpec`

nvmkuruc:
### Description of Change(s)
This updates `SdfCopySpec` to use `std::optional` instead of `boost::optional` as well as its usage in `usdUtils` and `sdffilter`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8718

PR #2705, opened 2023-09-18
Replace `boost::variant` with `std::variant` in `pxr/base/js`

nvmkuruc:
### Description of Change(s)
This replaces `boost::variant` and `boost::get` with `std::variant` and `std::get` for `JsValue`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8721

PR #2706, opened 2023-09-18
Replace `boost::any` with `std::any` in `base/js`

nvmkuruc:
### Description of Change(s)
This replaces `boost::any` and `boost::any_cast` with `std::any` and `std::any_cast` for `testJsConverter`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8722

PR #2707, opened 2023-09-18
Replace `boost::optional` with `std::optional` in `trace` and `js`

nvmkuruc:
### Description of Change(s)

This PR changes `JsOptionalValue` from `boost::optional` to `std::optional` and updates its usage in `pxr/base/trace`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8723

PR #2708, opened 2023-09-18
Replace `boost::variant` with `std::variant` in `pxr/base/trace`

nvmkuruc:
### Description of Change(s)
This replaces `boost::variant`, `boost::apply_visitor`, and `boost::get` with `std::variant`, std::visit`, and `std::get`.  This also removes `boost::static_visitor` as it's not required for `std::visit`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8724

PR #2709, opened 2023-09-18
Replace `boost::optional` with `std::optional` in `usdGeom`

nvmkuruc:
### Description of Change(s)
This replaces `boost::optional` with `std::optional` in `bboxCache.h`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8725

PR #2710, opened 2023-09-18
Replace `boost::variant` with `std::variant` in `usdGeom`

nvmkuruc:
### Description of Change(s)
Replace `boost::variant`, `boost::apply_visitor`, and `boost::get` with `std::variant`, `std::visit`, and `std::get`.  It also removes `boost::static_visitor` as its not needed for `std::variant`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8726

PR #2713, opened 2023-09-18
Make `sstream` dependency explicit in `hgiMetal`

nvmkuruc:
### Description of Change(s)
`hgiMetal` is transitively including `sstream` through one of the headers slated for removal in #2684. Making the dependency explicit will unblock #2684.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8730

PR #2714, opened 2023-09-19
Replace `boost::ptr_set` with `std::set<std::unique_ptr>>`

nvmkuruc:
### Description of Change(s)

This replaces `boost::ptr_set` with a `std::set` of `unique_ptr`s. To achieve this, `insert` has been replaced by `emplace` and a custom compare function has been created to allow `_Node` and `std::unique_ptr<_Node>` to compared by the value of the node's key interchangeably.

This appears to fix a bug in the recovery when the wrong node was located by key. Previously, when the wrong node was released from the pointer set, `this` was re-inserted instead of the released node.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8731

PR #2715, opened 2023-09-19
Replace `boost::optional` with `std::optional` for `sdf` list operations

nvmkuruc:
### Description of Change(s)
This PR replaces `boost::optional` with `std::optional` for list operations in `sdf` and its downstream dependencies in `usd/flattenUtils.cpp` and `usdUtils/stitch.cpp`.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8736

PR #2716, opened 2023-09-19
Remove unused `boost/optional.hpp` header from `usdSkel` and `trace`

nvmkuruc:
### Description of Change(s)
#2703 removed usage of `boost::optional` but didn't remove the header include in `cacheImpl.h`.  This completes that change.

This also removes the optional header from `trace/eventTree.cpp` as it is currently unused.

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8737

PR #2717, opened 2023-09-20
[usdview] Add the option to step through authored samples in usdview.

nxkb:
### Description of Change(s)
+ Adds a new feature to `usdview` that allows users to step through authored samples instead of a fixed time step
+ Adds a utility method to `UsdStage` to get all of the authored samples across all prims of a stage. A wrapped command is also made available to python.

Credit to @sirpalee for the original implementation :)

### Use Cases
This has been an essential feature at our studio for debugging sample related issues in caches. The feature really shines when using renderers with "frame relative samples". 
- If artists need to troubleshoot any motion blur issues, they can use this mode to play through an export and quickly determine if the issue has to do with a bad cache sample or the interpolation between samples. 
- It is also a very handy way to quickly determine where the samples exist in a cache using the timeline. This can be a lot more fluid and intuitive for artists than inspecting the attribute data.

### Example Screenshots
Screenshot 2023-09-19 at 5 55 34 PM
Here I am loading in a cache with frame relative samples at -.25, 0, and +.25. 
With the feature disabled (default), you can scrub by a constant frame step and see the .5 frame (which does not have an actual sample). 

Screenshot 2023-09-19 at 5 56 12 PM
With the feature enabled, you can scrub through and only see authored frames (If you input 1005.5 it will be rounded to 1005.75). 

Let me know what you think, and whether we can contribute this back to the core.

### DIFF
--- a/pxr/usd/usd/stage.cpp
+++ b/pxr/usd/usd/stage.cpp
@@ -9110,6 +9110,30 @@ UsdStage::SetFramesPerSecond(double framesPerSecond) const
     SetMetadata(SdfFieldKeys->FramesPerSecond, framesPerSecond);
 }
 
+std::vector<double>
+UsdStage::GetTimeSamplesFromPrims()
+{
+    // Since we are doing tons of lookups and relatively small amounts of pushes
+    // a sorted vector and lower_bound is way more efficient
+    // than push_backs and std::finds or std::sets or anything
+    std::vector<double> ret;
+    for (auto prim: Traverse()) {
+        for (const auto& attr: prim.GetAuthoredAttributes()) {
+            static thread_local std::vector<double> samples;
+            attr.GetTimeSamples(&samples);
+            for (const auto& sample: samples) {
+                const auto it = std::lower_bound(ret.begin(), ret.end(), sample);
+                if (it == ret.end() || *it != sample) {
+                    ret.push_back(sample);
+                    std::sort(ret.begin(), ret.end());
+                }
+            }
+        }
+    }
+    return ret;
+}
+
+
 void 
 UsdStage::SetColorConfiguration(const SdfAssetPath &colorConfig) const
 {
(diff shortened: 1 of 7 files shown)


jesschimein:
Filed as internal issue #USD-8738

pxr/usd/usd/stage.cpp
```diff
@@ -9110,6 +9110,30 @@ UsdStage::SetFramesPerSecond(double framesPerSecond) const
     SetMetadata(SdfFieldKeys->FramesPerSecond, framesPerSecond);
 }
 
+std::vector<double>
+UsdStage::GetTimeSamplesFromPrims()
```
nxkb, line 9114: Some notes on this method:
1. The main argument for having this method here is to keep it near the other `_GetTimeSamples*()` family of methods. But we could also put it somewhere else, maybe as a utility, since it's use case might not be fully generic and its cost is a bit high due to the stage traversal. However, accessing it from the stage is pretty convenient.
2. Currently this doesn't take any arguments, since that wasn't needed for our feature. It may make sense to provide a few more options if this method stays here... For example, we could add an option for a `Usd_PrimFlagsPredicate` to guide the stage traversal.

spiffmon:
Hi @nxkb !  This definitely seems like a useful feature - I have a few concerns/questions, though:

1. The method added to UsdStage could be _very_ expensive, and is easily parallelizable by prim, with proper concurrent gathering of results
2. Although this could change with the future introduction of constraints and other "non-local" dependencies, does it seem more useful that you'd want this analysis for a selected prim and its ancestors **only**, rather than always considering samples from all prims in the scene?  I realize that would complicate the use of the results in usdview, but wanted to explore the space
3. I'm concerned about the fact that once computed, the set-of-all-timeSamples is fixed.  If you change stage load-state or switch a variant (or of course mutate the stage), that could change.  usdview _can_ listen for authoring changes (including variant selections made in the UI), which should also cover the load/unload case, as that triggers ObjectsChanged notices.

PR #2718, opened 2023-09-20
Replace `boost::variant` with `std::variant` in `sdf`

nvmkuruc:
### Description of Change(s)
This replaces `boost::variant` with `std::variant` in `sdf`.  The API uses `boost::bad_get` to communicate a failure to parse due to a bad variant access or an out of range error.  This change remaps the STL `std::bad_variant_access` to a `boost::bad_get` to preserve this behavior.  As visitors used via `std::visit` do not specify `result_type`, the return type for several helper methods has been converted to `auto`.

One or more follow up changes will be needed to remove a spurious include of `boost/variant.hpp` from the parser logic and replace `boost::bad_get` with a new proposed `Sdf_BadGet` exception.

### Fixes Issue(s)

### DIFF
--- a/pxr/usd/sdf/namespaceEdit.cpp
+++ b/pxr/usd/sdf/namespaceEdit.cpp
@@ -31,10 +31,10 @@
 #include "pxr/base/tf/stringUtils.h"
 
 #include <boost/ptr_container/ptr_set.hpp>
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-8739

PR #2724, opened 2023-09-28
Autodesk: Support float type matrix (GfMatrix4f) for instanceTransform

erikaharrison-adsk:
### Description of Change(s)

Support VtArray<GfMatrix4f> as the value type for token instanceTransform

A instancer may contain huge number of matrices and occupy lots of memory. Some data models only need float precision
matrices for instancers, however USD only accepts VtArray<GfMatrix4d>, doubling the memory size needed.
To save the memory, this fix allows USD to accept instancer matrices of type VtArray<GfMatrix4f> directly.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/instancer.cpp
+++ b/pxr/imaging/hdSt/instancer.cpp
@@ -83,19 +83,30 @@ HdStInstancer::_SyncPrimvars(HdSceneDelegate *sceneDelegate,
         VtValue value = sceneDelegate->Get(instancerId, primvar.name);
         if (!value.IsEmpty()) {
             HdBufferSourceSharedPtr source;
-            if (primvar.name == HdInstancerTokens->instanceTransform &&
-                TF_VERIFY(value.IsHolding<VtArray<GfMatrix4d> >())) {
-                // Explicitly invoke the c'tor taking a
-                // VtArray<GfMatrix4d> to ensure we properly convert to
-                // the appropriate floating-point matrix type.
-                HgiCapabilities const * capabilities =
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8762

erikaharrison-adsk:
@jesschimein - I notice this is marked as "merged" but I'm not seeing the changes in instancer.cpp nor its commit history. How does this work? Is there a delay? https://git.autodesk.com/autodesk-forks/USD/commits/dev/pxr/imaging/hdSt/instancer.cpp

jesschimein:
Hi Erika!  I'm also seeing this change in dev in instancer.cpp and in
history.  Let me know if you're still not seeing it?

wingfiring:
Hi @jesschimein,

I noticed that the following statements are removed from the merge commit

image

The original conditional expression in if and the else branch are 
```
if ((primvar.name == HdInstancerTokens->instanceTransform ||
                 primvar.name == HdInstancerTokens->instanceTransforms) &&
                TF_VERIFY(value.IsHolding<VtArray<GfMatrix4d> >())) {
...
}
else {
                source.reset(new HdVtBufferSource(primvar.name, value));
```
if `TF_VERIFY(value.IsHolding<VtArray<GfMatrix4d> >())` is false, the control flow will jump to else branch. In my PR, I moved the value.IsHolding into if branch, so I added the else branch as above to do the same thing as original. It doesn’t affect our current work, but it might affect extensibility (for example, set a non-matrix data and would be consumed by customized render delegate).
Could you please confirm that the change is intentional? Thanks a lot!

PR #2733, opened 2023-10-13
usdGeom: avoid de-referencing iterators of an empty set when validating subset families with no indices

mattyjams:
This avoids undefined behavior when `UsdGeomSubset::ValidateFamily()` is called to validate a subset family that does not actually have any indices authored. In this case, we wind up with an empty set and the index bounds checking would try to de-reference iterators on it.

I think subset families with no indices would be considered invalid, so the function now deems them as so and adds a descriptive reason. If that's not appropriate, I can easily remove that part and only skip the bounds checking.

I also added newline characters to two of the existing reason messages to ensure that each reason is on its own line in the case of multiple failures.

### DIFF
--- a/pxr/imaging/hdSt/CMakeLists.txt
+++ b/pxr/imaging/hdSt/CMakeLists.txt
@@ -505,18 +505,6 @@ pxr_build_test(testHdStPrimvars
     CPPFILES
         testenv/testHdStPrimvars.cpp
 )
-if (PXR_ENABLE_PTEX_SUPPORT)
-pxr_build_test(testHdStPtex
-    LIBRARIES
-        hdSt
-        hd
-        sdf
-        tf
-        gf
-    CPPFILES
-        testenv/testHdStPtex.cpp
-)
-endif()
 pxr_build_test(testHdStQuadrangulation
     LIBRARIES
         hdSt
@@ -542,16 +530,6 @@ pxr_build_test(testHdStSamplerObjectRegistry
         tf
     CPPFILES
         testenv/testHdStSamplerObjectRegistry.cpp
-)
-pxr_build_test(testHdStShaders
-    LIBRARIES
-        hdSt
-        hd
-        sdr
-        tf
-        gf
-    CPPFILES
-        testenv/testHdStShaders.cpp
 )
 pxr_build_test(testHdStSubdivision
     LIBRARIES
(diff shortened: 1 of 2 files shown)


pxr/usd/usdGeom/subset.cpp
```diff
@@ -622,14 +622,24 @@ UsdGeomSubset::ValidateFamily(
             }
         }
 
+        if (indicesInFamily.empty()) {
+            valid = false;
```
mattyjams: I don't *think* there's any utility in a subset family with no indices, so they get marked invalid here, but I'm happy to take this out if anyone thinks otherwise.

The important part is that we avoid the de-reference of the iterators on lines 649 and 659 below.

spiffmon: Oh wow, @mattyjams ... in looking over the code above here to answer your question, I realized the logic here (reviewed by me!) is deeply problematic (and `ValidateSubsets()` above has a similar problem) in that it assumes that the `faceVertexIndices` property is not animated, which is a bad assumption.  _This_ method is already set up pretty well to accommodate the proper logic, though `ValidateSubsets()` is trickier because it is relying on the caller to pass in "the" faceCount, then doing its own iteration over `indices` timeSamples.

But to your question, I don't think it should necessarily be invalid to have no participating indices... it was the case of animation and varying topology that I was thinking about.  It seems perfectly valid to me that you might have a Family that _some_ of the time has Subsets that modify the base Mesh, but at some other points in time, leave it unmodified?  Correctness is the first goal of validation, but alerting the user to possibly surprising/unintended things is also useful, so maybe it makes sense to consider the Family invalid if there are **no** points in time at which there are non-empty (valid) indices in one or more Subsets?

If you'd be willing to fix up at least this method wrt varying topology, we'd be eager to take it in, as we have a project coming up to extend the allowable `elementTypes` for the physics community.

Thanks alot, @mattyjams !

mattyjams: Thanks for the info @spiffmon!

I took a more aggressive pass at fixing up `ValidateFamily()` so that it properly accounts for varying topology. The original fix to avoid de-referencing iterators of an empty set is in e648a6a057e22d73580d5581967f6226b74df378, and the bulk of the interesting new stuff is in 4d32257b15407f4794598d69366196fcb79fa7c9, but I broke up the commits so you can see the progression along the way.

One other issue I noticed was that for invalid subsets where *all* of the indices are negative, we would properly identify it as invalid but append an incorrect reason due to some integer overflow casting an `int` to a `size_t`. The fix for that is in 7bea2ff235f3165f22c444628c1c113b54f6072e.

Let me know how this is looking now!

Thanks!

jesschimein:
Filed as internal issue #USD-8832

mattyjams:
Looks like this was merged as of 6575a265a73de63a50559b22626ee723e3ba2059, and I see it was already extended in da4df841ce11d56dd2f50862bec37483a36357a5 to add support for point subsets, which is awesome.

Thanks all!

PR #2734, opened 2023-10-14
Fix timeVaryingBits for lights, light filters, and cameras for timesampled transform attribute

sdawson-nvidia:
### Description of Change(s)

Change calls to `UsdGeomXformable::IsTransformationAffectedByAttrNamed` to use `GetName` instead of `GetBaseName` so the namespace can be checked against.

### Fixes Issue(s)
-

Tested on Windows, these tests fail:
The following tests FAILED:
        430 - testUsdZipFile (Failed)
        678 - testUsdviewNavigationKeys (Failed)
        694 - testUsdResolverExample (Failed)
Errors while running CTest
Based on other issues, these seem to be failing for a while, on windows. testUsdZipFile fails as an example due to windows line endings vs unix line endings in the write portion of the test (crc doesn't match the expected value).


jesschimein:
Filed as internal issue #USD-8833

sdawson-nvidia:
The failure seems to be a timeout and likely unrelated - how do I rerun the checks?

PR #2553, opened 2023-07-28
Autodesk: Stabilizing HgiVulkan

erikaharrison-adsk:
### Description of Change(s)

This PR is an implementation to providing stability to HGIVulkan. Certain broken features have been disabled when Vulkan is enabled. Fixes for them will be followed in subsequent implementations.

For simplifying the merge process, there are dedicated sub-branches for
1. [Vulkan Build Support](https://github.com/autodesk-forks/USD/tree/adsk/feature/hgivulkan_build_support)
2. [Hgi Backend Selection](https://github.com/autodesk-forks/USD/tree/adsk/feature/hgivulkan_backend_selection)
3. [Unit test support for HgiVulkan](https://github.com/autodesk-forks/USD/tree/adsk/feature/hgivulkan_unit_test_support)

Please note, some changes related to hgiVulkan bugs and fixes are local to this branch only. 

Please refer to proposal ([PR-15](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/15)) for more details.

### Fixes Issue(s)
- N/A

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #2736, opened 2023-10-16
Autodesk: Hgi Backend Selection

erikaharrison-adsk:
### Description of Change(s)

So far, HgiGL has been the default render backend on Windows and Linux. Now that there is Vulkan support introduced into the stack, through this PR users can choose a render backend at initialization time. This PR provides relevant support to render delegate creation as well.

Note: This PR is child of https://github.com/PixarAnimationStudios/OpenUSD/pull/2553. Please refer it for more context.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-8836

brechtvl:
We ran into this problem in Blender, where it tries to initialize OpenGL while we want Vulkan. The environment variable `HGI_ENABLE_VULKAN` almost lets us work around it, but it gets read on program startup before we can override it.

Is there any plan to merge this pull request, or an alternative solution? The implementation makes sense to me.

spiffmon:
@brechtvl , really wish github had a "dupe" mechanism.  Internally we duplicated and closed this PR in favor of https://github.com/PixarAnimationStudios/OpenUSD/pull/2850, which was merged about a year ago.  Can you give that a try and see if it works for you?

brechtvl:
Ah, thanks! Though it seems that was merged into the `feature-hgi-vulkan` branch, not an official release?

brechtvl:
Or rather, only part of the PR was merged. The part in `pxr/imaging/hd/rendererPlugin.cpp` where it gets `HgiBackend` from settingsMap is missing.

clach:
We decided not to move forward with that aspect of the change (you can even see a comment about that specific part in PR [#2850](https://github.com/PixarAnimationStudios/OpenUSD/pull/2850) [here](https://github.com/PixarAnimationStudios/OpenUSD/pull/2850/files#r1528886800)). We did include the new function `CreateNamedHgi` that allows you to create a specific Hgi backend based on what token you provide, which may be of use in your situation.

brechtvl:
We did try `CreateNamedHgi`, but it's not enough. The problem is that Storm calls `Hgi::IsSupported()` without specifying the backend. Since it defaults to OpenGL and we have no OpenGL context when using Vulkan, that check fails.

I could try implementing a different solution, but it's not clear how it should be done. I didn't understand the comment about doing it at the `UsdImagingGLEngine` level, since as far as I can tell that's not available at the moment the `HdRendererPlugin::IsSupported` check needs to run.

Maybe a slight improvement would be to add a `settingsMap` argument to `HdRendererPlugin::IsSupported`, instead of a `hgiToken`. So that it's at least easier to make it consistent with `HdRendererPlugin::CreateRenderDelegate`. If you wanted to force them to be consistent more strongly, it could be a required argument instead of an optional one. Perhaps the `gpuEnabled` argument could be replaced by an entry in `settingsMap` too. But then it all becomes quite API breaking.

brechtvl:
I proposed an alternative solution in #3604, which hopefully sidesteps the concerns.

spiffmon:
We are hoping to have a solution for this in the next release - thanks for your patience!

PR #2740, opened 2023-10-17
[build] make it possible to import build_usd.py without executing it

pmolodo:
### Description of Change(s)

Allow importing of `build_usd.py` without immediately executing.

This opens a lot doors for build tools to either query some of the data in this file, or alter it in some way before execution, providing a lot of additional flexibility.

**NOTE FOR EASIER REVIEW**
I highly recommend turning on the option to ignore / hide whitespace differences when viewing the diff for this, as the vast majority of changes are just and additional level of indentation.

github_hide_whitespace


jesschimein:
Filed as internal issue #USD-8841

pmolodo:
Though I still think this would be a useful change in general, it no longer aligns with our needs at NVIDIA, so I'm closing it.

Issue #2741, opened 2023-10-17
Use Boost 1.82 for Python 3.11


dgovil:
USD can't build for Python 3.11 due to older Boost version usage. 

We should use Boost 1.82 when detecting Python 3.11 or higher in accordance with VFX Platform 2023

jesschimein:
Filed as internal issue #USD-8842

ze897c:
Folks, sorry for comment out of the blue. I am hacking (badly) build_usd.py to build for 3.11 for various reasons. On OSX I was able to change boost to 83 & build/run (mostly) just fine. Tried same on my ubuntu box: boost built (as yet untested) but thier is an issue building USD. Some digging leads to error with `sysconfig`. Call to `sysconfig.get_config_var("LDLIBRARY")` gives, 'libpython3.11.a', which does not exist. `sysconfig.get_makefile_filename()` gives `<path-to-conda-env>/lib/python3.11/config-3.11-x86_64-linux-gnu/Makefile' and the (auto-generated) Maklefile is wrong (LDLIBRARY=      libpython$(VERSION)$(ABIFLAGS).a).

This is just FYI/heads-up, and a note to keep me in-looped.

Thanks for the libs, all!

sunyab:
Hi @ze897c, which version of Ubuntu are you running? And what method did you use to install Python 3 -- did you go through `apt` or are you using `conda` or some other package manager?

ze897c:
Hello,Installed 3.11 through conda, while creating a new env. ubuntu```bash$ lsb_release -aNo LSB modules are available.Distributor ID:	UbuntuDescription:	Ubuntu 20.04.6 LTSRelease:	20.04Codename:	focal```Cheers,XeateOn Oct 24, 2023, at 12:08 PM, Sunya Boonyatera ***@***.***> wrote:Hi @ze897c, which version of Ubuntu are you running? And what method did you use to install Python 3 -- did you go through apt or are you using conda or some other package manager?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

nvmkuruc:
@dgovil This seems like it's been addressed.  Do you think we can close?

PR #2742, opened 2023-10-18
Enable Python 3.11 builds with Boost 1.82

dgovil:
### Description of Change(s)

This PR resolves https://github.com/PixarAnimationStudios/OpenUSD/issues/2741 by using Boost 1.82 for Python 3.11 or above, in accordance with VFX Platform CY2024.

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/OpenUSD/issues/2741

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -717,7 +717,7 @@ BOOST_VERSION_FILES = [
 def InstallBoost_Helper(context, force, buildArgs):
     # In general we use boost 1.76.0 to adhere to VFX Reference Platform CY2022.
     # However, there are some cases where a newer version is required.
-    # - Building with Python 3.10 requires boost 1.76.0 or newer.
+    # - Building with Python 3.10 requires boost 1.76.0 or newer, while 3.11 requires 1.82.
     #   (https://github.com/boostorg/python/commit/cbd2d9)
     #   XXX: Due to a typo we've been using 1.78.0 in this case for a while.
     #        We're leaving it that way to minimize potential disruption.
@@ -727,7 +727,9 @@ def InstallBoost_Helper(context, force, buildArgs):
     #   compatibility issues on Big Sur and Monterey.
     pyInfo = GetPythonInfo(context)
     pyVer = (int(pyInfo[3].split('.')[0]), int(pyInfo[3].split('.')[1]))
-    if context.buildPython and pyVer >= (3, 10):
+    if context.buildPython and pyVer >= (3, 11):
+        BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.82.0/source/boost_1_82_0.zip"
+    elif context.buildPython and pyVer >= (3, 10):
         BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.zip"
     elif IsVisualStudio2022OrGreater():
         BOOST_URL = "https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.zip"


jesschimein:
Filed as internal issue #USD-8843

sunyab:
@dgovil Hey Dhruv, I'm looking at merging this in but I'm running into issues building OpenImageIO on my Linux machine with this change. I get a bunch of undefined reference errors with boost::filesystem, like below. Do you see anything similar on your machine?

```
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::detail::directory_iterator_construct(boost::filesystem::directory_iterator&, boost::filesystem::path const&, unsigned int, boost::system::error_code*)'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::extension() const'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::root_directory() const'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::parent_path() const'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::detail::dir_itr_close(void*&, void*&)'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::replace_extension(boost::filesystem::path const&)'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::filename() const'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: ../../lib/libOpenImageIO_Util.so.2.3.21: undefined reference to `boost::filesystem::path::operator/=(boost::filesystem::path const&)'
```

dgovil:
Oh interesting. I must have built without OIIO. I'll check that out today and get back to you

dgovil:
Still looking for OIIO, but I was hitting an issue when I updated my Toolchain and Boost's fallback to C++03 was using deprecated symbols. I've set Boost to Cxx17 on macOS, though I think at some point we should do that across the board? I don't have a Win/Linux build setup to test if they also need 17 or not.

dgovil:
@sunyab , at least on macOS with Clang 15, I'm not getting a compile error with OIIO. I did have to make a change to Boost for macOS to use C++17 and I wonder if adding those cxxflags to Linux as well would fix things for you?

sunyab:
I was able to get past my issue by updating my version of CMake to latest. I was running 3.25.3 which didn't have explicit support for boost 1.82. 

So what I'd propose is:
- File a separate PR for the boost C++17 changes.

- For this PR, let's go back to what you had originally, but also add a check for CMake version here so that users are required to have 3.27 or above in this case: https://github.com/PixarAnimationStudios/OpenUSD/blob/0b18ad3f840c24eb25e16b795a5b0821cf05126e/build_scripts/build_usd.py#L2275

If you'd like, I'd be happy to take this off your plate and make these changes locally since I'm already in the middle of testing builds anyway. What do you think?

dgovil:
Ah interesting. Yeah, if you're already in there, it's probably faster if you do it since I won't be able to adequately test Linux. Thank you!

Split out the CXX17 PR here https://github.com/PixarAnimationStudios/OpenUSD/pull/2868

sunyab:
I've checked in an updated version of this PR internally and it'll go out with the next updates to the dev branch. Closing this out, thanks!

dgovil:
Thanks, Sunya!

PR #2748, opened 2023-10-19
Replace `BOOST_PP_SEQ` defined stream overloads in `vt/wrapArray.h`

nvmkuruc:
### Description of Change(s)
`wrapArray.h` provided overloads for `streamValue` so that integral and floating point values don't have to go through `TfPyRepr` as an optimization. Overloads were defined by `BOOST_PP_SEQ` operating on a list of floating point and integral type sequences.

This PR uses `if constexpr` and `TfMetaList` to provide a boost free implementation without SFINAE.

This additionally changes `_IsFinite` to be `inline`, as the `static` `GfHalf` overload was generating unused function warnings.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8857

PR #2751, opened 2023-10-20
Add locale independent `TfStringToLowerAscii`

nvmkuruc:
### Description of Change(s)
To more robustly support UTF-8 encoding of strings, this promotes a utility function from `fileFormatRegistry.cpp` used for locale independent case folding to `pixar/base/tf/stringUtils.h` as `TfStringToLowerAscii`. This will replace `TfStringToLower` in many cases where one of the strings is known to be ASCII (or UTF-8 where a specification explicitly callouts of only casefolding of `[A-Z]`).

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/stringUtils.cpp
+++ b/pxr/base/tf/stringUtils.cpp
@@ -1201,4 +1201,16 @@ TfGetXmlEscapedString(const std::string &in)
     return result;
 }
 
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-8862

PR #2752, opened 2023-10-20
Replace `BOOST_PP_ITERATE` with templates in `makePyConstructor.h`

nvmkuruc:
### Description of Change(s)
`BOOST_PP_ITERATE` was being used to expand macros into implementations of the various `TfMakePy...` utilities. Many of these can be replaced by variadic argument templates with minimal changes.  `InitCtorWithVarArgs` used by `TfMakePyCtorWithVarArgs` required adding a little indirection to create an `std::index_sequence` capable of splitting the boost python args tuple into the C++ constructor named positional args and unnamed variable length args.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8863

PR #2212, opened 2023-01-27
Replace boost traits and mpl with STL traits in pxr/base/vt

nvmkuruc:
### Description of Change(s)
Depends on #2214
- `std::is_trivially_default_constructible`, `std::is_trivially_copyable`, `std::is_trivially_assignable`, and `std::is_trivially_destructible` are used to replace `boost::has_trivial_constructor`, `boost::has_trivial_copy`, `boost::has_trivial_assign`, and `boost::has_trivial_destructor`
- Test coverage has been added for `VtValueTypeHasCheapCopy` to guard against regressions

### Fixes Issue(s)
- #2211

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2754, opened 2023-10-24
Replace `std::is_trivially_assignable` with `std::is_trivially_copy_assignable`

nvmkuruc:
### Description of Change(s)
#2212 and #2214 replaced `boost::has_trivial_assign<T>` with `std::is_trivially_assignable<T&, const T&>`.  C++17 has introduced `std::is_trivially_copy_assignable` which more compactly maps to the original `boost::has_trivial_assign`.

This change also replaces some usage of `{TRAIT}::value` with the equivalent `{TRAIT}_v` representation.

### Fixes Issue(s)
-


sunyab:
Filed as internal issue #USD-8876

Issue #2722, opened 2023-09-27
ERROR: Failed to run 'cmake --build . --config Release --target install -- -j16'


chuanyune:
There was a problem with the configuration, and the logs showed the following

[ 69%] Linking CXX shared library libusd_hd.so
[ 72%] Built target hd
[ 72%] Linking CXX shared library _sdf.so
[ 72%] Built target _sdf
gmake: *** [Makefile:136：all] 错误 2

jesschimein:
Filed as internal issue #USD-8751

OlivierArgentieri:
I think i got the same issue:

```
rez env python pyside2 uic PyOpenGL -- python D:\CWDOUSD\OpenUSD\build_scripts\build_usd.py D:\USD --generator "Visual Studio 16 2019" --python
D:\CWDOUSD\OpenUSD\build_scripts\build_usd.py:37: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils

Building with settings:
  USD source directory          D:\CWDOUSD\OpenUSD
  USD install directory         D:\USD
  3rd-party source directory    D:\USD\src
  3rd-party install directory   D:\USD
  Build directory               D:\USD\build
  CMake generator               Visual Studio 16 2019
  CMake toolset                 Default
  Downloader                    curl

  Building                      Shared libraries
    Variant                     Release
    Target
    Imaging                     On
      Ptex support:             Off
      OpenVDB support:          Off
      OpenImageIO support:      Off
      OpenColorIO support:      Off
      PRMan support:            Off
    UsdImaging                  On
      usdview:                  On
    Python support              On
      Python Debug:             Off
      Python docs:              Off
    Documentation               Off
    Tests                       Off
    Examples                    On
    Tutorials                   On
    Tools                       On
    Alembic Plugin              Off
      HDF5 support:             Off
    Draco Plugin                Off
    MaterialX Plugin            On

  Dependencies                  zlib, boost, TBB, MaterialX, OpenSubdiv
STATUS: Installing zlib...
2023-11-15 16:16
cmake -DCMAKE_INSTALL_PREFIX="D:\USD" -DCMAKE_PREFIX_PATH="D:\USD" -DCMAKE_BUILD_TYPE=Release     "D:\USD\src\zlib-1.2.13"
-- Building for: Visual Studio 17 2022
[...]

PR #2755, opened 2023-10-25
fix: use universal -j parameter for cmake build

### LINKED ISSUE #2486: Windows 10 build fail with `MSBUILD : error MSB1001: Unknown switch`
### 
        Python 3.8.10
        cmake 3.26.4
        PySide6

## Command
     python build_usd.py "E:\USD"

[log.txt](https://github.com/PixarAnimationStudios/USD/files/11699622/log.txt)

sssooonnnggg:
### Description of Change(s)
Due to cmake document, `-j` parameter is compatible for all platforms, there is no need to pass different tag to different platform build tools
https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project

### Fixes Issue(s)
This should fix #2486 #2722

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -350,14 +350,7 @@ def AppendCXX11ABIArg(buildFlag, context, buildArgs):
         flag=buildFlag, flags=" ".join(cxxFlags)))
 
 def FormatMultiProcs(numJobs, generator):
-    tag = "-j"
-    if generator:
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-8877

MustafaJafar (commented):
Hey, I faced the same error you mentioned in the description (although I have no idea why). 
I'm just letting you know that your solution worked for me. Thanks.

PR #2760, opened 2023-10-26
Autodesk: OpenVDB build error on RelWithDebugInfo

erikaharrison-adsk:
### Description of Change(s)

Fixes openvdb build error that happens with build type relwithdebuginfo

### Fixes Issue(s)
- N/A

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1231,6 +1231,13 @@ def InstallOpenVDB(context, force, buildArgs):
         # Add on any user-specified extra arguments.
         extraArgs += buildArgs
 
+        # Patch files to support relwithdebinfo and minsizerel build
+        PatchFile('cmake/FindTBB.cmake',
+                  [('        IMPORTED_LOCATION_RELEASE "${Tbb_${COMPONENT}_LIBRARY_RELEASE}")',
+                    '        IMPORTED_LOCATION_RELEASE "${Tbb_${COMPONENT}_LIBRARY_RELEASE}"\n' +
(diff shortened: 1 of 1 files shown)


meshula:
Has a corresponding issue been filed for OpenVdb? Or has this been fixed in a later OpenVdb version?

sunyab:
Filed as internal issue #USD-8881

zhangha182:
> Has a corresponding issue been filed for OpenVdb? Or has this been fixed in a later OpenVdb version?

We didn't file an issue for OpenVdb. As it still exists on its master branch, I don't think it has been fixed in later version.

I agree that this kind of issue is better to be fixed in the upstream, and having it fixed in USD just to unblock typical build types, especially RelWithDebInfo.

meshula:
Ok, independent of resolving this issue, I filed a request at the OpenVdb repo so they are aware.

e4lam:
I suggest that you submit the pull request upstream instead of just as a bug.

zhangha182:
I'll submit the PR in openvdb to address the issue.

erikaharrison-adsk:
https://github.com/AcademySoftwareFoundation/openvdb/pull/1763

Note: I'm unclear whether to close this PR, or instead we can adjust this PR to simply update the new OpenVDB version once its accepted there.

sunyab:
Let's leave this PR open for now, I think we may want to take it anyway. Since the dependencies in build_usd.py are currently tied to the CY2022 VFX Reference Platform, it may be some time before we pull in an official release that contains that fix. So taking this patch seems like a reasonable thing to do in the interim.

My only request is that we add a comment with a link to the OpenVDB issue and PR for reference. Thanks!

PR #2766, opened 2023-10-28
WIP: Port usdchecker from Python to C++

dgovil:
### Description of Change(s)

This is a WIP port of `usdchecker.py` and `complianceChecker.py` to C++. This includes every check as of USD 23.11.
I know Pixar are planning a more architectural rewrite of things, but a need came up so I spent a couple days doing a 1:1 port. I just wanted to post it here so I could see what interest is like in having something before the new checker architecture is ready (also totally happy for this PR to be throwaway work on my part since I don't want to step on any toes or existing ongoing work).

I haven't exposed it to Python yet because I wanted to see if people wanted every Rule exposed or just have some kind of abstract interface over them? The lack of Python bindings is the only thing I think is left on my TODO.

This would also allow https://github.com/PixarAnimationStudios/OpenUSD/pull/2101 to run checks without Python being built.


sunyab:
Filed as internal issue #USD-8887

FlorianZ:
@dgovil, thanks for taking a stab at this. Great work! However, we are currently working in this code, so we aren't ready to accept this PR. Do you mind if we put it on hold for a while? Thank you!!

dgovil:
No worries at all. I expected as much and was just putting it up so it's there even if it's throwaway work, since we needed it for something else.

Issue #2771, opened 2023-10-31
USD bakes full SDK path to INTERFACE_LINK_LIBRARIES in pxrTargets.cmake 


dgovil:
pxrTargets.cmake bakes in the full path to an SDK library during the build which means that the final build isn't portable to build other stuff against unless all the machines have the same SDK versions available. 

Unfortunately that means that if someone upgrades their OS, they won't have the same SDK versions available, and will therefore be unable to compile without modifying the targets file.

It would be great if USD used generic linkage for system libs instead of exact SDK versions

e.g 

```
set_target_properties(arch PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "/Library/Developer/CommandLineTools/SDKs/MacOSX13.5.sdk/usr/lib/libm.tbd"
)
set_target_properties(garch PROPERTIES
  INTERFACE_LINK_LIBRARIES "arch;tf;/Library/Developer/CommandLineTools/SDKs/MacOSX13.5.sdk/System/Library/Frameworks/OpenGL.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX13.5.sdk/System/Library/Frameworks/AppKit.framework"

)
```

jesschimein:
Filed as internal issue #USD-8891

dgovil:
Put up a PR for this

PR #2772, opened 2023-10-31
Use macOS linkage style for frameworks and system libs

dgovil:
### Description of Change(s)

Change macOS platform linkage to not use find_library, so that INTERFACE_LINK_LIBRARIES doesn't hardcode SDK versions

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/2771


cmake/defaults/Packages.cmake
```diff
@@ -227,6 +230,9 @@ if (PXR_BUILD_IMAGING)
             cmake_policy(SET CMP0072 OLD)
         endif()
         find_package(OpenGL REQUIRED)
```
meshula: shouldn't the find_package be removed to avoid getting the full path and also the -l bit?

pxr/imaging/garch/CMakeLists.txt
```diff
@@ -11,7 +11,6 @@ if(APPLE)
     set(GARCH_GLPLATFORMCONTEXT       glPlatformContextDarwin)
     set(GARCH_GLPLATFORMDEBUGWINDOW   glPlatformDebugWindowDarwin)
     set(GARCH_SOURCE_EXTENSION        mm)
-    FIND_LIBRARY(APPKIT_LIBRARY AppKit)
```
meshula, line 14: I think it might make sense to add `set(GARCH_PLATFORM_LIBRARIES "-framework AppKit") here and ...

pxr/imaging/garch/CMakeLists.txt
```diff
@@ -22,13 +21,14 @@ elseif(WIN32)
...
 
+
 pxr_library(garch
     LIBRARIES
         arch
         tf
         ${X11_LIBRARIES}
         ${OPENGL_gl_LIBRARY}
-        ${APPKIT_LIBRARY}
+        "-framework AppKit"
```
meshula: put `${GARCH_PLATFORM_LIBRARIES}` here

meshula:
I see you figured out how to specify frameworks in a non-full-path-name way :)

cmake/defaults/Packages.cmake
dgovil: I'd left it for Linux/Windows and then special cased the specific variable used for macOS. 

I could move the find_package behind a platform check so it never runs for macOS, but it shouldn't be meaningfully different I think?

pxr/imaging/garch/CMakeLists.txt
dgovil: Ah yeah that's a good idea. I'll make the change and update tomorrow.

dgovil:
Updated with your notes

jesschimein:
Filed as internal issue #USD-8896

PR #2682, opened 2023-09-12
Replace `BOOST_PP_IIF` and `BOOST_PP_EXPR_IIF` using `preprocessorUtilsLite.h`

nvmkuruc:
### Description of Change(s)

#2672 provided `_TF_PP_IFF` as an internal macro. This PR decouples array and non-array initialization to remove `BOOST_PP_EXPR_IIF` and replaces all other usage of `BOOST_PP_IIF` with `_TF_PP_IFF`.

### Fixes Issue(s)
-

- [ ] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2778, opened 2023-11-02
Replace `BOOST_PP_IIF` with `_TF_PP_IFF` in python bindings

nvmkuruc:
### Description of Change(s)

#2682 removed the usage of `BOOST_PP_IIF` from the main code base (sans the unused `preprocessorUtils.{h,cpp}`.  This completes its removal by removing it from the python bindings.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8905

PR #2779, opened 2023-11-02
Fix USDZ docs mime type

hybridherbst:
### Description of Change(s)

MIME type was incorrect. Now it matches the IANA registration and the link to that. 
See also https://github.com/google/model-viewer/issues/350#issuecomment-1221176630


spiffmon:
Thank you, @hybridherbst - good catch!

jesschimein:
Filed as internal issue #USD-8906

PR #2780, opened 2023-11-03
Autodesk: [hdSt] Compact drawingcoords interstage attributes

erikaharrison-adsk:
### Description of Change(s)

Pass drawingcoors attributes in ivec instead of single int attributes to save some attribute slots.

Right now, we are beyond the limits of the 16 interstage attributes for WebGPU (https://www.w3.org/TR/webgpu/#dom-supported-limits-maxinterstageshadervariables) but this can be solved easily by just packing them in vectors inbetween stages as it is done for this PR. Shaders already use the intermediate function `GetDrawingCoord`, so this change doesn't affect the logic of other components.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/codeGen.cpp
+++ b/pxr/imaging/hdSt/codeGen.cpp
@@ -1531,7 +1531,7 @@ HdSt_CodeGen::_PlumbInterstageElements(
 
     // Interstage variables of type "int" require "flat" interpolation
     TfToken const &qualifier =
-        (dataType == _tokens->_int) ? _tokens->flat : _tokens->_default;
+        (_GetFlatType(dataType) == _tokens->_int) ? _tokens->flat : _tokens->_default;
 
     // Vertex attrib input for VS, PTCS, PTVS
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8907

tcauchois:
Hi.  This change doesn't seem to be complete; it's referencing some metadata changes that aren't included in the PR.  Can you check that the complete PR is here?  Thanks!

kohakukun:
Hello @tcauchois .

The PR is complete. It only changes how things are accessed in the shader, basically compacting several ints to a couple of ivec4. I'm attaching the shaders generated with and without the change. The important thing is that since we are using GetDrawingCoord in the shader code, we only had to change codeGen.cpp.

This is a couple of images on how they differ in the shader
Vertex:
image
image
image



Fragment:
image




The shaders generated are when rendering a sphere
[shaders.zip](https://github.com/user-attachments/files/15971411/shaders.zip)

erikaharrison-adsk:
Addressed/took a different direction. Related: https://github.com/PixarAnimationStudios/OpenUSD/pull/3170

PR #2781, opened 2023-11-03
Support `TfHashAppend` leveraging of `std::hash` for unspecialized types

nvmkuruc:
### Description of Change(s)

A C++17 friendly implementation of leveraging `std::hash` in `TfHashAppend` had been left commented out. During the removal of `boost::hash`, specializations were added (#2298, #2304) for some types that explicitly forwarded the result of `std::hash` as their `TfHashAppend` implementation.

Now that OpenUSD requires C++17, `TfHashAppend` can hash any type that supports `std::hash` and those specializations can be removed.  To avoid `stdext::hash_value` from being prioritized over `std::hash` on some platforms, the lookup order is reordered so that `std::hash` is second and `hash_value` is last.

This also removes SFINAE for the `std::vector` specialization in favor of a `static_assert` guard against using `AppendContiguous` on `std::vector<bool>`.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -92,9 +92,12 @@ TfHashAppend(HashState &h, std::pair<T, U> const &p)
 
 // Support std::vector. std::vector<bool> specialized below.
 template <class HashState, class T>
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-8911

PR #2180, opened 2023-01-20
Replace boost hash usage with TfHash in pxr/base/sdf

nvmkuruc:
### Description of Change(s)
Requires #2176
To remove the dependency of pxr/base/sdf on boost's hashing functions
* Python bindings for `__hash__` added to `SdfAssetPath`, `SdfPayload`, `SdfListOp`, and `SdfReference`.
* Test coverage for hashing added for `SdfListOp`, `SdfReference`, `SdfPayload`, and `SdfValueTypeName`
* `boost::hash_combine` usage replaced with `TfHash::Combine`

### Fixes Issue(s)
-#2172 (additional PRs forthcoming)

- [ ] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #2785, opened 2023-11-06
Replace `boost::hash` with `TfHash` in `usd/sdf` python bindings

nvmkuruc:
### Description of Change(s)

#2180 replaced `boost::hash` and `boost::hash_combine` with `TfHash`, but excluded updating the python bindings.  This completes the removal of `boost::hash` from the entire `sdf` library.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8912

PR #2786, opened 2023-11-06
Remove unused `BoostHasher` from `base/tf` test case

nvmkuruc:
### Description of Change(s)

`BoostHasher` is commented out and unused in `pxr/base/tf/testenv/hash.cpp`.  With #2785 removing the last boost dependency in `pxr`, this unused code can be safely removed.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8913

PR #2687, opened 2023-09-13
Support for UTF-8 Character Class Processing

erslavin:
### Description of Change(s)
    
- Added UTF-8 utility functions to tf to read UTF-8 encoded characters and check if code points belong in XID_Start / XID_Continue character classes
- Added pre-processing script to generate character class static data structures for identifying code points in XID_Start / XID_Continue from source DerivedCoreProperties.txt
- Added tests for XID_Start / XID_Continue code point validity
- Added reference Unicode 15.1.0 database files as reference for pre-processing script

This PR is stacked on top of "Support for UTF-8 chars in TfDictionaryLessThan" https://github.com/PixarAnimationStudios/OpenUSD/pull/2673

### Fixes Issue(s)

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #2788, opened 2023-11-06
Add an iterator to tf to decode UTF-8 strings

nvmkuruc:
### Description of Change(s)
#2687 will be introducing support for identifying code points in the `Xid` unicode class.  This foundational change adds `Utf8CodePointIterator` and `Utf8CodePointView` to convert UTF-8 encoded strings to code points.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8914

crevilla:
With the making of the IsPastTheEnd and the constructor private, this is good to me

PR #2789, opened 2023-11-06
Autodesk: Redirecting shader parameters

erikaharrison-adsk:
### Description of Change(s)

Revert change [68dec0df](https://github.com/PixarAnimationStudios/OpenUSD/commit/68dec0dff7791b74fbbc679bed3350e897c4bcb8) and use the correct way to resolve the issue for redirecting shader parameters.

The change [68dec0df](https://github.com/PixarAnimationStudios/OpenUSD/commit/68dec0dff7791b74fbbc679bed3350e897c4bcb8) is because if a shader parameter has the same name as the primvar, it will redirect to the primvar. If the primvar doesn't emit scalar accessor, the shader parameter must define the scalar accessor for Mac platform. This change has a problem when the primvar has already emitted the scalar accessor. For example, if EmitAccessor is called, it will then call EmitScalarAccessor. In this case, there will be both the scalar accessor for primvar and the scalar accessor for shader parameter. As the name for the primvar and the shader parameter is the same, the two scalar accessor function will be duplicated. The correct solution is that every primvar that has the potential scalar access must provide the scalar accessor. In this change, we add EmitScalarAccessor to EmitStageAccessor and EmitStructAccessor. We may also need to add to _EmitComputeAccessor and _EmitBufferAccessor, but we can wait until we meet with the problem.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/codeGen.cpp
+++ b/pxr/imaging/hdSt/codeGen.cpp
@@ -256,6 +256,10 @@ static void _EmitAccessor(std::stringstream &str,
                           TfToken const &type,
                           HdStBinding const &binding,
                           const char *index=NULL);
+
+static void _EmitScalarAccessor(std::stringstream &str,
+                                TfToken const &name,
+                                TfToken const &type);
 /*
   1. If the member is a scalar consuming N basic machine units,
   the base alignment is N.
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8915

erikaharrison-adsk:
Hi @jesschimein or is there another person I should tag? Any update on this PR? We've a shipping product that's looking to incorporate this work. Let us know if there's anything on our end we can do to help this along.

PR #2794, opened 2023-11-07
Autodesk: WaitUntilCompleted on Metal

erikaharrison-adsk:
### Description of Change(s)

Correct the behavior when we want to submit a command. This will also make WaitTypeWaitUntilCompleted work properly.

The issue was introduced in change [28116f1](https://github.com/PixarAnimationStudios/OpenUSD/commit/28116f1390ee756cfcd25f4e26cd4e2707cb77f2). Before the change, the command buffer is committed in HgiMetal::_SubmitCmds(), and before the buffer commit, HgiMetal::_workToFlush was set to indicate if there is dispatching work in compute encoder or drawing work in graphics encoder. So when command buffer is committed, if _workToFlush is true, it will commit the commands to GPU. But in change [28116f1](https://github.com/PixarAnimationStudios/OpenUSD/commit/28116f1390ee756cfcd25f4e26cd4e2707cb77f2), because they want to introduce the secondary command buffer, the commit of the command buffer is moved to HgiCmds::_Submit (in _hgi->CommitPrimaryCommandBuffer(waitType)), which is before the set of _workToFlush. So even if the encoder contains dispatching work or drawing work, the command buffer will not be committed to GPU. This has two drawbacks. For HgiSubmitWaitTypeNoWait waitType, we push the work to GPU too late. This didn't make the most use of GPU. For HgiSubmitWaitTypeWaitUntilCompleted waitType, it doesn't wait at all.

As you can see, the problem is that _workToFlush is set after the commit of command buffer. So in this change, I add an API HgiMetal::HasWork(). It will set _workToFlush to true. When graphics encoder encodes the draw command or the compute encoder encodes the dispatch command, it can call the HasWork API to indicate there is already work command. So when command buffer is committed, _workToFlush is correctly set.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hgiMetal/computeCmds.h
+++ b/pxr/imaging/hgiMetal/computeCmds.h
@@ -101,7 +101,6 @@ private:
     id<MTLBuffer> _argumentBuffer;
     id<MTLComputeCommandEncoder> _encoder;
     bool _secondaryCommandBuffer;
-    bool _hasWork;
     HgiComputeDispatch _dispatchMethod;
 };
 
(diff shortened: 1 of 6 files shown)


jesschimein:
Filed as internal issue #USD-8920

PR #2795, opened 2023-11-07
Update usdrecord documentation to include all argument options

beersandrew:
### Description of Change(s)
Discovered from [this issue](https://github.com/usd-wg/assets/pull/64#issuecomment-1796446394), some of the `usdrecord` argument options were not in the docs, I've tried to add them to the best of my ability, using the help text from the python code as the descriptions

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-8921

spiffmon:
Thank you, @beersandrew !  For 24.08, we actually attacked this problem systemically, as we now have a script we run to update **all** of `toolset.rst` at once, each release.

PR #2801, opened 2023-11-09
Document dangling layers through new test

nvmkuruc:
### Description of Change(s)

Resolver context changes combined with layer updating can result in layer handles that exist outside of the registry. This PR documents how to construct such a case and some of the resulting behavior.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8938

PR #2802, opened 2023-11-10
Replace `boost::optional` with `std::optional` in `usd/usd`

nvmkuruc:
### Description of Change(s)

Replaces usage of `boost::optional` throughout `pxr/usd/usd`. This leaves some usage to be addressed by #2715 due to interlibrary dependencies.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8939

PR #2810, opened 2023-11-14
Removing unneeded trailing zero from `sdf` max units expression

nvmkuruc:
### Description of Change(s)

#2590 replaced some preprocessor expressions with a `constexpr` use of `std::max` to compute the maximum number of units. `std::max` takes an initializer list populated by a macro that appends a `,` after each element.

The original implementation added `0` to follow the last `,`.  However, as `std::initializer_list`s are allowed to have trailing commas, this was unneeded.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-8953

PR #2811, opened 2023-11-14
Use in class initializers to simplify token macros

nvmkuruc:
### Description of Change(s)
In class initializers reduce the number of internal macros required as well as the number of boost headers required to produce static tokens objects. This introduces a dependency onto `boost/preprocessor/seq/enum.hpp` to convert sequences into standard variadic arguments. This retains the dependency on `boost/preprocessor/seq/for_each.hpp` and `boost/preprocessor/seq/filter.hpp` to continue support for arrays.

While this change does not update the interface to support this, in class member initializers also allow for the public token macros to define tokens inline akin to the private macros.

```
/** header **/
#define MY_TOKENS \
   (abc)                       \
   (xyz)

TF_DECLARE_PUBLIC_TOKENS(MyTokens, MY_TOKENS);

/** source **/
TF_DEFINE_PUBLIC_TOKENS(MyTokens, MY_TOKENS);
```

could become

```
/** header **/
TF_DECLARE_PUBLIC_TOKENS(MyTokens,
   (abc)
   (xyz)
);

/** source **/
TF_DEFINE_PUBLIC_TOKENS(MyTokens);
```

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/staticTokens.h
+++ b/pxr/base/tf/staticTokens.h
@@ -82,11 +82,9 @@
 
 #include <vector>
 
-#include <boost/preprocessor/punctuation/comma_if.hpp>
+#include <boost/preprocessor/seq/enum.hpp>
 #include <boost/preprocessor/seq/filter.hpp>
 #include <boost/preprocessor/seq/for_each.hpp>
-#include <boost/preprocessor/seq/for_each_i.hpp>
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8954

PR #2812, opened 2023-11-15
Support up to 100 arguments in `preprocessorUtilsLite.h`

nvmkuruc:
### Description of Change(s)

#2811 seeks to replace `BOOST_PP_SEQ_FOR_EACH` with `TF_PP_FOR_EACH` by converting sequences to variadics.  

`HdTokens` contains about 84 elements, but `TF_PP_FOR_EACH` only supports up to 64. This increases that support to 100 arguments to support `HdTokens` and other long sequences.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/preprocessorUtilsLite.h
+++ b/pxr/base/tf/preprocessorUtilsLite.h
@@ -48,9 +48,9 @@
 #ifdef ARCH_COMPILER_MSVC
 
 /// Expand to the number of arguments passed.  For example,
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-8955

nvmkuruc:
The need to convert sequences to variadics has been dropped in favor of #2833.  There's no need to extend the range of `TF_PP_FOR_EACH`.

PR #2815, opened 2023-11-15
Support transparent hashing of `std::optional` for `TfHash`

nvmkuruc:
### Description of Change(s)

#2781 adds support for `TfHash` to fallback to `std::hash` when a `TfHashAppend` implementation isn't available. This provided support for `std::optional`.  However, `std::hash` will forward a valued `std::optional<T>` to `std::hash<T>{}(optional.value())` instead of `TfHash`.  This provides a specialization that forwards a valued `std::optional` to `TfHash` (following the lookup ordering of `TfHashAppend`, `std::hash`, and unqualified `hash_value`).

Transparent hashing of a valued `std::optional` matches the specification of `std::hash`.

An unvalued `std::optional` will derive its hash from its contents' `typeid`.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -33,6 +33,7 @@
 
 #include <cstring>
 #include <string>
+#include <optional>
 #include <map>
 #include <memory>
 #include <set>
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-8958

PR #2816, opened 2023-11-15
Guard against unnecessary array detaches in attribute creation

nvmkuruc:
### Description of Change(s)
Consider the following snippet--

```
# testDetach.py
from pxr import Usd
from pxr import UsdGeom

stage = Usd.Stage.CreateInMemory()
points = UsdGeom.Points.Define(stage, "/points")
points.CreatePointsAttr()
```

Running 
```
VT_LOG_STACK_ON_ARRAY_DETACH_COPY=1 python3 testDetach.py
```
produces 16 `VtArray` detaches due to non-const access of override property names in `schemaRegistry.cpp`.  While not verified, we suspect that there are two detaches per schema in its type hierarchy (`Points` has 7 ancestors plus itself => 8 total schema types).

This guards against detaches by both making the `VtTokenArray`s `const`.  Rerunning `testDetach.py` with this change and `VT_LOG_STACK_ON_ARRAY_DETACH_COPY=1` shows zero detaches.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/usd/schemaRegistry.cpp
+++ b/pxr/usd/usd/schemaRegistry.cpp
@@ -1330,7 +1330,7 @@ BuildPrimDefinition(_SchemaDefInitHelper *defInitHelper)
 
     // Get the list of names of any override properties this schema may have as
     // we want to skip these at first when initializing the prim definition.
-    VtTokenArray overridePropertyNames = _GetOverridePropertyNames(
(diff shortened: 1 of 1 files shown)


nvmkuruc:
Thanks to @andrewkaufman for discovering this issue.

jesschimein:
Filed as internal issue #USD-8959

spiffmon:
Wow, nice find, @andrewkaufman !

PR #2819, opened 2023-11-19
DRAFT: PEGTL Text File Format Parser

erslavin:
### Description of Change(s)

- Replaced existing lex/yacc based parser with PEGTL
- Moved path parser to independent file set to support direct inclusion in text file format parser
- Added specific tests for various grammar rules

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/CMakeLists.txt
+++ b/pxr/usd/sdf/CMakeLists.txt
@@ -116,6 +116,7 @@ pxr_library(sdf
         pathParser
         subLayerListEditor
         textParserContext
+        textParserHelpers
         valueTypeRegistry
         variableExpressionImpl
         variableExpressionParser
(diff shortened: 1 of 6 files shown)


erslavin:
Draft PR for initial review.  A few things still left to be done:

- Second pass at whitespace rules needs to be done to make them more robust
- Multi-line C++ style comments are not yet accounted for
- Additional specific grammar tests should be added, particularly around comments
- Issue with debug assertion when running tests under debug
- New output of parse errors needs to be introduced to baseline of testUsdviewFileFormatArguments5
- Small issues with evaluating whether certain characters can be included in certain rules such that when they are parsed these characters are accounted for because they are the same as the 'split' character
- Proper export of template specializations to be used downstream in derived parsers

jesschimein:
Filed as internal issue #USD-8970

erslavin:
Closed and replaced by #2879

PR #2822, opened 2023-11-20
PEGTL Path Parser Extraction

erslavin:
### Description of Change(s)

- Separated out existing PEGTL path parser into its own set of header / source files in preparation for it to be used directly in a PEGTL based text file format parser

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/pathParser.cpp
+++ b/pxr/usd/sdf/pathParser.cpp
@@ -1,0 +1,211 @@
+//
+// Copyright 2023 Pixar
+//
+// Licensed under the Apache License, Version 2.0 (the "Apache License")
+// with the following modification; you may not use this file except in
+// compliance with the Apache License and the following modification to it:
+// Section 6. Trademarks. is deleted and replaced with:
+//
+// 6. Trademarks. This License does not grant permission to use the trade
+//    names, trademarks, service marks, or product names of the Licensor
+//    and its affiliates, except as required to comply with Section 4(c) of
+//    the License and to reproduce the content of the NOTICE file.
+//
+// You may obtain a copy of the Apache License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-8973

pxr/usd/sdf/pathParser.cpp
```diff
@@ -0,0 +1,212 @@
+//
+// Copyright 2016 Pixar
```
crevilla: Need to update the year in the copyright since this is a new file

pxr/usd/sdf/pathParser.h
```diff
@@ -0,0 +1,182 @@
+//
+// Copyright 2016 Pixar
```
crevilla: Same here with the copyright year update

pxr/usd/sdf/pathParser.h
```diff
@@ -0,0 +1,182 @@
...
+PXR_NAMESPACE_OPEN_SCOPE
+
+bool
+Sdf_ParsePath(std::string const &pathStr, SdfPath *path, std::string *errMsg);
+
+namespace Sdf_PathParser {
+
+////////////////////////////////////////////////////////////////////////
+// SdfPath grammar:
+
```
crevilla, line 44: With this being in a header, I'm glad to see the use of the explicit namespace of tao::TAO_PEGTL_NAMESPACE.
But it's a bit verbose and hard to read. I wondering if a `using PEGTL = tao::TAO_PEGTL_NAMESPACE;` here (or PEGTL_NS) would give us the best of both worlds yielding a shorter but still explicit namespace for all the declarations.

erslavin: shortend namespace with `namespace PEGTL_NS = tao::TAO_PEGTL_NAMESPACE;`

PR #2823, opened 2023-11-20
Replace boost with std library in UsdAbc

tylerm-nv:
### Description of Change(s)
Replaces all boost usage except for python bindings in UsdAbc.

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/OpenUSD/issues/2211

- [x]  I have verified that all unit tests pass with the proposed changes


jesschimein:
Filed as internal issue #USD-8974

PR #2827, opened 2023-11-22
cmake: fix absolute path to GL libraries in pxrConfig.cmake

pmolodo:
### Description of Change(s)

The absolute path to GL makes pxrConfig.cmake non-portable.

### Fixes Issue(s)
- can't use pxrConfig.cmake on computers where location of GL library differs.

### DIFF
--- a/pxr/imaging/garch/CMakeLists.txt
+++ b/pxr/imaging/garch/CMakeLists.txt
@@ -27,7 +27,7 @@ pxr_library(garch
         arch
         tf
         ${X11_LIBRARIES}
-        ${OPENGL_gl_LIBRARY}
+        OpenGL::GL
         ${GARCH_PLATFORM_LIBRARIES}
 
     INCLUDE_DIRS


jesschimein:
Filed as internal issue #USD-8982

pmolodo:
I've confirmed that an absolute path sneaks in on Linux (tested on Ubuntu 22.04) - tried on Windows, and apparently it's not an issue there, for whatever reason.

Unsure of behavior on MacOS...

tallytalwar:
On MacOS I think "OPENGL_gl_LIBRARY" gets substituted by "-framework OpenGL" explicitly... https://github.com/PixarAnimationStudios/OpenUSD/pull/2772, but I think like you said linking to "OpenGL::GL" target should also work on MacOS.

Also will go ahead and apply the same fix for _pxr/imaging/glf/CMakeLists.txt_

pmolodo:
Actually - now that the Metal storm delegate exists, does MacOS even use / link to OpenGL anymore?

tallytalwar:
Hi @pmolodo , I confirmed with David and our imaging tests still depend on GL (via garch and glf) even on macOS. So I believe we still need this. The change is checked in internally and will be available on next dev push.

pmolodo:
Awesome, good to hear!

Issue #2653, opened 2023-09-06
HgiInteropMetal is leaking memory when buffer size changes


mnaehrig:
This can be reproduced with usdview 23.08 on macOS. If you do some window resizing you can watch in Activity Monitor how process memory is raising very quickly. Following error is raised if you spend some more time with window resizing:

`Fatal coding error: Failed to create the shared OpenGL color texture for Metal/GL interop [Python], in _SetAttachmentSize(), /Users/mnaehrig/OpenUSD/pxr/imaging/hgiInterop/metal.mm:572`

jesschimein:
Filed as internal issue #USD-8659

PR #2828, opened 2023-11-22
Remove use of CV*TextureCache in GL/Metal HgiIntrop

jasonfielder:
### Description of Change(s)

CVMetalTextureCache and CVOpenGLTextureCache was used to create the OpenGL and Metal resource handles to a shared GPU texture. This contained a memory leak, and was ultimately not necessary for the GL/Metal interop as the textures were only recreated when the OpenGL FBO resolution changed.

This PR removes the use of the CV*TextureCaches, simplifying the implementation and fixing the reported memory leak. 

### Fixes Issue(s)
- #2653


jesschimein:
Filed as internal issue #USD-8983

PR #2831, opened 2023-11-27
Replace `BOOST_PP_SEQ_FOR_EACH_I` with `constexpr` expressions in `vt`

nvmkuruc:
### Description of Change(s)

`BOOST_PP_SEQ_FOR_EACH_I` is used currently to generate compile time indices for all vt value types.  This PR generates a `TfMetaList` from the `VT_VALUE_TYPE` sequence which is leveraged by a recursive `constexpr` implementation of `GetIndex`.

Both changes still require `BOOST_PP_SEQ_FOR_EACH`, but a forthcoming change will provide a boost free implementation of sequence iteration.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/vt/types.h
+++ b/pxr/base/vt/types.h
@@ -33,11 +33,11 @@
 #include "pxr/base/arch/inttypes.h"
 #include "pxr/base/gf/declare.h"
 #include "pxr/base/gf/half.h"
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9000

PR #2832, opened 2023-11-28
Fix 'usdcat --flattenLayerStack' output

jackohoeg:
### Description of Change(s)

This PR addresses 2 issues: `usdcat --flattenLayerStack ...` outputting the wrong data, and a crash when calling the command on a non-existant file.

The first issue was due to the logic guiding what to export. If the `stage` was initialized, it would export it. Otherwise, it would export the `layer`. This doesn't work because, in the presence of the `--flattenLayerStack` flag, both the `stage` and the `layer` are initialized, but the **`layer`** should be exported. Instead, it was outputting the flattened, unloaded stage. To fix this, I simply flipped the order of the conditional blocks so that the `stage` is only exported in the absence of the `layer`. Since the only conditional branches that can result in both the `stage` and `layer` being initialized are bound to the `--flattenLayerStack` flag, we can verify that this change does not impact the behavior of any of the other cases.

The second issue is the result of an unsafe assumption that the input file is valid. Since the `--flattenLayerStack` command generates the layer from the input stage, we need to be sure that the stage successfully opened before creating the layer. `UsdUtilsFlattenLayerStack`expects a valid stage as input, so giving it a `TfNullPtr` results in a crash. I addressed this by requiring that the stage exist before running `UsdUtilsFlattenLayerStack`.

I also added a couple of new test cases that should help prevent similar issues from happening in the future.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/usd/bin/usdcat/CMakeLists.txt
+++ b/pxr/usd/bin/usdcat/CMakeLists.txt
@@ -50,6 +50,11 @@ pxr_register_test(testUsdCatMissingOrInvalidFiles2
     EXPECTED_RETURN_CODE 1 
 )
 
+pxr_register_test(testUsdCatMissingOrInvalidFiles3
+    COMMAND "${CMAKE_INSTALL_PREFIX}/bin/usdcat --flattenLayerStack foo.usda"
+    EXPECTED_RETURN_CODE 1
+)
+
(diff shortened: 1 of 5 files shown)


jesschimein:
Filed as internal issue #USD-9001

PR #2833, opened 2023-11-28
Add `TF_PP_SEQ_FOR_EACH` and `TF_PP_SEQ_SIZE` macros

nvmkuruc:
### Description of Change(s)

This PR adds `TF_PP_SEQ_FOR_EACH` and `TF_PP_SEQ_SIZE` which can operate on sequences of parenthetically separated elements.  (```(x)(y)(z)(w)``` is a sequence of four elements: `x`, `y`, `z`, and `w`).  Their implementation is similar to their variadic counterparts. They are aided by two new private helpers `_TF_PP_SEQ_HEAD` and `_TF_PP_SEQ_TAIL` which can be used to pop the first element off the sequence and apply an operation like a macro.

`pyStaticTokens.h` requires a global data argument for `TF_PP_SEQ_FOR_EACH` even though it's generally unused.  A future PR will update the static token macros to use `TF_PP_SEQ_FOR_EACH`.

As the sequence operations support the static token macros, they need to support more elements than their variadic equivalents.  (`HdTokens` currently contains over eighty tokens.)  The chosen macro's size limit of 229 is underneath the MSVC's recursive macro limit of 256.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9002

PR #2841, opened 2023-11-29
Text file format parser prep

erslavin:
### Description of Change(s)

In preparation for a new PEGTL Text File Format parser that uses the same helper functions as the existing ones
embedded in the current yacc parser, these helper methods are extracted into a new file and properly namespaced.

The existing yacc parser was not modified and will continue to function as normal, using the helper functions already embedded in the .yy file.

The purpose of this change is to make it easier to see where (if at all) these helper methods change when used by a new PEGTL based text file format parser.

All templated functions remain in the header.  Any functions marked with `inline static` had the `static` qualifier removed, otherwise they remain in the header.  Any functions not marked `inline` had the `static` qualifier removed, forward declared in the header and implemented in the cpp.

Partially addresses #USD-8970.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/CMakeLists.txt
+++ b/pxr/usd/sdf/CMakeLists.txt
@@ -116,6 +116,7 @@ pxr_library(sdf
         pathParser
         subLayerListEditor
         textParserContext
+        textParserHelpers
         valueTypeRegistry
         variableExpressionImpl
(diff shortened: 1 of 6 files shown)


jesschimein:
Filed as internal issue #USD-9010

PR #2842, opened 2023-11-30
Replace `BOOST_PP` macros in `testTraceData` with `std::tuple` utilities

nvmkuruc:
### Description of Change(s)

C++17 offers guaranteed copy elision, allowing for methods like `std::make_from_tuple` to create non-copyable and non-movable types.

This lets us bring many transformations in `testTraceData` out of the preprocessor space and into the tuple space, avoiding relying on specific indexing provided by `BOOST_PP_SEQ_FOR_EACH_I`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9016

PR #2848, opened 2023-12-07
Support for UTF-8 Identifiers

erslavin:
### Description of Change(s)

- Added UTF-8 character class support to Path parser
- Added UTF-8 based validation rules for SDF identifier methods
- Added UTF-8 support to lex-based text file format lexer
- Added tests for UTF-8 based identifiers

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-9056

pxr/usd/sdf/path.cpp
```diff
@@ -1787,101 +1791,140 @@ SdfPath::MakeRelativePath(const SdfPath & anchor) const
...
+
     // A valid C/Python identifier except we also allow the namespace delimiter
     // and if we tokenize on that delimiter then all tokens are valid C/Python
-    // identifiers.  That means following a delimiter there must be an '_' or
-    // alphabetic character.
-    constexpr char delim = SDF_PATH_NS_DELIMITER_CHAR;
-    for (char const *p = name.c_str(); *p; ++p) {
-        if (!_IsAlpha(*p) && *p != '_') {
+    // identifiers.
+    std::vector<std::string> tokens = SdfPath::TokenizeIdentifier(name);
```
gitamohr: I'm pretty concerned about the performance impact of this -- having to tokenize to a vector here, which we hit on most every AppendProperty().  Have we done perfeval runs on this change yet?

pxr/usd/sdf/path.cpp
```diff
@@ -1787,101 +1791,140 @@ SdfPath::MakeRelativePath(const SdfPath & anchor) const
...
     }
-    return false;
+
+    return true;
 }
 
 std::vector<std::string>
 SdfPath::TokenizeIdentifier(const std::string &name)
 {
+    constexpr uint32_t SDF_NAMESPACE_DELIMITER_CODE_POINT = 0x003Au;
```
gitamohr: I think this should go in `tokens.h` alongside the other ones like `SDF_PATH_NS_DELIMITER_CHAR`

pxr/usd/sdf/pathParser.h
```diff
@@ -51,21 +106,31 @@ struct ReflexiveRelative : Dot {};
...
+// valid identifiers start with an '_' character or anything in the XidStart
+// character class, then continue with zero or more characters in the
+// XidContinue character class
+struct Utf8IdentifierStart : PEGTL_NS::sor<
+    PEGTL_NS::one<'_'>,
+    XidStart> {};
+struct Utf8Identifier : PEGTL_NS::seq<
+    Utf8IdentifierStart,
+    PEGTL_NS::star<XidContinue>> {};
+
```
gitamohr, line 118: If it turns out this slows down the parser, one way we could speed it up perhaps is to make an 'Identifier' rule that first tries an ASCII ident first and only invokes the UTF-8 rule if that doesn't match (along the lines of sor<PEGTL_NS::identifier, Utf8Identifier>),

pxr/usd/sdf/pathParser.h
```diff
@@ -103,7 +168,7 @@ struct RelationalAttributeName : PropertyName {};
 
 struct MapperKW : TAO_PEGTL_KEYWORD("mapper") {};
 
-struct MapperArg : PEGTL_NS::identifier {};
+struct MapperArg : Utf8Identifier {};
```
gitamohr: Let's not change this for now please.  USD doesn't have mapper args and we want to limit exposure to Presto to start.

pxr/usd/sdf/textFileFormat.ll
```diff
@@ -134,24 +164,42 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+  * read an 'identifier'.  Note this rule will also match
+  * standard ASCII strings because the UTF-8 encoded byte 
+  * representation is the same for these characters.
+  * However, unlike the path lexer, we can guarantee that 
+  * prim names aren't something special to be called out here
+  * so we can be a little more specific about the kinds of strings
+  * we match, particularly to not collide with the pure digit match rule
+  * below
+  */
+{UTF8NODIGU}{UTF8U}* {
```
gitamohr, line 190: It seems like there ought to be code added to the text file parser to actually validate that identifiers are XID_Start/XID_Continue -- is that coming in a future change?

pxr/usd/sdf/testenv/testSdfParsing.py
```diff
@@ -64,6 +64,7 @@ def test_Basic(self):
         # This will mean that your new test runs first and you can spot
         # failures much quicker.
         testFiles = '''
+        217_utf8_identifiers.sdf
         216_bad_variant_in_relocates_path.sdf
         215_bad_variant_in_specializes_path.sdf
         214_bad_variant_in_inherits_path.sdf
```
gitamohr, line 72: Could we add tests that validate that the text file parser rejects "identifiers" that aren't actually XID_Start/XID_Continue identifiers?  We look for 'identifier's in several places other than prim/property names, and it seems like with this change as-is, we might accept non-identifiers in those places?

pxr/usd/sdf/path.cpp
nvmkuruc: We should probably make it a `TfUtf8CodePoint` too.

pxr/usd/sdf/textFileFormat.ll
nvmkuruc: @gitamohr, are you thinking that Ed should call `SdfPath::IsValidIdentifier` and `SdfPath::IsValidNamespacedIdentifier` here?

pxr/usd/sdf/textFileFormat.ll
```diff
@@ -134,24 +164,42 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+    (*yylval_param) = std::string(yytext, yyleng);
+    return TOK_IDENTIFIER;
+}
+
+ /* unquoted C++ namespaced identifiers are identifiers separated
+  * by the '::' character -- see bug 10775.
+  */
+{UTF8NODIGU}{UTF8U}*(::{UTF8NODIGU}{UTF8U}*)+ {
+    (*yylval_param) = std::string(yytext, yyleng);
+    return TOK_CXX_NAMESPACED_IDENTIFIER;
```
nvmkuruc: @gitamohr In the spirit of your note about not adding UTF-8 support to mappers should, we leave CXX_NAMESPACED_IDENTIFIERs alone too?

pxr/usd/sdf/path.cpp
phuchau1989: Phúc hậu

erslavin: The perf runs came back ok, but if we need to further optimize we can use @nvmkuruc suggestion with `string_view`

pxr/usd/sdf/pathParser.h
erslavin: The perf runs came back ok, the cost is really just the cost of decoding and class validation

pxr/usd/sdf/textFileFormat.ll
erslavin: The original changes to the path lexer did indeed have to do an extra validation check before reducing as a `TOK_IDENTIFIER`.  The text file format parser did not - I can go back and verify why.

pxr/usd/sdf/testenv/testSdfParsing.py
erslavin: yeah ok, it will accept unexpected type names, for example, which I added a test for.  I can add back in the extra validation that wsa done in the original modifications I made to the lexer last year, as long as we are aware that we pay a double cost in validation in some cases.

pxr/usd/sdf/path.cpp
erslavin: Updated to use `string_view` instead

pxr/usd/sdf/path.cpp
erslavin: Moved definition out to `tokens.h` (along with underscore code point)

pxr/usd/sdf/textFileFormat.ll
erslavin: Added additional validation to the three lexer rules

pxr/usd/sdf/testenv/testSdfParsing.py
erslavin: Added a few more tests and updated validation rules

pxr/usd/sdf/path.cpp
```diff
@@ -1787,48 +1789,86 @@ SdfPath::MakeRelativePath(const SdfPath & anchor) const
...
+
+        // can't end with ':'
+        if (index == remainder.size() - 1) {
+            return false;
         }
-        if (*p != delim) {
-            return !*p;
+
+        // substring must be a valid identifier
+        if (!SdfPath::IsValidIdentifier(std::string(remainder.substr(0, index)))) {
```
nvmkuruc: In lieu of creating a `std::string`, can we just make the internal `_IsValidIdentifier` (which `SdfPath::IsValidIdentifier` calls) take a `string_view`?

pxr/usd/sdf/textFileFormat.ll
gitamohr: I'm okay with that, I think.

pxr/usd/sdf/textFileFormat.ll
```diff
@@ -134,25 +164,93 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+  * below
+  */
+{UTF8NODIGU}{UTF8U}* {
+    std::string matched = std::string(yytext, yyleng);
+
+    // we perform an extra validation step here
+    // to make sure what we matched is actually a valid
+    // identifier because we can overmatch UTF-8 characters
+    // based on this definition
+    if (!SdfPath::IsValidIdentifier(matched)) {
```
gitamohr: Hmm -- I think we need a non-SdfPath specific function that checks if an identifier is the utf-8 style.  This TOK_IDENTIFIER thing is used for other things in the parser like dictionary keys, or metadata keys, not just prim & property names.  Which raises the question, are we intending to add utf-8 support for those elements too?  Or would it be preferable to scope this just to prim & property names?

pxr/usd/sdf/path.h
```diff
@@ -45,6 +45,21 @@
...
+SDF_API
+bool SdfIsValidIdentifier(const std::string& name);
+
+/// \name Valid identifiers separated by ':' characters.
+/// @{
+
+/// Returns whether \p name is a legal namespaced identifier.
+SDF_API
+bool SdfIsValidNamespacedIdentifier(const std::string& name);
+
```
gitamohr: Sorry I missed this before, but I think the right concept for these already exists as `SdfSchema::IsValid[Namespaced]Identifier()`.  The current implementations of those just call through to the equivalently-named `SdfPath` functions so hopefully that should work for us.

PR #2849, opened 2023-12-07
Autodesk: Providing Vulkan build support in build script and cmake files.

erikaharrison-adsk:
### Description of Change(s)

This PR provides Vulkan build support to OpenUSD by making changes to the build scripts and cmake files.

### Fixes Issue(s)
- N/A

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1611,6 +1611,11 @@ def InstallUSD(context, force, buildArgs):
         else:
             extraArgs.append('-DPXR_BUILD_TUTORIALS=OFF')
 
+        if context.buildVulkan:
+            extraArgs.append('-DPXR_ENABLE_VULKAN_SUPPORT=ON')
+        else:
+            extraArgs.append('-DPXR_ENABLE_VULKAN_SUPPORT=OFF')
+
         if context.buildTools:
             extraArgs.append('-DPXR_BUILD_USD_TOOLS=ON')
         else:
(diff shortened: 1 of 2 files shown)


build_scripts/build_usd.py
```diff
@@ -1878,6 +1883,11 @@ def InstallUSD(context, force, buildArgs):
 subgroup.add_argument("--no-tutorials", dest="build_tutorials", action="store_false",
                       help="Do not build tutorials")
 subgroup = group.add_mutually_exclusive_group()
+subgroup.add_argument("--vulkan", dest="build_vulkan", action="store_true",
+                      default=False, help="Build vulkan (default)")
+subgroup.add_argument("--no-vulkan", dest="build_vulkan", action="store_false",
```
clach, line 1888: The "(default)" should be on this help description, since we do not build Vulkan by default.

build_scripts/build_usd.py
```diff
@@ -1878,6 +1883,11 @@ def InstallUSD(context, force, buildArgs):
 subgroup.add_argument("--no-tutorials", dest="build_tutorials", action="store_false",
                       help="Do not build tutorials")
 subgroup = group.add_mutually_exclusive_group()
+subgroup.add_argument("--vulkan", dest="build_vulkan", action="store_true",
```
clach, line 1886: These options should go under the "Imaging and USD Imaging Options" argument group.

build_scripts/build_usd.py
```diff
@@ -2390,6 +2401,7 @@ def _JoinVersion(v):
     Tests                       {buildTests}
     Examples                    {buildExamples}
     Tutorials                   {buildTutorials}
+    Vulkan                      {buildVulkan}
```
clach, line 2404: Similar to above, this should go under "Imaging".

clach (approved):
Paraphrasing from another Pixar engineer who's more experienced in this area, but eventually we'll need better documentation about how to properly build with Vulkan i.e. provide clarification about what's expected re: the Vulkan install. But for now, this LGTM.

jesschimein:
Filed as internal issue #USD-9057

PR #2850, opened 2023-12-07
Autodesk: Provided initialization hgi backend selection for render plugin.

erikaharrison-adsk:
### Description of Change(s)

This PR provides the ability to select Vulkan/OpenGL as a Hgi backend from the render delegate and plugin's perspective.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hd/rendererPlugin.cpp
+++ b/pxr/imaging/hd/rendererPlugin.cpp
@@ -60,7 +60,14 @@ HdRendererPlugin::~HdRendererPlugin() = default;
 HdPluginRenderDelegateUniqueHandle
 HdRendererPlugin::CreateDelegate(HdRenderSettingsMap const& settingsMap)
 {
-    if (!IsSupported()) {
+    TfToken backendToken;
+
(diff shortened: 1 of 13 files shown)


jesschimein:
Filed as internal issue #USD-9058

pxr/imaging/hd/rendererPlugin.cpp
```diff
@@ -60,7 +60,14 @@ HdRendererPlugin::~HdRendererPlugin() = default;
 HdPluginRenderDelegateUniqueHandle
 HdRendererPlugin::CreateDelegate(HdRenderSettingsMap const& settingsMap)
 {
-    if (!IsSupported()) {
+    TfToken backendToken;
```
clach, line 63: On the whole, we're not a fan of the changes to this function. For one, the changes mean that CreateDelegate can fail even though IsSupported might succeed (such as if you gave it no hgi token).

Would it be possible to put the switch at the UsdImagingGLEngine level? For  example, you can pass your Hgi of choice to UsdImagingGLEngine in an HdDriver.

PR #2852, opened 2023-12-07
Autodesk: General HgiVulkan stability

erikaharrison-adsk:
### Description of Change(s)

This PR provides the following support to the hgiVulkan backend
1. Fixes Color Correction
2. Provides Layout Transition infrastructure
3. Fixes most Vulkan validation errors
4. Fixes debug layer struct issue
5. Fixes Pipeline Vertex Divisor struct issue

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/codeGen.cpp
+++ b/pxr/imaging/hdSt/codeGen.cpp
@@ -147,8 +147,14 @@ TF_DEFINE_PRIVATE_TOKENS(
     (early_fragment_tests)
 );
 
+// For now, looks like for Vulkan path, we require HGI Resource Generation path enabled
+#ifdef PXR_VULKAN_SUPPORT_ENABLED
+TF_DEFINE_ENV_SETTING(HDST_ENABLE_HGI_RESOURCE_GENERATION, true,
+    "Enable Hgi resource generation for codeGen");
+#else
 TF_DEFINE_ENV_SETTING(HDST_ENABLE_HGI_RESOURCE_GENERATION, false,
-                      "Enable Hgi resource generation for codeGen");
+    "Enable Hgi resource generation for codeGen");
+#endif
 
 /* static */
 bool
--- a/pxr/imaging/hdSt/interleavedMemoryManager.cpp
+++ b/pxr/imaging/hdSt/interleavedMemoryManager.cpp
@@ -462,8 +462,11 @@ HdStInterleavedMemoryManager::_StripedInterleavedBuffer::Reallocate(
     // Skip buffers of zero size.
     if (totalSize > 0) {
         HgiBufferDesc bufDesc;
+        bufDesc.debugName = curRangeOwner_->GetResources().begin()->first.GetString();
         bufDesc.byteSize = totalSize;
-        bufDesc.usage = HgiBufferUsageUniform;
+        // Vulkan Validation layer is raising an error here because
+        // a usage flag of Uniform is applied to a buffer bound as storage
+        bufDesc.usage = HgiBufferUsageUniform | HgiBufferUsageStorage;
         newBuf = hgi->CreateBuffer(bufDesc);
     }
 
--- a/pxr/imaging/hdSt/interleavedMemoryManager.cpp
+++ b/pxr/imaging/hdSt/interleavedMemoryManager.cpp
@@ -33,6 +33,7 @@
 #include "pxr/imaging/hgi/blitCmdsOps.h"
 #include "pxr/imaging/hgi/buffer.h"
 #include "pxr/imaging/hgi/capabilities.h"
+#include "pxr/imaging/hgi/tokens.h"
 
 #include "pxr/base/arch/hash.h"
 #include "pxr/base/tf/diagnostic.h"
@@ -458,6 +459,7 @@ HdStInterleavedMemoryManager::_StripedInterleavedBuffer::Reallocate(
     HgiBufferHandle newBuf;
 
     Hgi* hgi = _resourceRegistry->GetHgi();
+    bool const isVulkanEnabled = (hgi->GetAPIName() == HgiTokens->Vulkan);
 
     // Skip buffers of zero size.
     if (totalSize > 0) {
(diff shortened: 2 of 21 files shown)


pxr/imaging/hdx/colorCorrectionTask.cpp
```diff
@@ -877,11 +877,19 @@ HdxColorCorrectionTask::_CreateBufferResources()
     }
 
     // A larger-than screen triangle made to fit the screen.
-    constexpr float vertData[][6] =
+    float vertData[][6] =
             { { -1,  3, 0, 1,     0, 2 },
               { -1, -1, 0, 1,     0, 0 },
               {  3, -1, 0, 1,     2, 0 } };
 
+    // Vulkan back-end needs the UVs inverted along the y axis
```
schevrel, line 885: Instead of fixing this here, it might be better to bring in the fix at the shader generation level.  See https://github.com/PixarAnimationStudios/OpenUSD/pull/2504

jesschimein:
Filed as internal issue #USD-9060

pxr/imaging/hgiVulkan/texture.cpp
```diff
@@ -514,6 +522,73 @@ HgiVulkanTexture::CopyBufferToTexture(
...
+
+    HgiVulkanCommandQueue* queue = _device->GetCommandQueue();
+    HgiVulkanCommandBuffer* cb = queue->AcquireResourceCommandBuffer();
+
+    VkAccessFlags srcAccessMask, dstAccessMask;
+
+    // The following cases are based on few initial assumptions to provide
+    // an infrastructure for access mask selection based on layouts.
+    // Feel free to update depending on need and use cases.
+    switch (GetImageLayout()) {
```
clach, line 539: This and the following switch statement need a default case, otherwise they fail to compile for me.

pxr/imaging/hgiVulkan/graphicsPipeline.cpp
```diff
@@ -132,21 +132,19 @@ HgiVulkanGraphicsPipeline::HgiVulkanGraphicsPipeline(
         vertBufs.push_back(std::move(vib));
     }
 
+    VkPipelineVertexInputDivisorStateCreateInfoEXT vertexInputDivisor =
```
clach: I'm guessing this work was done on an earlier version of the dev/feature branch, since I believe this change is no longer needed.

pxr/imaging/hgiVulkan/instance.cpp
```diff
@@ -67,15 +67,15 @@ HgiVulkanInstance::HgiVulkanInstance()
         VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME,
     };
 
+    std::vector<const char*> instanceLayerList;
```
clach: This I think is also unneeded now.

pxr/imaging/hdSt/interleavedMemoryManager.cpp
```diff
@@ -462,8 +462,11 @@ HdStInterleavedMemoryManager::_StripedInterleavedBuffer::Reallocate(
     // Skip buffers of zero size.
     if (totalSize > 0) {
         HgiBufferDesc bufDesc;
+        bufDesc.debugName = curRangeOwner_->GetResources().begin()->first.GetString();
         bufDesc.byteSize = totalSize;
-        bufDesc.usage = HgiBufferUsageUniform;
+        // Vulkan Validation layer is raising an error here because
+        // a usage flag of Uniform is applied to a buffer bound as storage
+        bufDesc.usage = HgiBufferUsageUniform | HgiBufferUsageStorage;
```
clach: If you want this to run with HgiGL, too, then can you add a Vulkan-specific check here like you did for vboMemoryManger.cpp?

pxr/imaging/hdSt/vboSimpleMemoryManager.cpp
```diff
@@ -292,7 +292,9 @@ HdStVBOSimpleMemoryManager::_SimpleBufferArray::Reallocate(
 
         HgiBufferDesc bufDesc;
         bufDesc.byteSize = bufferSize;
-        bufDesc.usage = HgiBufferUsageUniform | HgiBufferUsageVertex;
+        // Vulkan Validation layer is raising an error here because
+        // a usage flag of Uniform is applied to a buffer bound as storage
+        bufDesc.usage = HgiBufferUsageUniform | HgiBufferUsageStorage;
```
clach: Same here re: a Vulkan-specific check so this works for HgiGL too.

pxr/imaging/hdSt/codeGen.cpp
```diff
@@ -147,8 +147,14 @@ TF_DEFINE_PRIVATE_TOKENS(
     (early_fragment_tests)
 );
 
+// For now, looks like for Vulkan path, we require HGI Resource Generation path enabled
```
clach, line 150: This too is already fixed in the current branch, just FYI.

pxr/imaging/hgiVulkan/texture.cpp
VipulKapoorAutodesk: Fixed.

pxr/imaging/hgiVulkan/graphicsPipeline.cpp
VipulKapoorAutodesk: Sure. This will be undone in the oncoming push

pxr/imaging/hdx/colorCorrectionTask.cpp
schevrel: My PR had some issues with HdStorm, I have updated it to remove some shader compilation errors.

pxr/imaging/hgiVulkan/device.cpp
```diff
@@ -253,7 +254,7 @@ HgiVulkanDevice::HgiVulkanDevice(HgiVulkanInstance* instance)
 
     VkPhysicalDeviceFeatures2 features =
         {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2};
-    features.pNext = &vulkan11Features;
+    features.pNext = (instance->GetVulkanVersion() >= VK_API_VERSION_1_2) ? &vulkan11Features : nullptr;
```
clach, line 257: Can you elaborate on the need for this change? I'm assuming this is because you're potentially targeting Vulkan versions < 1.2 for Android support. Do you have an idea of what other changes will be needed for that goal?

Right now we use vulkan11Features here to potentially enable shaderDrawParameters (you can see the associated hgi capability flag set in HgiVulkanCapabilities). If we potentially disapply vulkan11Features here, would that need a corresponding modification to HgiVulkanCapabilities?

PR #2853, opened 2023-12-07
Fix plugInfo reading when path contains regex characters

touraill-adobe:
### Description of Change(s)

The previous code was failing to glob plugInfo files when the path contained special characters such as parentheses. For instance, given the path
```
/foo(bar)/**/plugInfo.json
```
the previous code was generating a regex
```
/foo(bar)/.*/plugInfo.json
```
which was not matching anything. This commit escapes the entire path, including the non-wildcarded prefix. It also adds the missing special characters in the escape function.


jesschimein:
Filed as internal issue #USD-9061

PR #2673, opened 2023-09-11
Support for UTF-8 chars in TfDictionaryLessThan

erslavin:
### Description of Change(s)

- Added support for UTF-8 characters in sorting algorithm
- Added additional tests with strings containing UTF-8 characters

### Fixes Issue(s)
-

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #2855, opened 2023-12-09
Validate continuation bytes during code point iteration

nvmkuruc:
### Description of Change(s)

This change ensures that only continuation bytes (bytes that start with `10`) are consumed when incrementing. Consider an algorithm that searches for the ASCII character `.` in a string.  An algorithm that iterates over a UTF-8 string as bytes (using say `std::string::find`) would check each byte individually and always find any `.` characters.  An algorithm iterating over code points could miss a valid `.` following an invalid incomplete 2, 3, or 4 character UTF-8 character sequence if continuations are not checked before incrementing.

This change also changes encoding length checks to use ranges instead of shifting. In many compilers, the same or similar assembly is generated, but in gcc 9.3, shifting introduces an additional `movzx`. As that's the current compiler targeted by OpenUSD's linux build (and #2673 already made a similar switch from shift to range checks), it seems prudent to make a similar switch.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/testenv/unicodeUtils.cpp
+++ b/pxr/base/tf/testenv/unicodeUtils.cpp
@@ -26,6 +26,8 @@
 #include "pxr/base/tf/regTest.h"
 #include "pxr/base/tf/unicodeUtils.h"
 
+#include <algorithm>
+#include <array>
 #include <string_view>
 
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9066

PR #2857, opened 2023-12-11
Convert xid ranges classes into compile time lookup tables

erslavin:
### Description of Change(s)

This change converts the original `TfStaticData` based Unicode XidStart / XidContinue ranges into compile time lookup tables through the use of `constexpr` in order to minimize the initialization overhead of creating these ranges (and associated flag data) on startup.
 
### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9068

PR #2858, opened 2023-12-11
Add class to facilitate serialization and validation of code points

nvmkuruc:
### Description of Change(s)

Depends on (#2855)

#2788 was added to decode UTF-8 strings into code points.  This change repurposes the encoding logic from #2120 into a wrapper around a `uint32_t`.  This wrapper validates on construction and only stores valid UTF-8 code points (ie. surrogates are excluded). Invalid code points are replaced (silently) with the `ReplacementValue`. A stream operator for serialization is provided.

The `TF_INVALID_CODE_POINT` `uint32_t` has been replaced with a `TfUtf8InvalidCodePoint` `constexpr` instance.

To aid testing, `EndAsIterator()` has been added to the `TfUtf8CodePointView` class which returns an iterator of the same type as `begin()`.  While the language (via range based for loops) supports sentinel `end()` types as of C++17, many STL algorithms don't until C++20. While `EndAsIterator()` this creates redundant copies of the `string_view`'s `end()` value, in non-performance critical code, it may be preferable to enable usage of the STL.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/unicodeUtils.h
+++ b/pxr/base/tf/unicodeUtils.h
@@ -32,10 +32,60 @@
 #include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/unicodeCharacterClasses.h"
 
+#include <ostream>
 #include <string>
 #include <string_view>
 
 PXR_NAMESPACE_OPEN_SCOPE
+
+/// Wrapper for a code point value that can be encoded as UTF-8
+class TfUtf8CodePoint {
+public:
+    /// Code points that cannot be decoded or outside of the valid range are
+    /// may be replaced with this value.
+    static constexpr uint32_t ReplacementValue = 0xFFFD;
+
+    /// Values higher than this will be replaced with the replacement
+    /// code point.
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9069

pxr/base/tf/unicodeUtils.h
```diff
@@ -32,11 +32,60 @@
...
+    }
+
+private:
+    uint32_t _value{ReplacementValue};
+};
+
+TF_API std::ostream& operator<<(std::ostream&, const TfUtf8CodePoint);
+
+/// The replacement code point can be used to signal that a code point could
+/// not be decoded and needed to be replaced.
```
crevilla, line 86: It's not a big deal, but I think it would be a little cleaner if this was just
`constexpr TfUtf8CodePoint TfUtf8InvalidCodePoint = TfUtf8CodePoint();`
since the default constructed TfUtf8CodePoint is the invalid code point getting rid of another hardcoded 0xFFFD

PR #2814, opened 2023-11-15
Remove leftover minus sign

tamaskenezlego:
Remove an unneeded minus sign at the start of the line, probably a leftover from a diff.

### Fixes Issue(s)
-

- [ ] I have verified that all unit tests pass with the proposed changes

- [ ] I have submitted a signed Contributor License Agreement

PR #2859, opened 2023-12-11
Specify invalid UTF-8 characters in URI scheme registration errors

nvmkuruc:
### Description of Change(s)

(Depends on #2858 and #2889)

#2814 observed that a stray `-` from a `diff` ended up in the error reporting code for scheme validation. The conditional partitioning ASCII and UTF-8 code paths does not work, and the broken ASCII code path appears to never be exercised.

Now that #2858 has provided a class for encoding code points into strings, the same code path can be used for both.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9070

PR #2824, opened 2023-11-21
Cleaner exception handling in usdresolve.py

pshinners-blz:
### Description of Change(s)

Cleaner exception handling in `usdresolve.py`

If an exception happens while loading the resolver or resolving then this tool adds an additional Python exception and stack trace to the output. This situation can commonly happen when developing custom resolver plugins and using this tool to validate on the command line. In these cases the more concise and accurate the error message, the better. 

This is the error message I'd see before making these changes. Notice the unrelated Python exception trailing the information about the failed library.
```
Failed to resolve 'demo:/til:example' -
    Error in 'PXR_NS::PlugPlugin::_Load' at line 257 in file build/usd/pxr/base/plug/plugin.cpp : 'Failed to load plugin 'demoresolve': install/demoresolve/libdemoresolve.so: undefined symbol: _ZNK13demoResolver21_CreateDefaultContextEv in 'install/demoresolve/libdemoresolve.so''
    Error in 'PXR_NS::{anonymous}::_PluginResolver<ResolverType, ResolverTypeFactory>::Create [with ResolverType = PXR_NS::ArResolver, ResolverTypeFactory = PXR_NS::Ar_ResolverFactoryBase]' at line 335 in file build/usd/pxr/usd/ar/resolver.cpp : 'Failed to load plugin demoresolve for DemoResolver'
Traceback (most recent call last):
  File "usdresolve", line 141, in <module>
    sys.exit(main())
  File "usdresolve", line 127, in main
    if not resolved:
builtins.UnboundLocalError, local variable 'resolved' referenced before assignment
```

With the changes the output becomes
```
Failed to resolve 'demo:/til:example' -
    Error in 'sfd_usd_v2302__pxrReserved__::PlugPlugin::_Load' at line 257 in file /tmp/jenkins-pkg-build-pipeline-24807/bfdpkg/bfd/ext/usd/src/usd/pxr/base/plug/plugin.cpp : 'Failed to load plugin 'tessaresolve': /home/pshinners/pkg/bfd/usd/plugins/tessaresolver/install/linux_x64_usd23.02python3.7/lib/usd/tessaresolve/libtessaresolve.so: undefined symbol: _ZNK13TessaResolver21_CreateDefaultContextEv in '/home/pshinners/pkg/bfd/usd/plugins/tessaresolver/install/linux_x64_usd23.02python3.7/lib/usd/tessaresolve/libtessaresolve.so''
[...]

PR #2860, opened 2023-12-11
Handle error cases when no resolvers are available

dgovil:
### Description of Change(s)

We were hitting a couple segfaults in a couple spots where resolver initialization was failing (due to other reasons). But this would help provide slightly better diagnostics when we encounter those situations instead.

### DIFF
--- a/pxr/usd/ar/resolver.cpp
+++ b/pxr/usd/ar/resolver.cpp
@@ -39,6 +39,7 @@
 #include "pxr/base/plug/registry.h"
 #include "pxr/base/js/utils.h"
 #include "pxr/base/js/value.h"
+#include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/envSetting.h"
 #include "pxr/base/tf/errorMark.h"
 #include "pxr/base/tf/pathUtils.h"
@@ -350,7 +351,11 @@ _GetAvailablePrimaryResolvers(
             break;
         }
     }
-    TF_VERIFY(availablePrimaryResolvers.back().type == defaultResolverType);
+    if (availablePrimaryResolvers.size() > 0) {
+        TF_VERIFY(availablePrimaryResolvers.back().type == defaultResolverType);
+    } else {
+        TF_CODING_ERROR("No primary resolvers could be found.");
+    }
 
     return availablePrimaryResolvers;
 }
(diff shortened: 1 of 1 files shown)


pxr/usd/ar/resolver.cpp
```diff
@@ -1297,10 +1302,16 @@ class _DispatchingResolver final
         if (uriResolver) {
             return *uriResolver;
         }
+
+        if (!_resolver)
+        {
+            TF_FATAL_CODING_ERROR("Resolver is uninitialized.");
```
spiffmon: We are very cautious with use of fatal errors in OpenUSD, because as a generally embedded subsystem, it means we're taking down somebody else's DCC.  I believe the *only* places where TF_FATAL_CODING_ERROR is issued in OpenUSD, currently, is in the Metal integration, and we might want to reconsider those, too?

Also, here and above, is it clear that it is a coding error?  I imagine a resolver might fail to initialize because, e.g. a database went down somewhere.  So maybe these should both be TF_WARN. (and in the db case, possibly the resolver itself might have already reported a TF_RUNTIME_ERROR if it could not contact the db).

I realize leaving this as a warning does still likely leave unguarded access to _resolver in some other sites, here, but perhaps we should harden those, knowing that an error/warning has already been issued?

jesschimein:
Filed as internal issue #USD-9071

pxr/usd/ar/resolver.cpp
dgovil: That's fair. I figured an explicit abort was better than an unintended segfault, but I changed it to a TF_WARN instead as requested.

I agree that it's unclear on why the resolver is uninitialized, though I was unsure what the best way to harden this access is without changing the return and therefore changing all the callosities. 
We tend to hit this one when we run any of the CLI targets from the IDE and it can't find the plugInfo.json etc to initialize (since it's not installed yet).

pshinners-blz:
This is mildly related to my #2824 which shows an unnecessary stack trace in the `usdresolve` command line tool when resolvers fail to initialize.

sunyab:
Closing this PR out in favor of https://github.com/PixarAnimationStudios/OpenUSD/commit/0e11f5a4c029863850f0747955802af6622bcf0b. 

After some discussion we decided that a fatal error is more appropriate in cases like this where the likely culprit is a build configuration issue, since that would likely cause crashes or other bad behavior down the line anyway. The above commit emits additional diagnostics to help track down the issue.

Issue #93, opened 2016-10-29
Failing link in debug on clang, and some gcc versions in Debug


meshula:
Hi there,

TF_INSTANTIATE_SINGLETON fails in debug on clang and some versions of gcc, e.g. redhat linux with gcc 4.8.4.

e.g. from gcc

```
/var/tmp/doNotRemove/BuildTarget/bundles/XXX/RHEL6_AMD64_GCC484_DEBUG/dso/libplug.so: undefined reference to `TfSingleton<PlugRegistry>::GetInstance()'
/var/tmp/doNotRemove/BuildTarget/bundles/XXX/RHEL6_AMD64_GCC484_DEBUG/dso/libplug.so: undefined reference to `TfSingleton<PlugRegistry>::SetInstanceConstructed(PlugRegistry&)'
collect2: error: ld returned 1 exit status
```

also from clang

```
Undefined symbols for architecture x86_64:
  "TfSingleton<TfDiagnosticMgr>::GetInstance()", referenced from:
      TfDiagnosticMgr::GetInstance() in wrapError.o
  "TfSingleton<TfRefPtrTracker>::GetInstance()", referenced from:
      TfRefPtrTracker::GetInstance() in wrapRefPtrTracker.o
  "TfSingleton<Tf_PyEnumRegistry>::GetInstance()", referenced from:
      Tf_PyEnumRegistry::GetInstance() in wrapDiagnostic.o
      Tf_PyEnumRegistry::GetInstance() in wrapTestTfPython.o
  "TfSingleton<TfScriptModuleLoader>::GetInstance()", referenced from:
      TfScriptModuleLoader::GetInstance() in wrapScriptModuleLoader.o
  "TfSingleton<Tf_PyWeakObjectRegistry>::GetInstance()", referenced from:
      Tf_PyWeakObjectRegistry::GetInstance() in wrapNotice.o
```

A fix that works is to modify the macro as follows.

```
#define TF_INSTANTIATE_SINGLETON(T)                               \
    template <> std::mutex* TfSingleton<T>::_mutex = 0;           \
    template <> T* TfSingleton<T>::_instance = 0;                 \
                                                                  \
    template T& TfSingleton< T >::_CreateInstance();              \
    template void TfSingleton< T >::_DestroyInstance();           \
                                                                  \
    template T& TfSingleton< T >::GetInstance();                  \
    template void TfSingleton< T >::SetInstanceConstructed(T& instance)
```

jtran56:
[...]

PR #2861, opened 2023-12-11
Allow multiple USD in the same process on MacOS (#93)

dlarsson-adobe:
### Description of Change(s)
This PR allows loading multiple usd libraries into the same process on macos (assuming they are uniquely namespaced and plugins are loaded from different locations etc).

It's implemented by driving the ArchConstructorEntry's version field through a preprocessor define so that the logic for calling them can distinguish entries from the two different builds and only call the appropriate ones.

This is arguably a hack abusing the field for something it should not be used for but if we change the layout of the struct it will break cases where we don't have control of the other usd instance than the one we are building ourselves since the struct is at play in both usd builds.

Note, I don't necessarily want this PR merged in this form, I just want to point out an issue through a workaround we had to apply to solve an issue.

### Fixes Issue(s)
- Adds the ability to load multiple usd libraries into the same process on macos.

### DIFF
--- a/cmake/defaults/Options.cmake
+++ b/cmake/defaults/Options.cmake
@@ -214,3 +214,18 @@ if (${PXR_BUILD_PYTHON_DOCUMENTATION})
         set(PXR_BUILD_PYTHON_DOCUMENTATION "OFF" CACHE BOOL "" FORCE)
     endif()
 endif()
+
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9072

PR #2863, opened 2023-12-11
Cleanup arkit checks

dgovil:
### Description of Change(s)

This updates the ARKit checker based on what is supported as of WWDC 2023 as per this document:
https://developer.apple.com/documentation/realitykit/validating-usd-files

Though as per that doc, there may be some inter-platform variance that cannot be accounted for in the checker. I've aimed for most permissive in those cases.

Changes include:
1. Allowing MaterialX shaders within the USD file, denoted by an `ND_*` prefix.
2. Support new Prim types, including physics schema, RealityKit components and Preliminary schemas
3. Removed the ARKitRootLayerChecker and left a deprecation warning for the flag that invoked it
4. Modified the help docs for usdchecker to specify the year of RealityKit that it is based around. This should help in future iterations to provide more clarity to users as both USD and RealityKit grow.

### DIFF
--- a/pxr/usd/bin/usdchecker/usdchecker.py
+++ b/pxr/usd/bin/usdchecker/usdchecker.py
@@ -81,7 +81,7 @@ def main():
                         'aggregate stages.')
     parser.add_argument('--arkit', dest='arkit', action='store_true', 
                         help='Check if the given USD stage is compatible with '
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9074

PR #2864, opened 2023-12-11
Replace `boost::optional` with `std::optional` in `SdfLayer` and `Sdf_TextParserContext`

nvmkuruc:
### Description of Change(s)

Most of the usage of `boost::optional` is addressed by #2702 and #2715. This PR removes the remaining usage in `SdfLayer` and `Sdf_TextParserContext`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9075

PR #2867, opened 2023-12-12
Validate code point ordering for non-ASCII range

nvmkuruc:
### Description of Change(s)

(Depends on #2858)

#2673 updated `TfStringUtils` to order non-ASCII characters by bytes. This was to produce code point ordering for non-ASCII characters.  With the serialization of code points provided by #2858, we can now validate this property with unit testing.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9092

PR #2868, opened 2023-12-12
Boost: Use CXX17

dgovil:
### Description of Change(s)
Splitting this out from https://github.com/PixarAnimationStudios/OpenUSD/pull/2742
Boost doesn't build on newer macOS toolchains because it make use of previous deprecated and now removed std lib functions.

In theory, all platforms should use CXX17 but Boost does some more platform aware logic for other toolchains that mitigate it to some degree. At least on macOS, its best to force Boost to use CXX17


jesschimein:
Filed as internal issue #USD-9093

PR #2873, opened 2023-12-15
Add the ability to request CMake to ignore paths

dgovil:
### Description of Change(s)

A common frustration when building USD is that the dependencies (especially OpenImageIO, Ptex, Alembic) eagerly pick up things that are used in the users homebrew directory on macOS. This can cause issues if you're trying to build multi-arch (brew is single arch) and also if you hope to make your distribution distributable with other users.

This change adds a generic multi-arg flag that allows builders to say what paths to ignore.
On Apple Silicon Macs, I also add the option to specifically ignore homebrew. I chose not to add this for Intel Macs because homebrew installs to `/usr/local` on those machines, and that directory may include other useful libs. If an intel user did want to exclude it, they could still pass it to the ignore-paths flag.

This PR is mostly taking Lee Kerley's work for our internal fork and applying it more generically for OpenUSD.

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -421,6 +421,11 @@ def RunCMake(context, force, extraArgs = None):
 
         extraArgs.append('-DCMAKE_OSX_ARCHITECTURES={0}'.format(targetArch))
 
+    if context.ignorePaths:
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9098

PR #2874, opened 2023-12-15
Replace `boost::optional` with `std::optional` in `SdfAllowed`

nvmkuruc:
### Description of Change(s)

#2864 was meant to replace the remaining usage of `boost::optional` in `sdf`, but missed `SdfAllowed`. This completes that work.

There is still a spurious include in the parser that can be addressed once the UTF-8 work (#2848) is integrated.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9099

PR #2875, opened 2023-12-18
Fix for skipARKitRootLayerCheck warning

dgovil:
### Description of Change(s)
I forgot to remove the `not` in the check when providing the deprecation warning.
https://github.com/PixarAnimationStudios/OpenUSD/pull/2863/

This fixes it so it only warns when the flag is actually supplied as True, since the default is False and would have always triggered. 


tallytalwar:
Oops sorry I missed that during review, will get this in right now. Thanks @dgovil for a quick fix.

jesschimein:
Filed as internal issue #USD-9101

Issue #2876, opened 2023-12-18
UsdShadeOutput bool() doesn't check for attr definition


dgovil:
### Description of Issue

We're seeing an issue where UsdShadeOutputs aren't properly validated after a stage is dropped. 

This appears to be because [UsdShadeOutput::IsOutput](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdShade/output.cpp#L182) doesn't mirror [UsdShadeInput::IsInput](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdShade/input.cpp#L200) validation. Specifically the `attr && attr.IsDefined()` is missing.

I can put up a PR shortly but wanted to file the issue first. 

(Thanks to Maddy Adams for catching this)

jesschimein:
Filed as internal issue #USD-9104

PR #2877, opened 2023-12-18
UsdOutput::IsOutput should check attr validity

dgovil:
### Description of Change(s)

UsdShadeOutput::IsOutput should be checking for attr validity like UsdShadeInput::IsInput does.
Thanks to Maddy Adams for catching this and suggesting the fix.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2876

### DIFF
--- a/pxr/usd/usdShade/output.cpp
+++ b/pxr/usd/usdShade/output.cpp
@@ -181,7 +181,8 @@ UsdShadeOutput::ClearSdrMetadataByKey(const TfToken &key) const
 bool 
 UsdShadeOutput::IsOutput(const UsdAttribute &attr)
 {
-    return TfStringStartsWith(attr.GetName().GetString(), 
+    return attr && attr.IsDefined() &&
+        TfStringStartsWith(attr.GetName().GetString(),
                               UsdShadeTokens->outputs);
 }
 


jesschimein:
Filed as internal issue #USD-9105

FlorianZ:
Hi @dgovil. As a pattern, we try to not do a lot of validity checking in these types of API calls. In this case, for example, we expect the client to supply an already valid UsdAttribute. If we made this change, we think there is a ton of other API that would have to be changed to now follow a similar pattern, or otherwise we run the risk of making things behave inconsistently and muddying the waters in terms of whether or not the call site needs to check for validity. Does that make sense? What are your thoughts?

FlorianZ:
Wait, I just saw your comment on https://github.com/PixarAnimationStudios/OpenUSD/issues/2876, and it looks like we are already being inconsistent about what I said we wanted to be consistent about ;-)

Okay, so ignore my previous comment. We'll discuss internally, but it sounds like accepting the PR would make this more consistent actually...

dgovil:
Edit: oops missed your follow up comment. 

----

I totally see where you're coming from but I would argue that this change would bring some symmetry and consistency to the API. 

IsInput already checks for validity, which in turn means that one can follow the same pattern as prims to pass the Input object from a GetInput call to an if check. 

GetOutput would be the outlier here since it doesn't follow its counterpart. 

I wouldn't advocate adding the check elsewhere based on this PR, but I do think this would make validation checking on an output feel more consistent with the rest of the USD API.

PR #2878, opened 2023-12-18
If should only have one f

dgovil:
### Description of Change(s)
I noticed that a lot of older USD files seemed to have had a find/replace accident where `if` was replaced with `iff`.
This is a rather boring PR that just finds all uses of `iff`, and drops the second `f`.

Edit: some folks mentioned that it might have meant [If and only if](https://en.wikipedia.org/wiki/If_and_only_if) which was not an abbreviation that I was aware of. Feel free to disregard if that's the case, but I might suggest adding a disclaimer in the docs somewhere to point to that?

### DIFF
--- a/pxr/base/gf/camera.h
+++ b/pxr/base/gf/camera.h
@@ -228,10 +228,10 @@ public:
     /// Returns the focus distance in world units.
     GF_API float GetFocusDistance() const;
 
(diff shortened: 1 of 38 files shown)


spitzak:
I think that is on purpose, and means "if and only if"

PR #2879, opened 2023-12-19
PEGTL Text File Format Parser

erslavin:
- Added additional text file format parser based on PEGTL
- Added CMake option and preprocessor definition for enabling legacy yacc parser
- Added tests specific to PEGTL parser

### Description of Change(s)

This change introduces a new sdf / usda text file format paser based on PEGTL.  While intended to replace the current lex / yacc parser, this PR enables both to sit side by side and can be compile-time enabled via a new CMake option (defaulting to use the PEGTL parser).  This change aligns the way text file scene descriptions are parsed with the way paths are now being parsed and removes the dependency on lex / yacc (with the compile option enabled).

- Added PEGTL Text File Format Parser
- Added CMake option and preprocessor definition for enabling yacc parser
- Added tests specific to PEGTL parser that allow testing of partial scene description rules

### Fixes Issue(s)
-

### DIFF
--- a/cmake/defaults/Options.cmake
+++ b/cmake/defaults/Options.cmake
@@ -51,6 +51,8 @@ option(PXR_ENABLE_NAMESPACES "Enable C++ namespaces." ON)
 option(PXR_PREFER_SAFETY_OVER_SPEED
        "Enable certain checks designed to avoid crashes or out-of-bounds memory reads with malformed input files.  These checks may negatively impact performance."
         ON)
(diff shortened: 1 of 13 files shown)


erslavin:
This replaces PR #2819 due to the original dropping support for the lex / yacc version (associated internal issue #USD-8970).

jesschimein:
Filed as internal issue #USD-9106

erslavin:
Replaced with #3005

PR #2880, opened 2023-12-19
Update to stb_image version 2.29 for stbi_info() fix

angusdavisadobe:
### Description of Change(s)
Update to the latest version of stb_image

### Fixes Issue(s)
There was a bug fix to stb_image where previously, stbi_info() (and sbti_info_from_memory()) would return the wrong value for nchannels for a grey scale png image with an alpha channel. For a particular png, stbi_info() would incorrectly return 1 but when later calling stbi_load() with a pre-allocated block of memory to fill, the load method would detect that there are two channels and would overflow memory when decoding, causing a crash in hydra.

The changes are to update to version 2.29 of stb_image and make the appropriate updates to pxr/image/hio/stbImage.cpp to account for api changes to the stb_image_resize functions.
 


jesschimein:
Filed as internal issue #USD-9107

tcauchois:
Hi Angus,

This PR isn't building for us; we get an array overrun in stb_image_resize2.h, that's statically detectable by the compiler.  Would you be able to update the PR so it builds?

Thanks!
Tom

PR #2883, opened 2023-12-20
Undo inlining of `TfIsUtf8CodePointXid{Start,Continue}`

nvmkuruc:
### Description of Change(s)

#2857 was focused on reducing the cost / instruction count of initializing the range tables for the `Xid` character classes. It also inlined their usage in `unicodeUtils.{h,cpp}`. However, `unicodeCharacterClasses` is private and cannot be included in the header for `unicodeUtils` to be used outside of `Tf`.

This rolls back just that part of #2857.

It also removes the dependency on the `TF_MAX_CODE_POINT` symbol inherited from `unicodeCharacterClasses` to avoid including the private header in the test case.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/testenv/unicodeUtils.cpp
+++ b/pxr/base/tf/testenv/unicodeUtils.cpp
@@ -24,13 +24,59 @@
 #include "pxr/pxr.h"
 #include "pxr/base/tf/diagnosticLite.h"
 #include "pxr/base/tf/regTest.h"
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-9114

dsyu-pixar (commented):
<deleted>
Please ignore -- my review comment (for PR 2888) was accidentally copied here as well.

PR #2885, opened 2023-12-22
Using get value producing attrs when checking texture assets in NormalMapTextureChecker

sirpalee:
### Description of Change(s)
The change modifies `NormalMapTextureChecker` to use GetValueProducingAttributes instead of querying the value of the asset parameters on the texture. This way, the NormalMapTextureChecker works better with material networks that expose asset parameters on the Material and connect them to shader parameters.

Questions:
- I'm checking the namespace of the producing attribute to see if it is an input parameter. Is there a better way to do this?
- What should we do about asset attributes produced by output parameters? Is that something we should allow, or generally, that's not expected to work? Could we configure the normal map texture checker to turn this feature on/off?

@asluk

### Fixes Issue(s)
No issue was created.

### DIFF
--- a/pxr/usd/usdUtils/complianceChecker.py
+++ b/pxr/usd/usdUtils/complianceChecker.py
@@ -470,7 +470,7 @@ Specifically:
         if not valueProducingAttrs or len(valueProducingAttrs) != 1:
             return None
         # We require an input parameter producing the value.
-        if not UsdShade.Tokens.inputs.startswith(valueProducingAttrs[0].GetNamespace()):
+        if not UsdShade.Input.IsInput(valueProducingAttrs[0]):
             return None
         return valueProducingAttrs[0].Get(Usd.TimeCode.EarliestTime())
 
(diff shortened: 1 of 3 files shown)


pxr/usd/usdUtils/complianceChecker.py
```diff
@@ -464,7 +464,15 @@ def _GetInputValue(self, shader, inputName):
...
         if not input:
             return None
-        return input.Get(Usd.TimeCode.EarliestTime())
+        # Query value producing attributes for input values.
+        # This has to be a length of 1, otherwise no attribute is producing a value.
+        valueProducingAttrs = ShadeUtils.GetValueProducingAttributes(input)
+        if not valueProducingAttrs or len(valueProducingAttrs) != 1:
+            return None
+        # We require an input parameter producing the value.
+        if valueProducingAttrs[0].GetNamespace() != 'inputs':
```
asluk: Maybe use `UsdShade.Tokens.inputs` here

sirpalee: Thanks!

asluk (commented):
Looks good to me; thanks @sirpalee ! (I don't have approval perms though)

tallytalwar:
Filed as internal issue #USD-9116

PR #2888, opened 2023-12-26
Update documentation for UTF-8 utilities

nvmkuruc:
### Description of Change(s)

This change adds some additional examples and corrects some errors in the documentation.

- `TfUtf8CodePointIterator::operator*` suggested that it might throw an `out_of_range` exception, but that won't happen under the current implementation.
- Add `group_tf_String` grouping for UTF-8 utilities
- Add code examples

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/unicodeUtils.h
+++ b/pxr/base/tf/unicodeUtils.h
@@ -26,6 +26,7 @@
 
 /// \file tf/unicodeUtils.h
 /// \ingroup group_tf_String
+/// Definitions of basic UTF-8 utilities in tf.
 
 #include "pxr/pxr.h"
 #include "pxr/base/tf/api.h"
@@ -37,11 +38,22 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-/// Wrapper for a code point value that can be encoded as UTF-8
+/// \class TfUtf8CodePoint
+/// \ingroup group_tf_String
+///
+/// Wrapper for a 32-bit code point value that can be encoded as UTF-8.
+///
+/// \code{.cpp}
+/// // Stream operator overload encodes each code point as UTF-8.
+/// std::stringstream s;
+/// s << TfUtf8CodePoint(8747) << " " << TfUtf8CodePoint(120);
+/// \endcode
+/// A single `TfUtf8CodePoint` may be converted to a string using
+/// `TfStringify` as well.
 class TfUtf8CodePoint {
 public:
-    /// Code points that cannot be decoded or outside of the valid range are
-    /// may be replaced with this value.
+    /// Code points that cannot be decoded or are outside of the valid range
+    /// will be replaced with this value.
     static constexpr uint32_t ReplacementValue = 0xFFFD;
 
     /// Values higher than this will be replaced with the replacement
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-9119

pxr/base/tf/unicodeUtils.h
```diff
@@ -26,63 +26,139 @@
...
+/// \code{.cpp}
+/// // Stream operator overload encodes each code point as UTF-8.
+/// std::stringstream s;
+/// s << TfUtf8CodePoint(8747) << " " << TfUtf8CodePoint(120);
+/// \endcode
+/// A single `TfUtf8CodePoint` may be converted to a string using
+/// `TfStringify` as well.
+class TfUtf8CodePoint {
+public:
+    /// Code points that cannot be decoded or outside of the valid range are
```
dsyu-pixar: Minor nit-pick: maybe change to "Code points that cannot be decoded or are outside of the valid range"

pxr/base/tf/unicodeUtils.h
```diff
@@ -26,63 +26,139 @@
...
+    /// may be replaced with this value.
```
dsyu-pixar: Another nit-pick: my interpretation of "may be replaced" suggests that there are scenarios where we won't replace invalid code points with ReplacementValue. If this isn't true, it might be clearer to say "will be replaced".

pxr/base/tf/unicodeUtils.h
```diff
@@ -164,38 +239,49 @@ class TfUtf8CodePointIterator final {
...
+    /// the current one.
     TfUtf8CodePointIterator& operator++ ()
     {
+        // The increment operator should never be called if it's past
+        // the end. The user is expected to have already checked this
+        // condition.
+        TF_DEV_AXIOM(!_IsPastTheEnd());
+        _EncodingLength increment = _GetEncodingLength();
         // note that in cases where the encoding is invalid, we move to the
         // next byte this is necessary because otherwise the iterator would
```
dsyu-pixar: Nit-pick that I realize is in a part of the PR that wasn't changed (and also isn't part of a documentation comment): Maybe break into separate sentences or otherwise add punctuation to separate things --> "we move to the next byte. This is necessary..."

pxr/base/tf/unicodeUtils.h
```diff
@@ -164,38 +239,49 @@ class TfUtf8CodePointIterator final {
...
-        // never be satisfied
-        _EncodingLength encodingLength = _GetEncodingLength();
-        std::advance(_it, (encodingLength != 0) ? encodingLength : 1);
-        if (_IsPastTheEnd()) {
-            _it = _end;
+        // never be satisfied. This means that we increment, even if the
+        // encoding length is 0.
+        ++_it;
+        // Only continuation bytes will be consumed after the the first byte.
+        // This avoid consumption of ASCII characters or other starting bytes.
```
dsyu-pixar: Minor: "avoid" -> "avoids"

pxr/base/tf/unicodeUtils.h
```diff
@@ -94,6 +170,18 @@ class TfUtf8CodePointView final {
         return _view.empty();
     }
 
+    /// Returns an iterator of the same type as begin that identifies the end
```
dsyu-pixar: Optional, but it might help readability to put "begin" in code-font or italics. Maybe also applies to "PastTheEndSentinel" below.

dsyu-pixar (commented):
Reviewed doc comment changes and had a few very minor suggestions. Otherwise, looks great!

PR #2889, opened 2023-12-26
Untangle `TfUtf8CodePointIterator` from view class

nvmkuruc:
### Description of Change(s)

(Depends on #2888)

The sentinel for the `TfUtf8CodePointIterator` class is defined in the `TfUtf8CodePointView` class, and the iterator's constructor is `private`, only indirectly constructable via its `TfUtf8CodePointView` friend class.  Complecting the two classes prevented direct usage of the `TfUtf8CodePointIterator` in cases like #2859 where standard STL utilities were used to find the first-non ASCII byte and didn't require explicit decoding of every preceding character.

This makes `TfUtf8CodePointIterator`'s constructor public and moves the sentinel to the iterator so it no longer has any dependencies on the view for construction or to hold the sentinel definition.  This reorders the code so the view now appears after the iterator, as it needs the iterator's sentinel definition.

### Fixes Issue(s)
-


tallytalwar:
Filed as internal issue #USD-9120

PR #2891, opened 2023-12-28
Add `TfDelegatedCountPtr` to support custom bookkeeping

nvmkuruc:
### Description of Change(s)

OpenUSD currently uses `boost::intrusive_ptr` to support customized bookkeeping of types.  Specialization is achieved through ADL (`intrusive_ptr_add_ref` and `intrusive_ptr_release`).

This introduces `TfDelegatedCountPtr` as an alternative to support customized bookkeeping via ADL overloads `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement`.

The `TfDelegatedCountPtr` avoids implicit conversion of raw pointers and requires explicit tag dispatch to specify whether the pointer should be retained or not.

This also provides a `TfMakeDelegatedCountPtr` creator function to avoid direct allocation.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/delegatedCountPtr.h
+++ b/pxr/base/tf/delegatedCountPtr.h
@@ -1,0 +1,280 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the Apache License, Version 2.0 (the "Apache License")
+// with the following modification; you may not use this file except in
+// compliance with the Apache License and the following modification to it:
+// Section 6. Trademarks. is deleted and replaced with:
+//
+// 6. Trademarks. This License does not grant permission to use the trade
+//    names, trademarks, service marks, or product names of the Licensor
+//    and its affiliates, except as required to comply with Section 4(c) of
+//    the License and to reproduce the content of the NOTICE file.
+//
+// You may obtain a copy of the Apache License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the Apache License with the above modification is
+// distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the Apache License for the specific
+// language governing permissions and limitations under the Apache License.
+//
+
+#ifndef PXR_BASE_TF_DELEGATED_COUNT_PTR_H
+#define PXR_BASE_TF_DELEGATED_COUNT_PTR_H
+
+#include "pxr/pxr.h"
+#include "pxr/base/tf/tf.h"
+#include "pxr/base/tf/api.h"
+
+#include "pxr/base/tf/diagnosticLite.h"
+
+#include <memory>
+#include <type_traits>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// When constructing a `TfDelegatedCountPtr` from a raw pointer, use the
+/// `TfDelegatedCountIncrementTag` to explicitly signal that the pointer's
+/// delegated count should be incremented on construction. This is the most
+/// common tag.
+constexpr struct TfDelegatedCountIncrementTagType{}
+    TfDelegatedCountIncrementTag{};
+
+/// When constructing a `TfDelegatedCountPtr` from a raw pointer, use the
+/// `TfDelegatedCountDoNotIncrementTag` to avoid incrementing the delegated
+/// count on construction. This must be carefully used to avoid memory errors.
+constexpr struct TfDelegatedCountDoNotIncrementTagType{}
+    TfDelegatedCountDoNotIncrementTag{};
+
+/// Stores a pointer to a `ValueType` which uses `TfDelegatedCountIncrement`
+/// and `TfDelegatedCountDecrement` to bookkeep.
+///
+/// The `TfDelegatedCountPtr` is responsible for calling
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` as needed in
+/// construction, assignment, and destruction operations.
+///
+/// The increment and decrement methods can assume pointers are non-null.
+/// Releasing resources (e.g. freeing memory) is delegated to the user defined
+/// `TfDelegatedCountDecrement` overloads.
+///
+/// This may be created from a user constructed raw pointer or via
+/// `TfMakeDelegatedCountPtr`.
+template <typename ValueType>
+class TfDelegatedCountPtr {
+public:
+    using RawPtrType = std::add_pointer_t<ValueType>;
+    using ReferenceType = std::add_lvalue_reference_t<ValueType>;
+    using element_type = ValueType;
+
+    static_assert(
+        std::is_same_v<
+            void,
+            decltype(TfDelegatedCountIncrement(std::declval<RawPtrType>()))>);
+    static_assert(
+        std::is_same_v<
+            void,
+            decltype(TfDelegatedCountDecrement(std::declval<RawPtrType>()))>);
+
+    using IncrementIsNoExcept =
+        std::integral_constant<
+            bool,
+            noexcept(TfDelegatedCountIncrement(std::declval<RawPtrType>()))>;
+    using DecrementIsNoExcept =
+        std::integral_constant<
+            bool,
+            noexcept(TfDelegatedCountDecrement(std::declval<RawPtrType>()))>;
+    using IncrementAndDecrementAreNoExcept =
+        std::integral_constant<
+            bool, IncrementIsNoExcept() && DecrementIsNoExcept()>;
+    using DereferenceIsNoExcept =
+        std::integral_constant<bool, noexcept(*std::declval<RawPtrType>())>;
+
+private:
+    template <typename ConvertibleType>
+    using _IsPtrConvertible = std::is_convertible<
+        std::add_pointer_t<ConvertibleType>, RawPtrType>;
+
+public:
+    /// Create a pointer storing `nullptr`
+    TfDelegatedCountPtr() noexcept = default;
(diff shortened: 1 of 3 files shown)


tallytalwar:
Filed as internal issue #USD-9122

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
+//
+// Copyright 2023 Pixar
```
gitamohr: Could you update this to 2024?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+/// Stores a pointer to a `ValueType` which uses `TfDelegatedCountIncrement`
+/// and `TfDelegatedCountDecrement` to bookkeep.
+///
+/// The `TfDelegatedCountPtr` is responsible for calling
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` as needed in 
+/// construction, assignment, and destruction operations.
+///
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` are expected to
+/// be `noexcept`. The increment and decrement methods can assume pointers are
+/// non-null. Freeing of memory is delegated to `TfDelegatedCountDecrement`.
```
gitamohr: Might be good to say smth like, "Releasing resources (e.g. freeing memory) is delegated to...".

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+        _pointer(ptr.get()) {
+        ptr._pointer = nullptr;
+    }
+
+    /// Assigns by copying the pointer from `ptr`.
+    ///
+    /// Retains only if the pointer is valid.
+    TfDelegatedCountPtr& operator=(const TfDelegatedCountPtr& ptr) noexcept {
+        if (std::addressof(ptr) != this) {
+            _ReleaseIfValid();
```
gitamohr: I think there's a bug here where if `*this` transitively owns `ptr`, then releasing `this` first can cause `ptr` to dangle.  I think we have to retain `ptr` first, then release `*this`.

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+    /// This is necessary in part to allow a `const` `ValueType` to be
+    /// constructed from from a non-`const` `ValueType`.
+    ///
+    /// Retains only if the pointer is valid.
+    template <typename OtherType>
+    TfDelegatedCountPtr& operator=(
+        const TfDelegatedCountPtr<OtherType>& ptr) noexcept {
+        static_assert(_IsPtrConvertible<OtherType>::value);
+        // Does this need to guard against self assignment?
+        _ReleaseIfValid();
```
gitamohr: Same bug as above here -- have to retain `ptr` before releasing `*this`.

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+        _pointer = ptr.get();
+        _RetainIfValid();
+        return *this;
+    }
+
+    /// Assigns by moving the pointer from `ptr`.
+    ///
+    /// `ptr` will be reset to its default state (ie. `nullptr`).
+    TfDelegatedCountPtr& operator=(TfDelegatedCountPtr&& ptr) noexcept {
+        if (std::addressof(ptr) != this) {
```
gitamohr: Same bug as above.

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+        return *this;
+    }
+
+    /// Releases if the underlying pointer is valid 
+    ~TfDelegatedCountPtr() noexcept {
+        _ReleaseIfValid();
+    }
+
+    /// Dereference the underlying pointer
+    ReferenceType operator*() const {
```
gitamohr: `noexcept`?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+
+/// Stores a pointer to a `ValueType` which uses `TfDelegatedCountIncrement`
+/// and `TfDelegatedCountDecrement` to bookkeep.
+///
+/// The `TfDelegatedCountPtr` is responsible for calling
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` as needed in 
+/// construction, assignment, and destruction operations.
+///
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` are expected to
+/// be `noexcept`. The increment and decrement methods can assume pointers are
```
gitamohr: Requiring these to be `noexcept` seems potentially too restrictive to me...  What's the motivation for this requirement?  Could we just transparently propagate the `noexcept`-ness of these so that it's a "pay for what you use" type of thing?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,268 @@
...
+    /// necessary.
+    void reset() noexcept {
+        _ReleaseIfValid();
+        _pointer = nullptr;
+    }
+
+    /// Swaps the pointer between two delegated count pointers
+    void swap(TfDelegatedCountPtr& other) noexcept {
+        using std::swap;
+        swap(other._pointer, _pointer);
```
gitamohr: Do we want ADL here?  I kind of think that we really want to literally swap the pointers.

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,271 @@
...
+
+/// Stores a pointer to a `ValueType` which uses `TfDelegatedCountIncrement`
+/// and `TfDelegatedCountDecrement` to bookkeep.
+///
+/// The `TfDelegatedCountPtr` is responsible for calling
+/// `TfDelegatedCountIncrement` and `TfDelegatedCountDecrement` as needed in
+/// construction, assignment, and destruction operations.
+///
+/// As it may be called in a destructor, `TfDelegatedCountDecrement` is expected
+/// to be `noexcept`. The increment and decrement methods can assume pointers
```
gitamohr: We can remove this comment now, no?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,271 @@
...
+    using RawPtrType = std::add_pointer_t<ValueType>;
+    using ReferenceType = std::add_lvalue_reference_t<ValueType>;
+    using element_type = ValueType;
+
+    static_assert(std::is_same_v<void, decltype(TfDelegatedCountIncrement(RawPtrType{}))>);
+    static_assert(std::is_same_v<void, decltype(TfDelegatedCountDecrement(RawPtrType{}))>);
+
+    using IncrementIsNoExcept =
+        std::integral_constant<
+            bool, noexcept(TfDelegatedCountIncrement(RawPtrType{}))>;
```
gitamohr: could you use `std::declval<RawPtrType>()` for the value here?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,271 @@
...
+        return *this;
+    }
+
+    /// Releases if the underlying pointer is valid 
+    ~TfDelegatedCountPtr() noexcept(DecrementIsNoExcept()) {
+        _DecrementIfValid();
+    }
+
+    /// Dereference the underlying pointer
+    ReferenceType operator*() const {
```
gitamohr: this can be `noexcept` I think.

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,271 @@
...
+        static_assert(_IsPtrConvertible<OtherType>::value);
+        // Implement copy assigment in terms of move assignment
+        return (*this = TfDelegatedCountPtr{ptr});
+    }
+
+    /// Assigns by moving the pointer from `ptr`.
+    ///
+    /// `ptr` will be reset to its default state (ie. `nullptr`).
+    TfDelegatedCountPtr& operator=(TfDelegatedCountPtr&& ptr)
+    noexcept(DecrementIsNoExcept()) {
```
gitamohr, line 182: ok if I had a `foo` that was a `TfDelegatedCountPtr<Foo>` and I did:
```
foo = std::move(foo);
```
That would blow up, right?

pxr/base/tf/delegatedCountPtr.h
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /// Swaps the pointer between two delegated count pointers
+    void swap(TfDelegatedCountPtr& other) noexcept {
+        std::swap(other._pointer, _pointer);
+    }
+
+private:
+    void _IncrementIfValid() noexcept(IncrementIsNoExcept()) {
+        if (get()) {
```
gitamohr: any reason not to just do `if (_pointer)` here?  we use it directly in the body immediately after -- it seems clearer to me that way.

pxr/base/tf/delegatedCountPtr.h
nvmkuruc: There's a test case for this specifically.  It doesn't blow up.  `foo` is left in a valid state. See `TestMovingSelf()`.

pxr/base/tf/delegatedCountPtr.h
nvmkuruc: A bunch of prior usages of `_pointer` had to be converted to `get()` to support operations on convertible types without `friend`-ship.  But there's no reason this usage had to be converted.

PR #2893, opened 2023-12-28
Replace `boost::intrusive_ptr` with `TfDelegatedCountPtr` for `Sdf_Identity`

nvmkuruc:
### Description of Change(s)

(Depends on #2891)

#2891 provided a `TfIntrusiveTrackingPtr` to support custom bookkeeping. This replaces the usage of `boost::intrusive_ptr` with `TfDelegatedCountPtr` for `Sdf_Identity`. As `TfDelegatedCountPtr` does not support implicit conversion from raw pointers, construction with explicit dispatch tags are now used.

### Fixes Issue(s)
-


tallytalwar:
Filed as internal issue #USD-9124

PR #2898, opened 2023-12-28
Add explicit `TfHash` overload for `std::string_view`

nvmkuruc:
### Description of Change(s)

#2781 allowed `TfHash` to leverage `std::hash`.  That change allowed `std::string_view` to be hashable by `TfHash` indirectly via `std::hash`.  However, to leverage the specialized `AppendContinuous` implementation that `std::string` and C-strings use, a `TfHashAppend` specialization is required. This change adds that specialization.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/hash.h
+++ b/pxr/base/tf/hash.h
@@ -32,10 +32,11 @@
 #include "pxr/base/tf/api.h"
 
 #include <cstring>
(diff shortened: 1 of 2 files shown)


tallytalwar:
Filed as internal issue #USD-9129

PR #2899, opened 2023-12-29
Make static `TfToken` macros usable outside of `PXR_NS`

nvmkuruc:
### Description of Change(s)

The static `TfToken` macros make unqualified lookups to various types in the `PXR_NS` namespace. This limits their usage in headers to the `PXR_NS` (source `.cpp` files can use the `PXR_NAMESPACE_USING_DIRECTIVE` macro). This prefixes all usage of types in macros with `PXR_NS` to enable usage in headers.

### Fixes Issue(s)
-


tallytalwar:
Filed as internal issue #USD-9130

PR #2904, opened 2024-01-06
Added materialx library as resources in usdMtlx

dlarsson-adobe:
### Description of Change(s)

This PR tries to make distribution of the MaterialX standard library relocatable
for usdMtlx.
This allows an application to find the MaterialX standard library relative to the plugin's
installed location rather than rely on system wide environment variables or paths burned
into the plugin library at compile time

It implements it by

* Installing the MaterialX standard library as resources in the usdMtlx plugin.
* Adds a lowest priority search path for materialx libraries pointing to the 
  resource location of the plugin

### Fixes Issue(s)
Distribution of materialx applications that can't rely on system environment variables
for the MaterialX standard library


jesschimein:
Filed as internal issue #USD-9138

pxr/usd/usdMtlx/CMakeLists.txt
```diff
@@ -1,8 +1,33 @@
...
+    foreach(file ${MATERIALX_STD_LIB_FILES})
+        # Create the relative path for the destination in the resource destination
+        file(RELATIVE_PATH _resource_relative ${MATERIALX_STDLIB_DIR} ${file})
+
+        # XXX Create the path for the source file relative to the current source
+        # directory. This is done to avoid absolute paths which is not compatible with
+        # _install_resource_files in Private.cmake which uses : as a separator. This breaks 
+        # for absolute paths on windows which uses them for drive letters (c:\blah)
+        # This will still fail if MATERIALX_STDLIB_DIR points to assets on a different
+        # drive than the usd source directory on windows
```
spiffmon, line 23: Do we know for sure this is broken on Windows? Looks like [we are attempting to make this work](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/cmake/macros/Private.cmake#L277-L282), and therefore if it's not, an Issue is warranted?

dlarsson-adobe: It's probably a stale comment by now. I see a change for this in USD from September 2024 (which after this PR was created) and it appears to adress this. I did hit the issue when I was developing it so I wanted to point it out in case anyone else runs into but it might be worth revisiting and remove the comment

PR #2905, opened 2024-01-11
Added nullptr check for _FindEntry (like _InsertEntry and related)

marcodiiga:
### Description of Change(s)

Added nullptr check for `_FindEntry` (`_InsertEntry` and others all assume that the operation might fail, makes sense to include the `_FindEntry` as well). This can also happen (albeit erroneously and unsupportedly) in multithreaded situations and possibly in resource constrained conditions.

### DIFF
--- a/pxr/usd/usdGeom/bboxCache.cpp
+++ b/pxr/usd/usdGeom/bboxCache.cpp
@@ -1120,6 +1120,9 @@ UsdGeomBBoxCache::_Resolve(
     
     // If the bound is in the cache, return it.
     entry = _FindEntry(primContext);
+    if (entry == nullptr) {
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9152

asluk:
@jesschimein  this would be a good one to get in for v24.03 -- let me know if the USD team at Pixar have any particular concerns-- ideally there would have been a unit test case as well, but we weren't able to quite nail down a repro. In any case it's good to guard against crashes here, and I don't think there is any downside to this. Happy to talk through it more though. Thanks!

pixar-oss:
Thanks for flagging!  We're looking at it internally for inclusion for 24.03!

PR #2906, opened 2024-01-17
Change value type of `TfUtf8CodePointIterator` to `TfUtf8CodePoint`

nvmkuruc:
### Description of Change(s)

`TfUtf8CodePointIterator` was originally written before the `TfUtf8CodePoint` class was introduced.  This change makes the value and reference type of the iterator a `TfUtf8CodePoint` directly instead of expecting clients to construct it themselves.

It also introduces a `TfUtf8CodePointFromAscii` helper function to make it easier to create code points from known ASCII characters.

This change also simplifies several test cases and usages.  For example, `_IsValidIdentifier` no longer has to branch.

Currently, both the iterator and the code point class employ validation to ensure that code points outside the unicode range or surrogate code points are not decoded. It may be desirable to avoid this redundancy so that the validation is delegated entirely to the code point class in a future change.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/testenv/unicodeUtils.cpp
+++ b/pxr/base/tf/testenv/unicodeUtils.cpp
@@ -75,6 +75,13 @@ TestUtf8CodePoint()
                  TfStringify(TfUtf8InvalidCodePoint));
 
     }
+    {
+        // Test ASCII character helper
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9180

PR #2908, opened 2024-01-18
Autodesk: Bugfix: Enable customizing the suffix of constructor and destructor section name

erikaharrison-adsk:
### Description of Change(s)

Background:

USD put some functions in a section 'pxrctor', which perform initialization work after loading the module.  Also, they are expected to be called only once.

Now say we have library A linked to USD build A; and executable B has its own USD build B with modified dylib name. USD build A and B are different.  USD build A would be loaded first, then get a function lists via iterating all sections named 'pxrctor' from all loaded modules, and then call each of them. ie. intialize them. After that, while loading USD build B, the initialization process would also get a function lists in the same way, and it would also contain functions lists collected from USD build A. Again, all functions would be called, including those from USD build A. That means those functions from USD build A would be called twice. This leads to a crash in our use case.

Solution:

Enable the ability to customize and append a suffix to the section name.  The corresponding build argument is
`--build-args=USD,"-DPXR_INIT_SECTION_NAME_SUFFIX=_v2"`

### Fixes Issue(s)
- N/A

### DIFF
--- a/cmake/defaults/Options.cmake
+++ b/cmake/defaults/Options.cmake
@@ -217,3 +217,9 @@ if (${PXR_BUILD_PYTHON_DOCUMENTATION})
         set(PXR_BUILD_PYTHON_DOCUMENTATION "OFF" CACHE BOOL "" FORCE)
     endif()
 endif()
+
+# Customize the suffix of constructor and destructor section names
+if (DEFINED PXR_INIT_SECTION_NAME_SUFFIX)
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9189

FlorianZ:
@erikaharrison-adsk, thanks for your PR! We are actually looking at this internally, and received another PR #2861 which appears to tackle this same issue. Thanks again!

spiffmon:
Closing, as we addressed this with commit 319d589 - thank you!

PR #2914, opened 2024-01-22
Build boost libs only when necessary

tylerm-nv:
### Description of Change(s)
- Removed boost atomic. The commit it was added in indicates that it was needed for OpenImageIO, but the current version doesn't link or include anything from boost atomic. None of the other libraries seem to use it either. boost filesystem has a dependency on boost atomic if not compiled with C++ 20, so it is still built anyways for now.
- Moved the boost regex switch to openvdb, could not find any other libraries using it.


jesschimein:
Filed as internal issue #USD-9200

PR #2918, opened 2024-01-25
ts: fix for TsSimplify + testTsSimplify

pmolodo:
### Description of Change(s)

On windows debug builds, was getting a "Debug Assertion Failed" on [simplify.cpp, line 379](https://github.com/PixarAnimationStudios/OpenUSD/blob/35da196173554a40517cfe673610f3f2b3705b8e/pxr/base/ts/simplify.cpp#L379).  This is because this we were using iterators that have become invalid because the underlying data was altered.

### Fixes Issue(s)
- failure of testTsSimplify on windows debug builds


jesschimein:
Filed as internal issue #USD-9206

PR #2919, opened 2024-01-25
Disable BOOST_UNORDERED_HAVE_PIECEWISE_CONSTRUCT for MacOS builds

dgovil:
### Description of Change(s)

When building with Boost 1.82 or prior on macOS with clang 15 or higher, Boost will fail to compile due to the missing `std::piecewise_construct` constant.

std::piecewise_construct_t is an empty class tag type used to disambiguate between different functions that take two tuple arguments, and is present. std::piecewise_construct is a constant instant of it that doesn't seem to exist.

Since this is just a sentinal type, boost provides falling back to its own implementation. In theory this patch doesn't need to be platform specific, but for now I've set it to macOS only.


jesschimein:
Filed as internal issue #USD-9207

Issue #1052, opened 2019-12-04
[usdview] Loading a file resets the renderer to GL but the menu is not updated


digicpictures:
### Description of Issue
When a USD file is loaded the renderer in the 3d viewport gets reset to GL but the Hydra renderer menu still displays the previously set renderer. It should either retain the set renderer or reset the menu to.

### Steps to Reproduce
1. Open a USD file
2. Change renderer
3. Open another USD file
4. The viewport uses GL but in the menu the previously active renderer is shown.

### System Information (OS, Hardware)
Win 10

### Package Versions
USD 0.19.11

jilliene:
Filed as internal issue #USD-5739

PR #2927, opened 2024-01-31
Update Selected Renderer on File load / reload

beersandrew:
### Description of Change(s)
- added a signal in stageView to call back to update the UI in the case the renderer has changed
- this happens on file load / reload when the renderer is set to null and reinitialized again when the first frame is drawn, it's reinitialized to Storm but the UI was not being notified of the change

### Fixes Issue(s)
- #1052


jesschimein:
Filed as internal issue #USD-9224

PR #2929, opened 2024-02-02
tf: aarch64 fix for cases where CNTFRQ_EL0 returns bogus value

pmolodo:
### Description of Change(s)

As noted in this commit in the linux kernel:

https://github.com/torvalds/linux/commit/c6f97add0f2ac83b98b06dbdda58fa47638ae7b0

...the value of CNTFRQ_EL0 is sometimes unreliable.  The linux kernel instead reads the tick rate from the device tree, and if that fails, only then falls back on CNTFRQ_EL0.

Since we already have measurement-based code, and reading from the device tree seemed tricky, we instead check if CNTFRQ_EL0 seems "sane" (ie, > 1Hz), and if not, fall back on the measurement code used in all other linux flavors.

I ran across this issue when running an aarch64 vm running on aarch64 hardware - I'm not sure if the issue was with the actual hardware or just the VMWare vm.  In this case, CNTFRQ_EL0 was returning a negative value.  Not sure if all "bad" values will be as obvious, but this change is at least an improvement.

### Fixes Issue(s)
- Arch timing functions and TfStopwatch on linux aarch64 VMs (and possibly other aarch64 platforms?)

### DIFF
--- a/pxr/base/arch/testenv/testTiming.cpp
+++ b/pxr/base/arch/testenv/testTiming.cpp
@@ -32,6 +32,9 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 int main()
 {
+    ARCH_AXIOM(ArchGetNanosecondsPerTick() > 0.0);
+    // If you're not even doing 1 tick a second, it's probably a bogus value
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9239

PR #2930, opened 2024-02-02
[al] Invalid Prims Validation in UsdImagingMaterialAdapter::GetMaterialResource()

NickWu:
### Description of Change(s)
We came crossed this issue in our pipeline where there is a possibility that the `UsdPrim` passed in to `UsdImagingMaterialAdapter::GetMaterialResource()` could be invalid so we'd like to propose a prim validation check here.

### Fixes Issue(s)
- crashes in maya when displaying textures in the viewport. 

### DIFF
--- a/pxr/usdImaging/usdImaging/materialAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/materialAdapter.cpp
@@ -335,8 +335,12 @@ UsdImagingMaterialAdapter::GetMaterialResource(UsdPrim const &prim,
                                                UsdTimeCode time) const
 {
     TRACE_FUNCTION();
-    if (!prim || !_GetSceneMaterialsEnabled()) {
+    if (!prim) {
         TF_RUNTIME_ERROR("Received prim is null.");
+        return VtValue();
+    }
+
+    if (!_GetSceneMaterialsEnabled()) {
         return VtValue();
     }
 
--- a/pxr/usdImaging/usdImaging/materialAdapter.cpp
+++ b/pxr/usdImaging/usdImaging/materialAdapter.cpp
@@ -336,6 +336,7 @@ UsdImagingMaterialAdapter::GetMaterialResource(UsdPrim const &prim,
 {
     TRACE_FUNCTION();
     if (!prim || !_GetSceneMaterialsEnabled()) {
+        TF_RUNTIME_ERROR("Received prim is null.");
         return VtValue();
     }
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9240

FlorianZ:
Hi @NickWu. Good find! Our theory is that if `prim` is invalid, `material` should be invalid, too, but then we "dereference" the prim in generating the TF_RUNTIME_ERROR. We might not accept the PR as is, but will take a look internally! Thanks again!

NickWu:
> Hi @NickWu. Good find! Our theory is that if `prim` is invalid, `material` should be invalid, too, but then we "dereference" the prim in generating the TF_RUNTIME_ERROR. We might not accept the PR as is, but will take a look internally! Thanks again!

No worries!

tcauchois:
Hey Nick,

In our current usage of this function, I'd consider it an error for this function to be passed a null prim.  If I replace the early out with an error and an early out, you'd get an error message but no crash.  Would that solve your issue?

Thanks!
Tom

NickWu:
> Hey Nick,
> 
> In our current usage of this function, I'd consider it an error for this function to be passed a null prim. If I replace the early out with an error and an early out, you'd get an error message but no crash. Would that solve your issue?
> 
> Thanks! Tom

Hey @tcauchois , I'll give it a try and get back to you. Thank you for the suggestion.

NickWu:
Hi @tcauchois , I tested it and it worked fine to me. I have updated the code based on your suggestion.

NickWu:
> and get back to you. Thank you for the suggestion.

Hey @tcauchois , I think I might have misunderstood what you meant. I think you meant we only issue an error and early out in the case of a null prim and keep the `_GetSceneMaterialsEnabled()` check a simple early out? 

```C++
if (!prim) {
    TF_RUNTIME_ERROR("Received prim is null.");
    return VtValue();
}

if (!_GetSceneMaterialsEnabled()) {
    return VtValue();
}
```

tcauchois:
Yeah exactly, the code in your comment was what I had in mind.  But if the current PR works for you, the code in your comment should as well.  Thanks!

tcauchois:
Hi Nick, could you update your PR with the code proposed above?  Thanks!

NickWu:
> Hi Nick, could you update your PR with the code proposed above? Thanks!

Sure thing. I'll update it. Thanks!

NickWu:
> > Hi Nick, could you update your PR with the code proposed above? Thanks!
> 
> Sure thing. I'll update it. Thanks!

Hi @tcauchois , I have updated the code.

PR #2931, opened 2024-02-05
Fix a potential crash when updating skinned primitives.

### LINKED ISSUE #2809: Crash when displaying skinned primitives and then switching to the BakeSkinning result
### Description of Issue
After upgrading from USD 22.05 to 23.08 we encountered crashes in Hydra after displaying skinned primitives and then displaying the result of applying BakeSkinning to those prims.

From doing some bisection, I tracked it down to this commit: 6b3afc2f29993dc34b8113d15f1179302cfe7528

This can be reproduced with the Embree delegate in `usdview`, and also with Karma and Houdini GL within Houdini (Renderman would likely reproduce this as well, but I haven't tested). I wasn't able to reproduce it in Storm, possibly from its different code paths for handling computed primvars.

### Steps to Reproduce
1. Extract [usdskelcrash.tar.gz](https://github.com/PixarAnimationStudios/OpenUSD/files/13354229/usdskelcrash.tar.gz) and enter the folder.
2. Run `usdview usdskel.usda -r Embree`
3. In the interpreter, run
```
l = Sdf.Layer.FindOrOpen("usdskel_baked.usda")
usdviewApi.stage.GetRootLayer().TransferContent(l)
```
4. This should result in a crash in `HdExtComputationUtils`. This seems to indicate some stale data being accessed, since there should no longer be any skinning (computed primvars)

Ubuntu 18.04, gcc 8

### Package Versions

USD 23.08

cameronwhite:
This fixes the crash reported in #2809, which contains the example file and steps to reproduce.
We've had this patch in Houdini for a couple weeks and haven't encountered any issues in testing

### Description of Change(s)

With 6b3afc2f, skinned primitives are now replaced by an entry with no type and no data source instead of being removed, since they have child prims (the skinning computation Sprims).
When doing so, the entry's cache of computed primvars was never cleared.

This could result in a crash in scenarios such as switching to displaying the result of a BakeSkinning operation. In this case, the skinned primitive changes type from e.g. a Mesh (with a computed primvar for the skinned positions) to empty, and then back to Mesh (which should have no computed primvars). Since the entry's primvar cache was never cleared, this incorrectly returns a stale list of computed primvars for the updated mesh which later produces a crash.

When the prim type of an existing entry is being changed, the cached primvars are now also cleared out while removing the old prim from the render index, to produce the same behaviour as if a new entry was inserted

### Fixes Issue(s)
- Fixes #2809

### DIFF
--- a/pxr/imaging/hd/sceneIndexAdapterSceneDelegate.cpp
+++ b/pxr/imaging/hd/sceneIndexAdapterSceneDelegate.cpp
@@ -188,6 +188,16 @@ HdSceneIndexAdapterSceneDelegate::_PrimAdded(
             } else if (existingType == HdPrimTypeTokens->instancer) {
                 GetRenderIndex()._RemoveInstancer(indexPath);
             }
+
+            // If the prim type of an existing entry changed, also clear any
+            // cached data associated with it, e.g. computed primvars.
+            entry.primvarDescriptors.clear();
+            entry.primvarDescriptorsState.store(
+                _PrimCacheEntry::ReadStateUnread);
+            entry.extCmpPrimvarDescriptors.clear();
+            entry.extCmpPrimvarDescriptorsState.store(
+                _PrimCacheEntry::ReadStateUnread);
+
         } else {
             insertIfNeeded = false;
         }


jesschimein:
Filed as internal issue #USD-9255

cameronwhite:

I should already be registered under the SideFX CLA, and have done a few prior PRs already

PR #2933, opened 2024-02-05
Incorporate change to set umask properly when creating USD files

marktucker:
### Description of Change(s)

Incorporate change to set umask properly when creating USD files, as
described here: https://github.com/PixarAnimationStudios/USD/issues/1604,
using bitwise negation rather than numeric subtraction (which is equivalent in
some cases, but not all).

### Fixes Issue(s)
- 1604


jesschimein:
Filed as internal issue #USD-9259

marktucker:
I have signed a CLA through SideFX.

PR #2934, opened 2024-02-05
Ignore reparse points on network file systems. The destination is unl…

marktucker:
### Description of Change(s)

Ignore reparse points on network file systems. The destination is unlikely
to be accessible from the remote machine. So for example, machine A points
to a network location on machine B (//B/share/path/to/some/directory). But on
machine B, the directory "C:/path/to/some" that corresponds to
"//B/share/path/to/some" is a reparse point that maps to "D:/another/location".
If machine A tries to "simplify" this path using knownledge of the reparse point,
it ends up with a path like "D:/another/location/directory", which is totally
meaningless on machine A. So in this case we have to just ignore the reparse
point and machine A can only access the directory using the full path 
"//B/shared/path/to/some/directory".

### DIFF
--- a/pxr/base/tf/fileUtils.cpp
+++ b/pxr/base/tf/fileUtils.cpp
@@ -83,8 +83,7 @@ Tf_HasAttribute(
     if (path.back() == '/' || path.back() == '\\')
         resolveSymlinks = true;
 
-    const DWORD attribs =
-        GetFileAttributesW(ArchWindowsUtf8ToUtf16(path).c_str());
+    DWORD attribs = GetFileAttributesW(ArchWindowsUtf8ToUtf16(path).c_str());
     if (attribs == INVALID_FILE_ATTRIBUTES) {
         if (attribute == 0 &&
             (GetLastError() == ERROR_FILE_NOT_FOUND ||
@@ -94,6 +93,21 @@ Tf_HasAttribute(
         }
         return false;
     }
+
+    // Ignore reparse points on network volumes. They can't be resolved
+    // properly, so simply remove the reparse point attribute and treat
+    // if like a regular file/directory.
+    if ((attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0) {
+        // Calling PathIsNetworkPath sometimes sets the "last error" to an
+        // error code indicating an incomplete overlapped I/O function. We
+        // want to ignore this error.
+        DWORD olderr = GetLastError();
+        if (PathIsNetworkPathW(ArchWindowsUtf8ToUtf16(path).c_str())) {
+            attribs &= ~FILE_ATTRIBUTE_REPARSE_POINT;
+        }
+        SetLastError(olderr);
+    }
+
     if (!resolveSymlinks || (attribs & FILE_ATTRIBUTE_REPARSE_POINT) == 0) {
         return attribute == 0 || (attribs & attribute) == expected;
     }
--- a/pxr/base/tf/fileUtils.cpp
+++ b/pxr/base/tf/fileUtils.cpp
@@ -108,6 +108,15 @@ Tf_HasAttribute(
         SetLastError(olderr);
     }
 
+    // Because we remove the REPARSE_POINT attribute above for reparse points
+    // on network shares, the behavior of this bit of code will be slightly
+    // different than for reparse points on non-network volumes. We will not
+    // try to follow the link and get the attributes of the destination. This
+    // will result in a link to an invalid destination directory claiming that
+    // the directory exists. It might be possible to use some other function
+    // to test for the existence of the destination directory in this case
+    // (such as FindFirstFile), but doing this doesn't seem to be relevent
+    // to how USD uses this method.
     if (!resolveSymlinks || (attribs & FILE_ATTRIBUTE_REPARSE_POINT) == 0) {
         return attribute == 0 || (attribs & attribute) == expected;
     }
--- a/pxr/base/tf/fileUtils.h
+++ b/pxr/base/tf/fileUtils.h
@@ -42,6 +42,11 @@ PXR_NAMESPACE_OPEN_SCOPE
 /// If \p resolveSymlinks is false (default), the path is checked using
 /// lstat(). if \p resolveSymlinks is true, the path is checked using stat(),
 /// which resolves all symbolic links in the path.
+///
+/// On Windows, if the path points to a reparse point symlink on a network
+/// share, even if resolveSymlinks is true we are unable to follow the
+/// symlink, and so will return true even if the destination of the symlink
+/// does not exist.
 TF_API
 bool TfPathExists(std::string const& path, bool resolveSymlinks = false);
 
@@ -50,6 +55,11 @@ bool TfPathExists(std::string const& path, bool resolveSymlinks = false);
 /// If \p resolveSymlinks is false (default), the path is checked using
 /// lstat(). if \p resolveSymlinks is true, the path is checked using stat(),
 /// which resolves all symbolic links in the path.
+///
+/// On Windows, if the path points to a reparse point symlink on a network
+/// share, even if resolveSymlinks is true we are unable to follow the
+/// symlink, and so will return true even if the destination of the symlink
+/// does not exist.
 TF_API
 bool TfIsDir(std::string const& path, bool resolveSymlinks = false);
 
@@ -58,10 +68,21 @@ bool TfIsDir(std::string const& path, bool resolveSymlinks = false);
 /// If \p resolveSymlinks is false (default), the path is checked using
 /// lstat(). if \p resolveSymlinks is true, the path is checked using stat(),
 /// which resolves all symbolic links in the path.
+///
+/// On Windows, if the path points to a reparse point symlink on a network
+/// share, even if resolveSymlinks is true we are unable to follow the
+/// symlink, and so will return true even if the destination of the symlink
+/// does not exist.
 TF_API
 bool TfIsFile(std::string const& path, bool resolveSymlinks = false);
 
 /// Returns true if the path exists and is a symbolic link.
+///
+/// On Windows, if the path points to a reparse point symlink on a network
+/// share, we are unable to follow the symlink, so we will return false for
+/// this directory even though it is a link. This is because any attempt to
+/// actually follow this link will fail, so it is safer to pretend it is not
+/// actually a link.
 TF_API
 bool TfIsLink(std::string const& path);
 


jesschimein:
Filed as internal issue #USD-9260

marktucker:
I have signed a CLA through SideFX.

pxr/base/tf/fileUtils.cpp
```diff
@@ -83,8 +83,7 @@ Tf_HasAttribute(
     if (path.back() == '/' || path.back() == '\\')
         resolveSymlinks = true;
 
-    const DWORD attribs =
-        GetFileAttributesW(ArchWindowsUtf8ToUtf16(path).c_str());
+    DWORD attribs = GetFileAttributesW(ArchWindowsUtf8ToUtf16(path).c_str());
```
pmolodo, line 86: Minor nitpick, but since we re-use this, might be nice to save the Utf16 path...

pxr/base/tf/fileUtils.cpp
```diff
@@ -94,6 +93,21 @@ Tf_HasAttribute(
         }
         return false;
     }
+
+    // Ignore reparse points on network volumes. They can't be resolved
+    // properly, so simply remove the reparse point attribute and treat
+    // if like a regular file/directory.
+    if ((attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0) {
```
pmolodo, line 100: This seems to make the assumption that a child of a reparse point still has the `FILE_ATTRIBUTE_REPARSE_POINT` attr set, but in my testing, it doesn't seem to:

https://gist.github.com/pmolodo/a48c1bbc96d7e2723b5ae2ffcdf29c1d

That is, if we have made a junction:

```
C:\path\to\some => D:\another\location
```
...and at the "real location", there is a subdirectory `D:\another\location\directory`, then

```
C:\path\to\some = has ReparsePoint attr
C:\path\to\some\directory = DOESN'T have ReparsePoint attr
```

It's possible that the behavior is different with network paths, though... or my assumption that the PowerShell attributes returned correspond exactly to the c function ones is wrong.  Or that I'm testing with "Junction" reparse points, and you were testing with "Symlink" reparse points?

pxr/base/tf/fileUtils.cpp
```diff
@@ -94,6 +93,21 @@ Tf_HasAttribute(
...
+        // Calling PathIsNetworkPath sometimes sets the "last error" to an
+        // error code indicating an incomplete overlapped I/O function. We
+        // want to ignore this error.
+        DWORD olderr = GetLastError();
+        if (PathIsNetworkPathW(ArchWindowsUtf8ToUtf16(path).c_str())) {
+            attribs &= ~FILE_ATTRIBUTE_REPARSE_POINT;
```
pmolodo, line 106: We're unconditionally altering attribs, regardless of the value of `resolveSymlinks`.

This means that if we were asking  "Hey, DON'T try to resolve symlinks, and just tell me if this path IS a symlink"... we'd now we'd be answering, "No, it isn't a symlink"... which feels wrong.

Also - in the case where we're just checking for file existence, the old code resolves symlinks recursively before providing an answer, which implies that is necessary, at least for non-network paths. For network paths, does it work differently? If not, we'd be answering that the file exists, when it may not (ie, there's a broken link, even on the local computer).  (In that case, we might need a different method to check file existence - some windows specific method for just checking existence, that handles reparse points automatically?)

pmolodo (commented):
Hi @marktucker - I can see how this would create problems, but I had a few questions about your approach - see inline comments.

pxr/base/tf/fileUtils.cpp
marktucker: What you say is correct, but I don't see what it have to do with this PR? Here is how I (perhaps incorrectly) see things:

For clarity, I'm going to change your example so that "S:\path\to\some" -> "D:\another\location" (to remind us that "S" must be a network drive for this PR to have any effect, and recall that "D" is drive "D" on the remote machine and therefore inaccessible to the machine that has mounted drive "S"). So, before this PR, `S:\path\to\some\directory` would report as "not a reparse point", and after this PR `S:\path\to\some\directory` would report as "not a reparse point". Which is what I want.

This function is not where we do the recursive searching up the path for reparse points (nor does TfReadLink march up the path). That kind of marching up the path happens somewhere else. `resolveSymlinks` set to true passed into this function never has (and continue to not) resolve symlinks of ancestors of the specifically requested directory. So to summarize, the old behavior (with resolveSymLinks=true) was:
```
S:\path\to\some -> attempt to return attribs of D:\another\location (and fail, and therefore return "file doesn't exist")
S:\path\to\some\directory -> return attribs of S:\path\to\some\directory, which reports "not a reparse point"
```
and the new behavior is:
```
S:\path\to\some -> return attribs of S:\path\to\some (with reparse point attrib stripped off)
S:\path\to\some\directory -> unchanged (return attribs of S:\path\to\some\directory, which reports "not a reparse point")
```
At least this was the intent of this PR, and I _think_ it is in fact doing this...

pxr/base/tf/fileUtils.cpp
marktucker: I think this is the correct thing to do here because other code is going to make decisions (such as altering the path string to point to the symlink location) based on the answer to the question "is this directory a symlink". The whole point of this change is that no code anywhere should treat reparse points on network drives as reparse points. So I intentionally completely obscure the fact that the reparse points on network drives even reparse points.

That's an interesting point about broken links and file existence. But here's the thing - prior to this change _any_ attempt to check the existence of a directory that was a reparse point on a network drive  would try to follow the symlink, not find it, and _always_ return false for the file existence of `S:\path\to\some`. Now, if we find a directory with a reparse point we always return true, which I think is totally valid because the requested directory does in fact exist. If the destination of the symlink doesn't exist does it have any children? No. Can you make a file inside it? No. But at least now you can try. And the whole basis for needing this PR is that just by looking at `S:\path\to\some` there is literally no way from the remote machine to determine if the destination of the reparse point exists. So even if we wanted to do what you suggest I don't think it's possible.

pxr/base/tf/fileUtils.cpp
pmolodo: Ah - my mistake. I misinterpreted what you said in the PR description, thinking that it implied that a call to `Tf_HasAttribute("//B/share/path/to/some/directory")` would now have different results - but I now see your intent. Apologies!

pxr/base/tf/fileUtils.cpp
pmolodo: tl;dr
-----
Overall, I think this PR is a good idea, and a move in the right direction, so I think it should be merged.

However, I do think we should at least add documentation to `Tf_HasAttribute`, `TfPathExists`, `TfIsDir`, `TfIsFile`, and `TfIsLink` explaining this otherwise hidden behavior - and possibly consider adding a `bool ignoreNetworkReparsePoints=true` parameter to them as well.

Also, in the case of existence checks with `resolveSymlinks=true`, I think we should use `FindFirstFile` as a way to validate if a network junction is "working".

But hopefully neither of those are particularly hard changes...?

Full Details
------------

> I think this is the correct thing to do here because other code is going to make decisions (such as altering the path string to point to the symlink location) based on the answer to the question "is this directory a symlink". The whole point of this change is that no code anywhere should treat reparse points on network drives as reparse points. So I intentionally completely obscure the fact that the reparse points on network drives even reparse points.

I see your point.  I think ignoring the "link nature" will give the behavior people probably want more often than not.

However - that assessment relies on some assumptions:
- People are mostly calling `TfIsLink` to know whether or not to call `TfReadLink`
- A result from `TfReadLink` that results in an inaccessible path will generally not be useful
- That people won't be using "true" windows symlinks (IO_REPARSE_TAG_SYMLINK) over a network share

I think all those assumptions will generally be true - but it's worth noting that there are some scenarios where each of them may not be:

- There are cases where you want to know if something is a link, and don't care what it's target is:
  - Some recursive crawls or copies may always avoid symlinks to prevent potential recursion or double-copying
  - You might record names and attributes (including whether it's a symlink) to see if they've changed - say, as part of a directory-diff tool.  You might expect such a diff generated locally and through a network share to be the same.
- There are cases where even an "inaccessible" path is useful:
  - You're working in a context with "known paths", and can do your own substitutions - ie, it's not hard to imagine that a VFX facility somewhere has written code that "knows" that `Z:\my_share` and `\\MYSERVER\zshare` are interchangeable
  - You are only interested in a portion of the resolved path - for instance, you might have a directory structure like this:
    ```
    Z:\shared\apps\FuzzyPuppyFactory
        v1.0.1\
        v1.3.2\
        v2.1.0\
        latest [junction: Z:\shared\apps\FuzzyPuppyFactory\v2.1.0]
    ```
     ...and you have code which calls `TfReadLink("\\AppServer\apps\FuzzyPuppyFactory\latest")` and grabs the last token to figure out the version
  - You want a "canonical version" of a path for comparison - ie, you don't care if you can't access `Z:\some\path\myfile.usd`, you just want to look to look it up in a database using that path as a key
- It is possible to use "true" windows symlinks over a network share - it's just not possible by default.
  - To even create the symlinks, you need to either be an Administrator, or have the "Create symbolic links" privilege granted
  - To allow symbolic links to work over shares, the local computer must have "symlinkevaluation" property set to enable remote links, ie:
    - `fsutil behavior set SymlinkEvaluation R2R:1 R2L:1`
  - While that's a pain, it's certainly possible that in a managed environment, like a VFX studio, companies might have granted these permissions, and be relying on them

Again - I think those assumptions are generally true, so I think this change does more overall help than harm.  But the chance of breaking workflows is something we should be aware of.

Also, it does makes me a bit uncomfortable to do "hidden magic" like this.  In the situations where someone IS in one of the corner cases like above, it could result in a hard-to-track-down bug.  Can we at least add a clear comment to the docstrings (of both `Tf_HasAttribute`, and the public functions that use it - `TfPathExists`, `TfIsDir`, `TfIsFile`, and `TfIsLink`) explaining this behavior?  Or, we could go one step farther, and expose a parameter to control this behavior - say, `bool ignoreNetworkReparsePoints=true` - on the above functions.  That would both make this behavior more discoverable, and provide a workaround if someone relies on old behavior.

> prior to this change any attempt to check the existence of a directory that was a reparse point on a network drive would try to follow the symlink, not find it, and always return false for the file existence of S:\path\to\some. Now, if we find a directory with a reparse point we always return true, which I think is totally valid because the requested directory does in fact exist.

I definitely agree that always returning false for existence of remote junctions is something that needs to be fixed - and is part of the reason why I think this change is overall a good thing.  However, always returning true for a network reparse point, even if `resolveSymlinks=true` and the link is actually broken, isn't ideal.  The purpose of checking existence with `resolveSymlinks=true` is to get a better idea of "is this going to work". Again, I agree that universally returning true is an improvement over universally returning false - as it's more likely that a junction is going to work than not - or at least, I would hope so!  But it would be best if we could actually get a meaningful answer to the "will it work" question.

> And the whole basis for needing this PR is that just by looking at S:\path\to\some there is literally no way from the remote machine to determine if the destination of the reparse point exists. So even if we wanted to do what you suggest I don't think it's possible.

It's true that, if you were starting just from `S:\path\to\some`, it would be impossible to know if that works. But in our case, we can just do a heuristic test - ie, see if we can list the contents of the network path, `//B/share/path/to/some`.  It's true that there are situations where the directory could exist, but you can't list it's contents - permissions issues, for one - but following the "do more good than harm" guidelines, this feels like a move in the right direction.  I did a test with `FindFirstFile` with a network share, and it seemed to work as a reasonable heuristic for whether the link is broken or not.

marktucker: Could you clarify how you are using FindFirstFile? What if the directory exists but doesn't contain anything? Are you relying on the error code from FindFirstFile to distinguish between these two cases?

I'll preface all this by saying I think you and I may have different objectives here :) It matters enormously to me that the USD library works the way a user expects it to. It does not matter to me at all that these Tf library functions may not be the right functions to use if someone wants to write a file browser or directory traversal or diff utility. So the theoretical scenarios you describe where this change might produce incorrect results don't really sway me. If you can't show me how USD will behave incorrectly you'll have trouble convincing me. So now I'll defend this change from that perspective, understanding that my arguments may be wholly unconvincing if your objective is to allow this function to be used for a wide variety of potential future purposes:

I'm very reluctant to alter this PR (can you tell?) This PR fixed a real bug in production reported by a Houdini user. It has been running in every Houdini USD library for at least 2.5 years. And although I can see the scenarios you describe where this function returns wrong results, I believe that at worst this wrongness will cause the USD library to attempt to read or write a file that it can't read or write. This always has been and always will be an eventuality the USD library must deal with no matter how much pre-checking we do. And in my experience it deals with it quite robustly. The only difference to the user is that the error reported will be "I couldn't write the file s:/path/to/some/file.usd" instead of "I didn't try to write to S:/path/to/some/file.usd because I think it is a bad symbolic link". The latter is not really more useful, and if there is any chance of reporting the latter when in fact an attempt to write the file might succeed, that is a much worse kind of wrong that will have a strong negative impact on users. It is this kind of wrongness that led to the recent removal of all Windows permission pre-checks. Now USD just tries to read or write the file, and it succeeds or it fails (and reports the error). I'm just trying to follow a similar approach here.

Documentation is always good. If you want to point to some docs that could be augmented with this additional detail about reparse points on network drives I'm happy to add something.

As for adding a new `ignoreNetworkReparsePoints=true`, I understand the impulse, but I think it is one that should be resisted. If no code in the USD library sets these parms to "false" in running production code, then it's just untested code serving no purpose. It adds a maintenance burden and provides a pitfall that future users of this function might fall into not knowing the full history of that untested parm. And yes, I know - unit tests. But this is not a test that can be faked. It can only be tested on a Windows test machine with access to a separate Windows-based server with a reparse point set to point to a missing destination. That's a lot of effort and administration to test a parm that no real code ever uses.

Thank you for all your time and deep thinking and detailed responses around this change.

pmolodo: Hey Mark - I think we're actually in closer alignment here than you might think.  After posting I had second thoughts about including the whole "potential problems" section, because ultimately, I agree with you - I think this change is a good one, and should go in.  I just wanted to point out some the potential pitfalls, to make sure we were aware of them.

But perhaps it unnecessarily muddies the waters to point those sort of things out, if I'm then just going to say, "...but I still think we should just do it anyway."  I think it made me seem opposed to this PR, which I'm not.  So I apologize for that.

As for `FindFirstFile("somedir\*")` - when I tested it, for existing/accessible directories, it always returns at least one entry, `.`, even if the directory is emtpy.  (And presumably `FindNextFile` would return `..` as well...) That seems to contradict [the docs](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilea#return-value), which say that it will return `INVALID_HANDLE_VALUE` and the `GetLastError()` will be `ERROR_FILE_NOT_FOUND (0x2)`, but in my testing I could never see that behavior.  Whenever I tried a "valid" path (local, local junction, remote, or remote junction), it always returned `.`.  And whenever I tried an invalid path (whether local or remote, plain-not-existing or broken-symlink), it always returned `INVALID_HANDLE_VALUE` and `GetLastError()` was `ERROR_PATH_NOT_FOUND (0x3)`.  (You can see the simple CLI tool I made to test it [here](https://github.com/pmolodo/cppFileFunctionTests/blob/b1ef6b4ef827256d06e0264b9f118d8b39dd65d6/findFirstFileMain.cpp).)

And as for documentation - I was just thinking of adding a description of behavior for network paths to the doxygen-parsed comments of [`TfPathExists`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.03/pxr/base/tf/fileUtils.h#L42-L44), [`TfIsDir`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.03/pxr/base/tf/fileUtils.h#L48-L52), [`TfIsFile`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.03/pxr/base/tf/fileUtils.h#L56-L60), and [`TfIsLink`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.03/pxr/base/tf/fileUtils.h#L64) in `fileUtils.h`, and to the (non doxygen parsed) comment of [`Tf_HasAttribute`](https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.03/pxr/base/tf/fileUtils.cpp#L66-L67) in `fileUtils.cpp`.  The last wouldn't make it into documentation, per-se, but still seems like it might be helpful there.

And I'm fine with not adding an extra parameter - I was 50-50 on that anyway.

pmolodo: A follow up note on the behavior of `FindFirstFile` - I was able to get it to error with `ERROR_FILE_NOT_FOUND (0x2)` if I passed it a directory that does exist, and a glob str that doesn't match anything - ie, `FindFirstFile("somedir\nomatch*")`.  So I think it just considers `.` and `..` as "valid matches" for `*`... which... sort of makes sense?

marktucker: I added some comments to the places you suggested, including a mention that in Tf_HasAttribute that we may want to use FindFirstFile at some point to test for existence of the destination, but I did not actually add that indFirstFile code for fear it may not work in all required cases (both files and directories as the targets of Tf_HasAttribute) and platforms (Windows 10, 11, Server?). I don't feel equipped to do this testing, but agree that if this functionality can be sufficiently validated it would be an improvement, perhaps as a separate PR.

pmolodo: Thanks! Sounds good to me!
✔️ 💯 👍

pxr/base/tf/fileUtils.cpp
Denise-Yang, line 100: Hi @marktucker, thanks for creating this PR! 
I was testing this change with symlinks that point to existing and non-existent locations on another network and noticed some behavior that does not follow what is outlined in TfFileUtils.h. It seems that `PathIsNetworkPathW` does not register symlink paths as network paths and that setting `attribs &= ~FILE_ATTRIBUTE_REPARSE_POINT` can sometimes cause `(attribs & attributes) == expected` to return false such that `TfIsFile` no longer returns true when checking a reparse point on a network drive.

I'm proposing the change below to check if the resolved symlink path is a network path and explicitly returning a value based on what attribute is being checked against. Does this proposed change still capture your intention and solve the bug mentioned by the Houdini user?
```
    // Ignore reparse points on network volumes. They can't be resolved
    // properly, so simply remove the reparse point attribute and treat
    // if like a regular file/directory.
    if ((attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0) {
        // Calling PathIsNetworkPath sometimes sets the "last error" to an
        // error code indicating an incomplete overlapped I/O function. We
        // want to ignore this error.
        DWORD olderr = GetLastError();
        std::string resolvedPath = TfReadLink(path.c_str());
        if (PathIsNetworkPathW(ArchWindowsUtf8ToUtf16(resolvedPath).c_str())) {
            // Return false for checking if a path is a symlink but true for all
            // other cases
            return expected != FILE_ATTRIBUTE_REPARSE_POINT;
        }
        SetLastError(olderr);
    }
```

marktucker: Hi @Denise-Yang , thanks for looking into this. But I'm not sure I follow what you're saying... I'm having trouble conceiving of a concrete case where `TfIsFile` would return `false` with this PR for a path where this would have previously returned `true` (and where the correct answer is `true`). Could you provide a concrete example of how you set things up to return this incorrect result?

I can say that the impetus behind this PR is the fact that `TfReadLink`, when called on a reparse point on a network drive, returns a value that is completely meaningless on the local machine. So any use of the `resolvedPath` in your suggested code is not going to give a meaningful result. For example, suppose "S:" is a mounted network drive and "S:\path\to\some" is a reparse point to "D:\another\location". But "D" is drive "D" on the remote machine. Your code will then call `PathIsNetworkPathW("D:\another\location")`. `PathIsNetworkPathW` will look at drive "D" on the _local_ machine, which is _not_ the right drive "D" (the local machine may not have a drive D). That's why the value returned by `PathIsNetworkPathW` is meaningless, and so I don't think your change can remain as is, even if there is a bug in the original PR.

Denise-Yang: Ah I see! I have realized I also misunderstood the intent of the PR thinking that we wanted to entirely ignore reparse points pointing _to_ network file systems (which in hindsight seems a little strange) and thus structured my test case accordingly.
I agree with your PR, and will work on testing it against a reparse point _on_ a network file system! Out of curiosity did you reproduce the bug using multiple Windows machines or were you able to find a workaround having multiple machines?

marktucker: Oh good, you had me worried there :)

I was able to reproduce the bug, but to do so I needed to have a separate "file server" machine and "test machine". If you use the same machine for the "server" and the "test machine", then the drive letters on both side of the network mount "match up". The reason this bug occurs is that the drive letters on the "server" differ from the drive letters on the "test machine".

Denise-Yang: Hi Mark!
I was finally able to reproduce the bug :) (or at least what I think is a similar example), and this PR works as documented except in the case where the file/directory on the remote machine does not actually exist.

For my specific case, I had a symlink on a network share `T:\local\path` that points to `C:\remote\pathA\` that then points to `C:\remote\pathB\` where `C:\` was on a different machine. Evaluating `T:\local\path\directory\`, where `directory` exists at `C:\remote\pathB\directory`, and `T:\local\path\` produce expected results until I deleted  `C:\remote\pathB\`, which then caused `GetFileAttributesW` to return the `error network path not found` when trying to evaluate `T:\local\path\directory\`.

Looking at the [docs for symlinks](https://learn.microsoft.com/en-us/windows/win32/fileio/symbolic-link-effects-on-file-systems-functions?redirectedfrom=MSDN), it seems that `GetFileAttributes` does not handle paths that contain symlinks. I'd like to confirm if this is a correct repro of the bug, and or if you've encountered a similar issue when making this PR (to my understanding, we should be handling this case)?

marktucker: Yep, that sounds like an accurate reproduction of the problematic situation.

Once you delete `C:\remote\pathB\`, I think it is correct for the `path not found` return for `T:\local\path\directory\`. Because indeed the requested path does not exist, right?

The known issue with this PR pointed out by @pmolodo was that for purely local symlinks, when the destination of the link gets deleted querying the source end of the symlink would return `path not found`. But with this PR, after deleting `C:\remote\pathB\`, querying `T:\local\path\` would return saying the path _does_ exist. Which is wrong according to the purely local equivalent case.

My argument for accepting this PR as-is in spite of this outstanding issue was:
1) Even if this one case is wrong, lots and lots of other cases that were wrong before will now work.
2) The harm of returning "this path exists" in this case is not relevant to the use of this function within USD. So in some other application using this method, we might care that this result is incorrect. But the functioning of the USD library will be unaffected by leaving this bug in place.
3) As far as we were able to determine, there is no sure fire way from the client end to detect if the destination of the symlink on the server exists. In other words Windows does not provide the capability to produce the truly "correct" answer in this scenario.

Denise-Yang: Ah so just to clarify it's also expected for `T:\local\path\directory` is a file and is a directory queries to return that it is neither? 

I interpreted this discussion and the functions comments, "_even if resolveSymlinks is true we are unable to follow the symlink, and so will return true even if the destination of the symlink does not exist._", to mean that it should return that the path is both a file and directory if the destination of the symlink does not exist.

marktucker: The comment in the code refers only to the case where the specified path point _directly_ to the symlink. It does not apply to a query about any file or directory _under_ the symlink. Because the function being modified here doesn't pay any attention to symlinks part way down the specified path. When you call this function with the path `T:\local\path\directory`, the code here neither knows nor cares that `T:\local\path\` is a symlink.

I think @pmolodo was similarly misled (maybe by my wording in that comment), as referred to in the thread that ends with https://github.com/PixarAnimationStudios/OpenUSD/pull/2934#discussion_r1534663283.

Denise-Yang: I see, thanks for all the clarifications Mark! Despite reading the thread, I still managed to misunderstand the comments to mean that `Tf_HasAttribute("//B/share/path/to/some/directory")` would now behave differently, but this makes a lot more sense!

Issue #2932, opened 2024-02-05
Slow change notice generation when adding many sublayers to a stage with many prims


marktucker:
### Description of Issue
On a stage with millions of prims, adding and removing a bunch of sublayers can take quite a long time. This time is spent not recomposing, but generating the change notice (which is all single threaded).

### Steps to Reproduce
1. Open usdview on root.usd included in this zip file: 
[slow_sublayers.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14169298/slow_sublayers.zip)
2. Run the commands from this text file in the interpreter window (this adds 20 sublayers, each of which just contain a single opinion): 
[slow_sublayers_commands.txt](https://github.com/PixarAnimationStudios/OpenUSD/files/14169308/slow_sublayers_commands.txt)
3. Notice that this takes 10-20 seconds of wall-clock time using a single thread. The vast majority of this is spent in `PcpChanges::_DidChangeSublayer` either generating the list of prim paths that changed, or adding those prim paths to a set (that contains all the prims after the first sublayer is processed). The actual recomposition (like the initial composition of the 1 million prims in root.usd) only takes a couple of seconds.

### System Information (OS, Hardware)
Tested on Windows and Linux.

### Package Versions
Problem seen in 23.08 and 23.11+ dev branch.

jesschimein:
Filed as internal issue #USD-9256

meshula:
I'm curious about the impact on the patch on your timings?

sunyab:
Closing out old issue since #2937 was merged in 24.05. However, I wanted to answer @meshula's question about perf impact and also highlight additional work in improving change processing times over the last year.

In the original repro case, each of the sublayers being added contained a `def` that contributed overrides for the `/root` prim. While this works, it's more typical to use `over` for these cases to indicate that those prims are intended to override (or provide) opinions on a prim rather than define the existence of that prim. I ran some timings based on the original repro case and an updated version using `over` instead:

**24.03 (without #2937)**
defs: 10 s
overs: 10 s

**24.05 (including #2937)**
defs: 1.2 s
overs: 1.2 s

**25.11**
defs: 1.1 s
overs: 0.27 s

We can see #2937 had pretty significant impact when it was introduced. We can also see that in the latest OpenUSD release, handling the `over` case has also improved greatly. This is thanks to the work to perform finer-grained change processing when handling sublayers that initially landed in 25.08.

PR #2937, opened 2024-02-06
Improve performance of generating the sublayer change notices

marktucker:
### Description of Change(s)

Improve performance of generating the change notice when adding or removing sublayers from a stage. We don't need to explicitly enumerate descendant prims in the face of a significant change because marking a prim as significantly change implies all its descendants have also changed significantly.

### Fixes Issue(s)
- #2932 


jesschimein:
Filed as internal issue #USD-9274

PR #2938, opened 2024-02-06
build_usd: make AnimX download url a fixed hash

pmolodo:
### Description of Change(s)

build_usd: make AnimX download url a fixed hash instead of a branch, so we don't suddenly have builds break if AnimX updates their repo.  (Which, admittedly, they haven't done since 2018, but it still might happen...)


jesschimein:
Filed as internal issue #USD-9275

PR #2940, opened 2024-02-07
Autodesk: Update OpenImageIO to v2.5.6.0

erikaharrison-adsk:
### Description of Change(s)

The current version v2.3.21.0 includes a bug that reports false positives when using the `-p` flag and having a very high value for the absolute tolerance. This is fixed in v2.5.6.0 in https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4061.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9283

Issue #2942, opened 2024-02-09
Crash when accessing a null PcpNodeRef object through python


marktucker:
### Description of Issue
There is no protection against null pointers in the Pcp.NodeRef implementation, making it easy to cause a crash with python code. Even worse, there is not any way to detect a null pointer in the python code, so there is no way to avoid the crash even if you are aware of the potential danger.

### Steps to Reproduce
1. Run the following code in a python shell with USD libraries available:
```
from pxr import Usd, UsdGeom
s = Usd.Stage.CreateInMemory()
p = UsdGeom.Xform.Define(s, "/foo").GetPrim()
n = p.GetAttributes()[0].GetResolveInfo().GetNode()
```
2. At this point, n refers to a null PcpNodeRef. But `bool(n)` returns `True` and `n.layerStack` will crash the process.

### System Information (OS, Hardware)
Windows, Linux

### Package Versions
23.08+, but I believe this has always been the case.

jesschimein:
Filed as internal issue #USD-9288

sunyab:
Closing out old issue, this was fixed by #2943 in the 24.05 release.

PR #2943, opened 2024-02-09
Make it safe for Pcp.NodeRef to refer to a null PcpNodeRef

marktucker:
### Description of Change(s)
Add a proper cast-to-bool operation to pxr.Pcp.NodeRef which returns false if the referenced PcpNodeRef cast to bool returns false. Override the implementation of __getattribute__ so that trying to call any method on a null PcpNodeRef raises a python exception (instead of crashing).

### Fixes Issue(s)
- #2942 


jesschimein:
Filed as internal issue #USD-9289

Issue #1023, opened 2019-11-05
Use dccFps in UsdAbc for Alembic 1.7.12 and above


spiffmon:
The latest version of Alembic includes official support for encoding framesPerSecond into an archive, addressing a longstanding friction point converting between USD and Alembic files.  Alembic does not require distinct timeCodesPerSecond and framesPerSecond because it stipulates all ordinates are in seconds.  Therefore,  Alembic's framesPerSecond should map to USD's timeCodesPerSecond, as we continue to also rescale the actual timesamples we read from an abc archive.

We should also then deprecate the use of the custom "timeCodePerSecond" metadata we author currently.

We are not scheduling this work yet, but wanted to file it for the community.

jilliene:
Filed as internal issue #USD-5677

PR #2944, opened 2024-02-09
Use "DCCFPS" metadata to set TCPS value in alembic reader code

marktucker:
### Description of Change(s)
If an alembic file has authored the "DCCFPS" metadata in its config info, use that value as the TimeCodesPerSecond value of the equivalent USD data. This prevents unintended time scaling of Alembic files when translating them from Alembic (which stores samples times in seconds) into USD (which stores sample times in samples). This change does not remove support for the old "Usd" metadata section, which takes precedence to avoid unnecessary behavior changes, and change only takes effect if building against Alembic 1.7.12 or later (when DCCFPS was introduced).

### Fixes Issue(s)
- #1023 

### DIFF
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -952,7 +952,12 @@ _ReaderContext::Open(const std::string& filePath, std::string* errorLog,
     // Get info.
     uint32_t apiVersion;
     std::string writer, version, date, comment;
+    double fps = 0.0;
+#if ALEMBIC_LIBRARY_VERSION >= 10712
+    GetArchiveInfo(archive, writer, version, apiVersion, date, comment, fps);
+#else
     GetArchiveInfo(archive, writer, version, apiVersion, date, comment);
+#endif
 
     // Report.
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9290

diyajoy:
Hi Mark! Could you add a small abc file that has DCCFPS set to 30 or anything other than 24? Then we can add a small test that opens the abc file and verifies that its timeCodesPerSecond matches.

marktucker:
Of course... This file has the DCC FPS value set to 30 (generated by an as-yet-unreleased version of Houdini - shipping builds of Houdini don't author this metadata):
[output_30.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14379507/output_30.zip)

diyajoy:
Awesome, thank you!

diyajoy:
Hi Mark! We want to check your test file in to source control on our end, but the file provided is larger than our other test files. Would it be possible for you to provide a file with less data or only metadata that we could use in its place?

marktucker:
Sorry this took so long... 
[dccfps_30.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14605012/dccfps_30.zip)

diyajoy:
That's perfect, thank you!

PR #2945, opened 2024-02-09
Autodesk: Custom pick region to search over for Hit Prims

erikaharrison-adsk:
### Description of Change(s)

We now add a new GfVec4i to specify a search area in the pickTaskContextParams. This is to narrow the search over the desired pick location. The search is now supported in the pickTask when we pick item across all modes - closest to camera, closest to center, etc.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdx/pickTask.cpp
+++ b/pxr/imaging/hdx/pickTask.cpp
@@ -603,6 +603,40 @@ HdxPickTask::Execute(HdTaskContext* ctx)
 
     GfVec2i dimensions = _contextParams.resolution;
     GfVec4i viewport(0, 0, dimensions[0], dimensions[1]);
+    
+    auto useSearchRect = [&](const GfVec4i& sRect, const GfVec4i& vRect )->bool {
+        
+        // invalid search size or not initialized
+        if(sRect[2] < 0 || sRect[3] < 0)
+            return false;
+
+        // sRect x cordinate extremes  contained in vRect
+        if(sRect[0] < vRect[0] || sRect[0] > vRect[0] + vRect[2])
+            return false;
+
+        if((sRect[0] + sRect[2]) < vRect[0] || (sRect[0] + sRect[2]) > (vRect[0] + vRect[2]))
+            return false;
+
+        // sRect y coordinate extremes  contained in vRect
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9291

tcauchois:
Hi Erika,

I'm not sure what the intent is here.  When we run HdxPickTask, we typically call GfFrustum::ComputeNarrowedFrustum (https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/base/gf/frustum.h#L531, e.g. https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usdImaging/usdviewq/stageView.py#L2237) to compute a camera frustum based on the viewport camera, with the same near/far planes, but where the top/bottom/left/right have been moved in to enclose an area of NxN pixels around the "pick point" (which is where the mouse clicked on the near plane).  In the stage view example, we're picking a region of size 1x1 pixels on the near plane (though the actual pick buffer is 128x128).  The different resolve algorithms will then sort through the prims in the pick buffer based on distance-to-center (i.e. mouse); distance-to-near-plane; etc.  This way of specifying the search target also means we can use the same algorithm for box-select and other picking methodologies just by changing which frustum we pass in.

This seems like it does everything that your subrect would need to do, making this PR redundant.  If this isn't the case, could you provide an example of where the current API doesn't work?

Thanks,
Tom

(edited to correct some technical details, should have finished my coffee before posting...)

rsanthanam:
Hi. Thanks for your response. The scenario we had was requiring the need to render over the entire viewport (for other reasons) without doing the narrowed frustum compute around the pick point (as you mention). Subsequently we wanted to narrow the search around the pick location. Hope that helps.

tcauchois:
Since the pick task runs its own renderpass which is only used for the pick results, what's stopping you from narrowing the frustum?  If you want to re-use a full-screen buffer generated earlier, you should be looking at HdxPickFromRenderBufferTask.

rsanthanam:
To be honest, the reason for rendering over the entire view port is we were having picking issues with a custom prim rendered on the GPU with a fixed pixel size on screen. To circumvent the pick issue, we rendered over the entire viewport. Notice you mention HdxPickFromRenderBufferTask and say generated earlier. Wasn't sure what exactly  you meant. Thanks.

PR #2946, opened 2024-02-10
Fix issues that prevent building for windows via MINGW

expenses:
### Description of Change(s)

Being able to build via MINGW allows cross-compiling OpenUSD to windows from linux and has a bunch of advantages, most notably for me is that I can integrate it into a nix build system: https://github.com/expenses/openusd-minimal-nix/tree/57ee5b26c2e51057603e7340e3108cb9c4c0234a.

Almost all these changes are converting the names of includes and libraries to lowercase, as `windows.h` is case-sensitive on linux while it's case insensitive on Windows. The other changes resolve around cases where it is assumed that building for windows always is done via MSVC and never by GCC. There's also a change from `GetProcAddress` to `(void*)GetProcAddress` and an added `share.h` header.

I'm using the [msys2](https://www.msys2.org/) builds of tbb (2020) and opensubdiv for this, see https://github.com/expenses/openusd-minimal-nix/tree/57ee5b26c2e51057603e7340e3108cb9c4c0234a/msys2-packages.

I'm not sure that you'd want to merge this as-is given that I wrote it against v23.11 and not master. I've tested the build but haven't tested compiling on windows itself yet.


jesschimein:
Filed as internal issue #USD-9292

pmolodo:
I'm a bit leery of universally changing the include file from `Windows.h` to `windows.h`.  I know that the windows API is generally case insensitive, so this change SHOULD generally work... but NTFS itself is case sensitive, and the actual file installed by the Windows SDK is `Windows.h`, so it feels wrong to deliberately reference the incorrect name for VS users.  It seems like Microsft has has been making moves to make it more possible to operate in a case sensitive fashion (ie, [like this](https://www.ortussolutions.com/blog/how-to-make-windows-folders-case-sensitive)), so this feels a move in the wrong direction.

If mingw deliberately named it's windows header `windows.h` instead of `Windows.h`, I'd prefer we have a flag for mingw support and then either have some sort of macro (ie, `#define WINDOWS_HEADER <windows.h>`,  `#include WINDOWS_HEADER`), or in cmake automatically generate a `Windows.h` which contains nothing but `#include <windows.h>`.

However - I'm more familiar with POSIX dev than Windows, so if anyone more familiar with windows is willing to vouch that doing `include <windows.h>` is pretty common, than that would quiet my fears...

meshula:
I don't wanna scope creep this effort, but we could make an `ArchPlatformWindows.h` to be used where Windows.h is currently included. It could centralize the minGW test, and also, we could add in the LEAN_AND_MEAN macros which would probably give us a build speed improvement.

I think these are all appropriate for USD:

```
// these ones improve compilation speed by stripped out things we don't need
#define WIN32_LEAN_AND_MEAN
#define VC_EXTRALEAN
#define NOGDI

// these ones suppress warnings related to the C string functions, mostly
#define _CRT_SECURE_NO_WARNINGS
#define _CRT_NONSTDC_NO_DEPRECATE

// for case sensitive file systems
#ifdef _MINGW
#include "windows.h"
#else
#include "Windows.h"
#endif
```

expenses:
> However - I'm more familiar with POSIX dev than Windows, so if anyone more familiar with windows is willing to vouch that doing `include <windows.h>` is pretty common, than that would quiet my fears...

I'm also not an expert, but it seems common enough, at least for projects that are small enough to not warrant a 'proper' solution. See https://stackoverflow.com/a/15466951.

> I don't wanna scope creep this effort, but we could make an `ArchPlatformWindows.h` to be used where Windows.h is currently included. It could centralize the minGW test, and also, we could add in the LEAN_AND_MEAN macros which would probably give us a build speed improvement.

This seems reasonable, I'll look into this when I have time.

PR #2455, opened 2023-06-01
Apple: iOS long term support

slingthor:
This PR enables iOS building of Hydra Storm and will be supported long term until an official release.
Allows building with the build script against core dependencies, and most external dependencies.
While Python can run on iOS, PySide6 is currently not supported and thus USDView is not.

The intention for this PR is to keep it up against latest release, so anyone who needs iOS support can bring it in easily and maintain. Not intended for a merge into official repository as a whole - there are smaller chunked PRs out for that.

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #2949, opened 2024-02-14
Apple: Core iOS support

dgovil:
### Description of Change(s)

This PR adds **Core** iOS support to the OpenUSD project. This does not include Imaging, and any Imaging related components at this time. Imaging will be added in a follow up PR. As requested, MaterialX support was also removed and will be uploaded in a separate PR.

While the build_usd.py script only supports iPhone, this change also does make it possible for the main CMake build to target other Apple platforms as long as the developer can build the dependencies themselves.

It is a minimal version of Thor's work in https://github.com/PixarAnimationStudios/OpenUSD/pull/2455 against the latest `dev` branch.
Changes include:
* Using latest dev branch
* No imaging support. Will be added in a follow up PR.
* Makes use of CMake's inbuilt iOS support, negating the need for toolchain support
* Structures the code in such a way that we can add support for other iOS derived platforms+simulator in future PRs
* Swaps `ARCH_OS_IOS` with `ARCH_OS_IPHONE` to align with compiler directives. IPHONE refers to all derivative platforms, whereas iOS refers to only iPhone/iPad (Confusing but the case for historical reasons as [documented here](https://chaosinmotion.com/2021/08/02/things-to-remember-compiler-conditionals-for-macos-ios-etc/) or in `TargetConditionals.h` within any Apple SDK)
* Upgrades MaterialX to 1.38.8 (from 1.38.7) as that adds support for iOS derivative platforms as well.
* I've moved `import apple_utils` out from under a platform check, because the module import shouldn't be platform specific, and it removes the need for unnecessary platform checks all over the place
* TBB requires SDKROOT be passed in to workaround a bug where it may find a random other toolchain and fail. (Valentin Roussellet figured out this edge case)
* Added `APPLE_EMBEDDED` cmake parameter in Options.cmake so that it can be used to configure options further down the build setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ then build and install USD into `/path/to/my_usd_install_dir`.
 > python OpenUSD/build_scripts/build_usd.py /path/to/my_usd_install_dir
 ```
 
-##### MacOS:
+##### macOS:
 
 In a terminal, run `xcode-select` to ensure command line developer tools are
 installed. Then run the script.
@@ -140,6 +140,16 @@ then build and install USD into `/path/to/my_usd_install_dir`.
 ```
 > python OpenUSD/build_scripts/build_usd.py /path/to/my_usd_install_dir
 ```
+
+###### iOS
+
+When building from a macOS system, you can cross compile for iOS based platforms.
+
+iOS builds currently do not support Imaging.
+Additionally, they will not support Python bindings or command line tools.
+
+To build for iOS, add the `--build-target iOS` parameter.
+
 
 ##### Windows:
 
--- a/build_scripts/apple_utils.py
+++ b/build_scripts/apple_utils.py
@@ -274,4 +274,11 @@ def ConfigureCMakeExtraArgs(context, args:List[str]) -> List[str]:
         args.append(f"-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH")
         args.append(f"-DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH")
         args.append(f"-DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=BOTH")
+
+
+    # Signing needs to happen on all systems
+    if context.macOSCodesign:
+        args.append(f"-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY={GetCodeSignID()}")
+        args.append(f"-DCMAKE_XCODE_ATTRIBUTE_DEVELOPMENT_TEAM={GetDevelopmentTeamID()}")
+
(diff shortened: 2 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9301

dgovil:
@jesschimein or @meshula for whenever y'all are back, I just wanted to make a note so I can check in on the build_usd.py and apple_utils.py changes in this PR that didn't make it into 24.5.

Should I split it out into a separate PR? I think I'd missed that 24.5 was targeting just the CMake + C++ side of things. I'm just not sure how y'all want to handle the remainder of this PR.

davidgyu:
@dgovil There are some additional notes from internal code review that we want to address before merging the build script changes, but I think we can keep those associated with this PR for now.

dgovil:
Sounds good. Thanks!

meshula:
Reopened, as the noted commit only partially addressed the PR.

dgovil:
@davidgyu and @meshula This PR has been rebased against the latest in `dev`.
It only contains the build script changes which mostly amount to:
* Adding the new build target flag
* Setting the build target and sdkroot for TBB. The SDKRoot is required because TBB has a tendency to try and discover the wrong SDK.

dgovil:
I'm just fixing up a few other oddities I'm finding. 

1.  Currently `tbb` seems to want to only respect the iOS build target for Debug and not release. Looking into that now.
2. I found an intermittent bug in cmake where setting the `CMAKE_SYSTEM_NAME` causes the cmake find logic to change. I found this [thread](https://discourse.cmake.org/t/find-package-stops-working-when-cmake-system-name-ios/4609/6) with a fix and it involves setting the `CMAKE_FIND_ROOT_PATH_MODE` variables to `BOTH` to work. I've updated the PR with that.

I'll report back when I fix the TBB thing. But right now it works if you set the build mode to debug.

dgovil:
Okay, should be good now. TBB needs an extra arg called `extra_inc=big_iron.inc` that I had missed in the rebase.
Essentially it makes TBB become statically compiled, which is better for iOS anyway.

With the two changes above, USD for iOS should build as easily as 

`build_usd.py ~/usd --build-target iOS`

dgovil:
@davidgyu I just added https://github.com/PixarAnimationStudios/OpenUSD/pull/2949/commits/3cc8aa4b88b551a70c3a464920d90731a8ebf7eewhich adds the CMake code signing attributes to the CMake arguments so that even dependencies get code signed during their build if code signing is enabled.

In theory it isn't strictly related to this PR (as it should have been happening for macOS as well) but this brings it inline here

dgovil:
Rebased on top of latest from dev

PR #2952, opened 2024-02-15
Apple: Upgrade MaterialX to 1.38.8

dgovil:
### Description of Change(s)

This PR upgrades MaterialX to 1.38.8. [CHANGELOG](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/CHANGELOG.md#1388---2023-09-08)

The big upgrade for us is the support for iOS builds and DisplayP3 support, but there are a ton of great changes in that ChangeLog

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1484,7 +1484,7 @@ DRACO = Dependency("Draco", InstallDraco, "include/draco/compression/decode.h")
 ############################################################
 # MaterialX
 
-MATERIALX_URL = "https://github.com/materialx/MaterialX/archive/v1.38.7.zip"
+MATERIALX_URL = "https://github.com/materialx/MaterialX/archive/v1.38.8.zip"
 
 def InstallMaterialX(context, force, buildArgs):
     with CurrentWorkingDirectory(DownloadURL(MATERIALX_URL, context, force)):


jesschimein:
Filed as internal issue #USD-9315

dgovil:
@spiffmon I realize that MaterialX 1.38.9 just came out https://github.com/AcademySoftwareFoundation/MaterialX/releases/tag/v1.38.9

The biggest changes that might be beneficial over 1.38.8 is some correctness fixes to the nodes. I didn't want to assume that was fine for y'all, but thought I'd mention it here in case you want to skip forward.

spiffmon:
Closing, as we _are_ going to go directly to 1.38.9, and handling the upgrade internally.  Thanks, @dgovil !

dgovil:
@spiffmon  FYI, you'll probably need this patch for macOS unless MaterialX releases a 1.38.10 with it

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1741

spiffmon:
@dgovil , since we don't do XCode builds in our internal builds, we didn't think we'd be affected?

dgovil:
Hmm I don't think it's specific to Xcode. I'm building with Ninja and get these build errors. 

It fails at the cmake step because it's looking up the MaterialX cmake exported config , but it has an invalid linker string exported out.

spiffmon:
OK, I reckon we'll sort it then - thanks!

PR #2953, opened 2024-02-16
Apple: Remove unused CREATE_FRAMEWORK parameters

dgovil:
### Description of Change(s)

We discovered this unused CREATE_FRAMEWORK argument in several CMake files. I'd like to remove these pre-emptively for possibly adding framework build support to OpenUSD in the future to avoid confusion.


jesschimein:
Filed as internal issue #USD-9316

PR #2955, opened 2024-02-16
Fix DeprecationWarning caused by a regex containing an invalid escape sequence

marktucker:
### Description of Change(s)
Fix DeprecationWarning caused by a regex containing an invalid escape sequence.

Prefix the regex with "r" so it is interpreted as a raw string.


jesschimein:
Filed as internal issue #USD-9318

PR #2956, opened 2024-02-16
Fix a bug where the skinning computation could produce an incorrect time sample

cameronwhite:
### Description of Change(s)
- In `SampleExtComputationInput()`, inputs to the skinning computation which aren't time-varying (e.g. the bind transform) reported a single sample at a time offset of 0.

  The choice of 0 (the current time) here could produce unexpected behaviour. If none of the time-sampled computation inputs (e.g. joint transforms) actually had a sample at that time, this would cause `HdExtComputationUtils::SampleComputedPrimvar()` to evaluate and return an additional time sample for the computation.

  For example, if the sampling interval was (-0.25, 0.25) and the joint transforms were also time-sampled at [-0.25, 0.25], `SampleComputedPrimvar()` would return three samples at [-0.25, 0, 0.25]
This could be confusing for users when configuring render options such as the maximum number of samples to use for motion blur, since this extra time sample doesn't actually exist anywhere on the stage.

  An easy solution is to instead report these constant inputs as having a single time sample at the beginning of the sampling interval. Sampling always includes the boundary times so this eliminates the possibility of adding an extra sample.

- This issue also revealed a related bug where this extra sample had an unexpected pose, and was identical to the previous time sample rather than interpolating the joint transforms. Using the example above, when `SampleComputedPrimvar()` evaluates the computation at `sampleTime=0`, inputs such as the skinning transforms would first be interpolated since they only have time samples at [-0.25, 0.25]. The skinning transforms actually ended up being held at -0.25 rather than interpolated, since `float` variants of `GfMatrix` types were not registered with `hdTimeSampleArray` for interpolation. 

### DIFF
--- a/pxr/imaging/hd/timeSampleArray.cpp
+++ b/pxr/imaging/hd/timeSampleArray.cpp
@@ -26,8 +26,11 @@
 
 #include "pxr/base/gf/half.h"
 #include "pxr/base/gf/matrix2d.h"
+#include "pxr/base/gf/matrix2f.h"
 #include "pxr/base/gf/matrix3d.h"
+#include "pxr/base/gf/matrix3f.h"
 #include "pxr/base/gf/matrix4d.h"
+#include "pxr/base/gf/matrix4f.h"
 #include "pxr/base/gf/quatd.h"
 #include "pxr/base/gf/quatf.h"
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9319

PR #2957, opened 2024-02-16
Apple: Introduce APPLE_EMBEDDED parameter

dgovil:
### Description of Change(s)

This PR adds a new `APPLE_EMBEDDED` variable in CMake to  allow for switching configuration logic accordingly.
This will be required for potential follow up PRs that introduce imaging and framework builds.

Requires https://github.com/PixarAnimationStudios/OpenUSD/pull/2949 to be merged in first. I could split this out, but it's not very functional for other PRs until that one is merged first. So I opted to base it on that to have a nice tree.

I've added checks for all the known cross compilation targets that CMake supports for Apple platforms , even if there aren't plans to support all of them in OpenUSD in the immediate future. This enables less churn in future PRs.

Changes are confined to the last commit in the tree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ then build and install USD into `/path/to/my_usd_install_dir`.
 > python USD/build_scripts/build_usd.py /path/to/my_usd_install_dir
 ```
 
-##### MacOS:
+##### macOS:
 
 In a terminal, run `xcode-select` to ensure command line developer tools are
 installed. Then run the script.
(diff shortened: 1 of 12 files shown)


jesschimein:
Filed as internal issue #USD-9320

dgovil:
After some thought, I think it's best to move this into the core iOS branch because I need to disable some options accordingly in the CMake for it. I'll close this one out and move the changes over.

PR #2969, opened 2024-02-26
[OLD] Apple: Enable building as an Apple Framework

dgovil:
Edit: As requested by @meshula , I'll be splitting up this PR into the smaller constituent parts so we can land them individually.

I will list them here so it's easy to follow along with progress. I am listing the proposed bits here, and will link them to actual PRs as they go up. These are listed in no particular order. They will be checked off as they are merged.

They are nested based on their dependencies

- [ ] [TempDirectory fixes](https://github.com/PixarAnimationStudios/OpenUSD/pull/3705) : Makes sure that USD operations write to the containerized temp directory locations
- [ ] [Extract TBB Patches](https://github.com/PixarAnimationStudios/OpenUSD/pull/3706) : TBB requires patches for each embedded platform. This may not be required after porting to OneTBB (Not sure), but we should move them into Apple Utils to avoid cluttering up the main build code
    - [ ] [Simulator Support](https://github.com/PixarAnimationStudios/OpenUSD/pull/3707) : Support for Simulator targets for people testing without devices. 
- [ ] [Code Signing Improvements](https://github.com/PixarAnimationStudios/OpenUSD/pull/3710) : Improve handling of automatic code signing, especially to handle frameworks
- [ ] [Build as Framework](https://github.com/PixarAnimationStudios/OpenUSD/pull/3713): Enable building targets as frameworks
    - [ ] [Build as XCFramework ](https://github.com/PixarAnimationStudios/OpenUSD/pull/3715): Enable building for a combined framework  across platforms
    - [ ] [Modify Headers post build](https://github.com/PixarAnimationStudios/OpenUSD/pull/3714): simplifies inclusion of the framework into Xcode projects, or simplifies linking to just `-framework OpenUSD`
- [ ] [CCache Support](https://github.com/PixarAnimationStudios/OpenUSD/pull/3142) : Enables faster builds which is important when building for multiple targets (this is optional but its a huge speedup)

------------------------------

### Description of Change(s)

This PR enables building of USD as a Framework.
A framework build puts the output files into a directory structure which allows them to be included easily in apps on macOS, iOS and visionOS. The steps to use it are included in the README but are repeated here in brief:

1. User drags the Framework directory into their project target. This automatically sets up linkage. 
2. User sets it to Embed and Sign in the app

Frameworks are enabled automatically for Apple embedded platforms (iOS, visionOS).
Framework builds default to Monolithic because a framework may only include one dylib within it. Shared builds can still be enabled for some dev purposes, but aren't officially supported.

**Experimental Feature**

I request that the feature be marked as Experimental in the README, because I anticipate changes might be needed in the future to streamline how this works, such as not requiring the dependencies to be statically built. I think the current state of the PR will unblock the majority of people asking for this feature however and the changes I anticipate might be quite long term.

This incorporates some work by Maddy Adams and @meshula . The changes are largely non-invasive and are listed below:

**Core Code Changes**
1. Adds a post build step with a zsh script that will construct the Framework bundle.
2. Changes to the Plugin and usdMtlx registry to find their respective resources within the Framework
3. Enable building statically against TBB and OpenSubdiv.
4. New options in the CMake to handle framework build configuration

**Build Script Changes**
1. Added argument to build as Framework (or disable) with `--build-apple-framework` and `--no-build-apple-framework`
2. Added `--prefix-framework-headers` to opt-in to prefixing the headers
3. Added support to `apple_utils.py` to orchestrate an `XCFramework` build, that combines multiple targets into one framework.
4. Added support for Simulator targets
5. Allow shared/monolithic builds arg to allow a default based on other args.
6. Build Dependencies (MaterialX, TBB) as static when building monolithic
7. Allow users to pass in code signing ID manually
8. Improved Codesigning identity lookup speeds+reliability for both framework and non-framework builds
9. Setup signing for dependencies automatically when using Xcode target

**Header Notes**
Framework headers are automatically discovered when prefixed by the framework name.
This requires the user to build with `--prefix-framework-headers`
E.g instead of `#include <pxr/pxr.h>` you would do `#include <OpenUSD/pxr/pxr.h>`

If someone wants to avoid the prefixed behavior, they have to manually add the Header directory to their targets `SYSTEM_HEADER_SEARCH_PATHS` . 


jesschimein:
Filed as internal issue #USD-9360

meshula:
@dgovil  Could I ask you to pop in a note that reports when you are building a framework here?

```
Building with settings:
  USD source directory          /Users/nick/dev/usd
  USD install directory         /Users/nick/usd-dyfw-240616
  3rd-party source directory    /Users/nick/usd-dyfw-240616/src
...
```

meshula:
Apologies @dgovil I was meaning to merge locally and got the <- -> icons backwards in the interface. Please force push to overwrite my inadvertent commit when you are ready to update.

dgovil:
@meshula I rebased the latest on top of dev+ios_core PR. I overwrote your changes to my branch.
I added the print out in the build for the `Building with settings:` section. I already had a description of how to pull in the framework. It also includes the MaterialX copy over as well.

meshula:
Thanks for fixing the mishap and adding the Mx bits!

dgovil:
Currently refactoring this on top of the iOS imaging branch and the latest dev, so in flux for the moment.

dgovil:
Rebased this PR on the imaging branch. 
I've also worked on making this a fully monolithic build so that it is a kosher framework. This means that you get a single (massive) framework that includes OpenUSD/OpenSubdiv/TBB/MaterialX. That may not be to everyones taste, but the alternative would be a framework per dylib which is a fairly sizable uplift to put in USD today (but something we could drive towards longterm)

It is still possible to build as a non-monolithic build, but I am not focusing on that setup because frameworks expect a single dylib per framework and we'd have something like ~70 frameworks going that route. Instead, long term, we can work with each dependency to add monolithic+framework builds to them as well, and then tease things apart here.

It also allows for building for simulator (but isn't super well tested)

My todos before I think this is ready is:

[] Make sure the USD build can load all the plugins that are part of it. Specifically HD etc...
[] Make sure that MaterialX can load its shaders from the resources directory 

Potentially, I want to see if its possible to modify the headers to avoid needing to add the header search path manually.

meshula:
Looking promising!

> Potentially, I want to see if its possible to modify the headers to avoid needing to add the header search path manually.

If you can solve that one, that would be fantastic!


Could I ask you to drop a link in to point us at 

> Rebased this PR on the imaging branch.

Given the evolution of the ios + imaging work, it would be helpful to have a link to the branch out here in the main thread.

dgovil:
Yep, added a link to the main body above but also adding it here https://github.com/PixarAnimationStudios/OpenUSD/pull/3215

meshula:
Thanks!!

dgovil:
With the latest changes to the PR, the USD plugins under `lib/usd` load, as do the MaterialX shader defs when called with 

```
SdrRegistry &shaderReg = SdrRegistry::GetInstance();
auto nodes = shaderReg.GetNodeNames();
```

I need to now check for loading of the plugins that come under `plugin/usd` like `hdStorm, hioAvif, sdrGlsfx, usdShaders`.

dgovil:
I think it's in a good place if folks want to try it out. 

I've confirmed that USD is able to register all the plugins like AVIF and Storm. 

Both this and https://github.com/PixarAnimationStudios/OpenUSD/pull/3215 are rebased on top of latest dev to remove conflicts.

I've also improved the README to better reflect the setup instructions for frameworks. 

After talking to Nick, we decided to punt the modification of the Header includes. This means users will currently need to add a single path manually to their System Header Search Paths. In the future, we may revisit this once some other future changes for dependencies come to fruition.

Feedback is welcome.

dgovil:
After talking with Maddy, I actually managed to put in a post process in the zsh file that will change the headers. So now users just need to drop the framework into their project and they're off to the races.

dgovil:
Rebased on top of latest dev now that https://github.com/PixarAnimationStudios/OpenUSD/pull/3215 is merged. This reduces the diff to just what's needed for framework builds

dgovil:
@meshula I added some code to the `apple_utils.py` to do a build of all the targets as an [XCFramework](https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle) .

I put this in the `apple_utils.py` since there wasn't a good way to add orchestration like this in the `build_usd.py`. Hopefully that's okay, as I couldn't think of a cleaner solution.

With an XCFramework, its easier to share code between multiple platform targets, and you don't need to setup separate simulator and device targets.

build_scripts/build_usd.py
```diff
@@ -1077,30 +1077,21 @@ def InstallTBB_MacOS(context, force, buildArgs):
                   "ifeq ($(arch),$(filter $(arch),armv7 armv7s {0}))"
                         .format(apple_utils.GetTargetArmArch()))])
 
-        if context.buildTarget == apple_utils.TARGET_VISIONOS:
-            # Create visionOS config from iOS config
+        if MacOSTargetEmbedded(context) and context.buildTarget != apple_utils.TARGET_IOS:
```
meshula, line 1080: When GCD is available as a work backend let's get rid of this patching on vision os

build_scripts/build_usd.py
```diff
@@ -550,13 +550,13 @@ def DownloadURL(url, context, force, extractDir = None,
     will raise a RuntimeError if the SHA256 checksum computed from
     the file does not match.
 
-    Returns the absolute path to the directory where files have 
+    Returns the absolute path to the directory where files have
```
meshula, line 553: let's get ws fixes out of this pr

build_scripts/build_usd.py
```diff
@@ -1595,9 +1589,7 @@ def InstallDraco(context, force, buildArgs):
 
 def InstallMaterialX(context, force, buildArgs):
     with CurrentWorkingDirectory(DownloadURL(MATERIALX_URL, context, force)):
-        cmakeOptions = ['-DMATERIALX_BUILD_SHARED_LIBS=ON',
-                        '-DMATERIALX_BUILD_TESTS=OFF'
-        ]
+        cmakeOptions = ['-DMATERIALX_BUILD_TESTS=OFF']
```
meshula, line 1592: rebase on 25.05 some of this will disappear

build_scripts/build_usd.py
```diff
@@ -2029,6 +2036,7 @@ def InstallUSD(context, force, buildArgs):
                        default=codesignDefault, action="store_true",
                        help=("Enable code signing for macOS builds "
                              "(defaults to enabled on Apple Silicon)"))
+    group.add_argument("--codesign-id", dest="macos_codesign_id", type=str)
```
meshula, line 2039: can we clean up code signing separately?

build_scripts/build_usd.py
```diff
@@ -2788,7 +2804,8 @@ def FormatBuildArguments(buildArgs):
     buildMaterialX=("On" if context.buildMaterialX else "Off"),
     buildMayapyTests=("On" if context.buildMayapyTests else "Off"),
     buildAnimXTests=("On" if context.buildAnimXTests else "Off"),
-    enableHDF5=("On" if context.enableHDF5 else "Off"))
+    enableHDF5=("On" if context.enableHDF5 else "Off"),
```
meshula, line 2807: we're dropping hdf5 support anyway

cmake/defaults/CXXDefaults.cmake
```diff
@@ -92,3 +92,19 @@ if (PXR_PREFER_SAFETY_OVER_SPEED)
...
    set(PXR_PREFER_SAFETY_OVER_SPEED "0")
 endif()
+
+# Set that Apple Framework is being build
+if (PXR_BUILD_APPLE_FRAMEWORK)
+    _add_define("PXR_BUILD_APPLE_FRAMEWORK")
+endif()
+
+# Setup CCache for C/C++ compilation
+if(PXR_ENABLE_COMPILER_CACHE)
```
meshula, line 102: different PR

pxr/base/arch/defines.h
```diff
@@ -24,6 +24,9 @@
 #else
 #define ARCH_OS_OSX
 #endif
+#if TARGET_OS_SIMULATOR
```
meshula, line 27: separate PR?

pxr/base/plug/initConfig.cpp
```diff
@@ -93,6 +93,14 @@ ARCH_CONSTRUCTOR(Plug_InitConfig, 2, void)
     _AppendPathList(&result, buildLocation, binaryPath);
     _AppendPathList(&result, pluginBuildLocation, binaryPath);
 
+#ifdef PXR_BUILD_APPLE_FRAMEWORK
```
meshula, line 96: can we configure this in cmake to eliminate the ifdef here

pxr/imaging/hgiMetal/graphicsCmds.mm
```diff
@@ -112,15 +112,10 @@
             hasClear = true;
         }
         
-        if (@available(macos 100.100, ios 8.0, *)) {
-            metalColorAttachment.loadAction = MTLLoadActionLoad;
-        }
-        else {
-            metalColorAttachment.loadAction =
+        metalColorAttachment.loadAction =
```
meshula, line 115: separate

build_scripts/build_usd.py
dgovil: The GCD work will still require TBB for the data constructs unfortunately

PR #2970, opened 2024-02-26
Autodesk: Revise test case comparison parameters to avoid false positive

erikaharrison-adsk:
### Description of Change(s)

Problems:
As comparing test case output with baseline, the OIIO tool idiff consume a few options, e.g. -fail, -failpercent, -p (perceptual) and so on. From very beginning, the options value inputs are not reasonable to catch all of pixel changes. The false positive is found when using an alternative hgi backend. That is, a number of cases were passed, but rendering images are not correct. The results are not real positive.

Solutions:
Change test input parameters to be stricter in CMake file to successfully detect errors in the generated images. Another side, remove some perceptual options they are not proper to test for specific cases.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9361

PR #2987, opened 2024-03-05
Ensure dependencies are traversed in `scriptModuleLoader`

nvmkuruc:
### Description of Change(s)
`_HasTransitiveSuccessor` iterates over a predecessor's successors with an unused variable `j` and then attempts to tag the predecessor to be visited when it should be tagging the successors.

This change replaces the `TF_FOR_ALL` with a range based for loop and changes the logic so that successors are now added to the predecessor stack.

This also ensures that the initial state of the visit tracking `TfHashSet` matches the initial state of the stack, avoiding unnecessary re-visitation of the initial predecessor.

### Fixes Issue(s)
- #80


jesschimein:
Filed as internal issue #USD-9388

PR #2988, opened 2024-03-05
Remove ZLib as mandatory dependency

dgovil:
### Description of Change(s)

build_usd.py has zlib as a mandatory dependency for USD. However it seems that nothing in the core of USD seems to use it.

Instead, I've moved this to only specific dependencies that say they need it in their READMEs

This further reduces the library dependency count for USD core to just TBB (and Boost if you want Python bindings)

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2361,7 +2361,10 @@ if extraPythonPaths:
 if context.buildOneTBB:
     TBB = ONETBB
 
-requiredDependencies = [ZLIB, TBB]
+requiredDependencies = [TBB]
+
+# Some platforms provide zlib out of the box, so only add it when needed
+zlib_provided = (Linux() or MacOS())
 
 if context.buildBoostPython:
     requiredDependencies += [BOOST]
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9389

build_scripts/build_usd.py
```diff
@@ -2281,7 +2281,7 @@ def ForceBuildDependency(self, dep):
 
 # Determine list of dependencies that are required based on options
 # user has selected.
-requiredDependencies = [ZLIB, BOOST, TBB]
+requiredDependencies = [BOOST, TBB]
```
nvmkuruc: @dgovil Checkout line 2325.  I think you need to add a check to make sure that `ZLIB` is in `requiredDependencies` now before trying to remove.

pmolodo:
+1 to this - I came to a similar conclusion recently, but never got around to making a PR for it. Thanks @dgovil!

build_scripts/build_usd.py
dgovil: Ah good catch. I made it check first now as suggested.

davidgyu:
I think you need to add ZLIB when building PTEX as well.

dgovil:
@davidgyu Updated to include it in Ptex.

dgovil:
@davidgyu  I just wanted to ping this PR again to see if we could perhaps get it in? ZLib is small, but I like the idea of continuing @sunyab 's work of reducing dependencies down. If this was in, the core USD build would only bring in TBB (and Opensubdiv if imaging is enabled)

dgovil:
@meshula I believe this addresses your note on only building Zlib on platforms that need it.
Even though the `zlib_provided` argument could be simplified to `Windows()` today, I felt that keeping it as `zlib_provided` would allow for more granular control if needed.

meshula:
Thanks, I definitely prefer adding it if needed, rather than adding it and yanking it out again.

mhamid3d:
This needs to be added to alembic as well. Now the build fails on Windows with USD 25.02 when building with alembic.

After looking through alembics cmake, it may only be required when with hdf5.

pixar-oss:
Hi @mhamid3d, thanks for reporting!  Could you please create a new ticket with repro details?  A PR would also be very welcome if you have the time and the inclination. 🙏   We don't really use alembic or Windows in the studio.

mhamid3d:
Sure, PR created #3501

PR #2994, opened 2024-03-08
Adds pre-processor definitions for llvm sanitizers

nvidia-jomiller:
### Description of Change(s)

1. Introduces pre-processor definitions for Google Sanitizers: address sanitizer, thread sanitizer and undefined behavior sanitizer. 

* The first set of pre-processor definitions in `arch/defines.h` detect if USD is being built with sanitizers. These definitions allow for conditional compilation if certain code needs to opt-out of sanitizer instrumentation.
* The second set of pre-processor definitions in `arch/attributes.h` provide function attributes to disable sanitizer instrumentation for an entire function.

2. Disable tests that trigger a false-positive assertion from address sanitizer

* tf/SIGSEGV.cpp
* tf/SIGFPE.cpp
* arch/testArchUtil.cpp

#### Notes with Address Sanitizer

These changes should allow for OpenUSD to be built with Address Sanitizer and pass all unit tests for release and debug builds. **The one caveat is with Address Sanitizer's memory leak detection which triggers LOTS of assertions but is really the responsibility of Leak Sanitizer**. So for this reason, testing an address sanitized build of OpenUSD should set the following option:

```bash
export ASAN_OPTIONS=detect_leaks=0
```

This can be also be defaulted within code by defining the following function (see [ASAN runtime flags](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags) )

```cpp
const char *__asan_default_options() {
  return "detect_leaks=0";
}
```

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/arch/attributes.h
+++ b/pxr/base/arch/attributes.h
@@ -169,6 +169,29 @@ PXR_NAMESPACE_OPEN_SCOPE
 /// [[no_unique_address]] tag.
 #   define ARCH_EMPTY_BASES
 
+/// Macro used to indicate that a function should not be instrumented
+/// with address santizers.
+///
+/// In general, this attribute should be used sparingly. It's purpose
+/// is mostly for tests that when built with address sanitizer will trigger
+/// a false-positive, meaning the test is checking something that address
+/// sanitizer is trying to catch
+///
+/// Its important to understand that this attribute will only disable
+/// address sanitizer instrumentation for the function it's applied to.
+/// Any sanitized functions called from the target function
+/// will still be instrumented.
+///
+/// This attribute is used as follows:
+/// \code
+///    ARCH_NO_SANITIZE_ADDRESS_FUNCTION void Func() {
+///        ...
+///    }
+/// \endcode
+///
+/// \hideinitializer
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION
+
 #elif defined(ARCH_COMPILER_GCC) || defined(ARCH_COMPILER_CLANG)
 
 #   define ARCH_PRINTF_FUNCTION(_fmt, _firstArg) \
@@ -181,6 +204,16 @@ PXR_NAMESPACE_OPEN_SCOPE
 #   define ARCH_USED_FUNCTION __attribute__((used))
 #   define ARCH_EMPTY_BASES
 
+#if defined(ARCH_SANITIZE_ADDRESS)
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION \
+        __attribute__((no_sanitize_address))
+#else
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION
+#endif
+
+// Function attributes for other sanitizers (thread, undefined behavior, etc.)
+// intentionally omitted until support is added
+
 #elif defined(ARCH_COMPILER_MSVC)
 
 #   define ARCH_PRINTF_FUNCTION(_fmt, _firstArg)
@@ -190,6 +223,16 @@ PXR_NAMESPACE_OPEN_SCOPE
 #   define ARCH_UNUSED_FUNCTION
 #   define ARCH_USED_FUNCTION
 #   define ARCH_EMPTY_BASES __declspec(empty_bases)
+
+#if defined(ARCH_SANITIZE_ADDRESS)
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION \
+        __declspec(no_sanitize_address)
+#else
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION
+#endif
+
+// Function attributes for other sanitizers (thread, undefined behavior, etc.)
+// intentionally omitted until support is added
 
 #else
 
--- a/pxr/base/arch/defines.h
+++ b/pxr/base/arch/defines.h
@@ -104,4 +104,61 @@
 #define ARCH_HAS_MMAP_MAP_POPULATE
 #endif
 
+//
+// Sanitizers
+//
+
+// For most compilers sanitizers are enabled with something similar to
+// -fsanitize={address,thread,undefined}.
+// But detecting if the compiler is currently tyring to make a sanitized build
+// can vary depending on the compiler (or between versions of the compiler).
+// The following checks will determine if the compiler is making a sanitized
+// build and set a definition
+//
+// These definitions can be used to conditionally compile code where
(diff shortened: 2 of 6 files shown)


extras/imaging/examples/hdTiny/CMakeLists.txt
```diff
@@ -44,7 +43,6 @@ if (APPLE)
 elseif (UNIX)
     pxr_register_test(testHdTiny
         ENV
-            LD_PRELOAD=${CMAKE_INSTALL_PREFIX}/share/usd/examples/plugin/hdTiny.so
```
nvidia-jomiller: LD_PRELOAD seemed to be unnecessary here. This can be reverted and opt-out of the `testHdTiny` test for address sanitizer since LD_PRELOAD is [required](https://github.com/google/sanitizers/wiki/AddressSanitizerAsDso#asan-and-ld_preload) to correctly run tests

pxr/imaging/hdx/unitTestDelegate.cpp
```diff
@@ -58,7 +58,49 @@ TF_DEFINE_PRIVATE_TOKENS(
...
+);
+
+
+// XXX: These mostly-empty subclasses are here to prevent a link-time
+// optimization with HdRenderIndex::InsertTask<T>(), hdx/taskController.cpp and
+// HD_TRACE_FUNCTION() for address sanitized builds of USD. Without these
+// classes the linker will incorrectly "optimize" calls to the expanded
+// HD_TRACE_FUNCTION() in GetRenderIndex()->InsertTask<HdxPickTask>() for
+// taskController.cpp.
+class Hdx_UnitTestPickTask : public HdxPickTask {
```
nvidia-jomiller: I welcome feedback here as I'm not sure what the best solution is. Effectively, `GetRenderIndex()->InsertTask<HdxPickTask>()` is defined in `unitTestDelegate.cpp` and `taskController.cpp`. The linker will, understandably, optimize out the symbol in `taskController.o` since `unitTestDelegate.o` defines it first. This causes a linker error since it's `GetRenderIndex()->InsertTask<HdxPickTask>()` has been instrumented with ASan inside of `taskController.cpp`

pxr/base/arch/defines.h
```diff
@@ -104,4 +104,45 @@
...
+        #endif
+    #else
+        #if defined(__SANITIZE_ADDRESS__)
+            #define ARCH_SANITIZE_ADDRESS
+        #endif
+        #if defined(__SANITIZE_THREAD__)
+            #define ARCH_SANITIZE_THREAD
+        #endif
+    #endif
+#elif defined(AARCH_COMPILER_MSVC)
```
nvmkuruc: Should this be `ARCH_COMPILER_MSVC`?

pxr/base/arch/defines.h
```diff
@@ -104,4 +104,45 @@
 #define ARCH_HAS_MMAP_MAP_POPULATE
 #endif
 
+//
+// Sanitizers
```
nvmkuruc, line 108: Can you explain a little more has these work in the comment? It's not immediately clear if these are checking if the sanitizer is enabled or just if the compiler version supports sanitizing.

pxr/base/arch/attributes.h
```diff
@@ -181,6 +181,25 @@ PXR_NAMESPACE_OPEN_SCOPE
 #   define ARCH_USED_FUNCTION __attribute__((used))
 #   define ARCH_EMPTY_BASES
 
+#if defined(ARCH_SANITIZE_ADDRESS)
```
nvmkuruc, line 207: Is there harm in just always setting these, not just when the sanitizer is enabled? It would reduce variability.

pxr/base/arch/attributes.h
```diff
@@ -191,6 +210,15 @@ PXR_NAMESPACE_OPEN_SCOPE
 #   define ARCH_USED_FUNCTION
 #   define ARCH_EMPTY_BASES __declspec(empty_bases)
 
+#if defined(ARCH_SANITIZE_ADDRESS)
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION \
+        __declspec(no_sanitize_address)
+#else
+#   define ARCH_NO_SANITIZE_ADDRESS_FUNCTION
+#endif
+#   define ARCH_NO_SANITIZE_UNDEFINED_FUNCTION
```
nvmkuruc: Given that we don't use `ARCH_NO_SANITIZER_{THREAD,UNDEFINED_BEHAVIOR}_FUNCTION` yet and we don't support them on `msvc`, perhaps we elide them for now (for all compilers)?

jesschimein:
Filed as internal issue #USD-9406

extras/imaging/examples/hdTiny/CMakeLists.txt
nvidia-jomiller: Moved into it's own PR:
https://github.com/PixarAnimationStudios/OpenUSD/pull/2999

pxr/base/arch/defines.h
nvidia-jomiller: Indeed, thanks for catching this!

pxr/imaging/hdx/unitTestDelegate.cpp
nvidia-jomiller: Moved to it's PR:
https://github.com/PixarAnimationStudios/OpenUSD/pull/3000

pxr/base/arch/defines.h
nvidia-jomiller: I added documentation to both `attributes.h` and `defines.h` to describe the role that both `ARCH_SANITIZE_ADDRESS` and `ARCH_NO_SANITIZE_ADDRESS_FUNCTION` play

pxr/base/arch/attributes.h
nvidia-jomiller: As far as I know setting these function attributes when the sanitizer is not enabled is fine. What I'm not sure about is older compiler versions that might not have `no_sanitize_{address,thread,undefined}` (looking mostly at GCC and MSVC). So I error-ed on the side of caution and only apply function attributes if the corresponding sanitizer has been enabled

pxr/base/arch/attributes.h
nvidia-jomiller: Good point, really easy to add once support for the corresponding sanitizer is available. Removed but left a comment to serve as information and a place holder

PR #2997, opened 2024-03-11
Require strict URI scheme validation for `ArResolver`

nvmkuruc:
### Description of Change(s)

#2453 validated URI schemes against the RFC 3986 and RFC 3987 specification by default. An environment variable could be set to make scheme validation more permissive. This change removes that environment variable, requiring all schemes identifiers to adhere to the specification.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9414

PR #2999, opened 2024-03-11
[hdTiny] Remove LD_PRELOAD from testHdTiny

nvidia-jomiller:
### Description of Change(s)

`testHdTiny` was setting the `LD_PRELOAD` environment variable to that path pointing at `tinyHd.so`. Setting `LD_PRELOAD` seems to be unnecessary as `hdTiny` produces a plugInfo.json which should correctly load `tinyHd.so`. The `testHdTiny` test also passes without the use of `LD_PRELOAD`

This was initially discovered from https://github.com/PixarAnimationStudios/OpenUSD/pull/2994. Address sanitizers require the use of `LD_PRELOAD` to properly patch memory allocation function. `testHdTiny` would always fail with address sanitizers enabled since it overwrites `LD_PRELOAD`

### Fixes Issue(s)
-

### DIFF
--- a/extras/imaging/examples/hdTiny/CMakeLists.txt
+++ b/extras/imaging/examples/hdTiny/CMakeLists.txt
@@ -35,7 +35,6 @@ pxr_install_test_dir(
 if (APPLE)
     pxr_register_test(testHdTiny
         ENV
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9435

PR #3001, opened 2024-03-12
Autodesk: Add support for controlling how floats and doubles are streamed out

erikaharrison-adsk:
### Description of Change(s)

This PR adds a few options to control the format in which floats and doubles are streamed out by USD through the `TfStreamFloat` and `TfStreamDouble` classes. This PR strictly adds extra functionality and does not break the existing API and behavior; if no custom configuration is provided, the defaults will behave the same as before.

The original motivation behind this change was to allow us to dump Hydra data sources to text using the `HdDebugPrintDataSource` method, but using a given precision, in order to account for platform and architecture discrepancies. This in turn allows us to use the data source text dumps as reference values for tests.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/base/tf/stringUtils.cpp
+++ b/pxr/base/tf/stringUtils.cpp
@@ -947,32 +947,115 @@ Tf_GetDoubleToStringConverter()
     return conv;
 }
 
-void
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9437

PR #3006, opened 2024-03-16
Run PySideUIC with Python executable

dgovil:
### Description of Change(s)

This change makes the UIC tool be run with the same python executable as the build is using.

The UIC command line tool bakes the shebang of the Python it was installed with. However PySide builds are flexible across multiple Python versions due to how shiboken works. This can create scenarios where UIC is unexpectedly run with a different Python, or worse can fail if that older Python executable has been removed.

This just makes sure everything runs as expected with the user specified build environment.


jesschimein:
Filed as internal issue #USD-9463

PR #3014, opened 2024-03-23
[arch] Arch_DebuggerIsAttachedPosix: use TracerPid in /proc/self/status

pmolodo:
### Description of Change(s)

The old implemenation worked by trying to attach a ptrace to the current process, and if it failed, then we assumed it was being debugged.

The problem is that we have no easy way to check if an `EPERM` result from ptrace is because it's being debugged, or an actual permissions error - and permissions checking process is somewhat complex - see the "Ptrace access mode checking" section here:

https://man7.org/linux/man-pages/man2/ptrace.2.html

On a basic Ubuntu-22.04 test system, ptrace would return `EPERM` even when the process was not being debugged, resulting in a "false positive". The new `TracerPid` method was more accurate in this case.

Checking the `/proc/self/status` for `TracerPid` is also the same method that gperftools and even gdb itself use to check if a process is being debugged:

- https://github.com/gperftools/gperftools/blob/02adc8ceab39bbeac1f65e10bde577e1753094fa/src/heap-checker.cc#L183
- https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/nat/linux-procfs.c;h=b17e3120792e0e0790271898212b69b0577847cc;hb=HEAD#l68
- https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/testsuite/gdb.threads/siginfo-threads.c;h=22c6038206ba77fc3432da5ee30284c80641f305;hb=HEAD#l373

### Fixes Issue(s)

- `ArchDebuggerIsAttached` would return a false positive on some Linux distros (ie, Ubuntu-22.04)

### DIFF
--- a/pxr/base/arch/debugger.cpp
+++ b/pxr/base/arch/debugger.cpp
@@ -45,8 +45,10 @@
 #include <cstring>
 #include <errno.h>
 #include <fcntl.h>
+#include <fstream>
 #include <unistd.h>
 #include <string>
+#include <string_view>
 #endif
 #if defined(ARCH_OS_DARWIN)
 #include <sys/sysctl.h>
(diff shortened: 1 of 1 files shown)


pxr/base/arch/debugger.cpp
```diff
@@ -342,50 +343,59 @@ Arch_DebuggerAttachExecPosix(void* data)
 
 #if defined(ARCH_OS_LINUX)
 
+// Reads /proc/self/status, finds the line starting with "field:", and
+// returns the portion following the ":".
+// Note that the returned string will generally include leading whitespace
 static
-bool
-Arch_DebuggerIsAttachedPosix()
+std::string Arch_ReadProcStatusField(const char* field)
```
nvmkuruc: If you use `std::string_view` here, you don't need to use `strlen` or `strcmp` and don't have to worry about potentially handling `nullptr` `field` values.

nvmkuruc (commented):
👍

jesschimein:
Filed as internal issue #USD-9478

pxr/base/arch/debugger.cpp
pmolodo: Good point - done: https://github.com/PixarAnimationStudios/OpenUSD/pull/3014/commits/ed2b39b4d6a428db954ed3dae5153cfb8a618e72

gitamohr:
I'm pulling this one in today.

pxr/base/arch/debugger.cpp
```diff
@@ -344,50 +346,58 @@ Arch_DebuggerAttachExecPosix(void* data)
...
+            continue;
         }
 
-        // Wait for the parent to stop as a result of the attach.
-        int status;
-        while (waitpid(parent, &status, 0) == -1 && errno == EINTR) {
-            // Do nothing
+        if (line.compare(0, field.size(), field) == 0 && line[field.size()] == ':') {
+            // We found our "field:" line
+            return line.substr(field.size() + 2);
```
gitamohr, line 368: This should be `+ 1` no?  We want the text after the colon?

pxr/base/arch/debugger.cpp
```diff
@@ -344,50 +346,58 @@ Arch_DebuggerAttachExecPosix(void* data)
...
-    // Parent process
-    int status;
-    while (waitpid(pid, &status, 0) == -1 && errno == EINTR) {
-        // Do nothing
+// Reads the "TracerPid:" field from /proc/self/status
+// Returns a result < 0 if there was an error.
+static
+int Arch_ReadTracerPid() {
+    try {
+        return std::stoi(Arch_ReadProcStatusField("TracerPid"));
```
gitamohr, line 386: I think I'm going to rework this bit to do manual whitespace skipping and `std::from_chars()` to make it locale-independent.

pxr/base/arch/debugger.cpp
pmolodo: Ah - good catch.  I think the what happened was I initially thought that what followed the `:` was (possibly multiple) space characters - ie, `"Field:    value"`, and decided not to deal with whitespace stripping, added [this comment](https://github.com/PixarAnimationStudios/OpenUSD/pull/3014/files#diff-3bf66466b125291e781fb6bb0328fffe857d6732e71e37103a8e3125f4095ddbR351): 

```
// Note that the returned string will generally include leading whitespace
```

But I then realized what followed was always a tab character - ie, `"Field:\tvalue"`, and I modified the code to just skip the tab character... but forgot to add an explanatory comment (or remove the old comment).

Probably what you've proposed (locale-independent whitespace stripping) is more robust, though.

Issue #3015, opened 2024-03-25
Unlit Surface does not work in Storm


dgovil:
### Description of Issue

Storm fails to shadergen the MaterialX unlit shader with the following error

```
Warning: in _ValidateCompilation at line 229 of /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:1619:5: error: use of undeclared identifier 'u_viewPosition'
    u_viewPosition = vec3(HdGet_worldToViewInverseMatrix() * vec4(0.0, 0.0, 0.0, 1.0));
    ^
```

This appears to be caused by the shadergen assuming that the uniform which is added in `MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp, line 64` is available.
However, the unlit surface is using `MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp` which doesn't add this uniform.

So it seems like unlit or other shaders that do not output that constant will not work as a result.

### Steps to Reproduce
1.  Open the attached file in usdview
[skydome.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14749475/skydome.zip)
2. Look in the command line to see the error above

### System Information (OS, Hardware)
macOS 14.4 M2 Macbook

### Package Versions
USD 23.11 , 24.3

jesschimein:
Filed as internal issue #USD-9483

dgovil:
This will be resolved by https://github.com/PixarAnimationStudios/OpenUSD/pull/3045

PR #3016, opened 2024-03-25
Fix rendering of Unlit shaders (Issue 3015)

dgovil:
### Description of Change(s)

This PR makes sure that the shadergen checks that the View Position constant is present before it writes code that depends on it.

This fixes cases for shaders like unlit that do not depend on view direction in the rest of their shader code. 

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3015


jesschimein:
Filed as internal issue #USD-9484

dgovil:
@jesschimein I'll close this one out since https://github.com/PixarAnimationStudios/OpenUSD/pull/3045 is a more thorough fix that covers the same issue.

PR #3017, opened 2024-03-26
OCIO DisplayViewTransform uses display name as colourspace if display_colorspace is <USE_DISPLAY_NAME> in OCIO config

jonahjnewton:
OCIO DisplayViewTransform uses display name as colourspace if display_colorspace is <USE_DISPLAY_NAME> in OCIO config

### Description of Change(s)
OCIO 2 allows View transforms in an OCIO config to set their display_colorspace to <USE_DISPLAY_NAME> to use the Display name to find the display colorspace. Usdview does not handle this, and therefore errors stating that it can't find a colorspace named <USE_DISPLAY_NAME>. 

This change adds a simple check to use the display name as the colorspace in Usdview if the incoming colorspace is <USE_DISPLAY_NAME>.

### Fixes Issue(s)
- No existing issues that I can find

### DIFF
--- a/pxr/usdImaging/usdviewq/viewSettingsDataModel.py
+++ b/pxr/usdImaging/usdviewq/viewSettingsDataModel.py
@@ -463,7 +463,9 @@ class ViewSettingsDataModel(StateSource, QtCore.QObject):
         """Specifies the OCIO settings to be used. Setting the OCIO 'display'
            requires a 'view' to be specified."""
 
-        if colorSpace:
+        if colorSpace and colorSpace == "<USE_DISPLAY_NAME>":
+                self._ocioSettings._colorSpace = display
+        elif colorSpace:
             self._ocioSettings._colorSpace = colorSpace
 
         if display:


jesschimein:
Filed as internal issue #USD-9485

tcauchois:
Hey Jonah,

My understanding from the OCIO folks is that <USE_DISPLAY_NAME> should be entirely an internal OCIO detail, and client code shouldn't need to worry about it at all.  Can you give us some more information, like where the "<USE_DISPLAY_NAME>" is coming from (is it OCIO.GetCurrentConfig().getColorSpaces()?) and which error you're hitting exactly?

Thanks,
Tom

jonahjnewton:
Hey Tom, I'm simply loading an OCIO config with a View that contains the <USE_DISPLAY_NAME> keyword, and then applying that view transform in the OCIO section of USDview. The error that comes up in the USDview terminal states that it can't find a colorspace named <USE_DISPLAY_NAME>, which I understand should be recognised by OCIO and then translated to the display colorspace.

The same OCIO install works in multiple other applications just fine, so I'm unsure why it would be happening only with USDview, but that's the issue I'm encountering.

tcauchois:
Got it, thanks.  Could you paste the terminal output?

jonahjnewton:
I just get this line as the terminal output:
`ERROR: Usdview encountered an error while rendering.DisplayViewTransform error. Cannot find source color space named '<USE_DISPLAY_NAME>'.`

tcauchois:
The folks on OCIO slack suggested an alternate fix that I have written up locally, but if you could share some kind of minimal OCIO config for me to validate against that would be great.

tcauchois:
Or, if you'd like to try their proposal yourself: in appController.py, setOcioConfig, remove the line where we compute colorspace, and only pass "display" and "view" to setOcioSettings.  Let me know how that works!

jonahjnewton:
> Or, if you'd like to try their proposal yourself: in appController.py, setOcioConfig, remove the line where we compute colorspace, and only pass "display" and "view" to setOcioSettings. Let me know how that works!

Hey Tom, I just replicated this proposal on my side and it definitely works better than my fix. It updates the colorspaces more consistently, whereas I have found my fix can cause the display transform to be applied inconsistently. 

Thanks for the update!

tcauchois:
Awesome, thanks for the confirmation!  I've checked in my fix and plan to take it instead of this PR, but once it lands please let me know if you have any further issues!  And even though we're going with the other approach, thanks for the PR :).  I'm happy to see the OCIO integration in a more robust place.

Once my fix lands in dev, it should get linked to this PR and the PR should be closed.

jonahjnewton:
> Awesome, thanks for the confirmation! I've checked in my fix and plan to take it instead of this PR, but once it lands please let me know if you have any further issues! And even though we're going with the other approach, thanks for the PR :). I'm happy to see the OCIO integration in a more robust place.
> 
> Once my fix lands in dev, it should get linked to this PR and the PR should be closed.

Thanks a lot mate! Glad to see it fixed!

PR #3021, opened 2024-03-27
Add missing <algorithm> include

touraill-adobe:
On newer msvc toolchains there is a C2039/C3861 error about `iter_swap` not being a member of `std`.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9494

PR #3025, opened 2024-03-28
Autodesk: Optimize RenderIndex removal

erikaharrison-adsk:
### Description of Change(s)

There are two parts to this optimization.

- Optimize SceneIndex removal:
   - Modify _Entry storage for better alignment;
   - Use double link for the tree in `SdfPathTable`.

- Optimize RenderIndex removal:
    - Add indices container to avoid linear search.

There are mainly two issues in `SdfPathTable` and `Hd_SortedIds` separately.

1. `SdfPathTable`:
The one-way linked list in the node results in a linear search. This will cause O(n^2) time when deleting items one by one. To fix this problem, we replace it with double-way linked lists.
Besides, storing the exact value in the data structure may trigger the page fault more frequently, especially when data size increases. We then replace that with a fixed-size pointer.

2. `Hd_SortedIds`:
The swap and pop-up strategy for deleting single items also causes problems. This will break the order for most cases, and as a result, subsequent removals will fall into the linear search, or a `HdRenderIndex::GetRprimIds()` will require sorting the items. To resolve this problem with limited influence, we add an index container for fast query, while keeping the current partial sort strategy for fast insert and ranged deletion.

**Result**
Following is a brief view of what our changes bring to USD. We gain a significant improvement (~3.5x) when data size goes to 100k+:
image

- Data Size: flatten item count to remove (i.e. with the same prefix thus in the same list).
- Time(ns): average time of 9 iterations running the benchmark case.
- With/Without SceneIndex: `HD_ENABLE_SCENE_INDEX_EMULATION` true/false.
- Test environment: Mac Mini M2 (4p core + 4e core, 24G), Ventura 13.5.

And now, the difference between with and without SceneIndex simulation looks reasonable, with a ratio of ~2 and both maintaining a linear relationship with the amount of data:
image

As this change is supposed to cause no regression, the following are the results in even tiny data size:

image
image

Considering the extra data and the unit (ns), some tiny differences are reasonable.

**Reference**

- Test code (with google benchmark)

```
const int testAmount = 1 << 16;
const std::string prefix = isFlattened ? "/prim_" : "/root/prim_";
SdfPathVector pathList;

// Prepare data set.
for (int i = 0; i < testAmount; ++i)
{
    SdfPath id(prefix + std::to_string(i));
    pRenderIndex->InsertRprim(HdPrimTypeTokens->mesh, pSceneDelegate, id);
    pathList.push_back(id);
}
pRenderIndex->SyncAll(&tasks, &taskContext);

// Timing block
{
    for (int i = 0; i < testAmount; ++i)
        pRenderIndex->RemoveRprim(pathList[i]);
}

pRenderIndex->SyncAll(&tasks, &taskContext);
```

Corresponding topic in AOUSD:
- [Performance of removing rprim from RenderIndex](https://forum.aousd.org/t/performance-of-removing-rprim-from-renderindex/1034)

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hd/sortedIds.cpp
+++ b/pxr/imaging/hd/sortedIds.cpp
@@ -38,6 +38,7 @@ static const size_t INSERTION_SORT_MAX_ENTRIES = 32;
 
 Hd_SortedIds::Hd_SortedIds()
  : _ids()
+ , _idIndices()
  , _sortedCount(0)
  , _afterLastDeletePoint(INVALID_DELETE_POINT)
 {
@@ -46,6 +47,7 @@ Hd_SortedIds::Hd_SortedIds()
 
 Hd_SortedIds::Hd_SortedIds(Hd_SortedIds &&other)
  : _ids(std::move(other._ids))
+ , _idIndices(std::move(other._idIndices))
  , _sortedCount(other._sortedCount)
  , _afterLastDeletePoint(other._afterLastDeletePoint)
 {
@@ -63,88 +65,29 @@ void
 Hd_SortedIds::Insert(const SdfPath &id)
 {
     _ids.push_back(id);
+    _idIndices[id] = _ids.size() - 1;
     _afterLastDeletePoint = INVALID_DELETE_POINT;
 }
 
 void
 Hd_SortedIds::Remove(const SdfPath &id)
 {
-    // The first implementation of this deletion code deleted the
-    // element in place.  This kept the list sorted, but was a
-    // performance issue on unloading a stage as a lot of prims
-    // get removed and the shifting of the vector become a bottleneck
-    // So instead, we do a more efficient removal (by swapping the
-    // element to be removed with the element on the end of the vector).
-    // The down side is that the list is now unsorted, so needs to be
-    // sorted again (which is deferred).
-    //
-    // However, this means that the list is now unsorted in mass removal.
-    // In order to use the binary search, we need a sorted list, but sorting
-    // again would be too expensive in this case.
-    //
-    // We still need to optimize for batch deletions.  Typically these
-    // will be a sort list where the path to delete will be the next one
-    // in the list after the path just deleted.
-    //
-    // If it is not that path, then check to see if it is in the sorted
-    // portion.  If it is not in the sorted portion, do a linear search
-    // through the unsorted portion.
+    // For better removal, we use an extra index list to find the
+    // position of the id in the list. This is a trade-off between
+    // former implementation which can be partially sorted on demand.
 
-    // Try last removal point
-    SdfPathVector::iterator idToRemove = _ids.end();
-
-    if (_afterLastDeletePoint != INVALID_DELETE_POINT) {
-        if (_ids[_afterLastDeletePoint] == id) {
-            idToRemove = _ids.begin() + _afterLastDeletePoint;
-        }
-    }
-
-    // See if we can binary search sorted portion
-    if (idToRemove == _ids.end()) {
-        if (_sortedCount > 0) {
-            // Check to see if id is somewhere within the sorted range
-            if (id <= _ids[_sortedCount - 1]) {
-
-                SdfPathVector::iterator endSortedElements =
-                                                    _ids.begin() + _sortedCount;
-                idToRemove = std::lower_bound(_ids.begin(),
-                                              endSortedElements,
-                                              id);
-
-                if (idToRemove != endSortedElements) {
-                    // Id could actually exist in the unsorted part
-                    // because of an insert.
-                    // Lower bound will then return an iterator to
-                    // the element after where id should be.
-                    if (*idToRemove != id) {
-                        idToRemove = _ids.end();
-                    }
-                } else {
-                    // We checked that id should be in the sorted range
-                    // so lower_bound should return an iterator between
-                    // begin and endSortedElements - 1.
-                    TF_CODING_ERROR("Id (%s) greater than all items in "
-                                    " sorted list",
-                                    id.GetText());
-                    idToRemove = _ids.end();
-                }
-            }
-        }
-    }
-
-    // If all else fail, linear search through unsorted portion
-    if (idToRemove == _ids.end()) {
-        idToRemove = std::find(_ids.begin() + _sortedCount,
-                               _ids.end(),
-                               id);
-    }
-
-    if (idToRemove != _ids.end()) {
-        if (*idToRemove == id) {
-            SdfPathVector::iterator lastElement = _ids.end();
-            --lastElement;
-
+    // Search in the index list for fast access.
+    auto idIndexIter = _idIndices.find(id);
+    if (idIndexIter != _idIndices.end()) {
+        if (idIndexIter->first == id) {
+            SdfPathVector::iterator idToRemove = _ids.begin() + idIndexIter->second;
+            SdfPathVector::iterator lastElement = --_ids.end();
+            
             if (idToRemove != lastElement) {
+                // Update index
+                _idIndices[*lastElement] = idIndexIter->second;
+                _idIndices.erase(id);
+                
                 std::iter_swap(idToRemove, lastElement);
 
                 if (std::distance(idToRemove, lastElement) == 1) {
@@ -164,6 +107,7 @@ Hd_SortedIds::Remove(const SdfPath &id)
                                         static_cast<size_t>(
                                                   (idToRemove - _ids.begin())));
             } else {
+                _idIndices.erase(id);
                 _ids.pop_back();
                 _afterLastDeletePoint = INVALID_DELETE_POINT;
 
@@ -195,10 +139,13 @@ void Hd_SortedIds::RemoveRange(size_t start, size_t end)
         return;
     }
 
+    for(size_t index = start; index != end; index++)
+        _idIndices.erase(_ids[index]);
+    
     SdfPathVector::iterator itStart = _ids.begin() + start;
     SdfPathVector::iterator itEnd   = _ids.begin() + (end + 1);
-
     _ids.erase(itStart, itEnd);
+
     _sortedCount -= numToRemove;
     _afterLastDeletePoint = INVALID_DELETE_POINT;
 }
@@ -207,6 +154,7 @@ void
 Hd_SortedIds::Clear()
 {
     _ids.clear();
+    _idIndices.clear();
     _sortedCount = 0;
     _afterLastDeletePoint = INVALID_DELETE_POINT;
 }
@@ -217,7 +165,6 @@ Hd_SortedIds::_InsertSort()
     SdfPathVector::iterator sortPosIt = _ids.begin();
     // skip already sorted items
     sortPosIt += _sortedCount;
-
     while (sortPosIt != _ids.end()) {
         SdfPathVector::iterator insertPosIt = std::lower_bound(_ids.begin(),
                                                                sortPosIt,
@@ -226,6 +173,11 @@ Hd_SortedIds::_InsertSort()
         std::rotate(insertPosIt, sortPosIt, sortPosIt + 1);
         ++sortPosIt;
     }
+    
+    // Update indices.
+    size_t indexValue = 0;
+    for(auto const& id : _ids)
+        _idIndices[id] = indexValue++;
 }
 
 void
@@ -238,8 +190,17 @@ Hd_SortedIds::_FullSort()
     // If needed, merge
     if (mid == _ids.begin() || *(mid-1) < *(mid)) {
         // List is fully sorted.
+        
+        // Update remaining indices
+        for(size_t indexValue = _sortedCount; indexValue < _ids.size(); ++indexValue)
+            _idIndices[_ids[indexValue]] = indexValue;
     } else {
         std::inplace_merge(_ids.begin(), mid, _ids.end());
+        
+        // Update all indices
+        size_t indexValue = 0;
+        for(auto const& id : _ids)
+            _idIndices[id] = indexValue++;
     }
 }
 
--- a/pxr/imaging/hd/sortedIds.h
+++ b/pxr/imaging/hd/sortedIds.h
@@ -75,6 +75,7 @@ public:
 
 private:
     SdfPathVector           _ids;
+    std::unordered_map<SdfPath, size_t, SdfPath::Hash> _idIndices;
     size_t                  _sortedCount;
     ptrdiff_t               _afterLastDeletePoint;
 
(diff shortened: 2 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9502

tcauchois:
Thanks for the submission!  I'm trying to do a close read of these because these changes make a lot of engineering tradeoffs that I'm unsure about, in terms of adding memory and indirection, especially with SdfPathTable since that datastructure is used in a lot of ways in a lot of places in our code.

I think your analysis of Hd_SortedIds is correct, though I wonder if just having a "remove list" would be more efficient; e.g. when you remove, queue up to "N" individual removals, and when you get above some threshold or try to read you process the removals by doing the correct memmoves.

For SdfPathTable, this datastructure is used in a number of very different places.  I'm curious if your analysis was specific to one of the path tables in hydra?  e.g. the HdRetainedSceneIndex path table, or the HdSceneIndexAdapterSceneDelegate path table?  I'm also wondering how the performance breakdown differed in terms of storing value on the heap vs adding the "parent" pointer to _Entry vs any kind of alignment differences?  Storing the value on the heap is something we can (and probably should) do at the callsite.  But it sounds like the slowdown that prompted the parent pointer was in SdfPathTable::erase, and not in any traversal in hydra code, correct?

adskWangl:
Thank you very much for your review! Certainly there is more than one solution for these technique problems and we are always open for discussion.

> I think your analysis of Hd_SortedIds is correct, though I wonder if just having a "remove list" would be more efficient; e.g. when you remove, queue up to "N" individual removals, and when you get above some threshold or try to read you process the removals by doing the correct memmoves.

The reason why we don't adopt "remove list" is that:

1.  It cannot completely resolve the bottleneck. As this problem is caused by disordered ids, to sort the ids everytime before applying the "remove list" would definitely be expensive and unaffordable. Evaluating and trading off for an appropriate acceleration strategy is quite complicated and use-case dependent, resulting in maintenance difficulties. Also, "remove list" is not free and itself may become another bottleneck.
2. USD already has similar pending lists in many other places like SceneIndex. Adding another "remove list" in a fundamental data structure with async processing introduces more complexity.

We actually have another idea to replace the current std::vector with a single std::set and offer range access through std::ranges. This could be efficient but requires changes in APIs. Thus we'd like to get your input on this idea.

> For SdfPathTable, this datastructure is used in a number of very different places. I'm curious if your analysis was specific to one of the path tables in hydra? e.g. the HdRetainedSceneIndex path table, or the HdSceneIndexAdapterSceneDelegate path table? I'm also wondering how the performance breakdown differed in terms of storing value on the heap vs adding the "parent" pointer to _Entry vs any kind of alignment differences? Storing the value on the heap is something we can (and probably should) do at the callsite. But it sounds like the slowdown that prompted the parent pointer was in SdfPathTable::erase, and not in any traversal in hydra code, correct?

Sorry if I mislead you with the test data. The final analysis is actually made in the data structures themselves once the bottlenecks are located. We write our own benchmark cases similar to test cases like _TestSdfPathTable_. For storing value in the _Entry as a pointer instead of raw data, the most improvement is that there would be fewer page faults and also some small improvements when traversing. As only "key" matters when traversing, loading actual data is not necessary. We've also tried to create _Entry pool as its size fixed and known at compile time but unfortunately there was no big difference. Another thing I want to highlight is that RenderIndex has dual data storages, one from original implementation and another one is for SceneIndex emulation. Sometimes no downgrade found in traversal doesn't mean the SdfPathTable has no problem but only because it uses another data structure.
Could you please offer more cases for analysis? Specific cases with source code would be really helpful. Thanks!

tcauchois:
Just to continue the conversation with Hd_SortedIds: this class is meant to provide a sorted set of paths for us to use with this class: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/imaging/hd/primGather.h.  The interesting functions there are Subtree() and PredicatedFilter().  These output (filtered) path sets are then usually used for other operations, for example:
All prims -> filter to certain include/exclude paths -> for filtered prims, generate draw items -> add to draw batch.

I think for Subtree, a tree structure might be preferable, although we've had a lot of issues with std::set in the past because its API coupled with its balancing invariants just make it really slow for mass edits.

For PredicatedFilter, we dispatch sections of the scene to TBB, so we need some kind of random access iterator, and that's the reason Hd_SortedIds was originally written (to provide a sorted vector).  I'm very open to other approaches, though.

Our typical remove calls involve unloading a sub-scene, so if we're not tearing down hydra entirely we'll be deleting tens of thousands of prims at a time.  They will share a common path prefix, so usually they will be continguous in all of these datastructures.

Other questions: which malloc implementation are you using?  And is it possible to share any tests with us, like a remove-performance unit test that has the patterns you care about?

This PR has three pieces:
- Hd_SortedIds update.
- SdfPathTable entry storage change.
- SdfPathTable extra node pointers.

Whatever I land will have to be a piece at a time, I think, so that's one of the reasons I'm interested in which pieces had the biggest performance effect.

Thanks!

tcauchois:
One more question about the use cases you typically see issues with.  Are these:
1.) Uncorrelated leaf deletions?  (e.g. /A/B, /A/X/Y/Z, /D/E/F)
2.) Deletions of siblings but not a whole subtree (e.g. /A/B/X, /A/B/Y, but not /A/B/Z)
3.) Deletions of a whole subtree, but a leaf at a time (e.g. /A/B/X, /A/B/Y, /A/B/Z)

Thanks!

adskWangl:
Sorry for the late response and thank you for explaining the `Hd_SortedIds` design purpose.

I think the proposed change won't break the original designs or worsen removing a whole subtree. When the scenes are quite flattened, however, removing quantities of rprims will slow down the performance significantly as I analyzed above. And the first and second cases you mentioned are some typical cases we've come across:

> 1) Uncorrelated leaf deletions.
> 2) Deletions of siblings but not a whole subtree.

(I believe there is no reason to delete leaves one by one if we could decide the whole subtree is no longer needed.)

The following is the complete remove-performance case based on [google benchmark](https://github.com/google/benchmark) we used to **reproduce the key performance issues we've addressed**:

```cpp
BENCHMARK_DEFINE_F(HydraFixture, TestRenderIndexRemoval)(benchmark::State& state)
{
    // Test options. You can switch them on/off as you need.
    bool isInverse = true;
    bool hasSceneIndex = _pRenderIndex->IsSceneIndexEmulationEnabled(); // switch through HD_ENABLE_SCENE_INDEX_EMULATION.

    // Setup.
    auto const testAmount = state.range(0);
    auto groupSize = state.range(1);
    const std::string prefix = (groupSize > 0) ? "/root/g" : "/root/prim_";
    SdfPathVector primList;
    if (state.thread_index() == 0)
    {
        // A really simple SceneDelegate for setting up only.
        // Won't be used as we've located the bottleneck is in RenderIndex.
        _pSceneDelegate = std::make_unique<TinyDelegate>(_pRenderIndex.get());

        for (int i = 0; i < testAmount; ++i)
        {
            std::string str_id;
            if (groupSize > 0)
            {
                // This can test cases like "2.) Deletions of siblings but not a whole subtree (e.g. /A/B/X, /A/B/Y, but not /A/B/Z)".
                str_id = prefix + std::to_string(i%groupSize) + std::string("/prim_") + std::to_string(i);
            }
            else
            {
                // This can test cases like "1.) Uncorrelated leaf deletions? (e.g. /A/B, /A/X/Y/Z, /D/E/F)".
                str_id = prefix + std::to_string(i);
            }

            SdfPath id(str_id);
            primList.push_back(id);
        }
    }

    // Fundamental elements for running. Also won't impact the performance results,
    // as no actual rendering would be involved in the test.
    HdRprimCollection collection(HdTokens->geometry, HdReprSelector(HdReprTokens->hull));
    HdRenderPassSharedPtr renderPass(new TestRenderPass(_pRenderIndex.get(), collection));
    HdTaskSharedPtrVector tasks;
    tasks.push_back(std::make_shared<TestTask>(renderPass));
    HdTaskContext taskContext;

    // Test body (timing part).
    for (auto _ : state)
    {
        // Pause timing to prepare data set for the current iteration.
        state.PauseTiming();
        _pRenderIndex->SyncAll(&tasks, &taskContext);

        for (int i = 0; i < testAmount; ++i)
            _pRenderIndex->InsertRprim(HdPrimTypeTokens->mesh, _pSceneDelegate.get(), primList[i]);
        _pRenderIndex->SyncAll(&tasks, &taskContext);
        state.ResumeTiming();

        if(isInverse)
        {
            for (int i = testAmount-1; i >=0 ; --i)
                _pRenderIndex->RemoveRprim(primList[i]);
        }
        else
        {
            for (int i = 0; i < testAmount; ++i)
                _pRenderIndex->RemoveRprim(primList[i]);
        }
    }

    // Tear down & results output.
    if (state.thread_index() == 0)
    {
        primList.clear();
        _pRenderIndex->Clear();
        _pSceneDelegate.reset();

        // Time will be automatically output by google benchmark.
        // Following are supplement information of case configs.
        if(groupSize > 0)
            state.counters["GroupSize"] = groupSize;

        std::stringstream sstr;
        sstr << "Test";
        if (groupSize > 0) sstr << ",grouped";
        if (isInverse) sstr << ",inverse";
        if (hasSceneIndex) sstr << ",SceneIndex";
        state.counters[sstr.str()] = 1;
    }
}
BENCHMARK_REGISTER_F(HydraFixture, TestRenderIndexRemoval)
    ->ArgsProduct({
        benchmark::CreateRange(1 << 3, 1 << 18, getEnvOr("TEST_MULTIPLIER", 2)), // Test amount.
        { 0, 1 << 8, 1 << 9, 1 << 10 }                                           // Group size.
    });
```

The `Hd_SortedIds` mainly contributes to the slowdown in manipulating the member variable in `HdRenderIndex`, while `SdfPathTable` contributes to the `HdSceneIndexAdapterSceneDelegate` part when `HD_ENABLE_SCENE_INDEX_EMULATION` is on. Both of them are required to improve the `HdRenderIndex::RemoveRprim` and that's why we put them together. But I would be willing to separate the `Hd_SortedIds` change from `SdfPathTable` changes as it will also impact other workflow.

tcauchois:
Hey all, to update you on progress:

Testing your Hd_SortedIds implementation, we found that it improved random insert/delete metrics, but it regressed initial population and big subtree inserts/deletes.  We were inspired to do some optimization of our own on Hd_SortedIds, and the top of tree version beats both our old version and your version on our test cases.  Let me know if it looks good on your benchmark test/on your hardware!

I did an audit of SdfPathTable payloads in hydra, and found that almost everywhere they are pretty small (usually a token and a shared_ptr), and didn't feel like that was worth allocating the payload separately, like you're doing in this PR.  (Let me know if you find different!).  However, HdSceneIndexAdapterSceneDelegate had a huge payload: 3 tokens and 2 std::maps.  I've reworked that to be 1 token and 2 shared_ptrs, which also made the atomic code considerably simpler and probably more robust, and we've seen some modest performance gains from that.

In short: thanks for the great investigation, let us know if our changes help at all, and let us know if there's further work we can do in pxr/imaging to improve this!  We don't have as wide of a hardware suite to run perf testing on, but if changing allocation patterns in other path tables (for example) is a big win on your hardware and a wash on our hardware that's still probably something worth taking.

We can't take the SdfPathTable changes, full stop, since they have a big impact on USD.  I'm also very resistant to forking SdfPathTable for Hydra use, since it's a complex bit of code.  I'm interested to follow up on why you're doing so many uncorrelated leaf deletions, and whether they can be subtree deletions instead?  I know someone mentioned it might be related to managing residency in hydra, but I'd rather see that handled within the render delegate instead (though I know the code for that is much trickier at the moment).  Or if you keep a path table of subtree resource usage, that might help you partition the scene in a simpler way or something so that residency can be managed with a few subtree deletes.

Thanks!
Tom

adskWangl:
Thank you Tom for your work! Would you mind to share the code of your improvement? (Is this one: https://github.com/PixarAnimationStudios/OpenUSD/commit/5d746f34247d5d0a17f98a10e7d300a8548e94f8?) I want to try and test how it performs in our cases.

Unfortunately, most of our input modeling data is uncorrelated geometries. I'd first try to see if things are better as you decrease HdSceneIndexAdapterSceneDelegate's payload. So how about the idea of replacing the current std::vector with a single std::set and offering range access through std::ranges?

tcauchois:
The SceneIndexAdapterSceneDelegate change is here: https://github.com/PixarAnimationStudios/OpenUSD/commit/9378f9ebb7053b51c22ff40dfc705b1186d45756

The Hd_SortedIds changes are here:
https://github.com/PixarAnimationStudios/OpenUSD/commit/dc88512f4ca385457381f4d4dae1ac7e0f72607d
https://github.com/PixarAnimationStudios/OpenUSD/commit/3f4e37c91b05ac14908adee7329bd690f851beb8
https://github.com/PixarAnimationStudios/OpenUSD/commit/71117771e095f56c1aef3f528628902b361aba8f
https://github.com/PixarAnimationStudios/OpenUSD/commit/5d746f34247d5d0a17f98a10e7d300a8548e94f8

adskWangl:
Hi Tom! Your changes do help and work well in our benchmarks. The improvement in `HdSceneIndexAdapterSceneDelegate`'s payload optimizes the cases with SceneIndex emulation from another aspect, though I still have concerns about the `SdfPathTable`'s implementation. I'll let you know if we have further issues with more complex workflows.

> I'm interested to follow up on why you're doing so many uncorrelated leaf deletions, and whether they can be subtree deletions instead? I know someone mentioned it might be related to managing residency in hydra, but I'd rather see that handled within the render delegate instead.

The main reason is that we could have millions of renderable items. It would take too much memory and time to send them to render delegate and let the renderer do culling for us, thus we made some customized optimization with SceneIndex/SceneDelegate. As one typical side effect, the optimized scene data would become uncorrelated and flattened. BTW. Considering the data size, I would still like to consider the idea of adopting `std::ranges` as this could also help to avoid copying `SdfPath`s when querying.

tcauchois:
We've had bad experiences with std::set performance.  I'm curious where you'd use the std::ranges but vfxplatform is still on c++17 so I don't think we can use those yet.

I'm up for trying out any further ideas you have!  But I can only land them if they don't regress other important metrics, and the SdfPathTable change, by making path table nodes larger, causes a lot of regressions in USD...

tcauchois:
I'd really prefer to handle memory pressure inside of hdSt by dropping buffers, or never uploading buffers for culled items, but I know there's a lot of complications there.  I know that this is something that other folks are looking at though, in the context of running Storm on memory-limited devices.

Anyway, I appreciate the great performance notes and I'm glad to hear our changes are working for you!

jesschimein:
cc: @AdamFelt  could you let us know if the changes we've implemented so far are sufficient to close out this PR?  Thank you!

erikaharrison-adsk:
We’ve tested using our own benchmarks and the improvements work well in removing rprims. Closing for now and will follow up other PRs if we have other optimization suggestions. Thanks again!

PR #3005, opened 2024-03-16
PEGTL USDA Parser

erslavin:
- Added additional text file format parser based on PEGTL
- Added tests specific to PEGTL parser
- Removed lex / yacc based parser

### Description of Change(s)

This change introduces a new sdf / usda text file format paser based on PEGTL.  This change aligns the way text file scene descriptions are parsed with the way paths are now being parsed and removes the need for a dependency on lex / yacc.

Added PEGTL Text File Format Parser
Added tests specific to PEGTL parser that allow testing of partial scene description rules

### Fixes Issue(s)
-

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #3027, opened 2024-04-01
Replace `boost::bad_get` with new exception type

nvmkuruc:
### Description of Change(s)

(Requires #3005)

`boost::variant` is no longer used in OpenUSD. Its associated exception `boost::bad_get` is being used to signal that a conversion error has occurred during the parse.  This replaces usage of `boost::bad_get` with a new exception type `ConversionError` and removes the last non-python usage of boost in `sdf`.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9506

spiffmon:
THanks, @nvmkuruc !  It turns out Alex had a change already sitting in his tree that deboostifies the existing sdf parser code, so we'll get that in ASAP, and plan on pulling this in after we've moved to PEGTL.  Not sure if our change will cause conflicts for yours.

nvmkuruc:
We'll just adapt the PEGTL change whenever you do the next dev push.

PR #3028, opened 2024-04-01
Make boost an optional dependency of OpenUSD

nvmkuruc:
### Description of Change(s)

This PR makes `boost` an optional dependency when not building with python, openvdb, or openimageio. This changes the `build_usd` script to explicitly create the `lib` directory with the correct permissions, as downstream builds (like `tbb`) expected this to have already been created by the `boost` step.

### Fixes Issue(s)
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,11 @@ Dependencies
 Required:
  - C/C++ compiler
  - [CMake](https://cmake.org/documentation/)
- - [Boost](https://boost.org)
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9507

nvmkuruc:
This no longer depends on the PEGLT PR.

nvmkuruc:
Removed note about PCH files now that they've been updated by 5000b4535513d4bc5d5437a292690c0008e2767f

Issue #2757, opened 2023-10-25
usdedit Error: 'C:\Program' is not an internal or external command


michsiu:
### Description of Issue
After installed the USD, it works well with `usdcat`tool, but failed on both `usdview` and `usdedit`. The issue with the `usdedit` is that it pop-out an error message with `'C:\Program' is not an internal or external command`

I have searched it on Google which says that its the issue of `os.system`, but I don't know how to resolve it

```
'C:\Program' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
```
### Steps to Reproduce
1. open `x64 dev command tool`
2. run command `usdedit C:\Users\download\3dtest\OpenUSD-release\pxr\usd\bin\usdedit\testenv\testUsdEditFilePermissions\write_protected.usda`

### System Information (OS, Hardware)
Windows 10
![Snipaste_2023-10-26_01-24-12](https://github.com/PixarAnimationStudios/OpenUSD/assets/18113569/ceb95475-0eb6-4473-886a-b3e267c607ee)

### Package Versions
python 3.8

asluk:
Can you print %PATH% and %PYTHONPATH% from your terminal and paste the output here?

What if you install OpenUSD to folders with no spaces?

Is the usdview error the same as the usdedit error, or is the usdview error the other one you reported about not having the expected OpenGL capabilities?

michsiu:
@asluk Hello Asluk, thanks for your reply
Sure, this is the PATH env Variables I added `C:\USD\bin`, `C:\USD\lib`
And this is PYTHONPATH env Variables I added `C:\USD\lib\python`
And this is the OpenUSD path I installed `C:\USD`

I run the build script strictly and I see the success declaration when finish the installation

The issue of `usdview` is different, it seems it is the issue of OpenGL version, but I checked, it is the newest version , I have submitted the issue, this is the link `https://github.com/PixarAnimationStudios/OpenUSD/issues/2756`

Thanks

asluk:
Thanks @michsiu , can you also share what the terminal outputs if you echo %PATH% and %PYTHONPATH% ? Not sure why it would be trying to invoke something from c:\Program

michsiu:
@asluk No Problem, below is for `echo %path%`:
```
[...]

PR #3038, opened 2024-04-10
usdedit: Handle spaces in the file path to executables

medakk:
### Description of Change(s)

Correctly handle the case when the path to the editor has a space in it. This can often happen because of paths starting with `C:\Program Files\...`

### Fixes Issue(s)
- #2757

Three failures, but they look unrelated?

```
        453 - testUsdZipFile (Failed)
        711 - testUsdviewNavigationKeys (Failed)
        727 - testUsdResolverExample (Failed)
```

### DIFF
--- a/pxr/usd/bin/usdedit/usdedit.py
+++ b/pxr/usd/bin/usdedit/usdedit.py
@@ -25,7 +25,7 @@
 
 from __future__ import print_function
 
-import os, sys
+import os, subprocess, sys
 from pxr.UsdUtils.toolPaths import FindUsdBinary
 
 # lookup usdcat and a suitable text editor. if none are available, 
@@ -67,10 +67,9 @@ def _generateTemporaryFile(usdcatCmd, usdFileName, readOnly, prefix):
     (usdaFile, usdaFileName) = tempfile.mkstemp(
         prefix=fullPrefix, suffix='.usda', dir=os.getcwd())
 
-    # No need for an open file descriptor, as it locks the file in Windows.
+    subprocess.call([usdcatCmd, usdFileName], stdout=usdaFile)
+
     os.close(usdaFile)
- 
-    os.system(usdcatCmd + ' ' + usdFileName + '> ' + usdaFileName)
 
     if readOnly:
         os.chmod(usdaFileName, 0o444)
@@ -87,7 +86,7 @@ def _generateTemporaryFile(usdcatCmd, usdFileName, readOnly, prefix):
 def _editTemporaryFile(editorCmd, usdaFileName):
     # check the timestamp before updating a file's mtime
     initialTimeStamp = os.path.getmtime(usdaFileName)
-    os.system(editorCmd + ' ' + usdaFileName)
+    subprocess.call([editorCmd, usdaFileName])
     newTimeStamp = os.path.getmtime(usdaFileName)
     
     # indicate whether the file was changed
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9548

pxr/usd/bin/usdedit/usdedit.py
```diff
@@ -87,7 +86,7 @@ def _generateTemporaryFile(usdcatCmd, usdFileName, readOnly, prefix):
 def _editTemporaryFile(editorCmd, usdaFileName):
     # check the timestamp before updating a file's mtime
     initialTimeStamp = os.path.getmtime(usdaFileName)
-    os.system(editorCmd + ' ' + usdaFileName)
+    subprocess.call([editorCmd, usdaFileName])
```
nvmkuruc: This seems like a great change to make.  One note-- I think `subprocess.run` is preferred over `subprocess.call` for new usages.

https://docs.python.org/3/library/subprocess.html#subprocess.run

pxr/usd/bin/usdedit/usdedit.py
```diff
@@ -67,10 +67,9 @@ def _generateTemporaryFile(usdcatCmd, usdFileName, readOnly, prefix):
     (usdaFile, usdaFileName) = tempfile.mkstemp(
         prefix=fullPrefix, suffix='.usda', dir=os.getcwd())
 
-    # No need for an open file descriptor, as it locks the file in Windows.
+    subprocess.run([usdcatCmd, usdFileName], stdout=usdaFile)
```
nvmkuruc, line 70: Should `usdFileName` be `usdaFileName`?

medakk: No. `usdFileName` is the source file, `usdaFileName` is the temporary file created that we write to

(The names aren't great/descriptive for sure though)

nvmkuruc:
The original code wasn't doing this, but I'm wondering if you should check the result of `subprocess.run`.

pxr/usd/bin/usdedit/usdedit.py
```diff
@@ -87,7 +86,7 @@ def _generateTemporaryFile(usdcatCmd, usdFileName, readOnly, prefix):
 def _editTemporaryFile(editorCmd, usdaFileName):
     # check the timestamp before updating a file's mtime
     initialTimeStamp = os.path.getmtime(usdaFileName)
-    os.system(editorCmd + ' ' + usdaFileName)
+    subprocess.run([editorCmd, usdaFileName])
```
pixar-oss, line 89: `editorCmd` may not contain a single executable. It may be something like `gvim -f`. If it contains "emacs" then lines 53-56 append ` --name "usdedit <something>"` to the user provided editor.

We probably need to use `shlex.split()` to parse `editorCmd` into a list of  shell tokens, then append `usdaFilename` to the list and run that.

spitzak: Obviously `split()` will have the same problem. Do backslashes in front of the spaces work to indicate they are part of the executable name? Ie `foo\ bar baz` will run the program `foo bar` and set `argv[1]` to `baz'.

spitzak: Or `"foo bar" baz`?

PR #3043, opened 2024-04-15
Specify `analyze_t` in `Xid` parsing rules

nvmkuruc:
### Description of Change(s)
PEGTL offers an `analyze` utility that can detect some errors in grammars. It requires a hint about how a grammar matches against its input via `analyze_t`. Rules that consume input that it matches against are classified as "any" rules.

This change updates the `XidStart` and `XidContinue` rules defined in the path parser and adds a call to `analyze()` in `testSdfPathParser`. `analyze()` returns 0, if there are no errors.  Updating the path grammar will allow the forthcoming PEGTL parser #3005 to similarly validate against `analyze()`.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/pathParser.h
+++ b/pxr/usd/sdf/pathParser.h
@@ -44,6 +44,9 @@ namespace PEGTL_NS = tao::TAO_PEGTL_NAMESPACE;
 // Helper rules for parsing UTF8 content
 struct XidStart
 {
+    using analyze_t =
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-9553

nvmkuruc:
This works with the old PEGTL implementation, but not the new one.  analyze has been moved to `contrib/analyze.hpp`

PR #1831, opened 2022-04-14
Ogs/selection

AdamFelt:
This submission adds deep selection capabilities to HdStorm.  This allows for selection of all objects within a pick-point or frustum to be selected even if the objects are obscured by other objects drawn on top. 

It uses a GPU rendering technique similar to ID based picking except that it accummulates all pick results into a one dimensional "pickBuffer" instead of storing a single selection per pixel like the ID buffer approach does.

It can be enabled by setting a new resolve mode defined as HdxPickTokens->resolveDeep.

A new helper function named TestIntersections (plural) has been added to UsdImagingGlEngine to exercise the new functionality.  A new Unit test is included to test the new resolve mode and UsdImagingGLEngine helper function.

- [x ] I have submitted a signed Contributor License Agreement

PR #3044, opened 2024-04-16
Autodesk: Deep Selection

erikaharrison-adsk:
### Description of Change(s)

Following up on the feedback from https://github.com/PixarAnimationStudios/OpenUSD/pull/1831 

This resolves all code review comments indicated in the original review. 
* MacOS support
* Better documentation
* Bug fixing, etc.

Background On Original PR:

This submission adds deep selection capabilities to `HdStorm`. This allows for selection of all objects within a pick-point or frustum to be selected even if the objects are obscured by other objects drawn on top.

It uses a GPU rendering technique similar to ID based picking except that it accummulates all pick results into a one dimensional "pickBuffer" instead of storing a single selection per pixel like the ID buffer approach does.

It can be enabled by setting a new resolve mode defined as `HdxPickTokens->resolveDeep`.

A new helper function named `TestIntersections `(plural) has been added to `UsdImagingGlEngine` to exercise the new functionality. A new Unit test is included to test the new resolve mode and `UsdImagingGLEngine` helper function.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9558

pxr/imaging/hdx/pickTask.cpp
```diff
@@ -650,6 +768,13 @@ HdxPickTask::Execute(HdTaskContext* ctx)
             {HdxRenderTagTokens->widget});
     }
 
+    // For 'resolveDeep' mode, read hits from the pick buffer.
+    if (_contextParams.resolveMode == HdxPickTokens->resolveDeep) {
+        _ResolveDeep();
+        _hgi->EndFrame();
```
tcauchois, line 774: What's the specific need for EndFrame() here?  I know we have broader issues about calling that at the right time, but I'm very surprised to see it in an if branch here.

pxr/imaging/hdx/shaders/renderPass.glslfx
```diff
@@ -176,11 +239,95 @@ void RenderOutput(vec4 Peye, vec3 Neye, vec4 color, vec4 patchCoord)
...
+    int edgeId = GetPrimitiveEdgeId();
+    int pointId = GetPointId();
+
+    elementIdOut = IntToVec4(elementId);
+    edgeIdOut = IntToVec4(edgeId);
+    pointIdOut = IntToVec4(pointId);
     
     neyeOut = IntToVec4(hd_vec4_2_10_10_10_set(vec4(Neye,0)));
+
+    // add the item to the pick buffer, but only unique
```
tcauchois, line 252: Not a blocker to merging, but this would be nice as a helper function.

Since the pick buffer appears to be always allocated, do you expect the ATOMIC_LOAD() below for the early out when we're *not* deep picking to be a performance issue?

pxr/imaging/hdx/shaders/renderPass.glslfx
```diff
@@ -176,11 +239,95 @@ void RenderOutput(vec4 Peye, vec3 Neye, vec4 color, vec4 patchCoord)
...
+        const int subBufferOffset = storageOffset + bufferNumber * subBufferCapacity * entrySize;
+
+        // loop through the item's sub-buffer to see if the item is already there
+        // if not, add it in atomic fashion
+        int entryNumber = 0;
+        do
+        {
+            // see if we need to add a new entry
+            if (entryNumber == ATOMIC_LOAD(PickBuffer[sizeOffset])) {
+                ATOMIC_COMP_SWAP(PickBuffer[sizeOffset], entryNumber, entryNumber + 1);
```
tcauchois, line 310: Again, not a blocker to merging, but:
1. ATOMIC_LOAD + ATOMIC_COMP_SWAP could just be ATOMIC_ADD?  With some slight re-ordering of the checks to see when you hit subBufferCapacity.
2. Right now if two threads race to increment entryNumber, two threads may be trying to write a tuple in compareOrSet at the same time.  Since they are doing a component by component COMP_SWAP, I don't think this is a correctness issue, but there's probably a chance to simplify the code a bit with knowledge of who won the COMP_SWAP here (though it's not totally trivial, since if the winner starts writing an entry, the loser can't deduplicate against the entry until the winner is done writing...).

pxr/imaging/hgiMetal/shaderGenerator.mm
```diff
@@ -303,9 +303,9 @@ void _Init(
         " atomic_fetch_add_explicit(&a, v, memory_order_relaxed)\n"
         "#define ATOMIC_EXCHANGE(a, desired)"
         " atomic_exchange_explicit(&a, desired, memory_order_relaxed)\n"
-        "#define ATOMIC_COMP_SWAP(a, expected, desired) \\"
-        "  atomic_compare_exchange_strong_explicit(&a, &expected, desired, "
-        "memory_order_relaxed, memory_order_relaxed)\n"
+        "#define ATOMIC_COMP_SWAP(a, expected, desired)"
+        "  atomic_compare_exchange_weak_explicit(&a, &expected, desired, "
+        "memory_order_relaxed, memory_order_relaxed) ? expected : expected;\n"
```
tcauchois, line 308: Thanks for fixing the return value here!

A few issues with this function:
- This predates this PR, but I'm a bit worried about this macro introducing side effects (namely that expected is an out value) that aren't present in GL or Vulkan.
- I'm unclear if the comparable GLSL semantics are weak or strong.  But if weak works like I expect, we need to handle spurious failures in client code and there's a subtle/minor bug in your deep picking shader, which will likely show up in other CAS uses as well.

pxr/imaging/hdx/shaders/renderPass.glslfx
tcauchois: Possibly a bigger issue, as identified in shaderGenerator.mm, if we're using weak CAS I think we need to do something with the return value.  It seems like we're not necessarily using it to control who gets to write to the buffer, but it's at least used in pickTask.cpp in read-back, so if we spuriously fail to update it here there's a small chance that pickTask.cpp will get an incorrect entry count.

tcauchois (commented):
This is awesome work!  Thanks for incorporating the last round of comments!

I left a few questions and I have some specific concerns about metal CAS; any details you can provide would be helpful.  I'm planning to discuss this further with the team, and hopefully we can talk next steps on Thursday.

pxr/imaging/hdx/pickTask.cpp
tcauchois: I missed that there was an EndFrame later in this function, nevermind.

pxr/imaging/hdx/shaders/renderPass.glslfx
tcauchois: And to get a bit deeper into this; I'm not up-to-date on how expensive an ATOMIC_LOAD will be if nobody's writing to the SSBO (because we're not deep picking), but if this might be a performance concern we could replace it with one of the following:

1. Removing the "pickBuffer" binding when we're not deep picking, which should be pretty fast.
2. Adding a uniform to tell us whether to run the deep picking code.

Would appreciate any information about whether you expect/have seen this to be a performance concern or a non-issue.

pxr/imaging/hgiMetal/shaderGenerator.mm
tcauchois: Note: GLSL has weak semantics too; so there's still some care needed, and I think there's still a latent bug in your usage of it, but as far as this file is concerned the only issue is the side effect.  Maybe we can make a scoped temporary copy of expected?

pxr/imaging/hdx/shaders/renderPass.glslfx
tcauchois: Got around to running my own timings and the ATOMIC_LOAD is a non-issue, at least on the hardware I have access to, so I think it's fine until someone comes up with a hardware configuration where it's problematic.

pxr/imaging/hdx/shaders/renderPass.glslfx
tcauchois: Note: I think this code is correct as long as we can guarantee strong semantics on ATOMIC_COMP_SWAP, which is true of GL.

pxr/imaging/hgiMetal/shaderGenerator.mm
tcauchois: Having a very hard time pinning this down, but based on specs and other projects cross-compiling GLSL to MSL, I think GLSL CAS is strong, and MSL is weak.  I have an alternate implementation of the macro for Metal that I'm going to test, to give it strong semantics and no side effects.

spitzak, line 308: Is `x() ? expected : expected` a typo? If the intention is to always return `expected` then I think `(x(), expected)` might work as well and be clearer.

PR #3051, opened 2024-04-22
Autodesk: Prevent undesired materialx garbage collection

erikaharrison-adsk:
### Description of Change(s)

Summary: Add a member variable for HdMaterialNetwork to hold the generated MaterialX Shader to prevent unnecessary garbage collection.

With the current code base, every time we re-apply an existing material, the MaterialX::ShaderPtr is still considered as "the first instance"

https://github.com/PixarAnimationStudios/OpenUSD/blob/0b18ad3f840c24eb25e16b795a5b0821cf05126e/pxr/imaging/hdSt/materialXFilter.cpp#L1008

even the underlying HdMaterialNetwork is cached on application side. The reason is that after every drawcall, there is a garbage collection process (https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/imaging/hdSt/resourceRegistry.cpp#L1098).
The previously unapplied material (MaterialX::ShaderPtr) is garbage collected because HdStorm considered that nobody needs it anymore even though, as I said, the HdMaterialNetwork is cached. This is not the case for PreviewSurace or any other custom glslfx because there is member variable called _surfaceGfx which holds a reference of the generated glslfx thus the reference count is at least 2 and the ResourceRegistry won't garbage collect it.

To address this issue, we can simply add a member variable in HdMaterialNetwork to hold the generated MaterialX glslfx shared_ptr just like the _surfaceGfx to prevent the unnecessary garbage collection for MaterialX shader as well.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/materialNetwork.cpp
+++ b/pxr/imaging/hdSt/materialNetwork.cpp
@@ -1107,7 +1107,7 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
 
 #ifdef PXR_MATERIALX_SUPPORT_ENABLED
         if (!isVolume) {
-            HdSt_ApplyMaterialXFilter(&surfaceNetwork, materialId,
+            _materialXGfx = HdSt_ApplyMaterialXFilter(&surfaceNetwork, materialId,
                                       *surfTerminal, surfTerminalPath,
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9573

PR #3052, opened 2024-04-23
Apple: Core visionOS Support

dgovil:
### Description of Change(s)

This PR builds on the [Core iOS support PR](https://github.com/PixarAnimationStudios/OpenUSD/pull/2949) to add support for building the USD core for visionOS.

Thanks to Valentin Roussellet for the basis of the TBB patch.

It includes the following changes:

* Use CMake 3.28 (which adds visionOS support)
* Patch TBB to support visionOS. Note that clang requires new platforms to be provided via the target triplet.
* Add visionOS to the accepted platforms.

Note that  visionOS's SDK is called xros, much like how the iOS SDK is iphoneos for historical reasons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ then build and install USD into `/path/to/my_usd_install_dir`.
 > python OpenUSD/build_scripts/build_usd.py /path/to/my_usd_install_dir
 ```
 
-##### MacOS:
+##### macOS:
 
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9587

dgovil:
Rebased on top of latest dev/ios_core

PR #3053, opened 2024-04-24
Support `string_view` initialization and lookup in `TfToken`

nvmkuruc:
### Description of Change(s)
`TfToken` is designed to optimize storage and comparison of commonly used strings like property names, values, and other path components. It currently provides two constructors, one for NULL-terminated C-strings and the other for `std::string`.

There are cases where a token may need to be constructed from a part of a `std::string`. This often necessitates an intermediate heap-allocated copy when no allocation or copy is needed at all if the token already exists in the registry.  The path parser currently optimizes this case by introducing a stack allocated `char[32]` and copying and NULL-terminating strings short enough to fit in the buffer.  However, this has not been generalized outside of the path parser, is limited to short strings, and still requires a copy.

This change introduces a `std::string_view` constructor and associated utilities. It benefits the current implementation by unifying the C-string and `std::string` code paths, avoiding templating and specialization.

It also replaces `Find(std::string const&)` with the more general `Find(std::string_view)`.

A note is left in the `TfToken::_Rep` class about using a set that supports transparent hashing.  This would allow lookups to transparently compare `std::string_view` against a `TfToken::_Rep` without `_Rep` storing an extra `std::string_view` field.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/tf/token.h
+++ b/pxr/base/tf/token.h
@@ -23,6 +23,7 @@
 #include <atomic>
 #include <iosfwd>
 #include <string>
+#include <string_view>
 #include <vector>
 #include <set>
 
@@ -132,11 +133,24 @@ public:
     // and release their memory.
     TF_API TfToken(char const* s, _ImmortalTag);
 
+    /// Acquire a token for the given string_view.
+    //
+    // This constructor involves a string hash and a lookup in the global
+    // table, and so should not be done more often than necessary.  When
+    // possible, create a token once and reuse it many times.
+    TF_API explicit TfToken(std::string_view sv);
+    /// \overload
+    // Create a token for \p sv, and make it immortal.  If \p sv exists in the
+    // token table already and is not immortal, make it immortal.  Immortal
+    // tokens are faster to copy than mortal tokens, but they will never expire
+    // and release their memory.
+    TF_API TfToken(std::string_view sv, _ImmortalTag);
+
     /// Find the token for the given string, if one exists.
     //
     // If a token has previous been created for the given string, this
     // will return it.  Otherwise, the empty token will be returned.
-    TF_API static TfToken Find(std::string const& s);
+    TF_API static TfToken Find(std::string_view sv);
 
     /// Return a size_t hash for this token.
     //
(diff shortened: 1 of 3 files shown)


pxr/base/tf/token.h
```diff
@@ -149,11 +150,24 @@ class TfToken
     // and release their memory.
     TF_API TfToken(char const* s, _ImmortalTag);
 
+    /// Acquire a token for the given string.
```
gitamohr: "string" -> "string_view"

pxr/base/tf/token.h
```diff
@@ -149,11 +150,24 @@ class TfToken
...
+    //
+    // This constructor involves a string hash and a lookup in the global
+    // table, and so should not be done more often than necessary.  When
+    // possible, create a token once and reuse it many times.
+    TF_API explicit TfToken(std::string_view const sv);
+    /// \overload
+    // Create a token for \p s, and make it immortal.  If \p s exists in the
```
gitamohr: "s" -> "sv"

pxr/base/tf/token.h
```diff
@@ -149,11 +150,24 @@ class TfToken
...
+    // token table already and is not immortal, make it immortal.  Immortal
+    // tokens are faster to copy than mortal tokens, but they will never expire
+    // and release their memory.
+    TF_API TfToken(std::string_view const sv, _ImmortalTag);
```
gitamohr: `const` here has no meaning, can you pls remove it?

pxr/base/tf/token.h
```diff
@@ -149,11 +150,24 @@ class TfToken
...
+
     /// Find the token for the given string, if one exists.
     //
     // If a token has previous been created for the given string, this
     // will return it.  Otherwise, the empty token will be returned.
-    TF_API static TfToken Find(std::string const& s);
+    TF_API static TfToken Find(std::string_view const sv);
```
gitamohr: Same here -- meaningless `const`.

pxr/base/tf/token.h
```diff
@@ -390,14 +405,14 @@ class TfToken
             , _setNum(rhs._setNum)
             , _compareCode(rhs._compareCode)
             , _str(rhs._str)
-            , _cstr(rhs._str.c_str() == rhs._cstr ? _str.c_str() : rhs._cstr) {}
+            , _view(std::string_view{rhs._str} == rhs._view ? std::string_view{_str} : rhs._view) {}
```
gitamohr, line 391: The comment above needs an update as there's no more `_cstr` member.

jesschimein:
Filed as internal issue #USD-9588

Denise-Yang:
Closing this PR out because this change causes `_Rep` to increase in size, and when we upgrade to C++23, we will be able to address the issue that this PR brings up without needing to bloat `_Rep`.

PR #3061, opened 2024-04-26
Add utilities for processing ies light files

pmolodo:
### Description of Change(s)

This adds ies.h and ies.cpp, for processing ies light files.

These are originally taken from blender, v4.1.1:

- https://github.com/blender/blender/blob/v4.1.1/intern/cycles/util/ies.h
- https://github.com/blender/blender/blob/v4.1.1/intern/cycles/util/ies.cpp

These files are provided by Blender under the [Apache 2.0](http://opensource.org/licenses/Apache-2.0) license.  (See the note [here](https://www.blender.org/about/license/) about how the "Blender Cycles rendering engine is available as Apache 2.0").

These are not yet built / used, but merely introduce the files for use in the codebase later (ie, as part of a reworked [UsdLux update](https://github.com/PixarAnimationStudios/OpenUSD/pull/2758)).

The intent is to allow the Pixar legal approval process to get started early.

### DIFF
--- a/pxr/imaging/plugin/hdEmbree/pxrIES/README.md
+++ b/pxr/imaging/plugin/hdEmbree/pxrIES/README.md
@@ -1,0 +1,31 @@
+# IES utilities
+
+Utilities for reading and using .ies files (IESNA LM-63 Format), which are used
+to describe lights.
+
+The files `ies.h` and `ies.cpp` are originally from
+[Cycles](https://docs.blender.org/manual/en/latest/render/cycles/index.html),
+which is part of the larger
+[Blender](https://projects.blender.org/blender/blender/) project, though with
+a the less restrictive Apache 2.0 license - see:
+
+- https://www.blender.org/about/license/
+
+## Version
+
+v4.1.1 ( 53c49589f311bd3c9e3ea8f62b3fa8fe8e5d2c8c )
+
+## Setup
+
+When updating IES, the following steps should be followed:
+
+1. Copy `intern/cycles/util/ies.h` and `intern/cycles/util/ies.cpp` over the
+   copies in `pxr/imaging/plugin/hdEmbree/pxrIES`.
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9596

pmolodo:
Updated the way that these files are vendored, to match how pxrCLI11 works - ie, with a patch file.

pmolodo:
This PR doesn't actually build/use ies.h or ies.cpp yet, as we don't want to slow down / alter the actual build until this code is used - this is to get the legal process started - but look at this pr/ies_build branch for one that will build ies.cpp, to validate it at least compiles:

https://github.com/pmolodo/OpenUSD/commits/pr/ies_build/

pmolodo:
If it helps to get approval from legal - I've since learned that the exact same source code is also hosted in a separate ["cycles" repository](https://projects.blender.org/blender/cycles/).  The two files in question are the exact same (which I confirmed), and both explicitly say in their headers that they're covered by Apache 2.0:

```
* SPDX-License-Identifier: Apache-2.0 */
```

However, the cycles repo has the additional nice feature that the license listed at the root of the project is the Apache 2.0 one.  
I can update this PR to reference the cycles repo, instead of the blender one, if it would be preferrable.

Relevant Links:
- Repo: https://projects.blender.org/blender/cycles/
- v4.1.1: https://projects.blender.org/blender/cycles/src/tag/v4.1.1
- ies.h: https://projects.blender.org/blender/cycles/src/tag/v4.1.1/src/util/ies.h
- ies.cpp: https://projects.blender.org/blender/cycles/src/tag/v4.1.1/src/util/ies.cpp
- LICENSE: https://projects.blender.org/blender/cycles/src/branch/main/LICENSE

pmolodo:
Closing in favor of:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3192

Issue #3060, opened 2024-04-26
UsdUtils.ModifyAssetPaths strips empty paths from asset array attributes


marktucker:
### Description of Issue
UsdUtils.ModifyAssetPaths strips empty asset paths from asset array attributes. This makes sense for composition arcs, but for attributes/primvars, this kind of arbitrary manipulation of array contents (and lengths!) is likely to break things. This change appears to be new, probably cause when ModifyAssetPaths moved over to the AssetLocalization framework.

### Steps to Reproduce
1. Extract the attached asset_array.usda: 
[asset_array.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/15132812/asset_array.zip)
2. Run the following script in python:
```
def cb(asset):
    return asset

from pxr import Usd, UsdUtils, Sdf
l = Sdf.Layer.FindOrOpen("asset_array.usda")
print(l.ExportToString())
UsdUtils.ModifyAssetPaths(l, cb)
print(l.ExportToString())
```
4. The result should be completely unchanged (since the callback doesn't change any asset paths), but instead the empty asset paths have been stripped from the asset array.

### Package Versions
24.03

jesschimein:
Filed as internal issue #USD-9595

spiffmon:
Thanks, @marktucker - agreed and we'll try to get this addressed for the next release.

marktucker:
Thanks. I have a PR that I just need to do a little more testing on before I submit it for your approval (hopefully).

PR #3063, opened 2024-04-29
Don't let UsdUtilsModifyAssetPaths change the length of asset array attributes/metadata

marktucker:
### Description of Change(s)
When modifying asset paths in a layer, don't remove empty elements from arrays. The empty elements may be meaningful, as in primvars that must be a specific length. Updated the output of the ModifyAssetPaths "removal" test, which validates the behavior of this API when modified asset paths are empty.

Arguably, this change should perhaps be limited to USD attributes (not metadata), but depending on the use case changing the array length is still throwing away what may be important information in a pipeline.

I also considered whether an array that has all values set to empty asset paths should remove the attribute. This seems more defensible to me, but again, I'm not sure the length of the array couldn't still be considered useful information that is worth preserving in some circumstances.

I'm open to different answers on these two questions, and I believe each could be addressed without too drastic changes to this PR.

### Fixes Issue(s)
- #3060 

### DIFF
--- a/pxr/usd/usdUtils/assetLocalization.cpp
+++ b/pxr/usd/usdUtils/assetLocalization.cpp
@@ -505,10 +505,19 @@ UsdUtils_LocalizationContext::_ProcessAssetValue(
 
                 const std::vector<std::string> processedDeps = 
                     _delegate->ProcessValuePathArrayElement(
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9598

Issue #457, opened 2018-04-10
build_usd.py: potential file name collisions


jcowles:
### Description of Issue
build_usd.py downloads several zip files, many of which are not prefixed with a library name. Here is a sample of what my $build/src directory contains:

**1.5.1.zip                   
1.7.1.zip                   
v2.2.0.zip                  
v3_1_1.zip**                 
           
These file names may one day collide, but more immediately, it makes it harder to determine which library each file contains.

### System Information (OS, Hardware)
Windows

### Package Versions
USD v0.8.4

jtran56:
Filed as internal issue #159450.

Cewein:
The possible collision is still here with https://github.com/PixarAnimationStudios/OpenUSD/blob/2864f3d04f396432f22ec5d6928fc37d34bb4c90/build_scripts/build_usd.py#L543

I'll will look into it. For gitlab and github url it should be ok.

For example this could work :
```py
url.split("/")[4] + "-" + url.split("/")[-1]
```
it would give for this type of url : `https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2020.3.zip`
This result : `oneTBB-v2020.3.zip`

PR #3065, opened 2024-05-01
Changed the way we name downloaded zip files in `build_usd.py` to avoid file name collisions.

Cewein:
### Description of Change(s)

Changed the way we name downloaded zip files in `build_usd.py` to minimise the chance of file collision. Files from a GitHub or GitLab repository previously only had their version numbers as their names. Now, they will incorporate the project name and the version.

I assume other URLs provide this (i.e., Boost and HDF5).

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/457

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -49,6 +49,7 @@ import sys
 import sysconfig
 import zipfile
 
+from urllib.parse import urlparse
 from urllib.request import urlopen
 from shutil import which
 
@@ -539,13 +540,16 @@ def DownloadURL(url, context, force, extractDir = None,
     Returns the absolute path to the directory where files have 
     been extracted."""
     with CurrentWorkingDirectory(context.srcDir):
-        # Extract filename from URL.
-        filename = url.split("/")[-1]
+        # Parse the information of the url
+        pURL = urlparse(url)
 
-        # Change filename for a file form a github/gitlab repository. 
+        # Extract filename from URL path.
+        filename = pURL.path.split("/")[-1]
+
+        # Change filename for a file form a github repository. 
         # Filename becomes projectName-version to avoid file collision.
-        if ("github" or "gitlab") in url:
-            filename = url.split("/")[4] + "-" + filename
+        if pURL.netloc == "github.com":
+            filename = pURL.path.split("/")[2] + "-" + filename
 
         # See if file already exists. 
         if force and os.path.exists(filename):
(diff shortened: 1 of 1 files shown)


build_scripts/build_usd.py
```diff
@@ -539,8 +539,15 @@ def DownloadURL(url, context, force, extractDir = None,
...
     been extracted."""
     with CurrentWorkingDirectory(context.srcDir):
-        # Extract filename from URL and see if file already exists. 
-        filename = url.split("/")[-1]       
+        # Extract filename from URL.
+        filename = url.split("/")[-1]
+
+        # Change filename for a file form a github/gitlab repository. 
+        # Filename becomes projectName-version to avoid file collision.
+        if ("github" or "gitlab") in url:
```
nvmkuruc: The expression `"github" or "gitlab"` will always evaluate to `"github"` so it's not actually checking for the presence of `gitlab` in the URL.

The only usage of `gitlab` in `build_usd.py` points to a named release as well, so you can probably just drop the check for `gitlab`.
```
> grep -r gitlab build_scripts/
build_scripts/build_usd.py:TIFF_URL = "https://gitlab.com/libtiff/libtiff/-/archive/v4.0.7/libtiff-v4.0.7.zip"
```

build_scripts/build_usd.py
```diff
@@ -539,8 +539,15 @@ def DownloadURL(url, context, force, extractDir = None,
     Returns the absolute path to the directory where files have 
     been extracted."""
     with CurrentWorkingDirectory(context.srcDir):
-        # Extract filename from URL and see if file already exists. 
-        filename = url.split("/")[-1]       
+        # Extract filename from URL.
+        filename = url.split("/")[-1]
```
nvmkuruc: I'd consider if it would be better to use `urllib.parse.urlparse` to breakdown the `url` into its components (https://docs.python.org/3/library/urllib.parse.html).  That should be more robust than splitting on `/`.

It might also make things more robust to just use the full `path` component of the URL when creating a target to download the files rather than trying to decompose based on the naming conventions of particular repositories.  I'm not sure if there's any drawbacks to that.

build_scripts/build_usd.py
Cewein: Thank you for pointing this out. I checked the GitLab too quickly, since the output name contains the name of the project. I looked into multiple GitLab repositories online, they always come with the project name and the check can be removed.

build_scripts/build_usd.py
Cewein: The only drawback that I can see with the name as the path is if the filename become too long, it might not save properly.

Using `urllib.parse.urlparse` is a good idea though.

jesschimein:
Filed as internal issue #USD-9603

PR #3070, opened 2024-05-06
Fix resource leak in hdSt 

dgovil:
### Description of Change(s)

Garbage collect the texture handle registry and resource registry to prevent memory leaks.

Submitting on behalf of Maddy Adams

### DIFF
--- a/pxr/imaging/hdSt/renderBuffer.cpp
+++ b/pxr/imaging/hdSt/renderBuffer.cpp
@@ -61,7 +61,9 @@ HdStRenderBuffer::HdStRenderBuffer(
 {
 }
 
-HdStRenderBuffer::~HdStRenderBuffer() = default;
+HdStRenderBuffer::~HdStRenderBuffer() {
+    _resourceRegistry->GarbageCollect();
+}
 
 void
 HdStRenderBuffer::Sync(HdSceneDelegate *sceneDelegate,
--- a/pxr/imaging/hdSt/resourceRegistry.cpp
+++ b/pxr/imaging/hdSt/resourceRegistry.cpp
@@ -1135,6 +1135,8 @@ HdStResourceRegistry::_GarbageCollect()
     _uniformSsboBufferArrayRegistry.GarbageCollect();
     _singleBufferArrayRegistry.GarbageCollect();
 
+    _textureHandleRegistry->GetTextureObjectRegistry()->GarbageCollect();
+
     // Garbage collection may reallocate buffers, so we must submit blit work.
     SubmitBlitWork();
 }


jesschimein:
Filed as internal issue #USD-9643

pxr/imaging/hdSt/renderBuffer.cpp
```diff
@@ -61,7 +61,9 @@ HdStRenderBuffer::HdStRenderBuffer(
 {
 }
 
-HdStRenderBuffer::~HdStRenderBuffer() = default;
+HdStRenderBuffer::~HdStRenderBuffer() {
+    _resourceRegistry->GarbageCollect();
```
tcauchois, line 65: Hi!  Can you expand on the intent of this line?  Is there a benefit to invoking this here, as opposed to ~HdStRenderDelegate or something?  Thanks!

dgovil: @tcauchois Just copy pasting on Maddy's behalf

> I tested the suggested change (moving the call to _resourceRegistry->GarbageCollect(); from HdStRenderBuffer::~HdStRenderBuffer to HdStRenderDelegate::~HdStRenderDelegate). ctest -C Release still shows all tests passing, and my simple repro for the memory leak doesn't show leaking with it. 

> When debugging the leak, I put the garbage collection in HdStRenderBuffer, because that's what Instruments and stack traces were pointing me to. (HgiMetalTexture objects were being leaked because they were owned by HdStUvTextureObjects, HdStUvTextureObjects were being leaked because they were created by HdStRenderBuffer and kept alive in the HdSt_TextureObjectRegistry even after HdStRenderBuffer dropped its shared pointers to _textureObject and _textureMSAAObject in its default destructor.) To me, that points to HdStRenderBuffer being responsible for the leaked memory, including cleaning it up. But I'm not familiar with the architecture of Usd and imaging, so if people find that it makes more sense for the clean-up to be in the HdStRenderDelegate instead, that's fine.

Hopefully that makes sense, but I can ping Maddy for more.

Let me know what you think and I can adjust the PR accordingly, or however you want to handle it

unhyperbolic:
I am missing some context about how this memory leak came about and what the repro steps are.

Looking at this again, I noticed that HdStResourceRegistry::GarbageCollect doesn't call GarbageCollect on the HdSt_TextureHandleRegistry. That seems a problem.

My suggested fix is this: Rename HdSt_TextureHandleRegistry::_GarbageCollectAndComputeTargetMemory to GarbageCollect and make it public. Call it from HdStResourceRegistry::GarbageCollect. HdStResourceRegistry::~HdStResourceRegistry calls GarbageCollect already which should be called from HdStRenderDelegate::~HdStRenderDelegate when dropping the shared_ptr to the resource registry.

That should fix the memory leak under the following assumptions:
1. We fully depopulate the render delegate so that no materials or render buffers are holding on to the textures.
2. No one but the render delegate is holding on to a shared ptr to the resource registry. HdStRenderDelegate::GetResourceRegistry should arguably return a raw pointer.
3. The Hgi instance is alive when we destruct the HdStRenderDelegate.

unhyperbolic:
In a couple of days, the change I described above and checked in should appear here. Can you confirm it fixes the leak?

dgovil:
I’ll be out for a couple weeks but I’ll confirm when I’m back.

tcauchois:
@dgovil the commit Matthias is referring to is here: https://github.com/PixarAnimationStudios/OpenUSD/commit/470623ac6484175f79cbc315c44d7568e9d7a6fa ...  Let us know if this fixes the leak!  If so I think we can close the PR out; otherwise we can iterate.

Thanks for finding this!

dgovil:
In my testing the latest patches seemed to have resolved this. I'll reopen if we encounter it again. 

Thank you!

PR #3071, opened 2024-05-06
Autodesk: Add render-selected-edge-from-face global env var

erikaharrison-adsk:
### Description of Change(s)

To improve the performance of using Geometry Shader in HdStorm, expose an environment variable to enable/disable render-selected-edge-from-face feature, which may trigger geometry shader.

In HdStorm, when rendering mesh using HdMeshGeomStyleSurf/HdMeshGeomStyleHull geomStyle, the Geometry Shader might be enabled to support the render-selected-edge-from-face feature, this could bring performance overhead if the application doesn't need that feature, so, adding an environment variable to enable/disable render-selected-edge-from-face feature in HdStorm,

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/meshShaderKey.cpp
+++ b/pxr/imaging/hdSt/meshShaderKey.cpp
@@ -25,6 +25,7 @@
 
 #include "pxr/imaging/hd/mesh.h"
 #include "pxr/imaging/hdSt/meshShaderKey.h"
+#include "pxr/base/tf/getenv.h"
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9644

tcauchois:
@erikaharrison-adsk we believe that David's commit above addresses the original issue, so we're going to close this out.  Please let us know if you have further issues here!

Issue #3004, opened 2024-03-16
Storm shader compilation errors when using MaterialX triplanarprojection node.


ld-kerley:
### Description of Issue
When using the MaterialX `<triplanarprojection>` in a material, then Storm reports shader compilation errors.
```
Warning: in _ValidateCompilation at line 228 of usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:3197:35: error: use of undeclared identifier 'triplanar_filex'
    NG_triplanarprojection_color3(triplanar_filex, triplanar_filey, triplanar_filez, triplanar_layerx, triplanar_layery, triplanar_layerz, triplanar_default, geomprop_Pobject_out1, geomprop_Nobject_out1, triplanar_upaxis, triplanar_blend, triplanar_filtertype, triplanar_framerange, triplanar_frameoffset, triplanar_frameendaction, triplanar_out);
                                  ^
program_source:3197:52: error: use of undeclared identifier 'triplanar_filey'
    NG_triplanarprojection_color3(triplanar_filex, triplanar_filey, triplanar_filez, triplanar_layerx, triplanar_layery, triplanar_layerz, triplanar_default, geomprop_Pobject_out1, geomprop_Nobject_out1, triplanar_upaxis, triplanar_blend, triplanar_filtertype, triplanar_framerange, triplanar_frameoffset, triplanar_frameendaction, triplanar_out);
                                                   ^
program_source:3197:69: error: use of undeclared identifier 'triplanar_filez'
    NG_triplanarprojection_color3(triplanar_filex, triplanar_filey, triplanar_filez, triplanar_layerx, triplanar_layery, triplanar_layerz, triplanar_default, geomprop_Pobject_out1, geomprop_Nobject_out1, triplanar_upaxis, triplanar_blend, triplanar_filtertype, triplanar_framerange, triplanar_frameoffset, triplanar_frameendaction, triplanar_out);
                                                                    ^
```

### Steps to Reproduce
1. `usdview triplanar.usda` (from [attached zip file](https://github.com/PixarAnimationStudios/OpenUSD/files/14621745/triplanar_bug.zip))

Loading the same MaterialX file in to MaterialXView displays the material as expected.
`MaterialXView --material triplanar.mtlx` renders

[...]

Issue #2330, opened 2023-03-13
MaterialX: Storm fails to reuse GLSL shaders for networks differing only on node names


JGamache-autodesk:
### Description of Issue

The function `HdSt_MaterialNetworkShader::CanAggregate()` will compare the hash of the GLSL shader text to know if it can aggregate two material networks.

MaterialX will use the node names when generating a uniform variable name when there are connected ancillary nodes. If these node names vary, the hash will differ and the merging will fail.

### Steps to Reproduce
1. Load attached scene in USDView [goldRamp.zip](https://github.com/PixarAnimationStudios/USD/files/10961755/goldRamp.zip)
[BUG] compiles 11 GLSL shaders where only one would be needed

The scene was designed to show the issue by having a `ND_dot_float` with a varying name connected in exactly the same way on each surface.

### Potential fix

One option would be to anonymize the node names in the network so they follow a consistent scheme. This will harmonize the uniform parameter names and allow more material networks to reuse the same shader.

sunyab:
Filed as internal issue #USD-8091

PR #3069, opened 2024-05-05
Implement MaterialX material tag detection for glTF PBR

pablode:
### Description of Change(s)
As described in the linked issue, MaterialX materials using the `<gltf_pbr>` with alpha blending currently render incorrectly:
AlphaBlendModeTest_before

This PR extends the detection heuristic to cover the inputs of the glTF PBR. The alpha blending is smooth:
AlphaBlendModeTest_after

Test asset: [AlphaBlendModeTest_mtlx.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/15214541/AlphaBlendModeTest_mtlx.zip)

### Fixes Issue(s)
#1882 

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement

PR #3073, opened 2024-05-06
Autodesk: Optimize MaterialX shader generation in Storm by comparing topologies

erikaharrison-adsk:
### Description of Change(s)

Fixes:

- MaterialX: Storm fails to reuse GLSL shaders for networks differing only on node names #2330
- Storm shader compilation errors when using MaterialX triplanar projection node. #3004
- Improve transparency detection and fix generation

Using a topo-trimmed network resulted in all shaders being generated with opaque semantics. The shaders are now created with the right transparent hardware settings.

Also improved transparency detection at the `_GetMaterialTag` level by adding more checks for known surfaces and delegating to MaterialX when dealing with custom shaders, but without generating the full MaterialX document.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2330
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3004

### DIFF
--- a/pxr/imaging/hdSt/materialXFilter.cpp
+++ b/pxr/imaging/hdSt/materialXFilter.cpp
@@ -29,6 +29,7 @@
 #include "pxr/imaging/hdMtlx/hdMtlx.h"
 #include "pxr/imaging/hgi/tokens.h"
 
+#include "pxr/usd/sdf/schema.h"
 #include "pxr/usd/sdr/registry.h"
 #include "pxr/imaging/hio/glslfx.h"
 #include "pxr/imaging/hgi/capabilities.h"
@@ -48,7 +49,6 @@ namespace mx = MaterialX;
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-
 TF_DEFINE_PRIVATE_TOKENS(
     _tokens,
     (mtlx)
@@ -61,9 +61,25 @@ TF_DEFINE_PRIVATE_TOKENS(
     ((defaultInput, "default"))
 
     // Opacity Parameters
+    (UsdPreviewSurface)
     (opacity)
     (opacityThreshold)
+
+    (standard_surface)
     (transmission)
+
+    (open_pbr_surface)
+    (transmission_weight)
+    (geometry_opacity)
+
+    (gltf_pbr)
+    (alpha_mode)
+    (alpha_cutoff)
+    (alpha)
+
+    (convert)
+    (ND_convert_color4_surfaceshader)
+    (ND_convert_vector4_surfaceshader)
 
     // Fallback Dome Light Tokens
     (domeLightFallback)
@@ -71,10 +87,48 @@ TF_DEFINE_PRIVATE_TOKENS(
     (file)
 
     // Colorspace Tokens
-    (colorSpace)
     (sourceColorSpace)
+
+    // Anonymization constants
+    (NG_Anonymized)
+
+    // Primvar detection constants
+    (geompropvalue)
 );
 
+TF_DEFINE_PRIVATE_TOKENS(
+    _topologicalTokens,
+    // This represents living knowledge of the internals of the MaterialX shader generator
+    // for both GLSL and Metal. Such knowledge should reside inside the generator class
+    // provided by MaterialX.
+
+    // Dot filename is always topological due to code that prevents creating extra OpenGL samplers
+    // this is the only shader node id required. All other tests are done on the shader family.
+    (ND_dot_filename)
+    // Topo affecting nodes due to object/model/world space parameter
+    (position)
+    (normal)
+    (tangent)
+    (bitangent)
+    // Topo affecting nodes due to channel index.
+    (texcoord)
+    (geomcolor)
+    // Geompropvalue primvar name is topo-affecting.
+    (geompropvalue)
+    // Swizzles are inlined into the codegen and affect topology.
+    (swizzle)
+    // Some conversion nodes are implemented by codegen.
+    (convert)
+    // Constants: they get inlined in the source.
+    (constant)
+);
+
+TF_DEFINE_PRIVATE_TOKENS(
+    _textureParamTokens,
+    (filtertype)
+    (uaddressmode)
+    (vaddressmode)
+);
 
 ////////////////////////////////////////////////////////////////////////////////
 // Shader Gen Functions
@@ -82,10 +136,11 @@ TF_DEFINE_PRIVATE_TOKENS(
 // Generate the Glsl Pixel Shader based on the given mxContext and mxElement
 // Based on MaterialXViewer Material::generateShader()
 static mx::ShaderPtr
-_GenMaterialXShader(mx::GenContext & mxContext, mx::ElementPtr const& mxElem)
+_GenMaterialXShader(
+    mx::GenContext & mxContext,
+    mx::ElementPtr const& mxElem)
 {
-    bool hasTransparency = mx::isTransparentSurface(mxElem,
-                               mxContext.getShaderGenerator().getTarget());
+    bool hasTransparency = mxContext.getOptions().hwTransparency;
 
     mx::GenContext materialContext = mxContext;
     materialContext.getOptions().hwTransparency = hasTransparency;
@@ -135,6 +190,135 @@ _CreateHdStMaterialXContext(
     }
 }
 
+using HdMtlxNodePathMap = std::unordered_map<SdfPath, SdfPath, SdfPath::Hash>;
+
+static bool
+_IsTopologicalShader(TfToken const& nodeId)
+{
+    static const TfToken::HashSet topologicalTokenSet(
+        _topologicalTokens->allTokens.begin(),
+        _topologicalTokens->allTokens.end());
+
+    if (nodeId == _topologicalTokens->ND_dot_filename) {
+        return true;
+    }
+
+    SdrRegistry &sdrRegistry = SdrRegistry::GetInstance();
+    const SdrShaderNodeConstPtr sdrNode = 
+        sdrRegistry.GetShaderNodeByIdentifierAndType(nodeId, _tokens->mtlx);
+
+    return sdrNode && topologicalTokenSet.count(sdrNode->GetFamily()) > 0;
+}
+
+size_t _BuildEquivalentMaterialNetwork(
+    HdMaterialNetwork2 const& hdNetwork,
+    HdMaterialNetwork2* topoNetwork,
+    HdMtlxNodePathMap* nodePathMap)
+{
+    // The goal here is to strip all local names in the network paths in order to produce MaterialX
+    // data that do not have uniform parameter names that vary based on USD node names.
+
+    // We also want to strip all non-topological parameters in order to get a shader that has
+    // default values for all parameters and can be re-used.
+
+    size_t nodeCounter = 0;
+    nodePathMap->clear();
+
+    // Paths will go /NG_Anonymized/N0, /NG_Anonymized/N1, /NG_Anonymized/N2...
+    SdfPath ngBase(_tokens->NG_Anonymized);
+
+    // We will traverse the network in a depth-first traversal starting at the
+    // terminals. This will allow a stable traversal that will not be affected
+    // by the ordering of the SdfPaths and make sure we assign the same index to
+    // all nodes regardless of the way they are sorted in the network node map.
+    std::vector<const SdfPath*> pathsToTraverse;
+    for (const auto& terminal : hdNetwork.terminals) {
+        const auto& connection = terminal.second;
+        pathsToTraverse.push_back(&(connection.upstreamNode));
+    }
+    while (!pathsToTraverse.empty()) {
+        const SdfPath* path = pathsToTraverse.back();
+        pathsToTraverse.pop_back();
+        if (!(*nodePathMap).count(*path)) {
+            const HdMaterialNode2& node = hdNetwork.nodes.find(*path)->second;
+            // We only need to create the anonymized name at this time:
+            (*nodePathMap)[*path] = ngBase.AppendChild(TfToken("N" + std::to_string(nodeCounter++)));
+            for (const auto& input : node.inputConnections) {
+                for (const auto& connection : input.second) {
+                    pathsToTraverse.push_back(&(connection.upstreamNode));
+                }
+            }
+        }
+    }
+
+    // Copy the incoming network using only the anonymized names:
+    topoNetwork->primvars = hdNetwork.primvars;
+    for (const auto& terminal : hdNetwork.terminals) {
+        topoNetwork->terminals.emplace(
+            terminal.first,
+            HdMaterialConnection2 { (*nodePathMap)[terminal.second.upstreamNode],
+                                    terminal.second.upstreamOutputName });
+    }
+    for (const auto& nodePair : hdNetwork.nodes) {
+        const HdMaterialNode2& inNode = nodePair.second;
+        HdMaterialNode2        outNode;
+        outNode.nodeTypeId = inNode.nodeTypeId;
+        if (_IsTopologicalShader(inNode.nodeTypeId)) {
+            // Topological nodes have parameters that affect topology. We can not strip
+            // them.
+            outNode.parameters = inNode.parameters;
+        } else {
+            // Parameters that are color managed are also topological as they
+            // result in different nodes being added in the MaterialX graph
+            for (const auto& param: inNode.parameters) {
+                const auto colorManagedFile = 
+                    SdfPath::StripPrefixNamespace(param.first.GetString(),
+                                                  SdfFieldKeys->ColorSpace);
+                if (colorManagedFile.second) {
+                    outNode.parameters.insert(param);
+                    // Need an empty asset as well:
+                    outNode.parameters.emplace(TfToken(colorManagedFile.first),
+                                               VtValue{SdfAssetPath{}});
+                }
+            }
+        }
+
+        for (const auto& cnxPair : inNode.inputConnections) {
+            std::vector<HdMaterialConnection2> outCnx;
+            for (const auto& c : cnxPair.second) {
+                outCnx.emplace_back(HdMaterialConnection2 { (*nodePathMap)[c.upstreamNode],
+                                                            c.upstreamOutputName });
+            }
+            outNode.inputConnections.emplace(cnxPair.first, std::move(outCnx));
+        }
+        topoNetwork->nodes.emplace((*nodePathMap)[nodePair.first], std::move(outNode));
+    }
+
+    // Build the topo hash from the topo network:
+    Tf_HashState topoHash;
+    for (const auto& terminal : topoNetwork->terminals) {
+        TfHashAppend(topoHash, terminal.first);
+        TfHashAppend(topoHash, terminal.second.upstreamNode.GetName());
+    }
+    for (const auto& node : topoNetwork->nodes) {
+        TfHashAppend(topoHash, node.first.GetName());
+        TfHashAppend(topoHash, node.second.nodeTypeId);
+        for (const auto& param : node.second.parameters) {
+            TfHashAppend(topoHash, param.first);
+            TfHashAppend(topoHash, param.second.GetHash());
+        }
+        for (const auto& connection : node.second.inputConnections) {
+            TfHashAppend(topoHash, connection.first);
+            for (const auto& source : connection.second) {
+                TfHashAppend(topoHash, source.upstreamNode.GetName());
+                TfHashAppend(topoHash, source.upstreamOutputName);
+            }
+        }
+    }
+
+    return topoHash.GetCode();
+}
+
 // Use the given mxDocument to generate the corresponding glsl shader
 // Based on MaterialXViewer Viewer::loadDocument()
 mx::ShaderPtr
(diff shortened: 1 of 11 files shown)


JGamache-autodesk:
Primary goal was to only reconstruct the MaterialX document if we are really going all the way to compiling the OpenGL shader. This means we prioritize gathering the Storm requirements from:
1- A previously generated MaterialX::Shader
2- The Sdr registry
3- The NodeDef info that can be found in the MaterialX library
4- A partially reconstructed MaterialX document as last resort (for resolving transparency)

To do this, we reduce the HdMaterialNetwork2 to a topology equivalent skeleton network that will produce the exact same shader code. We generate a hash of this skeleton network and use it to fetch a previously compiled MaterialX shader.

With this we can generate the uniform block required by Storm using the default values found in the shader, modified by the values found in the original network. Some extra work was required to make sure the final HdMaterialNetwork had all the requirements for correctly findings the textures to load.

Testing shows that for the scene submitted in #2330 we now only generate a single OpenGL shader. Time to first frame is 10X faster. Also, a scene with a simple animated material (animated `ND_dot_color3` connected to the `diffuse_color` input of a `standard_surface` shader), the frame rate went from 9.3 to 23.7 on my machine since we do not regenerate the MaterialX document at each frame. Could probably be improved by caching the fallback values as a VtValue dictionary.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -61,31 +61,86 @@ TF_DEFINE_PRIVATE_TOKENS(
...
 
     // Opacity Parameters
+    (UsdPreviewSurface)
     (opacity)
     (opacityThreshold)
+
+    (standard_surface)
     (transmission)
 
+    (open_pbr_surface)
```
JGamache-autodesk, line 71: Added programmatic detection for both OpenPBR and glTf surfaces to be able to get the materialTag without having to peek into the MaterialX NodeGraph implementation.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -61,31 +61,86 @@ TF_DEFINE_PRIVATE_TOKENS(
...
+
+    // Primvar detection constants
+    (geompropvalue)
 );
 
+TF_DEFINE_PRIVATE_TOKENS(
+    _topologicalTokens,
+    // This represents living knowledge of the internals of the MaterialX shader generator
+    // for both GLSL and Metal. Such knowledge should reside inside the generator class
+    // provided by MaterialX.
```
JGamache-autodesk, line 103: To be submitted to MaterialX for 1.39.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -61,31 +61,86 @@ TF_DEFINE_PRIVATE_TOKENS(
...
 // Based on MaterialXViewer Material::generateShader()
 static mx::ShaderPtr
-_GenMaterialXShader(mx::GenContext & mxContext, mx::ElementPtr const& mxElem)
+_GenMaterialXShader(
+    mx::GenContext & mxContext,
+    mx::ElementPtr const& mxElem)
 {
-    bool hasTransparency = mx::isTransparentSurface(mxElem,
-                               mxContext.getShaderGenerator().getTarget());
+    bool hasTransparency = mxContext.getOptions().hwTransparency;
```
JGamache-autodesk, line 143: Should have been correctly set in ` mxContext` .

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -135,6 +190,135 @@ _CreateHdStMaterialXContext(
...
+            // result in different nodes being added in the MaterialX graph
+            for (const auto& param: inNode.parameters) {
+                const auto colorManagedFile = 
+                    SdfPath::StripPrefixNamespace(param.first.GetString(),
+                                                  SdfFieldKeys->ColorSpace);
+                if (colorManagedFile.second) {
+                    outNode.parameters.insert(param);
+                    // Need an empty asset as well:
+                    outNode.parameters.emplace(TfToken(colorManagedFile.first),
+                                               VtValue{SdfAssetPath{}});
```
JGamache-autodesk, line 281: Might be incorrect if there are plans to color-correct color3 values on nodes that are not file based.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -431,217 +693,44 @@ _GetHdNodeTypeId(mx::NodePtr const& mxNode)
...
-    for (mx::InputPtr const& input : imageNode->getActiveInputs()) {
-        _GetHdTextureParameters(
-            input->getName(), input->getValueString(), hdParameters);
-    }
-    return true;
-}
-
-// Replace the filenameInput parameter with a connection to a new texture node
-static void
-_ReplaceFilenameInput(
```
JGamache-autodesk, line 457: Moved below at line 1158. Heavily modified to only require the Hydra network.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -690,223 +779,662 @@ _UpdatePrimvarNodes(
...
+    {
+        return true;
+    }
+    // Assume default value is equal to paramValue.
+    return false;
 }
 
 static std::string const&
-_GetMaterialTag(HdMaterialNode2 const& terminal)
+_GetUsdPreviewSurfaceMaterialTag(HdMaterialNode2 const& terminal)
```
JGamache-autodesk, line 807: Using a per nodetype function since generic code does not work all the time. Especially true for UsdPreviewSurface and gfTF since they both support `masked` mode and this prevents automatic detection from properly deducing the material tag.

pxr/imaging/hdSt/materialXShaderGen.cpp
```diff
@@ -508,11 +508,9 @@ HdStMaterialXShaderGen<Base>::_EmitMxVertexDataDeclarations(
...
 
     for (size_t i = 0; i < block.size(); ++i) {
-        line += _EmitMxVertexDataLine(block[i], separator);
-        // remove the separator from the last data line
-        if (i == block.size() - 1) {
-            line = line.substr(0, line.size() - separator.size());
-        }
+        auto const& lineSeparator =
+            (i == block.size() - 1) ? mx::EMPTY_STRING : separator;
+        line += _EmitMxVertexDataLine(block[i], lineSeparator);
```
JGamache-autodesk, line 513: Found while testing some assets. Make sure we do not add the commas in the first place because there might be more than one to remove afterwards.

pxr/imaging/hdSt/materialXShaderGen.cpp
```diff
@@ -548,19 +546,25 @@ HdStMaterialXShaderGen<Base>::_EmitMxVertexDataLine(
...
         
         // Wrap initialization inside #ifdef in case the object does not have 
         // the st primvar
         hdVariableDef = TfStringPrintf("\n"
                 "    #ifdef HD_HAS_%s\n"
-                "        HdGet_%s(),\n"
+                "        HdGet_%s()%s\n"
                 "    #else\n"
-                "        %s(0.0),\n"
+                "        %s(0.0)%s\n"
```
JGamache-autodesk, line 562: This is where the unremovable extra commas were added.

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen/baseline/shadergen_UsdPSdefault.out
```diff
@@ -70,7 +70,6 @@ float displacement;
 float occlusion;
 
 // Uniform block: PrivateUniforms
-float u_alphaThreshold = 0.001000;
```
JGamache-autodesk, line 73: These were incorrectly detected as transparent by MaterialX. The new code that enforces transparency using the materialTag is correct so I had to fix the baseline.

jesschimein:
Filed as internal issue #USD-9646

pxr/imaging/hdSt/materialXFilter.cpp
tcauchois: Ahh, just merged https://github.com/PixarAnimationStudios/OpenUSD/pull/3069 ...  Any quick opinions on how to merge those changes vs these?

pxr/imaging/hdSt/materialXFilter.cpp
tcauchois: Is the intent here to have bool MaterialXIsNodeTopological(std::string name) or something?

tcauchois: We'd definitely be interested in this code moving into MX, for the record :D.

pxr/imaging/hdSt/materialXFilter.cpp
tcauchois: We're planning to, in some form, but I think we can update this bit when we get to it.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -293,6 +480,57 @@ _AddDefaultMtlxTextureValues(std::map<TfToken, VtValue>* hdTextureParams)
...
+                                        mxInput->getValueString(),
+                                        hdTextureParams);
+            }
+        }
+    }
+
+    // Everything boils down to an <image> node. We might have to dig it out of
+    // the nodegraph. Unsure about triplanar that has 3 image nodes. Does Storm
+    // require per-image texture params? How does one specify that using a single
+    // token?
```
tcauchois, line 504: Are you talking about sampler parameters (e.g. wrap mode)?  Hydra takes them specified per-texture, and in the texture metadata we're gathering we should have a per-texture entry for everything.  If the triplanar node only has one set of params, we can just splat that to all of the referenced textures.

pxr/imaging/hdSt/materialXFilter.cpp
JGamache-autodesk: This PR also includes detection for glTf textures via a [_GetGlTFSurfaceMaterialTag()](https://github.com/PixarAnimationStudios/OpenUSD/pull/3073/files#diff-c5e0ffdbaf94bdfaffbc73413cb05018d1a3115640d2c53d51ddff9c886be346R856) function that properly differentiate between translucent and masked mode. The merge resolution strategy would be to let this PR fully win over #3069, equivalent to reverting it before merge.

I have opened Pablo's asset with my local build and it renders correctly.

pxr/imaging/hdSt/materialXFilter.cpp
JGamache-autodesk: Yes, the information being dependent on ShaderGen target, it should probably be `static bool MaterialX::GlslShaderGenerator::isNodeDefTopological(MaterialX::NodeDefPtr const&)`

I do love the fact that we can get all the required info from Sdr and compare via TfTokens without having to use the MaterialX library, but in this case correctness trumps convenience, so this will definitely be added to MaterialX. If performance matters, we can always cache the result since the return value will always be the same for a given nodeId.

pxr/imaging/hdSt/materialXFilter.cpp
JGamache-autodesk: Yes, sample parameters could theoretically vary from one texture to the other. This is not the case with triplanar, but a custom NodeDef/NodeGraph node with multiple texture could try having one texture repeat while the other clamps. This is a bit of a corner case though, so unsure how much energy we should invest into covering it.

tcauchois: I'm happy to leave it until we have a concrete asset we're trying to get working.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -293,6 +480,57 @@ _AddDefaultMtlxTextureValues(std::map<TfToken, VtValue>* hdTextureParams)
...
 }
 
+static void
+_AddDefaultMtlxTextureValues(
+    mx::NodeDefPtr const& nodeDef,
+    std::map<TfToken, VtValue>* hdTextureParams)
+{
+    _AddDefaultMtlxTextureValues(hdTextureParams);
+
+    if (nodeDef->getCategory() == mx::ShaderNode::IMAGE) {
```
klucknav, line 490: Isn't the category for nodeDef's 'nodeDef' should we be checking the category on the mtlx Node itself?

JGamache-autodesk: Agreed. Checking on the node would indeed work. The other option is to use `nodeDef->getNodeString()` instead. Do you want me to update the PR?

JGamache-autodesk: But, unit tests passing, as well as my test scenes rendering correctly, give a strong indication that this code block should be removed.

klucknav: I agree I think it should be removed. My next question was going to be how is it different from the other `_AddDefaultMtlxTextureValues()` above, which seems to be the reason the tests were ok. 
Am I understanding this function correctly in that the first `_AddDefaultMtlxTextureValues()` call handles the `<image>`and `<tiledimage>` nodes and the rest is intended to handle any other custom texture based nodes?

JGamache-autodesk: We are looking for default values and indeed those default are correct for `mx::ShaderNode::IMAGE`. One more reason to remove the `if (nodeDef->getCategory() == mx::ShaderNode::IMAGE)` block of code.

The first `_AddDefaultMtlxTextureValues()` can probably be merged into the second one since it is the only place where it is called.

JGamache-autodesk: I just browsed the MaterialX library, and all image nodes used in NodeGraphs are `repeat`, which mean the defaults are sufficient for all cases I tested.

The only way to insure this code actually does something useful would be to create a custom `untiledimage` NodeDef that does not have `repeat` as the default on its embedded image node.

klucknav: Just to verify my understanding of this function. Since the first `_AddDefaultMtlxTextureValues()` correctly gets the default values for both of the image based stdlib nodes, is the rest of the function (after that if block we agree should be removed) handling the case that we have a custom texture node that uses different default values for the wrap mode or filterType than the stdlib nodes?

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen.cpp
```diff
@@ -28,16 +28,73 @@
...
+    if (node && node->getType() == mx::MATERIAL_TYPE_STRING) {
+        // Use auto so can compile against MaterialX 1.38.0 or 1.38.1
+        auto mxShaderNodes = mx::getShaderNodes(node, mx::SURFACE_SHADER_TYPE_STRING);
+        if (!mxShaderNodes.empty()) {
+            renderableElem = *mxShaderNodes.begin();
+        }
+    }
+
+    auto renderableNode = renderableElem->asA<mx::Node>();
+    if (renderableNode && renderableNode->getCategory() == "UsdPreviewSurface")
```
klucknav, line 81: Why do we need special handling for UsdPreviewSurface materials? Would we not get the correct result by calling `mx::isTransparentSurface()` below?

pxr/imaging/hdSt/materialXFilter.cpp
klucknav: We actually already support that for MaterialX - there is an existing test case but it looks like the opensource thresholds were not tight enough to catch this. Let me know if you are good with this change to catch color corrected  values
```
const auto colorManagedInput = 
SdfPath::StripPrefixNamespace(param.first.GetString(), SdfFieldKeys->ColorSpace);

// If this parameter indicates a colorspace on a color 
// managed input, find and add that corresponding input
if (colorManagedInput.second) {
    outNode.parameters.insert(param);

    // Get the parameter value for the input
    const TfToken colorInputParam(colorManagedInput.first);
    const auto colorInputIt = inNode.parameters.find(colorInputParam);
    if (colorInputIt != inNode.parameters.end()) {
        VtValue colorInputValue = colorInputIt->second;

        // Use an empty asset for color managed files
        if (colorInputValue.IsHolding<SdfAssetPath>()) {
            colorInputValue = VtValue(SdfAssetPath());
        }
        outNode.parameters.emplace(colorInputParam, colorInputValue); 
    }
}
```

JGamache-autodesk: Sadly there will not be a value to be found in the data since it all gets pruned to create a pure "default" shader. But the answer we are looking for can be found in the Sdr registry with code pretty close to this:

```C++
const auto colorManagedInput = 
SdfPath::StripPrefixNamespace(param.first.GetString(), SdfFieldKeys->ColorSpace);

// If this parameter indicates a colorspace on a color 
// managed input, we need to find out if it manages an asset.
if (colorManagedInput.second) {
    outNode.parameters.insert(param);

    // Get the parameter type for the input
    const TfToken colorInputParam(colorManagedInput.first);
    SdrRegistry &sdrRegistry = SdrRegistry::GetInstance();
    const SdrShaderNodeConstPtr sdrNode = 
        sdrRegistry.GetShaderNodeByIdentifierAndType(inNode.nodeTypeId, _tokens->mtlx);
    auto sdrInput = sdrNode->GetShaderInput(colorInputParam);
    if (sdrInput->GetTypeAsSdfType().first == SdfValueTypeNames->Asset) {
        // Use an empty asset for color managed files
        colorInputValue = VtValue(SdfAssetPath());
        outNode.parameters.emplace(colorInputParam, colorInputValue); 
    }
}
```

The other solution would be to fix `_AddMaterialXNode()` in `usd\pxr\imaging\hdMtlx\hdMtlx.cpp` to change this block:

```C++
        // Skip Colorspace parameter, this is already captured in the paramData.
        // Note: Colorspace inputNames are of the form 'colorSpace:inputName'
        const std::pair<std::string, bool> result = 
            SdfPath::StripPrefixNamespace(mxInputName, SdfFieldKeys->ColorSpace);
        if (result.second) {
            continue;
        }
```

to:

```C++
        // Colorspace parameter: create an empty input if it does not already
        // exist to set colorspace metadata.
        // Note: Colorspace inputNames are of the form 'colorSpace:inputName'
        const std::pair<std::string, bool> result = 
            SdfPath::StripPrefixNamespace(mxInputName, SdfFieldKeys->ColorSpace);
        if (result.second) {
            mx::InputPtr mxInput = mxNode->addInputFromNodeDef(result .first);
            mxInput->setColorSpace(HdMtlxConvertToString(paramData.value));
            continue;
        }
```

Which would allow not having to create a dummy value in the first place.

JGamache-autodesk: Actually had to fix this on our side since it impacted our internal unit tests. The following function was added:
```C++
// MaterialX requires a value for color-spaced attributes. Fetch the default from Sdr shader:
void _AddMaterialXDefaultValue(HdMaterialNode2& node, std::string const& parameterName) {
    SdrRegistry &sdrRegistry = SdrRegistry::GetInstance();
    const SdrShaderNodeConstPtr sdrNode = sdrRegistry.GetShaderNodeByIdentifierAndType(
        node.nodeTypeId, _tokens->mtlx);
    if (sdrNode) {
        SdrShaderPropertyConstPtr sdrInput = sdrNode->GetShaderInput(TfToken(parameterName));
        if (sdrInput) {
            if (sdrInput->IsAssetIdentifier()) {
                // Replace empty filename with a placeholder ignored by MaterialX shadergen:
                node.parameters.emplace(TfToken(parameterName),
                                        VtValue{SdfAssetPath{"placeholder"}});
            } else {
                node.parameters.emplace(TfToken(parameterName),
                                        sdrInput->GetDefaultValueAsSdfType());
            }
        }
    }
}
```
and was called:
```diff
@@ -282,9 +303,8 @@ size_t _BuildEquivalentMaterialNetwork(
                                                   SdfFieldKeys->ColorSpace);
                 if (colorManagedFile.second) {
                     outNode.parameters.insert(param);
-                    // Need an empty asset as well:
-                    outNode.parameters.emplace(TfToken(colorManagedFile.first),
-                                               VtValue{SdfAssetPath{}});
+                    // We need a value as well:
+                    _AddMaterialXDefaultValue(outNode, colorManagedFile.first);
                 }
             }
         }
```
And this made sure there was a correct value for the MaterialX code.

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen.cpp
JGamache-autodesk: Sadly, `mx::isTransparentSurface` is only able to handle the most basic cases. It digs inside layers of NodeGraph until it finds either the opacity input of the ND_surface, or the weight input of a transparent Schlick node.

Then is has to traverse back up the connections to the calling node interface to check what was the actual value.

The problem is when there are intermediate nodes in that code path. MaterialX is able to deal with the luminance found in standard_surface that converts the color3 interface opacity input down to the float opacity input of the ND_surface node, but when the graph is more complex, as is the case with PreviewSurface and glTf, it becomes very hard to evaluate the expression in C++ to know or not if the surface is transparent.

In these cases, the `mx::isTransparentSurface` will always return transparent, and this is why we need a custom test. This is also why I had to fix the output of a few PreviewSurface unit tests because the surface was not transparent anymore.

PR #3077, opened 2024-05-07
[build_usd.py] Update OpenEXR dependency to 3.1.13

nvidia-jomiller:
### Description of Change(s)

Updates OpenEXR dependency from v3.1.11 -> v3.1.13

### Fixes Issue(s)
-

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -1160,7 +1160,7 @@ PNG = Dependency("PNG", InstallPNG, "include/png.h")
 ############################################################
 # IlmBase/OpenEXR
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9650

dgovil:
@meshula As I understand it, this would only affect USD's dependencies that depend on EXR right? Since you have your own EXR reader inside USD?

That said, 3.1.12 fixes a [CVE](https://takeonme.org/cves/CVE-2023-5841.html) so it would probably be good to update just to protect folks who transitively use the dependency.

meshula:
Yes, that's right. Updating OpenEXR won't affect OpenUSD directly.

PR #3085, opened 2024-05-14
Add exposure controls to UsdGeomCamera

anderslanglands:
This PR adds exposure controls to UsdGeomCamera, allowing to specify brightness as with a physical camera.  The calculated exposure value is fed to `HdCamera::GetExposure()` via `UsdImagingCameraAdapter`. This means that:
1. Existing scenes that use the existing `exposure` attribute continue to work as before
2. Renderers that already use `HdCamera::GetExposure()` get the exposure calculation automatically if the new exposure attibutes are changed from their defaults.

This is currently draft to start discussion as per @meshula

### Description of Change(s)

Adds exposure:time, exposure:iso, exposure:fNumber, exposure:responsivity attributes and repurposes exposure attribute to be interpreted as exposure compensation. Adds UsdGeomCamera::ComputeLinearExposureScale() method which allows users to calculate the imaging ratio (i.e. given a certain luminance on the sensor, what is the photometric exposure output) from these attributes.

Also adds a unit test to test the above.

Adds tokens `exposureTime`, `exposureIso`, `exposureFNumber`, `exposureResponsivity`, `exposureCompensation` to `HdCamera`. Adds `HdCamera::GetLinearExposureScale()` for delegates to be able to use the new controls with a single function call.


jesschimein:
Filed as internal issue #USD-9673

pxr/usd/usdGeom/camera.cpp
```diff
@@ -581,4 +653,23 @@ UsdGeomCamera::SetFromCamera(const GfCamera &camera, const UsdTimeCode &time)
     GetFocusDistanceAttr().Set(camera.GetFocusDistance(), time);
 }
 
+float
+UsdGeomCamera::GetExposureScale(UsdTimeCode time) const
```
tcauchois: Is there any reasonable way to make this function a pass-through to the "exposure" attribute if the other attributes are un-authored?  And is "exposure" (here exposureCompensation) just a non-physical modifier that you'd want to leave unauthored if you're trying to use the new controls?

anderslanglands: It should already do that: it defaults all the others so if they're not authored it will only take the exposure attribute into account (the defaults for the others combine to do nothing)

anderslanglands: Exposure compensation is a standard control on e.g. DSLRs. It's useful in production in the same way that exposure on the lights is useful for adjustment

pxr/usd/usdGeom/schema.usda
```diff
@@ -2353,16 +2353,38 @@ class Camera "Camera" (
         doc = """Frame relative shutter close time, analogous comments from
                  shutter:open apply. A value greater or equal to shutter:open
                  should be authored, otherwise there is no exposure and a
-                 renderer should produce a black image."""
+                 renderer should produce a black image. Used for motion blur."""
     )
```
meshula, line 2399: In the larger physical camera proposal, we've got this chunk of verbiage ~

```
Describes the exposure settings common to still and cinema cameras.

    These settings are used to scale the brightness of the illumination incident
    on the camera system. The scaling factor, S, is computed as:
       S = exposure_time * iso * 2^exposure_compensation * responsivity_compensation / (100 * fnumber^2)
```

Can we work that in somewhere?

spitzak: Can't the existing exposure be used as exposure_compensation? If all the others have appropriate fallback values then the result will be the exact same scaling factor without any need to switch between different math expressions.

spitzak: Nevermind, I see that was in the original proposal. The default (sorry, the "fallback") values for all the attributes are apparently set so that the math is 2^exposure.

I do recommend not using the term "exposure_compensation" anywhere, use "exposure" instead, so nobody is confused into thinking there are more parameters than there really are.

anderslanglands: Yeah can rename it here. 

I do use exposureCompensation as a token in HdCamera in order to be able to read the `exposure` attribute as opposed to getting the computed exposure value.

anderslanglands: @meshula I've added a bit of an explanation to the UsdGeomCamera description which renders like the attached
Screenshot_20240516_120837

If you think that fits the bill I'll push it

meshula: Yes, please :)

meshula: That's awesome, thank you!!

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,35 @@ class Camera "Camera" (
...
+    referred to as the __imaging ratio__) that converts from incident luminance at
+    the sensor in __nits__, to an output signal based on photometric exposure in 
+    __lux-seconds__. 
+    
+    The calculated scaling factor can be obtained from 
+    \\ref UsdGeomCamera::GetExposureScale(), which is also the value provided by 
+    \\ref HdCamera::GetExposure(). It is computed as:
+    \\code
+        exposureScale = 
+            (exposureTime * exposureIso * pow(2, exposure) * exposureResponsivity)
```
unhyperbolic: I consider the 100 an idiosyncracy of the iso number, so I would write it as (exposureIso / 100).

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,35 @@ class Camera "Camera" (
...
+    f-number, and exposure compensation, mirroring the controls on a real camera.
+    These parameters are provided by \\ref UsdGeomCamera::GetExposureTimeAttr(),
+    \\ref UsdGeomCamera::GetExposureIsoAttr(), \\ref UsdGeomCamera::GetExposureFNumberAttr(), 
+    and \\ref UsdGeomCamera::GetExposureAttr(), respectively. 
+    \\ref UsdGeomCamera::GetExposureResponsivityAttr() provides an additional scaling
+    factor to model the overall responsivity of the system, including response of 
+    the sensor and loss by the lens.
+
+    The combination of these parameters gives an exposure scaling factor (also 
+    referred to as the __imaging ratio__) that converts from incident luminance at
```
unhyperbolic: "at sensor" and "output signal" seems to imply that we do not take into affect things that happen before the sensor such as the lens and aperture size. Yet, we include the fStop, so this doesn't seem quite correct.

pxr/usd/usdGeom/schema.usda
```diff
@@ -2353,16 +2382,38 @@ class Camera "Camera" (
...
                  -1.0 will halve them."""
     )
+
+    # exposure controls
+    float exposure:iso = 100.0 (
+        doc = """The speed rating of the sensor or film when calculating exposure.
+                 Higher numbers give a brighter image, lower numbers darker."""
+    )
+
+    float exposure:time = 1.0 (
```
unhyperbolic, line 2414: How does this relate to the shutter interval? I assume that in a physical camera, the exposure time is the length of the shutter interval (modulo units of one using frames, the other seconds). I think the comment should say something about it.

Also, the default of 1.0s seems unusually high. I would say 1/60 is more typical. Though, you might subscribe to the philosophy that it is safer to have a non-sensical default value and forcing the user to author the actual value. That is, if we forget to author this value, 1s will stick out. Vs 1/60 as default looks reasonable and might be overlooked.

pxr/usd/usdGeom/schema.usda
```diff
@@ -2353,16 +2382,38 @@ class Camera "Camera" (
...
+        doc = """Time in seconds that the sensor is exposed to light when calculating exposure.
+                 Longer exposure times create a brighter image, shorter times darker."""
+    )
+
+    float exposure:fNumber = 1.0 (
```
unhyperbolic: Is this the so called tStop?

Given how it is used, it sure sounds like that. If it is the tStop, I think that it is an established enough term that we should just use it and we don't need to give more documentation how it relates to the fStop.

If not, this defintively needs a comment how it relates to the fStop.

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,35 @@ class Camera "Camera" (
     However, it follows that if even one property is authored in the correct
     scene units, then they all must be.
 
+    \\section UsdGeom_CameraExposure Camera Exposure Model
+
+    UsdGeomCamera models exposure by a camera in terms of exposure time, ISO,
+    f-number, and exposure compensation, mirroring the controls on a real camera.
```
unhyperbolic: From a high level view: if we don't author the f-number and exposure time, should they fallback to be the fStop and the shutter interval length (converted to seconds)? Or would that be undesirable?

pxr/usd/usdGeom/schema.usda
anderslanglands: Yeah that's clearer

pxr/usd/usdGeom/schema.usda
anderslanglands: Well the light has to get through the lens to reach the sensor :) I've added wording to clarify

pxr/usd/usdGeom/schema.usda
anderslanglands: The default values are specifically constructed such that the exposure scale is exactly 1, thus existing content will continue to work exactly as before. Or to look at it another way, the defaults are the values "currently being used" by a renderer that doesn't yet consider any of this.

The shutter interval is a measure of exposure time, yes, just in different units. Added a clarifying note.

pxr/usd/usdGeom/schema.usda
anderslanglands: f-stop and f-number are synonyms. I used f-number here to distinguish from f-stop, which is used for DOF.

pxr/usd/usdGeom/schema.usda
anderslanglands: That would be undesirable as it would mean that when users tweak the depth of field or their motion blur settings, the image would change brightness, not to mention the backward-compatibility I described in the other comment. 

You can imagine cases in which an application wanting to model a real camera might want these properties linked to "make it more real", but IMO that is a UX thing, not a schema thing.

pxr/usd/usdGeom/schema.usda
anderslanglands: Yet another way to look at it is that the default values are what I call "passthrough" values: the output signal (pixel value) is the nits arriving at the sensor.

pxr/usd/usdGeom/schema.usda
anderslanglands: t-stop is a measurement used on cinema cameras that is essentially a remapping such that transmission and other geometric effects of the lens are ignored.

pxr/usd/usdGeom/schema.usda
unhyperbolic: Makes sense.

pxr/usd/usdGeom/schema.usda
unhyperbolic: Since this is namespace anyway, I would consider going with fStop.

I have a larger picture question about what factor in this is taking into account how much light we use because of, e.g., lens coating.

anderslanglands: > I have a larger picture question about what factor in this is taking into account how much light we use because of, e.g., lens coating.

That's all rolled into "responsivity", which is a catch-all correction factor to model lens transmission loss, ISP highlight protection etc. - basically everything that's too complex to simulate or unknown. For instance, if I shoot a macbeth chart with my DSLR and my 55mm lens, I can compute that responsivity factor as ~0.65 by comparing the pixel values generated by the camera to what a spectrometer gives me.

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,35 @@ class Camera "Camera" (
...
+    \\ref UsdGeomCamera::GetExposureIsoAttr(), \\ref UsdGeomCamera::GetExposureFNumberAttr(), 
+    and \\ref UsdGeomCamera::GetExposureAttr(), respectively. 
+    \\ref UsdGeomCamera::GetExposureResponsivityAttr() provides an additional scaling
+    factor to model the overall responsivity of the system, including response of 
+    the sensor and loss by the lens.
+
+    The combination of these parameters gives an exposure scaling factor (also 
+    referred to as the __imaging ratio__) that converts from incident luminance 
+    passing through the lens system and reaching the sensor, in __nits__, to an 
+    output signal based on photometric exposure in __lux-seconds__.
```
unhyperbolic: For clarity, is the "imaging ratio" the factor that converts the luminance in nits=cd/m^2 (=lux/radians^2, I think) into exposure in lux-seconds (so [per-wavelength perceptually adjusted] energy per area) of the film? -So "imaging ratio" has unit radians^2 second(??)? Or is this the factor that converts it all the way from luminance to actual the unit-less floating point value that I want to write into some image file, e.g., OpenEXR? So it's unit is simply 1/nits?

Since the factor also includes the iso, it seems to me that it is the latter. In that case, I guess exposureResponsivity should have unit "1/(nits * 100ISO)" - if ISO were an actual unit.

Minor nitpick: "passing through" sounds like you only look at the light that makes it through, say, the coating of a lens. Would "Enter the lens system" be more correct? And where would you author what fraction of light doesn't make it through the lens system?

Note: It might be nice to remind the reader that nits = cd/m^2 in SI units. Bonus: lux and nits should actually be related by radians^2. Or even better, maybe say that ignoring units, the "imaging ratio" is the value of the output signal you get if you had nits luminance entering the lens or equivalently a large Lambertian radiator far from the lens radiating with f lux (where f should be 1/area of sphere, I think, I might be missing some factor here).

anderslanglands: > Minor nitpick: "passing through" sounds like you only look at the light that makes it through, say, the coating of a lens. Would "Enter the lens system" be more correct? And where would you author what fraction of light doesn't make it through the lens system?

This is the light that gets all the way through to the sensor. The fraction of light that doesn't make it due to being absorbed (plus other effects) is 1-responsivity.

unhyperbolic: Thanks.

Reading "image ratio", I want to know what the numerator and the denominator is. Currently, it says "factor that converts from [1] to [2]". So I should be able to read of the denominator from [1] and numerator from [2]. [1] ends in "reaching the sensor". So if I read this literally, "image ratio" does not account for anything that happens in front of the sensor such as light being absorbed by lens coating. But you are just saying that such an affect would be accounted for by the responsivity.

I know that an expert would probably not read it this way. But I think the wording in the usd schema should be unambiguous. Breaking up that sentence should achieve this.

anderslanglands: I'm sorry I'm not following, when you say "break up the sentence" what do you mean exactly?

unhyperbolic: I was thinking of first defining the "imaging ratio" by unambiguously saying what it is the ratio of or what it is converting from and to. And then listing intermediate steps or contributing factors.

"from incident luminance passing through the lens system and reaching the sensor" can be interpreted in multiple ways.

anderslanglands: OK I've added the units of the imaging ratio (sr.s) and noted that nits are cd/m^2.

anderslanglands: @unhyperbolic pushed another update that fixes my bad dimensional analysis and more explicitly clarifies the conversion and units involved. Let me know what you think.

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,36 @@ class Camera "Camera" (
...
+    \\ref UsdGeomCamera::GetExposureScale(), which is also the value provided by 
+    \\ref HdCamera::GetExposure(). It is computed as:
+    \\code
+        exposureScale = exposureResponsivity * 
+            (exposureTime * (exposureIso/100) * pow(2, exposure)) 
+            / (exposureFStop * exposureFStop)
+    \\endcode
+
+    This scaling factor is combined from two parts: The first, known as the __imaging ratio__ 
+    (in _steridian-second_), converts from incident luminance at the front
```
spiffmon: steradian with an "a" ?

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,36 @@ class Camera "Camera" (
...
+    \\code
+        exposureScale = exposureResponsivity * 
+            (exposureTime * (exposureIso/100) * pow(2, exposure)) 
+            / (exposureFStop * exposureFStop)
+    \\endcode
+
+    This scaling factor is combined from two parts: The first, known as the __imaging ratio__ 
+    (in _steridian-second_), converts from incident luminance at the front 
+    of the lens system, in _nit_ (_cd/m^2_), to photometric exposure at the sensor
+    in _lux-second_. The second, <tt>exposureResponsivity</tt>
```
spiffmon: since you're already using markdown for formatting, should be able to use backticks rather than `<tt>` html?

pxr/usd/usdGeom/schema.usda
```diff
@@ -2353,16 +2383,47 @@ class Camera "Camera" (
...
+                 Note that shutter:open and shutter:close model essentially the 
+                 same property of a physical camera, but are for specifying the 
+                 size of the motion blur streak which is for practical purposes
+                 useful to keep separate."""
+    )
+
+    float exposure:fStop = 1.0 (
+        doc = """f-stop of the aperture when calculating exposure. Smaller numbers
+                 create a brighter image, larger numbers darker.
+                 Note that the fStop attribute also models the diameter of the camera
```
spiffmon: Can we get some formatting on `fStop` ?

spiffmon (commented):
a few formatting notes.

pxr/usd/usdGeom/schema.usda
anderslanglands: lol I've been spelling (and pronouncing) it wrong this entire time. How embarassing...

pxr/usd/usdGeom/schema.usda
anderslanglands: @unhyperbolic polite bump here - does this address your concerns?

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,36 @@ class Camera "Camera" (
     However, it follows that if even one property is authored in the correct
     scene units, then they all must be.
 
+    \\section UsdGeom_CameraExposure Camera Exposure Model
+
+    UsdGeomCamera models exposure by a camera in terms of exposure time, ISO,
+    f-number, and exposure compensation, mirroring the controls on a real camera.
```
meshula: did we decide to normalize on "f-stop" rather than "f-number"? I think this is a lingering reference.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -259,9 +259,33 @@ UsdImagingCameraAdapter::Get(UsdPrim const& prim,
...
-        VtValue v;
-        cam.GetExposureAttr().Get(&v, time); // conversion n/a
-        return v;
+        // we hijack the "exposure" token here in order to pass through the calculated
+        // exposure and make it backwards-compatible.
+        // existing clients will be expected it as logarithmic exposure, so we need
+        // to convert from a scalar multiplier to logarithmic exposure here.
+        // To get the original exposure attribute value, see "exposureCompensation"
+        // below
+        return VtValue(log2(cam.GetExposureScale(time)));
```
tcauchois: We're missing the equivalent code in dataSourceCamera.cpp: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usdImaging/usdImaging/dataSourceCamera.cpp#L132

pxr/usd/usdGeom/schema.usda
tcauchois: As @meshula pointed out, I think we're using f-stop instead of f-number, and the comment and the doxygen ref should be updated.  (See comment in generatedSchema.usda as well).

pxr/usd/usdGeom/schema.usda
```diff
@@ -2292,6 +2292,36 @@ class Camera "Camera" (
...
+    \\code
+        exposureScale = exposureResponsivity * 
+            (exposureTime * (exposureIso/100) * pow(2, exposure)) 
+            / (exposureFStop * exposureFStop)
+    \\endcode
+
+    This scaling factor is combined from two parts: The first, known as the __imaging ratio__ 
+    (in _steradian-second_), converts from incident luminance at the front 
+    of the lens system, in _nit_ (_cd/m^2_), to photometric exposure at the sensor
+    in _lux-second_. The second, `exposureResponsivity`
```
tcauchois: Per @unhyperbolic , should this include a factor of pi/4 since the aperture is a disk with area pi / 4 * diameter^2?

pxr/usd/usdGeom/generatedSchema.usda
```diff
@@ -4194,6 +4194,36 @@ class Camera "Camera" (
     However, it follows that if even one property is authored in the correct
     scene units, then they all must be.
 
+    Camera Exposure Model
+
+    UsdGeomCamera models exposure by a camera in terms of exposure time, ISO,
+    f-number, and exposure compensation, mirroring the controls on a real camera.
+    These parameters are provided by \\ref UsdGeomCamera::GetExposureTimeAttr(),
+    
+    and respectively.
```
tcauchois: This section of the generated schema is kinda messed up, and drops a few lines from the doxygen.  As @meshula noted, schema.usda has a lingering reference to GetFNumberAttr which doesn't exist; possibly that's causing this?

pxr/imaging/hd/camera.h
```diff
@@ -74,6 +74,11 @@ PXR_NAMESPACE_OPEN_SCOPE
     (shutterOpen)                                                 \
     (shutterClose)                                                \
     (exposure)                                                    \
+    (exposureTime)                                                \
+    (exposureIso)                                                 \
+    (exposureFStop)                                               \
+    (exposureResponsivity)                                        \
+    (exposureCompensation)                                        \
```
tcauchois: These new params are missing from the hydra camera schema: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/imaging/hd/cameraSchema.h ... as well as the two modes of scene index emulation:
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/imaging/hd/dataSourceLegacyPrim.cpp#L1036
... and ...
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/imaging/hd/sceneIndexAdapterSceneDelegate.cpp#L1237

... although it looks like for those, since you aren't doing any weird type conversion, the fallbacks might be ok and if you just add the names to dataSourceLegacyPrim.cpp that'll be enough.

pxr/imaging/hd/camera.cpp
```diff
@@ -255,6 +255,41 @@ HdCamera::Sync(HdSceneDelegate * sceneDelegate,
             _exposure = vExposure.Get<float>();
         }
 
+        const VtValue vExposureCompensation =
```
tcauchois: We should add initialization for all of these member variables in the HdCamera constructor (following the pattern for _exposure).

tcauchois (commented):
Hey Anders!  I added a single discussion point on the math; also some plumbing stuff we'll need to address before merging, if you're inclined to get a head start on that.

pxr/usd/usdGeom/schema.usda
anderslanglands: I've pushed commits that fix this

pxr/usd/usdGeom/schema.usda
anderslanglands: I've pushed commits that add an explanation here, linking to the physLight pdf and explaining how we've combined terms to simplify the equation.

pxr/usd/usdGeom/camera.h
```diff
@@ -574,6 +711,19 @@ class UsdGeomCamera : public UsdGeomXformable
...
+    /// <tt>exposure:fnumber</tt>, <tt>exposure:time</tt> and <tt>exposure:responsivity</tt> 
+    /// attributes.
+    ///
+    /// This is expected to be applied as a multiplier to the brightness of the 
+    /// image generated by the renderer, and given physically meaningful lighting
+    /// values in the scene, allows the exposure controls on UsdGeomCamera to behave
+    /// like those of a real camera.
+    ///
+    USDGEOM_API
+    float GetExposureScale(UsdTimeCode time=UsdTimeCode::Default()) const;
```
tcauchois: @anderslanglands I'm going to change this to "Compute".  I was wondering a bit about the fact that it's linear, when at least hydra is expecting something log-scale.  Is it valuable to have a linear result, or would it be better to put the log inside the function for consistency?  (If there's a use for it in linear that would lose precision from the exp2(log2(*)) that's a compelling reason to leave it as is).

If we leave it linear, I might go a step further and call it ComputeLinearExposureScale...  Anyway, thoughts?

(Don't worry about updating this, I've pulled it and it's marked up a bunch for compatibility with internal code so it's easier for me to push the last few changes through).

pxr/usdImaging/usdImaging/cameraAdapter.cpp
tcauchois: (Note: I've implemented this internally).

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -259,9 +259,33 @@ UsdImagingCameraAdapter::Get(UsdPrim const& prim,
         cam.GetShutterCloseAttr().Get(&vShutterClose, time); // conversion n/a
         return vShutterClose;
     } else if (key == HdCameraTokens->exposure) {
-        VtValue v;
-        cam.GetExposureAttr().Get(&v, time); // conversion n/a
-        return v;
+        // we hijack the "exposure" token here in order to pass through the calculated
```
tcauchois: @anderslanglands This part worries me a bit, as it has the potential to be confusing forever.  If I swap us over to just transporting the attributes as named, compute the exposure scale in HdCamera, and update our renderers to look for exposureScale and fallback to exposure, would you have any major objections?  It would be a more manual upgrade process to get the benefit of this new schema, but the transport side wouldn't be confusing, and if we add a good imaging unit test that seems like it would help with compliance.

A similar variant of that is to keep the GetExposureScale here but transport it as "exposureScale", and otherwise do what I mentioned above.

anderslanglands: Hi Tom! No I wouldn't have major objections. I found it "nice" that it all just fed through automatically but I understand your concerns about long-term confusion just for the sake of a little ease now, particularly when a bunch of renderers don't actually pay attention to exposure anyway.

I can't find your other comment on here about logarithmic vs linear now... but in reference to that, personally I think linear makes more sense. I did have to convert it to logarithmic in the adapter just to make the cute "hijack exposure" trick work, so if we move away from that I would keep it linear as there's no good reason for it to be logarithmic and it saves a pow2 on the renderer side.

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -259,9 +259,33 @@ UsdImagingCameraAdapter::Get(UsdPrim const& prim,
...
+        // exposure and make it backwards-compatible.
+        // existing clients will be expected it as logarithmic exposure, so we need
+        // to convert from a scalar multiplier to logarithmic exposure here.
+        // To get the original exposure attribute value, see "exposureCompensation"
+        // below
+        return VtValue(log2(cam.GetExposureScale(time)));
+    } else if (key == HdCameraTokens->exposureTime) {
+        VtValue vExposureTime;
+        cam.GetExposureAttr().Get(&vExposureTime, time); // conversion n/a
```
pmolodo: I noticed that there's a bug here - it should be `GetExposureTimeAttr()`.

@tcauchois - given that you already have some outstanding internal changes, would you prefer I make an additional commit to this PR to fix this, or would that just complicate things for you?

pxr/imaging/hd/camera.h
pmolodo: Hi @tcauchois - since I see you've already implemented some of the other related stuff, I'm guessing you've already made the changes to hd/hdSchemaDefs.py that would address this?  If not, I'm happy to push these fixes myself.

One note - while testing this, I noticed that `hdGenSchema.py` requires `jinja2 >= 3.0` - I'll make a separate PR updating that in `VERSIONS.md`.

pmolodo: Here's the separate PR to update the jinja2 req:

https://github.com/PixarAnimationStudios/OpenUSD/pull/3314

pxr/usdImaging/usdImaging/cameraAdapter.cpp
tcauchois: I've fixed it already internally, in the copy of the PR I pulled down.

pxr/imaging/hd/camera.h
tcauchois: I've made the schemaDefs changes already; I think modulo the discussions downthread about ComputeLinearExposureScale and changing the transport tokens, this is almost ready to go.  Thanks for the VERSIONS.md update!

anderslanglands: Hi @tcauchois just to be really clear here: are you saying the only update to this PR you require from us is the rework of the exposure scale calculation to be moved to a separate function on HdCamera that gives the (linear) value from UsdGeomCamera::ComputeExposureScale(), and return HdCamera::GetExposure() to its original behaviour?

tcauchois: I've pulled this down and added the Hydra 2 functionality and started reconciling issues with closed source code.  My plan was to take the PR with my additions, plus the changes mentioned (ComputeExposureScale and the different tokens), if that's ok with you, in which case there's no further action required on your part.  If you'd like me to upload my changes as a new PR instead, for you all to look at or iterate on, let me know.

anderslanglands: Would be great if you could just so we can all see what’s actually going in

tcauchois:
Apologies for getting sidetracked; Anders had requested I post my edits back here for NVidia to look at:
https://github.com/anderslanglands/USD/pull/1

From PR description:

Pixar schema feedback, UsdImaging 2.0 implementation.

I tried to pull the merge artifacts out, since some of these files are out of date w.r.t. top of tree, apologies if I missed anything. The header license text changed everywhere.

Schema feedback: there were some 80 characters changes but the only substantial change was renaming "GetExposureScale" to "ComputeLinearExposureScale". "Compute" because it's idiomatic to call it compute when it's doing a computation; Get is only for data access. "Linear" because exposure elsewhere in USD is log2-encoded, and it seemed like a nice clarification.

I added UsdImagingStageSceneIndex support as well as fixed some invalidation issues in UsdImagingDelegate support.

Further discussion topics:

I'm worried about the token swap in usdImaging: hydra exposure -> usd computed exposure, hydra exposureCompensation -> usd exposure, as a source of future bugs. Other potential avenues of dealing with this would be:

- Revving the camera schema, e.g. adding UsdGeomCamera_1 which has no "exposure" but has an "exposureCompensation" attribute, and "exposure" is purely computed.
- Passing "exposure" as "exposure" to hydra, and exposureScale as "exposureScale", although this would require updating render delegates to conform.
- Only passing "exposure" (i.e. Computed Exposure Scale) to hydra, and treating it as a computed attribute that happens to have the same name as one of its scene inputs.

Let me know what you think of those three options.

We're happy with the math and the schema (with the one function rename), and just needed to iterate on the usdImaging code a bit.  There were some additional internal complications that are resolved as well.

anderslanglands:
I like option 1 - updating to UsdGeomCamera_1 and changing `exposure` to `exposureCompensation`. I like that `exposureCompensation` is the correct name for the attribute, and the version bump makes it clear that semantics have changed, while renderers could still seamlessly get the upgrade if they're using `exposure` already.

pmolodo:
Hi - so despite initially leaning toward option 1, after @nvmkuruc pointed out that versioning the schema would make all uses of the `UsdGeomCamera` C++ class need to be adapted to instead use, ie, `std::variant<UsdGeomCamera, UsdGeomCamera_1>`, we decided this would probably be too disruptive.

So we now think Option 2 would be best:

> - Passing "exposure" as "exposure" to hydra, and exposureScale as "exposureScale", although this would require updating render delegates to conform.

The need to update render delegates seems an acceptable cost, compared to the much larger scale changes we'd need with the schema version change.

So I've pushed some commits which make that switch (as well as merge in `dev`).  Sorry for the switch!

pxr/usdImaging/usdImaging/dataSourceCamera.cpp
```diff
@@ -109,6 +109,78 @@ class _Vec4fArrayToVec4dArrayDataSource
...
+            UsdGeomTokens->exposure,
+            UsdGeomTokens->exposureTime,
+            UsdGeomTokens->exposureIso,
+            UsdGeomTokens->exposureFStop,
+            UsdGeomTokens->exposureResponsivity
+        };
+
+        _inputs.resize(inputNames.size());
+        UsdPrim prim = _usdCamera.GetPrim();
+        for (TfToken inputName : inputNames) {
```
nvmkuruc: @pmolodo If possible, avoid a copy of `TfToken` as it increases the ref count--  It should be safe to use `const TfToken&` or `const auto&` here.

pxr/usd/usdGeom/wrapCamera.cpp
```diff
@@ -129,6 +129,34 @@ _CreateExposureAttr(UsdGeomCamera &self,
     return self.CreateExposureAttr(
         UsdPythonToSdfType(defaultVal, SdfValueTypeNames->Float), writeSparsely);
 }
+        
+static UsdAttribute
+_CreateExposureIsoAttr(UsdGeomCamera &self,
+                                      object defaultVal, bool writeSparsely) {
```
nvmkuruc, line 135: Is this spacing offset intentional?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15)
```
nvmkuruc: @pmolodo Was the `cmake` bump intentional?

pxr/usd/usdGeom/wrapCamera.cpp
pmolodo: Not sure - but every `_CreateXXX` function in that file has the same spacing.  Should probably fix all or none.

CMakeLists.txt
pmolodo: Nope: [undo accidental cmake version bump](https://github.com/PixarAnimationStudios/OpenUSD/pull/3085/commits/95e2cad44c9c329a221cd7c89f62a5d4de1a12d6)

pxr/usdImaging/usdImaging/dataSourceCamera.cpp
pmolodo: [avoid copying TfToken in loop](https://github.com/PixarAnimationStudios/OpenUSD/pull/3085/commits/7e94c68d78389b25c6fdab2fc1a56fc2ff657553)

pmolodo:
Made a follow-up PR, that adds Storm support:

- #3464 

To see just the changes vs THIS PR, see here:

- https://github.com/anderslanglands/USD/pull/2/files

pxr/usdImaging/usdImaging/cameraAdapter.cpp
```diff
@@ -265,7 +265,7 @@ UsdImagingCameraAdapter::Get(UsdPrim const& prim,
         return vExposureResponsivity;
     } else if (key == HdCameraTokens->exposureScale) {
         // The computed linear exposure multiplier ratio
-        return VtValue(log2(cam.ComputeLinearExposureScale(time)));
+        return VtValue(cam.ComputeLinearExposureScale(time));
```
tcauchois, line 268: This had been log2 throughout, to match the encoding of "exposure".  Is the loss of precision such that this is necessary for you all?  If so: (1) let's make the hydra token "linearExposureScale" and (2) let's update dataSourceCamera.cpp correspondingly.

anderslanglands: This was just log2 so that it backdoored into the existing exposure field (since that's log2). Since we're not doing that hijack any more this wants to be linear. @pmolodo can you update the token per Tom's comment?

pmolodo: The thinking behind linear wasn't avoiding precision loss, so much as simplicity and consistency:

- the value is most easily expressed and calculated in linear space (since of the 5 component attributes, only 1 is in log2 space)
- the value will generally be used programmatically in linear space (ie, as a pixel value multiplier)
- as @anderslanglands points out, we formerly only converted to log2 where necessary for shoe-horning back into the exposure attribute.  Now that we're making "exposure" and "exposureScale" completely separate attributes, and not worrying about existing implementations picking up the new behavior / attributes "for free", we have the opportunity to make ALL usages of exposureScale in linear space.

Regardless, I totally agree with you that the possibility for confusion is high - so, in addition to making a commit where the token names are changed to "linearExposureScale", I also added a commit where ALL references to "exposureScale" in the code are changed to "linearExposureScale", for clarity.  (You can leave that commit out if you feel it's too verbose, though!)

PR #3089, opened 2024-05-16
Remove `boost/noncopyable.h` from python related headers

nvmkuruc:
### Description of Change(s)
#3028 demonstrated that OpenUSD can be built without boost. However, some transitive inclusions through python only headers are causing boost headers to end up in the `pch.h` files.

This removes the unnecessary include of `boost/noncopyable` in favor of explicitly deleting the copy constructor and assignment operator.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9683

PR #3090, opened 2024-05-17
Support namespaces in `propertyNamespacePrefix` for `usdGenSchema`

nvmkuruc:
### Description of Change(s)

The current implementation of `usdGenSchema` assumes that `propertyNamespacePrefix` used in multiple apply schemas is a valid identifier and it can be used as a token name directly. This prevents property namespaces from containing the otherwise valid `:`.

This PR changes `propertyNamespacePrefix` to be a dataclass that stores both the prefix value a valid token name that can be threaded through the schema generation process.

With this change, a library's public static tokens can be used directly and there's no need for the private `_schemaTokens` internal to the class anymore.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/usd/usdGenSchema.py
+++ b/pxr/usd/usd/usdGenSchema.py
@@ -41,6 +41,7 @@ generated that will compile and work with USD Core successfully:
 """
 
 from __future__ import print_function
+import dataclasses
 import sys, os, re, inspect
 import keyword
 from argparse import ArgumentParser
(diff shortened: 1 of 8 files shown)


pxr/usd/usd/usdGenSchema.py
```diff
@@ -880,17 +894,17 @@ def ParseUsd(usdFilePath):
         # make sure that if we have a multiple-apply schema with a property
         # namespace prefix that the prim actually has some properties
         if classInfo.apiSchemaType == MULTIPLE_APPLY:
-            if classInfo.propertyNamespacePrefix and \
+            if classInfo.propertyNamespace and \
                 len(sdfPrim.properties) == 0:
                     raise _GetSchemaDefException(
                         "Multiple-apply schemas that have the "
```
crevilla, line 900: This exception text should still say propertyNamespacePrefix since that's the name of the field

pxr/usd/usd/usdGenSchema.py
```diff
@@ -880,17 +894,17 @@ def ParseUsd(usdFilePath):
...
-                        "propertyNamespacePrefix metadata fields must have at "
+                        "propertyNamespace metadata fields must have at "
                         "least one property", sdfPrim.path)
-            if not classInfo.propertyNamespacePrefix and \
+            if not classInfo.propertyNamespace and \
                 not len(sdfPrim.properties) == 0:
                     raise _GetSchemaDefException(
                         "Multiple-apply schemas that do not"
```
crevilla, line 906: Same here

jesschimein:
Filed as internal issue #USD-9691

PR #3092, opened 2024-05-21
Autodesk: Add HgiAttachmentResolveFilter for depth texture in HgiMetal

erikaharrison-adsk:
### Description of Change(s)

Depth Priority is broken after MSAA resolve of the depth buffer. This changeset adds HgiAttachmentResolveFilter enumeration to the HgiAttachmentDesc and add MSAA resolver support for the depth texture in HgiMetal.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9699

PR #3103, opened 2024-05-31
Add initial implementation of `usdSemantics`

nvmkuruc:
### Description of Change(s)
Implementation of https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/semantic_schema which adds a `SemanticsLabelsAPI` schema for labeling prims with tokens.

Note that the implementation diverges slightly with the proposal with the taxonomy being after `labels` instead of before in the property name.  (ie. `semantics:labels:<taxonomy>` instead of `semantics:<taxonomy>:labels`).

In addition to the schema, a utility `UsdSemanticsLabelsQuery` provides a reference for how ancestor labeling is expected to work and inherit.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9719

pxr/usd/usdSemantics/overview.md
```diff
@@ -0,0 +1,190 @@
...
+      `bookcase`, but not `shelf`.
+    * `/OfficeBookshelf` is not part of anything
+
+Workflows around semantic labels should specify whether they are
+capturing direct labels, ancestor labels, or either. It's expected
+but not required that most labeling workflows will check for both
+types of labels.
+
+## Taxonomy and Comparison to Model Hierarchy
+Model hierarchy is a strict hierarchical taxonomy tha
```
nvaszabo: tha -> that

pxr/usd/usdSemantics/labelsQuery.h
```diff
@@ -0,0 +1,129 @@
...
+    /// Requires that neither the `interval` nor `taxonomy` to be empty.
+    ///
+    /// \warning Finite minimum values of the interval will return the same
+    /// result regardless of closed or open state due to held interpolation
+    /// semantics and Zeno's paradox.
+    USDSEMANTICS_API UsdSemanticsLabelsQuery(const TfToken& taxonomy,
+                                             const GfInterval& interval);
+
+    /// Compute the union of values for `semantics:labels:<taxonomy>` in the
+    /// specified interval.
```
spiffmon: For this and other methods that return an array of labels, please specify whether there is a stable ordering of the returned values, or client must expect them to be disordered.

For this method specifically, it WBN to just reinforce the convention that "Direct" means this prim's contributions only.

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,229 @@
...
+    _taxonomy(taxonomy), _time(interval) {
+    TF_VERIFY(!taxonomy.IsEmpty());
+    if (!TF_VERIFY(!interval.IsEmpty())) {
+        _time = UsdTimeCode::Default();
+    }
+}
+
+bool
+UsdSemanticsLabelsQuery::_PopulateLabels(const UsdPrim& prim) {
+    TF_DEV_AXIOM(!prim.IsPseudoRoot());
```
spiffmon: TF_VERIFY is preferred to TF_DEV_AXIOM

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,229 @@
...
+    UsdSemanticsLabelsAPI schema(prim, _taxonomy);
+    if (!schema.GetPrim().HasAPI<decltype(schema)>(schema.GetName()))
```
spiffmon: Constructing a schema object is not trivial in cost; can we do this check first, using the given `prim`, `UsdSemanticsLabelsAPI`, and `_taxonomy` ?

pxr/usd/usdSemantics/labelsQuery.h
```diff
@@ -0,0 +1,129 @@
...
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// The `UsdSemanticsLabelsQuery` can be used to query a prim's
+/// labels for a specified taxonomy and time from the `UsdSemanticsLabelsAPI`.
+/// Time may be an individual time code or an interval.
+///
+/// This utility requires that all prims are on the same stage.
+///
+/// The query caches certain reads and computations and should be discarded
+/// when the state of the stage changes.
```
spiffmon: Helpful to mention that it is safe to use in a multi-threaded context... not all caches in Usd schemas are currently, regrettably.

pxr/usd/usdSemantics/schema.usda
```diff
@@ -0,0 +1,35 @@
...
+    customData = {
+        string libraryName = "usdSemantics"
+        string libraryPath = "pxr/usd/usdSemantics"
+    }
+) {
+}
+
+class "SemanticsLabelsAPI" (
+    inherits = </APISchemaBase>
+    doc = """Applied set of labels for a prim for a taxonomy specified by the
```
spiffmon: Given implied meanings of "set", can we be specific on what we mean, here?  Is it an ordered list of labels, or is authored order insignificant?

pxr/usd/usdSemantics/labelsQuery.h
```diff
@@ -0,0 +1,129 @@
...
+    ///
+    /// If no time samples are authored, the default and fallback values will
+    /// be queried.
+    ///
+    /// \sa UsdSemanticsLabelsQuery::ComputeUniqueInheritedLabels
+    USDSEMANTICS_API
+    VtTokenArray ComputeUniqueDirectLabels(const UsdPrim& prim);
+
+    /// Compute the union of values for `semantics:labels:<taxonomy>` in the
+    /// specified interval including any labels inherited from ancestors.
```
spiffmon, line 87: Further question on ordering... I thought I remembered discussion within the proposal that when computing all inherited labels, more ancestrally-distant labels would appear _later_ in the result than closer/direct ones.  Is that no longer a necessary/desirable property?

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: I don't think we should be doing this check at all-- we should just be validating the schema.

pxr/usd/usdSemantics/labelsQuery.h
nvmkuruc: I found the discussion here-- https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/19/files#r1399829067

I'll follow up about ordering.

pxr/usd/usdSemantics/labelsQuery.cpp
spiffmon: The HasAPI check is a faster check, though, and given that we expect, in many scenes, to have few if any applications of the schema, it may be worth doing the fast check up front?

pxr/usd/usdSemantics/labelsQuery.h
```diff
@@ -0,0 +1,133 @@
...
+    ///
+    /// The returned order is the determnistic ordering of `TfToken`.
```
spiffmon, line 89: Do you have a problem with "lexicographic" rather than deterministic?  Since TfToken has two possible ordering functors, it's worth being specific, I feel.

spiffmon: We can fix this locally, if you agree.

nvmkuruc: Is `TfToken`'s ordering lexicographic for UTF-8?

spiffmon: Ah good point, @nvmkuruc - I, and the token.h documentation, are still stuck in an ascii mindset.  It simply uses std::string's comparator, claiming that's lexicographic.

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+            VtTokenArray labelsAtTime;
+            if (!labelsAttr.Get(&labelsAtTime, queryTime)) {
+                TF_RUNTIME_ERROR(
+                    "Failed to read tokens from %s",
+                    UsdDescribe(labelsAttr).c_str());
+            }
+            return {std::cbegin(labelsAtTime), std::cend(labelsAtTime)};
+        }
+        if constexpr (std::is_same_v<TimeType, GfInterval>) {
+            TF_DEV_AXIOM(!queryTime.IsEmpty());
```
matthewcpp, line 85: maybe this could use TF_VERIFY? Unless checking this condition is expensive enough to have a real impact on performance?

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+            // querying at this time sample will return default or fallback
+            // values. No special handling required.
+            TF_DEV_AXIOM(!times.empty());
+
+            _UnorderedTokenSet uniqueLabels;
+            for (const auto timeInInterval : times) {
+                VtTokenArray labelsAtTime;
+                if (!labelsAttr.Get(&labelsAtTime, timeInInterval)) {
+                    TF_RUNTIME_ERROR("Failed to read value at %s",
+                                     UsdDescribe(labelsAttr).c_str());
```
matthewcpp, line 119: I'm curious when we would hit the two conditions here that cause runtime errors to be emitted. Are those really runtime conditions that make that happen?

(Sorry to be so focused on error policy here, rather than what I'm supposed to be weighing in on!!)

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+bool
+UsdSemanticsLabelsQuery::_PopulateInheritedLabels(const UsdPrim& prim) {
+    const auto stage = prim.GetStage();
+    bool hasInheritedLabel = false;
+    for (const auto& path : prim.GetPath().GetAncestorsRange()) {
+        // Note that _PopulateLabels must run every ancestor to update the
+        // cache. Attempting to collapse this expression using |= or the
+        // STL may result in some population being skipped and incorrect
+        // results.
+        if (_PopulateLabels(stage->GetPrimAtPath(path)))
```
matthewcpp, line 147: We could do this population in parallel?

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+VtTokenArray
+UsdSemanticsLabelsQuery::ComputeUniqueDirectLabels(const UsdPrim& prim)
+{
+    // If the prim was not labeled, we can early exit without
+    // locking.
+    if (!_PopulateLabels(prim)) {
+        return {};
+    }
+
+    std::shared_lock lock{_cachedLabelsMutex};
```
matthewcpp, line 162: I'm curious why we need to lock here; I assume the locking is only to protect against simultaneous writes (or reads while writing is happening). But at this point, we know the cache has already been populated, so presumably it is safe to allow multiple readers to proceed in parallel?

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+    const auto it = _cachedLabels.find(prim.GetPath());
+    if (it == std::cend(_cachedLabels)) {
+        return {};
+    }
+    VtTokenArray result{std::cbegin(it->second), std::cend(it->second)};
+    std::sort(result.begin(), result.end());
```
matthewcpp, line 168: Could we sort as part of population?

Ah, I guess we can't because we use sets so that we can combine inherited results. It seems like it would be nice to frontload the sorting, somehow, but there are tradeoffs to consider, and we'd want to actually measure performance to understand the impact, most likely.

pxr/usd/usdSemantics/labelsQuery.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+    // If no ancestors were labeled, we can early exit without
+    // locking.
+    if (!_PopulateInheritedLabels(prim)) {
+        return {};
+    }
+
+    _UnorderedTokenSet uniqueElements;
+    {
+        std::shared_lock lock{_cachedLabelsMutex};
```
matthewcpp, line 183: Again, I'm not yet understanding why we need to lock once population is complete.

matthewcpp (commented):
The comments contained herein came from our internal code review.

pxr/usd/usdSemantics/labelsQuery.cpp
spiffmon: Thanks for doing so, @nvmkuruc , as I was just about to.  RUNTIME_ERRORs should be reserved for system-level errors.  Here and elsewhere, if we really expected there should have been something authored but there wasn't, it would be more of a TF_WARN, I think?

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: Sure.  Shouldn't be performance critical.

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: @matthewcpp The following layer can trigger that runtime error

```
def "test" (apiSchemas = ["SemanticsLabelsAPI:test"])
{
    int[] semantics:labels:test.timeSamples = {100: [1]}
}
```

with the following code.

```
from pxr import Usd, UsdSemantics
s = Usd.Stage.Open('sl.usda')
query = UsdSemantics.LabelsQuery('test', 100)
print(query.ComputeUniqueDirectLabels(s.GetPrimAtPath('/test')))
```

@spiffmon I'm happy to use whatever diagnostic makes the most sense, but it did seem like an error to me.

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: If another thread is querying the cache, a rehash could be triggered, invalidating the iterator.

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: It's possible it would be more efficient to use `std::merge` on sorted token arrays.  But perhaps we could address later?

pxr/usd/usdSemantics/labelsQuery.cpp
nvmkuruc: See the previous note about the iterator lifetime.

pxr/usd/usdSemantics/labelsQuery.cpp
spiffmon: Right, so that's "bad scene description", similar to having something like
```
over "mySphere"
{
    string radius = "five"
}
```
which is not considered a runtime error.  In Presto (and future USD) it would be a validation error, but at scene-interrogation-time, I think we are silent about it, actually.

PR #1654, opened 2021-10-13
Add USD.natvis file for debugging

aloysbaillet:
### Description of Change(s)
This simply adds a new `USD.natvis` file that helps Visual Studio (and VSCode on both Windows and Linux) make sense of USD types.
This file is the result of the collaboration of a few NVIDIA engineers.

See before:
natvis-off

and after:
natvis-on

N.B. the namespace is hardcoded for now, I think it would make sense to rename it to `USD.natvis.in` and set the right namespace at build time, but the file can be used by developers easily enough without automation.

Another step that could be taken is to embed the natvis file into each library generated by the build, this can be done simply in cmake by adding `${CMAKE_SOURCE_DIR}/pxr/USD.natvis` to the list of targets in `_pxr_library` macro.

PR #3104, opened 2024-05-31
Autodesk: Add native visualisation for versions 23 and 24.

erikaharrison-adsk:
### Description of Change(s)

Developers using Visual Studio find it very helpful to have the .natvis files for debugging. This changeset provides them for USD versions 23 and 24.

Note: This is an alternative to https://github.com/PixarAnimationStudios/OpenUSD/pull/1654 , supporting different versions instead of just one.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9720

Issue #3105, opened 2024-06-03
hdSt fails to render EDF


dgovil:
### Description of Issue

We're attempting to add MaterialX support in Blender, and hitting this issue where `ND_surface` includes a `EDF` input. However , hdSt always fails with the following.

I've tried removing it from the output usda file, and it still seems to try and pick it up. 
It looks like this is because the ND_standard_surface lofts up the EDF attribute and hdSt fails to deal with it. Instead of just ignoring it during the compile, it just outright fails.

Is there a workaround for this that anyone might know of in the meantime?

```
Warning: in _ValidateCompilation at line 229 of /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:1200:5: error: unknown type name 'EDF'
    EDF edf;
    ^

Warning: in _ValidateCompilation at line 229 of /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:1137:5: error: unknown type name 'EDF'
    EDF edf;
    ^

Warning: in _ValidateCompilation at line 229 of /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:2457:5: error: unknown type name 'EDF'
    EDF edf;
    ^

Warning: in _ValidateCompilation at line 229 of /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/glslProgram.cpp -- Failed to compile shader (FRAGMENT_SHADER): program_source:2394:5: error: unknown type name 'EDF'
    EDF edf;
    ^

ERROR: Usdview encountered an error while rendering.
	Error in '&pxrInternal_v0_24__pxrReserved__::HdSt_DrawBatch::_GetDrawingProgram' at line 408 in file /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/drawBatch.cpp : 'Failed to compile shader for prim /root/Plane/Plane.'
	Error in '&pxrInternal_v0_24__pxrReserved__::HdSt_DrawBatch::_GetDrawingProgram' at line 408 in file /Users/dhruvgovil/Projects/usd/pxr/imaging/hdSt/drawBatch.cpp : 'Failed to compile shader for prim /root/Sphere/Sphere.'
```

[...]

PR #3106, opened 2024-06-03
Apple: Fix shader compile bug when using EDF in Storm

dgovil:
Fixes the handling of MaterialX nodes that expose an EDF parameter

Thanks to Lee Kerley for the fix

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3105


jesschimein:
Filed as internal issue #USD-9728

Issue #3095, opened 2024-05-23
Cannot `Sdf.CopySpec` relationship without targets to relationship with targets in crate layers


nvmkuruc:
### Description of Issue
When `Sdf.CopySpec` is run with the following conditions
a) on a crate layer
b) with a relationship without a target as a source
c) and a relationship with a target as the destination, 

`TF_VERIFY` fails with the following error.
```
pxr.Tf.ErrorException:
        Error in 'pxrInternal_v0_24__pxrReserved__::_ProcessChildren' at line 153 in file copyUtils.cpp : 'Failed verification: ' oldDstChildrenValue.IsHolding<ChildrenVector>() ''
```

Change any of the above conditions (ie. use `usda` instead of `usdc`), and the `Sdf.CopySpec` will succeed.

@roggiezhang-nv traced this back to divergent behaviors for how `targetChildren` is handled in the crate vs. text file formats.  The text file format returns a `VtValue` with a valid type that is non-empty.  The crate file format returns an empty `VtValue`.

We'd recommend making the text file format and crate behave the same way, and we're happy to contribute a fix if that's the agreed upon best approach. An alternative fix we considered is whether or not `Sdf.CopySpec` should be more permissive and filter out empty `VtValue`s, but that might be riskier.

### Steps to Reproduce
1. Create a file `copyrel.usda` with the following contents
```
def "target" {}

def "source" {
   rel myrel
}

def "dest" {
   rel myrel = </target>
}
```
2. Run `usdcat copyrel.usda -o copyrel.usdc`
3. Run the following python snippet
```
from pxr import Sdf

text = Sdf.Layer.FindOrOpen("./copyrel.usda")
Sdf.CopySpec(text, '/source', text, '/dest')

crate = Sdf.Layer.FindOrOpen("./copyrel.usdc")
Sdf.CopySpec(crate, '/source', crate, '/dest')
```

Note `Sdf.CopySpec` works for `usda` but not for `usdc` for the same content.

### System Information (OS, Hardware)
Linux

### Package Versions
24.05

jesschimein:
Filed as internal issue #USD-9703

gitamohr:
> We'd recommend making the text file format and crate behave the same way, and we're happy to contribute a fix if that's the agreed upon best approach.

Thanks for the great report! I completely agree that consistency is key and I'd be happy to take a PR, cheers!

roggiezhang-nv:
[...]

PR #3117, opened 2024-06-11
NVIDIA: Fix SdfCopySpec issues about overwriting field SdfFieldKeys->TargetPaths/ConnectionPaths

roggiezhang-nv:
### Description of Change(s)
The whole process of SdfCopySpec can be summarized as the following steps:

* Getting all fields to be copied, including value fields and children fields.
* Copying value fields firstly.
* Then copying children fields.

It's possible that the children fields fetched in step 1 is cleared in step 2, so step 3 is failed with copying a non-existent children field. This fix is to delay the fetch of children fields after step 2 so step 3 will be done on the correct children fields.

### Fixes Issue(s)
- [SdfCopySpec issues about overwriting field SdfFieldKeys->TargetPaths/ConnectionPaths](https://github.com/PixarAnimationStudios/OpenUSD/issues/3095)

### DIFF
--- a/pxr/usd/sdf/copyUtils.cpp
+++ b/pxr/usd/sdf/copyUtils.cpp
@@ -69,14 +69,10 @@ std::vector<TfToken>
 _GetValueFieldNames(
     const SdfLayerHandle& layer, const SdfPath& path)
 {
-    std::vector<TfToken> valueFields;
     const SdfSchemaBase& schema = layer->GetSchema();
-    const std::vector<TfToken> allFields = layer->ListFields(path);
-    for (const TfToken& field : allFields) {
-        if (!schema.HoldsChildren(field)) {
-            valueFields.push_back(field);
-        }
-    }
+    std::vector<TfToken> valueFields = layer->ListFields(path);
+    valueFields.erase(std::remove_if(std::begin(valueFields), std::end(valueFields),
+        [&schema](const auto& field) { return schema.HoldsChildren(field); }), std::end(valueFields));
 
     TfTokenFastArbitraryLessThan lessThan;
     std::sort(valueFields.begin(), valueFields.end(), lessThan);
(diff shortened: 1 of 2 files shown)


pxr/usd/sdf/copyUtils.cpp
```diff
@@ -61,30 +61,50 @@ namespace {
...
+// TfTokenFastArbitraryLessThan comparator.
+static
+std::vector<TfToken>
+_GetChildrenFieldNames(
+    const SdfLayerHandle& layer, const SdfPath& path)
+{
+    std::vector<TfToken> childrenFields;
+    const SdfSchemaBase& schema = layer->GetSchema();
+    const std::vector<TfToken> allFields = layer->ListFields(path);
+    for (const TfToken& field : allFields) {
```
nvmkuruc: Minor suggestion.  If you use `std::remove_if` / `erase`, it'll avoid the extra copies of the tokens.

```
std::vector<TfToken> fields = layer->ListFields(path);
// In C++20, this can be just erase_if
fields.erase(
    std::remove_if(std::begin(fields), std::end(fields), [&schema](const auto& field) { return !schema.HoldsChildren(field); },
    std::end(fields));
```

jesschimein:
Filed as internal issue #USD-9746

roggiezhang-nv:
@jesschimein It looks to me windows build is failed with infra issue. How to restart it?

nvmkuruc:
> @jesschimein It looks to me windows build is failed with infra issue. How to restart it?

The windows build just timed out.  This is a known issue right now and not much you can do about it.  When a PR is selected for merging, it'll go through a different CI/CD process before integrating; it shouldn't time out there.

An update to the PR will signal that the checks should be rerun. If you make this update (for example)-- https://github.com/PixarAnimationStudios/OpenUSD/pull/3117/files#r1634702654 the azure jobs should rerun.

Hope this helps!

pixar-oss:
This should be addressed by the fix we're pushing in 24.08 for #3095!  I'm going to close this for now, but if you still have issues later, please feel free to reopen!

PR #3119, opened 2024-06-12
Autodesk: Adopt size_t and add checks to fix image loading/processing crash

erikaharrison-adsk:
### Description of Change(s)

* Use `size_t` instead of `int` to fix crash caused by `int` overflow.
* Add more checks as IO may fail due to users' environment.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9753

PR #3121, opened 2024-06-14
Autodesk: Lavapipe support on macOS

erikaharrison-adsk:
### Description of Change(s)

⚠ Note: This PR is for feature-hgi-vulkan branch.

The main goal is to enable Lavapipe (Vulkan driver) as a software rasterizer on all supported platforms, for testing purposes. This PR addresses macOS support by fixing some Vulkan instance and device creation issues/bugs.

Enabling MoltenVK support could be useful for testing, but it's not a priority, and there are issues with geometry shaders support preventing that, which this PR does not address. `HgiVulkan` currently assumes that geometry shaders are always supported, although it is actually an [optional feature](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceFeatures.html). If a feature check is added, then this could be fixed.

Lavapipe doesn't support Metal surfaces (`VK_EXT_metal_surface`), but display still works thanks to the `HgiInterop` Vulkan path:
- After building the list of desired instance extensions, remove any that are not actually available. This mostly serves to remove the `VK_EXT_metal_surface` extension when using Lavapipe, since it doesn't support it.
- Make the `VK_KHR_swapchain` device extension optional. It is not available if Lavapipe is built without any selected platforms, as it won't support any surface output.

Update Vulkan support to the latest recommendations for macOS, based on [this document](https://www.lunarg.com/wp-content/uploads/2024/03/The-State-of-Vulkan-on-Apple-LunarG-Richard-Wright-03-18-2024.pdf) from LunarG:
- Do no statically link with the MoltenVK libraries. The desired Vulkan driver should instead by loaded through the ICD loader. This is mandatory for loading the Lavapipe drivers.
- Replace the deprecated `VK_MVK_macos_surface` instance extension with `VK_EXT_metal_surface`.
- Add the `VK_KHR_portability_enumeration` instance extension on macOS. This is [required for MoltenVK](https://github.com/KhronosGroup/MoltenVK/blob/main/Docs/MoltenVK_Runtime_UserGuide.md?plain=1#L389-L395), but not for Lavapipe (it's a [Vulkan 1.3 conformant implementation](https://www.khronos.org/conformance/adopters/conformant-products#submission_696)).
- `VK_EXT_descriptor_indexing` is available on Lavapipe and MoltenVK (partial support), remove the conditional compilation for macOS.

With these changes, using Lavapipe, USDView launches and can display some test scenes, although more testing is required to identify remaining issues. Unit tests also all pass, although many graphics tests are skipped on macOS.

### Fixes Issue(s)
- N/A

### DIFF
--- a/cmake/defaults/Packages.cmake
+++ b/cmake/defaults/Packages.cmake
@@ -265,10 +265,7 @@ if (PXR_BUILD_IMAGING)
             endforeach()
 
             # Find the OS specific libs we need
-            if (APPLE)
-                find_library(MVK_LIBRARIES NAMES MoltenVK PATHS $ENV{VULKAN_SDK}/lib)
-                list(APPEND VULKAN_LIBS ${MVK_LIBRARIES})
-            elseif (UNIX AND NOT APPLE)
+            if (UNIX AND NOT APPLE)
                 find_package(X11 REQUIRED)
                 list(APPEND VULKAN_LIBS ${X11_LIBRARIES})
             elseif (WIN32)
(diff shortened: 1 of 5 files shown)


jesschimein:
Filed as internal issue #USD-9759

PR #2739, opened 2023-10-17
usdUtils: add a GeomSubsetsChecker for validating UsdGeomSubsets with usdchecker

mattyjams:
This new rule checker can be used to validate various aspects of `UsdGeomSubset`s:
- `GeomSubset`s must be authored as direct children of an `Imageable` prim.
- The set of all subset family names is fetched for each subset parent `Imageable`, and each family is checked for validity (using `UsdGeomSubset::ValidateFamily()`).
- If the `materialBind` subset family is authored on an `Imageable`, it is checked to ensure that it is of a restricted type (either `nonOverlapping` or `partition`), since it is invalid for an element of geometry to be bound to multiple materials.
- If a subset has authored material bindings but no authored subset family name, it is suggested that the family name should be set to `materialBind` to ensure that the material bindings are visible to renderers. The material bindings will have no effect otherwise.

I have run the unit tests successfully with the changes here, but they depend on fixes and improvements to `UsdGeomSubset::ValidateFamily()` from #2733.

- [X] I have verified that all unit tests pass with the proposed changes
- [X] I have submitted a signed Contributor License Agreement

PR #3123, opened 2024-06-14
usdGeom, usdShade: add validators related to UsdGeomSubsets

mattyjams:
The validators added in `usdGeom` and `usdShade` here check various aspects of `UsdGeomSubset`s:
- The set of all subset family names is fetched for an `Imageable`, and each family is checked for validity (using `UsdGeomSubset::ValidateFamily()`).
- `GeomSubset`s must be authored as direct descendants of an `Imageable` prim.
- If the `materialBind` subset family is authored on an `Imageable`, it is checked to ensure that it is of a restricted type (either `nonOverlapping` or `partition`), since it is invalid for an element of geometry to be bound to multiple materials. Those subsets are also checked to ensure that they are of element type `face`, since material bindings may
only be applied to geometric faces.
- If a `GeomSubset` has authored material bindings but no authored subset family name, it is suggested that the family name should be set to `materialBind` to ensure that the material bindings are visible to renderers. The material bindings will have no effect otherwise.

This replaces #2739, reworking what was there in the older compliance checker Python framework into the new C++ validation framework.

### DIFF
--- a/pxr/usd/usdGeom/subset.cpp
+++ b/pxr/usd/usdGeom/subset.cpp
@@ -803,8 +803,11 @@ UsdGeomSubset::ValidateFamily(
         return false;
     }
 
+    // Get subsets of *all* element types in the family so we can detect
+    // whether any are authored that do not match the given element type.
     const std::vector<UsdGeomSubset> familySubsets =
-        UsdGeomSubset::GetGeomSubsets(geom, elementType, familyName);
+        UsdGeomSubset::GetGeomSubsets(
+            geom, /* elementType = */ TfToken(), familyName);
     const TfToken familyType = GetFamilyType(geom, familyName);
     const bool familyIsRestricted = (familyType != UsdGeomTokens->unrestricted);
 
--- a/pxr/usd/usdGeom/subset.cpp
+++ b/pxr/usd/usdGeom/subset.cpp
@@ -749,7 +749,7 @@ UsdGeomSubset::ValidateSubsets(
                     valid = false;
                     if (reason) {
                         *reason += TfStringPrintf("Found overlapping index %d "
-                            "in GeomSubset at path <%s> at time %s.\n", index,
+                            "in GeomSubset at path <%s> at time %s.", index,
                             subset.GetPath().GetText(), TfStringify(t).c_str());
                     }
                 }
@@ -773,7 +773,7 @@ UsdGeomSubset::ValidateSubsets(
             valid = false;
             if (reason) {
                 *reason += TfStringPrintf("Found one or more indices that are "
-                    "greater than the element count %ld at time %s.\n", 
+                    "greater than the element count %ld at time %s.",
                     elementCount, TfStringify(t).c_str());
             }
         }
@@ -781,7 +781,7 @@ UsdGeomSubset::ValidateSubsets(
             valid = false;
             if (reason) {
                 *reason += TfStringPrintf("Found one or more indices that are "
-                    "less than 0 at time %s.\n", TfStringify(t).c_str());
+                    "less than 0 at time %s.", TfStringify(t).c_str());
             }
         }
     }
@@ -798,7 +798,7 @@ UsdGeomSubset::ValidateFamily(
     std::string * const reason)
 {
     if (!_ValidateGeomType(geom, elementType)) {
-        *reason += TfStringPrintf("Invalid geom type for elementType %s.\n", 
+        *reason += TfStringPrintf("Invalid geom type for elementType %s.",
                 elementType.GetText());
         return false;
     }
@@ -821,7 +821,7 @@ UsdGeomSubset::ValidateFamily(
         valid = false;
         if (reason) {
             *reason += TfStringPrintf("Unable to determine element count "
-                "at earliest time for geom <%s>.\n", geom.GetPath().GetText());
+                "at earliest time for geom <%s>.", geom.GetPath().GetText());
         }
     }
 
@@ -865,7 +865,7 @@ UsdGeomSubset::ValidateFamily(
                     if (reason) {
                         *reason += TfStringPrintf("Indices attribute has an "
                             "odd number of elements in GeomSubset at path <%s> "
-                            "at time %s with elementType edge.\n",
+                            "at time %s with elementType edge.",
                             subset.GetPath().GetText(), TfStringify(t).c_str());
                     }
                 }
@@ -882,7 +882,7 @@ UsdGeomSubset::ValidateFamily(
                         valid = false;
                         if (reason) {
                             *reason += TfStringPrintf("Found duplicate index %d "
-                                "in GeomSubset at path <%s> at time %s.\n", index,
+                                "in GeomSubset at path <%s> at time %s.", index,
                                 subset.GetPath().GetText(), TfStringify(t).c_str());
                         }
                     }
@@ -902,7 +902,7 @@ UsdGeomSubset::ValidateFamily(
             if (reason) {
                 *reason += TfStringPrintf("Geometry <%s> has no elements at "
                     "time %s, but the \"%s\" GeomSubset family contains "
-                    "indices.\n", geom.GetPath().GetText(),
+                    "indices.", geom.GetPath().GetText(),
                     TfStringify(t).c_str(), familyName.GetText());
             }
         }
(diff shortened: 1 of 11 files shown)


pxr/usd/usdGeom/subset.cpp
```diff
@@ -749,7 +749,7 @@ UsdGeomSubset::ValidateSubsets(
                     valid = false;
                     if (reason) {
                         *reason += TfStringPrintf("Found overlapping index %d "
-                            "in GeomSubset at path <%s> at time %s.\n", index,
```
mattyjams, line 752: I think the newline characters here were a vestige from when we had planned to use this from the Python framework. We weren't consistent about including or not including the newline though, and it somewhat complicated things when trying to write the unit tests for the `usdGeom` subset families validator. I ended up just taking them all out so that clients can decide whether or not a trailing newline makes sense.

pxr/usd/usdGeom/validators.cpp
```diff
@@ -0,0 +1,136 @@
...
+    UsdValidationErrorVector errors;
+
+    for (const TfToken& subsetFamilyName : subsetFamilyNamesVec) {
+        const std::vector<UsdGeomSubset> familySubsets =
+            UsdGeomSubset::GetGeomSubsets(
+                imageable,
+                /* elementType = */ TfToken(),
+                /* familyName = */ subsetFamilyName);
+
+        // Determine the element type of the family by looking at the
```
mattyjams, line 59: I wasn't completely certain about this. Is it possible/useful for the subsets of a given family to mix element types, or must they all have the same element type? I *think* the intention is the latter given what `ValidateFamily()` does here:
https://github.com/PixarAnimationStudios/OpenUSD/blob/5332d627cb28991e812acd67b8971e87367d9191/pxr/usd/usdGeom/subset.cpp#L831

But I don't actually think it's possible to end up in the code at the moment, since the `GetGeomSubsets()` call above it specifies an `elementType` which I don't think can be empty?

pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
```diff
@@ -23,40 +23,75 @@
...
 
 void
 TestUsdShadeValidators()
 {
-    // This should be updated with every new validators added with
+    // This should be updated with every new validator added with the
     // UsdShadeValidators keyword.
-    UsdValidationRegistry &registry = UsdValidationRegistry::GetInstance();
-    UsdValidatorMetadataVector metadata = 
+    const std::set<TfToken> expectedUsdShadeValidatorNames = {
```
mattyjams, line 35: I ended up re-organizing this a little bit since we're interested in both all of the validators for the `UsdShadeValidators` keyword *and* the `UsdGeomSubset` keyword. The latter is a keyword we re-use from `usdGeom`'s validator keyword tokens.

jesschimein:
Filed as internal issue #USD-9764

mattyjams:
Rebased on the current dev branch (c42f6d8a808156bc40734575f6db41b2acccf487) and resolved conflicts in `testUsdShadeValidators.cpp`.

pxr/usd/usdGeom/validators.cpp
diyajoy: The intention is for all the subsets in a family to have the same elementType!

The elementType in `GetGeomSubsets` can be empty, in which case all subsets in the family regardless of elementType will be returned. That being said, the `GetGeomSubsets()` call does pass in elementType, so great catch about that code never being hit! I'll update this code and the test for it separately.

mattyjams: Great, thanks for clarifying @diyajoy!

Yeah, `GetGeomSubsets()` does allow for `elementType` to be empty, but the `_ValidateGeomType()` call at the top of `ValidateFamily()` basically ensures that it won't be, so our call to `GetGeomSubsets()` will be given a non-empty `elementType` and filter out all other element type subsets, preventing us from seeing whether there's a mix of element types.

Once we have that sorted out, I can add a test case for this in a follow-up PR to make sure we catch the case where a subset family contains subsets with differing element types and that we flag those as validation errors.

Thanks again!

diyajoy: Edit: Please ignore, this comment was posted before your reply above!
Actually if it's easy enough, would you mind including a fix for the `ValidateFamily` issue in this PR? It should just be passing in an empty `TfToken` for the `elementType` in the call to `GetGeomSubsets()`!

diyajoy: A follow-up PR sounds great, thank you!

Since `elementType` will always be non-empty, I'm thinking that if we pass in an empty `TfToken` directly instead of the `elementType`, we'll get all subsets in the family regardless of their type.

mattyjams: It was indeed easy enough, so I just went ahead and made the change. :)

The fix is exactly as you described in d9cc6f344ee3c537cf597bc7dc5648f4777b183b, and then I incorporated a `mixedElementTypes` subset family in the unit test in c3d111878f93754775c483504f34e8a91cbdbeb7.

Thanks!

pxr/usd/usdShade/validators.cpp
```diff
@@ -192,13 +199,160 @@ _ShaderPropertyTypeConformance(const UsdPrim &usdPrim)
...
+
+    const std::vector<UsdProperty> matBindingProperties =
+        usdPrim.GetProperties(
+            /* predicate = */ [](const TfToken& name) {
+                return TfStringStartsWith(
+                    name, UsdShadeTokens->materialBinding);
+            }
+        );
+    for (const UsdProperty& matBindingProperty : matBindingProperties) {
+        if (matBindingProperty.Is<UsdRelationship>()) {
```
tallytalwar, line 270: Will be surprising to see a material:binding property on a prim which isn't a relationship. @mattyjams what do you think? I so we can clean this bit internally.

Thanks

tallytalwar: Rather, we should have a core usdShade validator just for this ... check if a prim has a property starting with "material:binding" which isn't a relationship and if found, report a validation error.

And clean this check here.

mattyjams: Good call, thanks @tallytalwar!

I added a `MaterialBindingRelationships` validator that issues errors about non-relationship properties in 77ef087ab883ff9264236ab860b4dacb78ef87df, and a test for it in 7556b56b9c772343191abf70b2267ef614357c12.

I think we need to keep the `if (matBindingProperty.Is<UsdRelationship>()) {` here though, because this subset validator is trying to determine if there are any valid material binding relationships. If there are any properties, this particular validator doesn't care about those and wants to ignore them, but the new validator will spit errors about that.

Does that sound ok to you?

tallytalwar: Yeah that makes sense to me, thanks @mattyjams for jumping on the suggestion and implementing the new validator right away!

pxr/usd/usdShade/validators.cpp
```diff
@@ -192,13 +243,164 @@ _ShaderPropertyTypeConformance(const UsdPrim &usdPrim)
...
+                UsdShadeTokens->materialBind.GetText(),
+                materialBindFamilyType.GetText(),
+                UsdGeomTokens->nonOverlapping.GetText(),
+                UsdGeomTokens->partition.GetText())
+        );
+    }
+
+    // Check that all subsets belonging to the "materialBind" family are of
+    // element type "face", since material bindings may only be applied to
+    // geometric faces.
```
diyajoy, line 354: Since usdPhysics [also uses](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdPhysics/overview.dox#L403-L408) the materialBindingAPI, we now have use cases for binding materials to non-face subsets! Since [createMaterialBindSubsets](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdShade/materialBindingAPI.h#L909-L927) allows any elementType as well, I don't think we need this validation check anymore.

If you agree with that @mattyjams, I can go ahead and update your PR internally before pushing it!

mattyjams: Oh, interesting! Sure, that sounds good to me, thanks @diyajoy!

PR #3124, opened 2024-06-17
Autodesk: Remove the duplicate code in basisCurves.glslfx

erikaharrison-adsk:
### Description of Change(s)

Fix alpha channel is missing in ShadingTerminal function of basisCurves.glslfx.

In change 56e3997c678b154f875ada0bd16af61505e20d1a, some common shading code are moved to terminals.glslfx. But he forgot to remove the corresponding snippets in BasisCurves.Fragment of BasisCurves.glslfx. And then in the following change, the removed snippets for Curves.Fragment.Wire is incorrectly returned back at change b3e6feeb69c905b6d74e0a9d668dd9e6690fa33e. In this change, we remove the redundant code in BasisCurves.glslfx, so that ApplyColorOverrides and LightBlending will not be twicely called.

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hdSt/shaders/basisCurves.glslfx
+++ b/pxr/imaging/hdSt/shaders/basisCurves.glslfx
@@ -1186,8 +1186,6 @@ void main(void)
 #ifdef HD_HAS_displayOpacity
     color.a = HdGet_displayOpacity();
 #endif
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9765

Issue #3122, opened 2024-06-14
Convert rule to Validation Framework: StageMetadataChecker


pixar-oss:

jesschimein:
Filed as internal issue #USD-9763

PR #3125, opened 2024-06-17
feat: Add StageMetadataChecker Validator for core USD

beersandrew:
### Description of Change(s)
- adding validation for defaultPrim
- including necessary updates to associated plugInfo.json & CMakeLists.txt files

### Fixes Issue(s)
- #3122

### DIFF
--- a/pxr/usd/usd/CMakeLists.txt
+++ b/pxr/usd/usd/CMakeLists.txt
@@ -299,6 +299,13 @@ pxr_build_test_shared_lib(TestUsdResolverChangedResolver
         testenv/TestUsdResolverChangedResolver.cpp
 )
 
+pxr_build_test(testCoreUsdValidators
+        LIBRARIES
+        usd
+        CPPFILES
+        testenv/testCoreUsdValidators.cpp
+)
+
 pxr_build_test(testUsdAttributeBlockingCpp
     LIBRARIES
         tf
@@ -761,6 +768,11 @@ pxr_install_test_dir(
 pxr_install_test_dir(
     SRC testenv/testUsdZipFile.testenv
     DEST testUsdZipFile_CPP
+)
+
+pxr_register_test(testCoreUsdValidators
+        COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testCoreUsdValidators"
+        EXPECTED_RETURN_CODE 0
 )
 
 pxr_register_test(testUsdAppliedAPISchemas
--- a/pxr/usd/usd/coreValidators.cpp
+++ b/pxr/usd/usd/coreValidators.cpp
@@ -10,8 +10,6 @@
 #include "pxr/usd/usd/validatorTokens.h"
 #include "pxr/usd/usd/validator.h"
 
-//#include "pxr/usd/usdGeom/tokens.h"
-
 PXR_NAMESPACE_OPEN_SCOPE
 
 static
@@ -41,19 +39,7 @@ _GetStageMetadataErrors(const UsdStagePtr &usdStage)
                             UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
                             "Stage has missing or invalid defaultPrim.");
     }
-//    if (!usdStage->HasAuthoredMetadata(
-//            UsdGeomTokens->metersPerUnit)) {
-//        errors.emplace_back(UsdValidationErrorType::Error,
-//                            UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
-//                            "Stage does not specify its linear scale in "
-//                            "metersPerUnit.");
-//    }
-//    if (!usdStage->HasAuthoredMetadata(
-//            UsdGeomTokens->upAxis)) {
-//        errors.emplace_back(UsdValidationErrorType::Error,
-//                            UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
-//                            "Stage does not specify an upAxis.");
-//    }
+
     return errors;
 }
 
--- a/pxr/usd/usd/coreValidators.cpp
+++ b/pxr/usd/usd/coreValidators.cpp
@@ -10,28 +10,60 @@
 #include "pxr/usd/usd/validatorTokens.h"
 #include "pxr/usd/usd/validator.h"
 
+//#include "pxr/usd/usdGeom/tokens.h"
+
 PXR_NAMESPACE_OPEN_SCOPE
+
+static
+UsdValidationErrorVector
+_GetCompositionErrors(const UsdStagePtr &usdStage)
+{
+    UsdValidationErrorVector errors;
+    const PcpErrorVector pcpErrors = usdStage->GetCompositionErrors();
+    errors.reserve(pcpErrors.size());
+    for (const PcpErrorBasePtr &pcpError : pcpErrors) {
+        UsdValidationErrorSites errorSites = {
+                UsdValidationErrorSite(usdStage, pcpError->rootSite.path)};
+        errors.emplace_back(UsdValidationErrorType::Error,
+                            std::move(errorSites),
+                            pcpError->ToString());
+    }
+    return errors;
+}
+
+static
+UsdValidationErrorVector
+_GetStageMetadataErrors(const UsdStagePtr &usdStage)
+{
+    UsdValidationErrorVector errors;
+    if (!usdStage->GetDefaultPrim()) {
+        errors.emplace_back(UsdValidationErrorType::Error,
+                            UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
+                            "Stage has missing or invalid defaultPrim.");
+    }
+//    if (!usdStage->HasAuthoredMetadata(
+//            UsdGeomTokens->metersPerUnit)) {
+//        errors.emplace_back(UsdValidationErrorType::Error,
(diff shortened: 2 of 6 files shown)


jesschimein:
Filed as internal issue #USD-9770

pxr/usd/usd/coreValidators.cpp
```diff
@@ -12,26 +12,44 @@
...
 
+static
+UsdValidationErrorVector
+_GetStageMetadataErrors(const UsdStagePtr &usdStage)
+{
+    UsdValidationErrorVector errors;
+    if (!usdStage->GetDefaultPrim()) {
+        errors.emplace_back(UsdValidationErrorType::Error,
+                            UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
+                            "Stage has missing or invalid defaultPrim.");
```
tallytalwar: Maybe include the path to the rootLayer here in the reported error string.

pxr/usd/usd/plugInfo.json
```diff
@@ -148,7 +148,10 @@
                 "Validators": {
                     "CompositionErrorTest": {
                         "doc": "Validator aims at providing all composition errors, which were generated while composing the stage."
-                    }, 
+                    },
+                    "StageMetadataChecker": {
+                        "doc": "Stages that can be consumed as referencable assets should have a valid 'defaultPrim' declared."
```
tallytalwar: Its a error and not a warning, so maybe "must have a valid 'defaultPrim' specified."?

pxr/usd/usd/testenv/testCoreUsdValidators.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <iostream>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+static
+void TestCoreUsdStageMetadata()
+{
+
+    // Get stageMetadataChecker
```
tallytalwar, line 21: This is good, but I will also add a test of a valid stage which provides a defaultPrim.

tallytalwar (commented):
Thanks @beersandrew 

This looks great, just a couple of nitpicks. I will get this synced internally within Pixar code as soon as these are addressed. Really appreciate the help here.

pxr/usd/usd/CMakeLists.txt
```diff
@@ -299,6 +299,13 @@ pxr_build_test_shared_lib(TestUsdResolverChangedResolver
         testenv/TestUsdResolverChangedResolver.cpp
 )
 
+pxr_build_test(testCoreUsdValidators
```
tallytalwar, line 302: Also this makes me realize I should have included my CompositionErrorValidator test in something like this, instead of bundling it in the validationRegistry test. I will do move it internally. But thanks!

tallytalwar:
Also, I would also update this test: http://grok.pixar.com/xref/mainline/pxr/usd/usd/testenv/testUsdValidationRegistry.cpp?r=2331214#225-240 to include this StageMetadata test now.

I also realize this test should be a subset test like I did for usdShade: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp#L22-L46 (May I suggest while you are updating the test to include StageMetadata validator, to also modify it to be a subset test like with usdShade). Please let me know if this doesn't make sense and we can chat about it. Thanks

beersandrew:
> Also, I would also update this test: http://grok.pixar.com/xref/mainline/pxr/usd/usd/testenv/testUsdValidationRegistry.cpp?r=2331214#225-240 to include this StageMetadata test now.
> 
> I also realize this test should be a subset test like I did for usdShade: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp#L22-L46 (May I suggest while you are updating the test to include StageMetadata validator, to also modify it to be a subset test like with usdShade). Please let me know if this doesn't make sense and we can chat about it. Thanks

I think this is what you meant? : https://github.com/PixarAnimationStudios/OpenUSD/pull/3125/commits/be2078039b0c3fa9f9f4a7b8e48759266029609a

I'm getting an error when I run the validator registry test locally before making that change:

`testUsdValidationRegistry crashed. FATAL ERROR: Failed axiom: ' !PlugRegistry::GetInstance().RegisterPlugins(testDir).empty() '`

Let me know if this is what you meant and if not I may need help getting my local setup to pass the `testUsdValidationRegistry` test to make sure my changes are working before committing

pxr/usd/usd/coreValidators.cpp
beersandrew: Updated

pxr/usd/usd/plugInfo.json
beersandrew: Updated

pxr/usd/usd/testenv/testCoreUsdValidators.cpp
beersandrew: Updated

pxr/usd/usd/testenv/testUsdValidationRegistry.cpp
```diff
@@ -310,6 +294,27 @@ void TestUsdValidationRegistry()
...
+    // The following test keeps track of all the available validators within
+    // UsdCoreValidators keyword, hence as new validators are added under
+    // this keyword this unit test will have to be updated.
+    const UsdValidatorMetadataVector coreValidatorMetadata =
+            registry.GetValidatorMetadataForKeyword(
+                    UsdValidatorKeywordTokens->UsdCoreValidators);
+    TF_AXIOM(coreValidatorMetadata.size() == 1);
+    const std::vector<TfToken> expectedValidatorNames =
+            {UsdValidatorNameTokens->compositionErrorTest,
+             UsdValidatorNameTokens->stageMetadataChecker};
```
tallytalwar, line 317: Thanks. But since you are touching this code, can I request you to update this to the subset check instead of equality check (Similar to this https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp#L22-L46).

Reasoning being, if a client is running this test in their own environment and they have validators available in their plugins which are listing "UsdCoreValidators" as a keyword, the equality will fail, so all we can do here is to make sure the validator provides by openUSD under UsdCoreValidators keyword are found.

tallytalwar: And you mentioned the following assert is failing for you?

`TF_AXIOM(!PlugRegistry::GetInstance().RegisterPlugins(testDir).empty());`

- This could be because the "pxr/usd/usd/testenv/testUsdValidationRegistry" dir is not being installed property for you .. I don't know why that will be though.
- Are you running the test via ctest? 
- The test uses [ArchGetCwd()](https://github.com/PixarAnimationStudios/OpenUSD/blob/be2078039b0c3fa9f9f4a7b8e48759266029609a/pxr/usd/usd/testenv/testUsdValidationRegistry.cpp#L322C33-L322C43), which should have returned the path where testUsdValidationRegistry is installed.

I checked all build configurations internally and this test does pass in all.

Please let me know if you still need help with this.

tallytalwar (commented):
Provided notes on testUsdValidationRegistry inline. Let me know (here or on slack) if you are still running into test failure without your change. Thanks

pxr/usd/usd/testenv/testUsdValidationRegistry.cpp
beersandrew: updated to do the subset check rather than equality, sorry i missed that the first time around.

following up on slack about testing setups

pxr/usd/usd/testenv/testCoreUsdValidators.cpp
```diff
@@ -0,0 +1,59 @@
...
+    UsdValidationRegistry &registry = UsdValidationRegistry::GetInstance();
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdValidatorNameTokens->stageMetadataChecker);
+    TF_AXIOM(validator);
+
+    // Create an empty stage
+    SdfLayerRefPtr rootLayer = SdfLayer::CreateAnonymous();
+    UsdStageRefPtr usdStage = UsdStage::Open(rootLayer);
```
tallytalwar, line 29: You could have directly done UsdStage::CreateInMemory here, and then you do not need to use the SdfLayer at all.

tallytalwar (approved):
Just 1 tiny cleanup, and all great. Thanks for the patience and doing this work @beersandrew .

pxr/usd/usd/testenv/testCoreUsdValidators.cpp
beersandrew: I wanted to create a layer here to verify that the root layer string is in the error message (used lower on line 40). when I did it without a root layer, the error message prints "<>" which didn't feel like a great test. Happy to remove the layer if you'd like

tallytalwar: I see, yeah makes sense. Sorry I missed the check on line 40.

pxr/usd/usd/testenv/testCoreUsdValidators.cpp
tallytalwar, line 1: Sorry noticed it when internalizing. (Taking care of it internally). But FYI

As a convention we prefix the test with the library name: so testUsdCoreValidators instead of testCoreUsdValidators.

Issue #3113, opened 2024-06-06
Capsule_1 rendering breaks when top/bottom radii are different


crydalch:
### Description of Issue
The new `Capsule_1` primitive seems to have a issues drawing correctly. Compared to adjusting radiusTop and radiusBottom with `Cylinder_1`, the new cylinder prim has unexpected results that look buggy.

### Steps to Reproduce
1. Download attachment from
[Bug_Capsule_1.usda.zip](https://github.com/user-attachments/files/15686610/Bug_Capsule_1.usda.zip)
, and open in `usdview`
2. With radiusBottom of 1.0 and radiusTop of 0.3 on both primitives, it seems reasonable to expect the capsule to resemble the cylinder, except for the rounded ends. But it looks unexpectedly different, maybe broken:
![image](https://github.com/PixarAnimationStudios/OpenUSD/assets/458856/a69d0768-794e-48bb-ace5-0517d06bca6b)

3. Open up an Interpeter window, to use for next steps
4. In usdview, select _primitive1_, and then select its *radiusBottom* property, and in the interpreter window, run `usdviewApi.property.Set(0.3)`
5. Then select its *radiusTop* property, and in the interpreter window, run `usdviewApi.property.Set(1.0)`
6. This looks entirely broken, like there are assumptions about the values/order of radiusTop and radiusBottom:
![image](https://github.com/PixarAnimationStudios/OpenUSD/assets/458856/3169b2b3-07d6-4358-b9f0-48ecab8813f3)

### Package Versions

USD 24.05

spiffmon:
Thanks, @crydalch - do you get the same results with Hydra 2 as with default Hydra 1?

crydalch:
Hi @spiffmon!

Yes, it appears the same with `USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1`

jesschimein:
Filed as internal issue #USD-9737

marktucker:
Is this something that is already being worked on by Pixar? I was thinking of tackling this, but didn't want to duplicate any work that may already be ongoing.

marktucker:
This was easier than I thought... PR incoming soon. I'll just cross my fingers that there hasn't been any duplicated effort.

pixar-oss:
Thank you @marktucker!  We had not started on this yet, so the PR is very welcome!

crydalch:
Closing as this has been fixed. Thank you!

PR #3129, opened 2024-06-19
Fix the code for calculating mesh points on Capsule_1 prims

marktucker:
Fix the code for calculating mesh points on Capsule_1 prims with different
top and bottom radii. The code to calculate the angle between the spheres was incorrect.

### Description of Change(s)
The code used to use acos(rdiff/height) to calculate the angle between the top and bottom spheres, but rdiff/height is opposite/adjacent for the angle of interest, and so needed to use atan to return the correct value.

### Fixes Issue(s)
- #3113 


jesschimein:
Filed as internal issue #USD-9778

PR #3133, opened 2024-06-20
Fix Tf_RegistryManagerImpl singleton instance not being deleted

moshev:
### Description of Change(s)

Delete the dynamically allocated Tf_RegistryManagerImpl singleton instance when the static TfRegistryManager singleton instance which uses it is destroyed.

### Fixes Issue(s)
- 3088

### DIFF
--- a/pxr/base/tf/registryManager.cpp
+++ b/pxr/base/tf/registryManager.cpp
@@ -170,6 +170,10 @@ public:
          return TfSingleton<Tf_RegistryManagerImpl>::GetInstance();
     }
 
+    static void DeleteInstance() {
+        TfSingleton<Tf_RegistryManagerImpl>::DeleteInstance();
+    }
+
     /// Stores the active library's registration functions and runs those
     /// that are subscribed to then makes no library active. Returns an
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9782

meshula:
Hi @moshev !  I think you meant to make this PR against the dev branch, not the release branch. That is why there are a couple of hundred extra commits in it :)  Could you try to recreate it against dev please?

moshev:
> Hi @moshev ! I think you meant to make this PR against the dev branch, not the release branch. That is why there are a couple of hundred extra commits in it :) Could you try to recreate it against dev please?

Oops. Yes, of course.

moshev:
I changed the base to dev branch in this review, let me know if there's still an issue.

meshula:
Much better :) Thanks!

spiffmon:
Hi @moshev , we don't think we can accept this PR.  Nothing in OpenUSD currently deletes the singleton, and it would not be safe, generally, to do so, in our codebase.  Of particular concern is static destruction ordering issues, as there are many other destructors that may access the registry.

PR #3134, opened 2024-06-20
feat: adding stageMetadata validator for usdGeom

beersandrew:
### Description of Change(s)
- Part 2 of #3122, including upAxis & metersPerUnit checks in a new usdGeom validator structure
- added unit tests
- added associated changes to plugInfo.json & CMakeLists.txt

### Fixes Issue(s)
- #3122 

### DIFF
--- a/pxr/usd/usdGeom/plugInfo.json
+++ b/pxr/usd/usdGeom/plugInfo.json
@@ -393,7 +393,7 @@
                 },
                 "Validators": {
                     "StageMetadataChecker": {
-                        "doc": "All stages should declare their 'upAxis' and 'metersPerUnit' and stages meant for consumer-level packaging should always have upAxis set to 'Y'."
+                        "doc": "All stages must declare their 'upAxis' and 'metersPerUnit'."
                     },
                     "keywords": [
                         "UsdGeomValidators"
--- a/pxr/usd/usdGeom/plugInfo.json
+++ b/pxr/usd/usdGeom/plugInfo.json
@@ -390,6 +390,14 @@
                         ], 
                         "schemaKind": "abstractTyped"
                     }
+                },
+                "Validators": {
+                    "StageMetadataChecker": {
+                        "doc": "All stages should declare their 'upAxis' and 'metersPerUnit' and stages meant for consumer-level packaging should always have upAxis set to 'Y'."
+                    },
+                    "keywords": [
+                        "UsdGeomValidators"
+                    ]
                 }
             }, 
             "LibraryPath": "@PLUG_INFO_LIBRARY_PATH@", 
--- a/pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
+++ b/pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
@@ -1,0 +1,62 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the terms set forth in the LICENSE.txt file available at
+// https://openusd.org/license.
+//
+
+#include "pxr/usd/usd/validator.h"
+#include "pxr/usd/usd/validationError.h"
+#include "pxr/usd/usdGeom/validatorTokens.h"
+#include "pxr/usd/usd/validationRegistry.h"
+#include "pxr/usd/usdGeom/metrics.h"
+#include "pxr/usd/usdGeom/tokens.h"
+
+#include <iostream>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+static
+void TestUsdStageMetadata()
+{
+    // Get stageMetadataChecker
+    UsdValidationRegistry &registry = UsdValidationRegistry::GetInstance();
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdGeomValidatorNameTokens->stageMetadataChecker);
+    TF_AXIOM(validator);
+
+    // Create an empty stage
+    UsdStageRefPtr usdStage = UsdStage::CreateInMemory();
+
+    UsdValidationErrorVector errors = validator->Validate(usdStage);
+
+    // Verify both metersPerUnit and upAxis errors are present
+    TF_AXIOM(errors.size() == 2);
+    TF_AXIOM(errors[0].GetType() == UsdValidationErrorType::Error);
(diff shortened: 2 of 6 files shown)


jesschimein:
Filed as internal issue #USD-9787

pxr/usd/usdGeom/plugInfo.json
```diff
@@ -390,6 +390,14 @@
                         ], 
                         "schemaKind": "abstractTyped"
                     }
+                },
+                "Validators": {
+                    "StageMetadataChecker": {
+                        "doc": "All stages should declare their 'upAxis' and 'metersPerUnit' and stages meant for consumer-level packaging should always have upAxis set to 'Y'."
```
tallytalwar: The secondary bits of "consumer-level packaging should always have upAxis set to 'Y', wasnt that a arkit rule requirement from ComplianceChecker. I would remove this statement. (https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L213-L217)

It looks like you code is also not doing this check, which is correct!

pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
```diff
@@ -0,0 +1,62 @@
...
+    TF_AXIOM(validator);
+
+    // Create an empty stage
+    UsdStageRefPtr usdStage = UsdStage::CreateInMemory();
+
+    UsdValidationErrorVector errors = validator->Validate(usdStage);
+
+    // Verify both metersPerUnit and upAxis errors are present
+    TF_AXIOM(errors.size() == 2);
+    TF_AXIOM(errors[0].GetType() == UsdValidationErrorType::Error);
```
tallytalwar: Other than GetMessage, you can probably loop to size of errors. And then for GetMessage you can explicitly check the 0th and 1st index?

pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
```diff
@@ -0,0 +1,62 @@
...
+    TF_AXIOM(errors[1].GetMessage() == expectedUpAxisErrorMsg);
+
+    // Fix the errors
+    UsdGeomSetStageMetersPerUnit(usdStage, 0.01);
+    UsdGeomSetStageUpAxis(usdStage, UsdGeomTokens->y);
+
+    errors = validator->Validate(usdStage);
+
+    // Verify the errors are fixed
+    TF_AXIOM(errors.size() == 0);
```
tallytalwar, line 57: Nice!

pxr/usd/usdGeom/validators.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "pxr/usd/usd/validationError.h"
+#include "pxr/usd/usd/validationRegistry.h"
+#include "pxr/usd/usdGeom/validatorTokens.h"
+#include "pxr/usd/usdGeom/tokens.h"
+#include "pxr/usd/usd/validator.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+    static
```
tallytalwar: Extra indentation here.

pxr/usd/usdGeom/validators.cpp
```diff
@@ -0,0 +1,45 @@
...
+    UsdValidationErrorVector
+    _GetStageMetadataErrors(const UsdStagePtr &usdStage)
+    {
+        UsdValidationErrorVector errors;
+        if (!usdStage->HasAuthoredMetadata(
+                UsdGeomTokens->metersPerUnit)) {
+            errors.emplace_back(UsdValidationErrorType::Error,
+                                UsdValidationErrorSites{UsdValidationErrorSite(usdStage, SdfPath("/"))},
+                                "Stage does not specify its linear scale in "
```
tallytalwar: Might make sense to provide the rootLayer of the Stage in the error string.

tallytalwar (commented):
Thanks @beersandrew 

Left some notes. Also might make sense to also add a test which is similar to what I have in usdShade: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp#L22-L46

Will sync this in as soon as these notes are addressed. Thanks.

pxr/usd/usdGeom/plugInfo.json
beersandrew: good catch, updated

pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
beersandrew: added these messages to their own list as well so everything can be done in the loop. let me know what you think

pxr/usd/usdGeom/validators.cpp
beersandrew: updated

pxr/usd/usdGeom/validators.cpp
beersandrew: updated

tallytalwar (approved):
Syncing this in.

pxr/usd/usdGeom/testenv/testUsdGeomValidators.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    // Verify both metersPerUnit and upAxis errors are present
+    TF_AXIOM(errors.size() == 2);
+    auto rootLayerIdentifier = rootLayer->GetIdentifier().c_str();
+    std::vector<std::string> expectedErrorMessages = {
+            TfStringPrintf("Stage with root layer <%s> does not specify its linear scale in metersPerUnit.", rootLayerIdentifier),
+            TfStringPrintf("Stage with root layer <%s> does not specify an upAxis.", rootLayerIdentifier)
+    };
+
+    for(int i = 0; i < errors.size(); ++i)
```
tallytalwar, line 42: size() returns a size_t (unsigned_int) and the index here is an int. (I corrected this in the internal sync already, just fyi).

tallytalwar:
synced in, will be available in the next dev push after the release. Thanks a lot @beersandrew for this PR and other validator contributions.

PR #3137, opened 2024-06-21
Fix compiler warning return type is int and trying to return a size_t

evanburkert:
### Fixes Issue(s)
- Fixes compiler warning pxr\usd\ndr\property.h(108,30): warning C4267: 'return': conversion from 'size_t' to 'int', possible loss of data


jesschimein:
Filed as internal issue #USD-9792

PR #3142, opened 2024-06-27
Apple: Enable the use of ccache and other compiler cache utilities

dgovil:
### Description of Change(s)

This PR adds the ability to use a compiler cacher like [ccache](https://ccache.dev) when building USD.
This can be a huge speed boost when iterating on USD. Lee Kerley had added this internally and it's been a huge accelerant for us. It can make iterative builds take a fraction of the time, after the first build.

I control this with a boolean option `PXR_ENABLE_COMPILER_CACHE` which is OFF by default. This would be harmless to leave on by default, but msvc is considered a secondary tier, so I turned it off.

There is also `PXR_COMPILER_CACHE_NAME` which allows a developer to override the name of the compiler cache program. This can be useful if a team is using [sccache](https://github.com/mozilla/sccache) which is a distributed version of ccache, but some larger companies sometimes have their own drop in versions of ccache.

If the option is enabled and the compiler cache is not found, it just prints a status message but does not impede anything.


jesschimein:
Filed as internal issue #USD-9809

pmolodo:
This would be awesome - I use ccache on Linux for my OpenUSD development builds, and it helps a lot when iterating.  Adding ccache support directly in the CMake files would simply our ccache setup.

Also - I've recently been experimenting with sccache on Windows, and so far it's been helping / working.  It does require:
- turning off pre-compiled headers (PCH)
  - no big loss, since it's trying to solve the same iterative-build-time issue as sccache, but not as efficiently
- changing settings for debug builds
  - to embed debug symbols in the `.o` files, instead of in separate external debug `.pdb` databases which are shared by multiple compilation units
  - note that the per-dll `.pdb` databases can still be generated, but later in the process, at link time
  
In theory, you can also use ccache on windows, but MSVC is only [B level /  Partially Supported by ccache](https://ccache.dev/platform-compiler-language-support.html), whereas it's [fully supported by sccache](https://github.com/mozilla/sccache?tab=readme-ov-file#usage).

meshula:
@pmolodo good to know it can help on Windows as well. I'm curious, if debug symbols are embedded instead of in .pdb files, is there any difference in disk footprint? Or does the .o simply increase as if the pdb itself was embedded?

dgovil:
@meshula I've rebased this PR on latest dev and added the build flags to build_usd.py

It's enabled by default for macOS and Linux, but can be disabled. If it is not found, it will continue as normal but emit a message.

PR #3146, opened 2024-06-28
Apple: Usdrecord C++ port

ld-kerley:
### Description of Change(s)
C++ port of usdrecord.

There are a few minor changes in the command line arguments due to CLI11 not supporting arguments of the form `-cam`, instead we provide `--cam`. 

GLFW is used to provide an OpenGL context on non macOS platforms, macOS doesn't require an OpenGL context to render with usdrecord.

`build_usd.py` has been updated to build GLFW when required. 

Tests run and pass on both macOS and linux.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9815

dgovil:
@jesschimein sorry, had a snafu and I'll get Lee added to our CLA asap. Might be a hot minute just because the necessary folks are on vacation.

pxr/usdImaging/bin/usdrecord/usdrecord.cpp
```diff
@@ -0,0 +1,709 @@
...
+            // otherwise the entire original filename is in the prefix string.
+            std::stringstream ss;
+            ss << frameNumberFormater;
+            ss << timeCode << "." << outputImagePathSuffix;
+            outputImagePath += ss.str();
+        }
+
+        try {
+            frameRecorder->Record(usdStage, usdCamera, timeCode, outputImagePath);
+        } catch (TfBaseException e) {
```
nvmkuruc: Should this be a reference?
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e15-throw-by-value-catch-exceptions-from-a-hierarchy-by-reference

pxr/usdImaging/bin/usdrecord/usdrecord.cpp
```diff
@@ -0,0 +1,709 @@
...
+                      << e.what() << std::endl;
+            return 1;
+        }
+    }
+
+    // Release our reference to the frame recorder so it can be deleted before other resources are freed
+    frameRecorder = nullptr;
+
+#if PXR_USE_GLFW
+    glfwTerminate();
```
nvmkuruc: I noticed there's a few return code paths that don't terminate the `glfw` context.

I'm wondering if an RAII pattern could be used here so that the `glfw` context is always terminated.

nvmkuruc (commented):
It's great to have a version of `usdrecord` available with less dependencies.

I was wondering if it's possible to provide a deprecation warning for users using the `-cam` argument if it's no longer going to be supported.

pxr/usdImaging/bin/usdrecord/usdrecord.cpp
ld-kerley: fixed - thanks

pxr/usdImaging/bin/usdrecord/usdrecord.cpp
ld-kerley: I added a RAII container for the GLFW resources - I'm not totally convinced if this is the best route for usdrecord, if thats the only app that needs to manage this OpenGL context creation in C++.  It's a lot more code to add, instead of just fixing up the missing `glfwTerminate()`.  I do understand the RAII approach being better in general, but here maybe the application is simple enough?  

Anyway, I added it so we can compare approaches, and it's easy enough to revert if thats deemed preferred.  I don't really have strong opinions either way.

ld-kerley:
I think any `-cam` deprecation warning would need to be in a separate PR applied to `usdrecord.py`, and would only be useful if released before this C++ port. 

CLI11 doesn't not parse "short" options with more than one character, so we can't have `-cam`, but I have added `--cam` as the closest fallback.

dgovil:
I wonder if we should include a `-cam` deprecation warning in the changelog for 24.8? 
Technically, even in argparse, multi character short-commands are bug prone (-cam is ambiguously -c -a -m or -cam)

and yeah, like Lee suggests, it would have to be deprecated on the Python side. @meshula or @tcauchois do you have any thoughts?

pxr/usdImaging/bin/usdrecord/usdrecord.cpp
nvmkuruc: It might be a reasonable case for a generalized "finally" utility in `Tf`.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction

nvmkuruc:
@dgovil @ld-kerley Could you check for `-cam` in `argv` and output a warning before passing it to `cli11`?

Issue #3100, opened 2024-05-29
usdMtlx does not consistently preserve MaterialX node names


dgovil:
### Description of Issue

When I convert a `.mtlx` document to USD, the resulting graph does not preserve the name.

Take this example here and note that the `UsdPreviewSurface` node has a specified name `SR_brass1`.
Also note that the `tileimage` nodes both have a name `image_roughness` and `image_color`.

```xml
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
  <nodegraph name="NG_brass1" fileprefix="../../../Images/">
    <tiledimage name="image_color" type="color3">
      <input name="file" type="filename" value="brass_color.jpg" colorspace="srgb_texture" />
      <input name="uvtiling" type="vector2" value="1.0, 1.0" />
    </tiledimage>
    <tiledimage name="image_roughness" type="float">
      <input name="file" type="filename" value="brass_roughness.jpg" />
      <input name="uvtiling" type="vector2" value="1.0, 1.0" />
    </tiledimage>
    <output name="out_color" type="color3" nodename="image_color" />
    <output name="out_roughness" type="float" nodename="image_roughness" />
  </nodegraph>
  <UsdPreviewSurface name="SR_brass1" type="surfaceshader">
    <input name="diffuseColor" type="color3" nodegraph="NG_brass1" output="out_color" />
    <input name="metallic" type="float" value="1" />
    <input name="roughness" type="float" nodegraph="NG_brass1" output="out_roughness" />
  </UsdPreviewSurface>
  <surfacematerial name="USD_Tiled_Brass" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_brass1" />
  </surfacematerial>
</materialx>
```

When converted to USD, I get the following hierarchy (from usdtree)

```
/
 `--MaterialX [def]
     |--Materials [def]
     |   `--USD_Tiled_Brass [def Material]
     |       |--ND_UsdPreviewSurface_surfaceshader [def Shader]
     |       `--NG_brass1 [def NodeGraph]
     |           |--image_color [def Shader]
     |           `--image_roughness [def Shader]
     |--Shaders [def]
     |   `--ND_UsdPreviewSurface_surfaceshader [def Shader]
     `--NodeGraphs [def]
         `--NG_brass1 [def NodeGraph]
             |--image_color [def Shader]
             `--image_roughness [def Shader]
```

[...]

PR #3147, opened 2024-06-28
Apple: Use Node name instead of NodeDef name for SurfaceShaders

dgovil:
### Description of Change(s)

Change the translated name of surface shaders from the node defs name to the Node's name, toggleable by setting an env var of `PXR_MTLX_SUPPORT_MATERIAL_INHERITS`

UsdMtlx was using the node defs name for surface shaders as an attempt to support [material inheritance](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#material-inheritance), the logic was using the most generic name possible to allow easier overriding.

Unfortunately this causes issues where the imported names for just the surface shader does not match the input mtlx file.

While I understand the original rationale, I think that until we find a better heuristic, there is more value in keeping the naming consistent and intuitive. I do not believe material inheritance is a common feature in MtlX libraries, and certainly does not feature in the set of mtlx files provided by the MaterialX library or with OpenUSD.

For pipelines that require this, I suggest they instead set the PXR_MTLX_SUPPORT_MATERIAL_INHERITS boolean env var, until a better heuristic can be found.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3100

### DIFF
--- a/pxr/usd/usdMtlx/reader.cpp
+++ b/pxr/usd/usdMtlx/reader.cpp
@@ -26,6 +26,7 @@
 #include "pxr/usd/usd/editContext.h"
 #include "pxr/usd/usd/specializes.h"
 #include "pxr/usd/usd/stage.h"
+#include "pxr/base/tf/getenv.h"
 #include "pxr/base/tf/stringUtils.h"
 #include "pxr/base/tf/pathUtils.h"
 
@@ -1511,15 +1512,22 @@ _Context::AddShaderNode(const mx::ConstNodePtr& mtlxShaderNode)
         return UsdShadeShader();
     }
 
-    // Choose the name of the shader.  In MaterialX this is just
+    // Choose the name of the shader.
+    //
+    // In MaterialX this is just
     // mtlxShaderNode->getName() and has no meaning other than to uniquely
-    // identify the shader.  In USD to support materialinherit we must
-    // ensure that shaders have the same name if one should compose over
+    // identify the shader.  In USD to support Material Inheritance
+    // https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#material-inheritance
+    // we must ensure that shaders have the same name if one should compose over
     // the other.  MaterialX composes over if a shader node refers to the
-    // same nodedef so in USD we use the nodedef's name.  This name isn't
+    // same nodedef so in USD we use the nodedef's name.  This name wasn't
     // ideal since it's just an arbitrary unique name;  the nodedef's
-    // node name is more meaningful.
-    const auto name = _MakeName(mtlxNodeDef);
+    // node name was more meaningful.
+    //
+    // However this naming isn't always intuitive for conversions.
+    // Therefore you may toggle between them with this env variable.
+    const auto name = (TfGetenvBool("PXR_MTLX_SUPPORT_MATERIAL_INHERITS", false)) ?
+        _MakeName(mtlxNodeDef) : _MakeName(mtlxShaderNode);
 
     // Create the shader if it doesn't exist and copy node def values.
     auto shaderImplPath = _shadersPath.AppendChild(name);
--- a/pxr/usd/usdMtlx/reader.cpp
+++ b/pxr/usd/usdMtlx/reader.cpp
@@ -26,7 +26,6 @@
 #include "pxr/usd/usd/editContext.h"
 #include "pxr/usd/usd/specializes.h"
 #include "pxr/usd/usd/stage.h"
-#include "pxr/base/tf/getenv.h"
 #include "pxr/base/tf/stringUtils.h"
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9816

dgovil:
I confirmed with the MaterialX project that Material Inherits are not a supported feature in MaterialX currently. 
Based on that, I wonder if I should remove the env-var switch here and just use the new proposed behaviour? Or should we keep the switch in case people need to transition to the new naming scheme?

spiffmon:
I personally like the idea of just improving the behavior unconditionally, but we're also not using MaterialX in production yet, so no archival issues or overrides to worry about.  Maybe ask the question on the forum and the slack channel Dhruv, with the alternative being to default the env var to the _new_ behavior, still allowing folks to disable it if they do have important overrides on older translations of still-live mtlx files?

dgovil:
Will do, though just to clarify, the PR does have the default for the env var to OFF, in which case the new behaviour is picked by default.

JGamache-autodesk:
I plan to go have a look as well in that usdMtlx codebase because we came up with a list of things we want to either fix or implement differently, and in the pre-investigations I have checked to see if the registry and or syntax allows multiple readers (it doesn't) and if there was a way to pass parameters to the loader to make this explicit instead of implicit (there is, look for `SDF_FORMAT_ARGS`).

I think having these arguments saved in the USD data would solve the issue when loading a stage without having set the right combination of env vars, and would solve the composition problem of assembling assets from different studios requiring different combination of usdMtlx env vars.

So, my recommendation on this PR would be to allow passing a `materialInherits` format argument in the referencing call itself. I think there is an implementation example in the [Alembic reader](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/plugin/usdAbc/testenv/testUsdAbcSDFArguments/testUsdAbcSDFArguments.usda#L8), and I have seen it used also in some of the [file format plugins](https://github.com/adobe/USD-Fileformat-plugins/tree/main/gltf#file-format-arguments) managed by Adobe.

@spiffmon, does this make sense?

dgovil:
@JGamache-autodesk in general I like your idea but for this specific issue I have a few thoughts. 

I think we should aim to remove the old behaviour in this case and use the env var to help people transition over a predetermined period. 

I worry that if it's a file format arg, that means we're encouraging keeping the old behaviour around for longer. 

Since MaterialX itself doesn't support this feature, my thinking is that we do the following:

1. Switch to the new behaviour by default, and have the env variable to help people transition. 

2. Eventually remove the environment variable all together. 

3. If MaterialX introduces material inherits for real, we evaluate then whether it needs a change to the conversion and if so, introduce a file format argument at that time.

spiffmon:
Deciding which approach to take seems to depend on whether we think the MaterialX-USD-integration is mature enough that people are expecting the "archival document" promise of MaterialX to apply to MaterialX-consuming-USD documents, as well?  I don't know the answer to that question...

tcauchois:
If we're hoping to fully delete the old behavior, it seems better to keep this out of assets, and avoid plumbing it through.

While I appreciate the utility of environment variables in facilitating a soft transition, they also complicate deployment and so I'm trying to crack down on some of ours internally.  Two questions then:
(1) Is there a process that would give us confidence in a variant of this change where we just switch to the new behavior with no rollback flag?  i.e. asking relevant companies, posting on the forums, bringing it up in an ASWF meeting?
(2) If #1 feels too risky, can we change the env var to USDMTLX_OLD_SURFACE_NODE_NAMES (or something) and put in a // deprecated label and a warning that it will go away in 25.02 or 25.05?

It seems like the benefit of the nodedef names is that, whenever material inheritance is added to MaterialX that will facilitate that feature in UsdMtlx, right?  Is the benefit of node names just that they are slightly less arcane in the shadergen?  I wonder if we could pass some off-label node data with the node name and add that as a comment somewhere in the shader or something instead.  (Please excuse my shallow understanding of the technical meat here...)

dgovil:
We had discussed it in the MaterialX meetings and the consensus seems to be that the new behaviour is good.

Nobody could talk to whether the old behaviour needed to stay for deprecation reasons. It would only come up if someone was referencing a MTLX file and also applying an override. 

I think pragmatically, it's safe to remove the env var and just move to the new behaviour. UsdMtlx currently doesn't have a wide adoption rate in pipelines, and the deprecation period is something people would ignore. 

If that sounds reasonable, I can update the PR to remove the old behaviour.

tcauchois:
This is definitely personal judgement here, and I'm fine going with your change either way/going with a consensus, now that I have a better understanding of the context, but I'd feel slightly better dropping the env var and am also happy to be on record about that and assume the risk.  Let me know if you intend to rev it and I can hold off on checking it in.

dgovil:
Sounds good to me. I just updated with a commit to remove the env var and previous behaviour.

PR #3148, opened 2024-07-03
Apple: Added HioImageIO as optional feature on macOS

AlexSchwank:
### Description of Change(s)

Added HioImageIO as optional feature to replace OIIO with ImageIO on macOS. Credit goes to Nicholas Blasingame.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9827

jesschimein:
Disregard the note above about missing a signed CLA; we do have your CLA on file, Alex.   Sorry for any confusion!

build_scripts/build_usd.py
```diff
@@ -2058,6 +2064,10 @@ def InstallUSD(context, force, buildArgs):
                       help="Build OpenImageIO plugin for USD")
 subgroup.add_argument("--no-openimageio", dest="build_oiio", action="store_false",
                       help="Do not build OpenImageIO plugin for USD (default)")
+if MacOS():
+    group.add_argument("--imageio", dest="build_imageio", action="store_true", 
+                      default=False,
+                      help="Build ImageIO plugin for USD (Uses ImageIO.framework). Overrides OpenImageIO")
```
spiffmon: "Overrides" is a little vague, since it could refer to runtime behavior.  Instead, could we say "Disables building OpenImageIO" ?

dgovil:
I've updated the PR:

1. Rebased on latest `dev` commits
2. It is now enabled by default and gets disabled when OpenImageIO is enabled. This feels more intuitive to me.
3. It now better handles textures inside packages based on the HioAVIF code.

build_scripts/build_usd.py
```diff
@@ -2367,6 +2380,8 @@ def __init__(self, args):
         self.buildOIIO = ((args.build_oiio or (self.buildUsdImaging
                                                and self.buildTests))
                           and not embedded)
+        if MacOS():
+            self.buildImageIO = args.build_imageio and not self.buildOIIO
```
tcauchois: Can we change this so that ImageIO and OpenImageIO don't turn each other off?  i.e. get rid of the "and not self.buildOIIO" here...

build_scripts/build_usd.py
```diff
@@ -2154,6 +2160,13 @@ def InstallUSD(context, force, buildArgs):
                       help="Build OpenImageIO plugin for USD")
 subgroup.add_argument("--no-openimageio", dest="build_oiio", action="store_false",
                       help="Do not build OpenImageIO plugin for USD (default)")
+if MacOS():
+    group.add_argument("--imageio", dest="build_imageio", action="store_true", 
+                      default=True,
+                      help="Build the ImageIO.framework plugin for USD (default). "
+                           "Will be turned off if the OpenImageIO plugin is enabled")
```
tcauchois: Comment update.

pxr/imaging/plugin/CMakeLists.txt
```diff
@@ -5,7 +5,10 @@ endif()
 if (${PXR_BUILD_PRMAN_PLUGIN} AND "${RENDERMAN_VERSION_MAJOR}" GREATER "22")
     add_subdirectory(hdPrmanLoader)
 endif()
+
 if (PXR_BUILD_OPENIMAGEIO_PLUGIN)
     add_subdirectory(hioOiio)
+elseif (APPLE AND PXR_BUILD_IMAGEIO_PLUGIN)
```
tcauchois: elseif -> if

pxr/imaging/plugin/hioImageIO/plugInfo.json
```diff
@@ -0,0 +1,20 @@
+{
+  "Plugins": [
+    {
+      "Info": {
+        "Types": {
+          "HioImageIO_Image" : {
+            "bases": ["HioImage"],
+            "imageTypes": ["tif", "tiff", "zfile", "tx"],
+            "precedence": 10
```
tcauchois: If we build both OpenImageIO and ImageIO, we choose the plugin to use based on precedence.  Since hioOiio is "10", we could bump this to like "15" and then it would be used preferentially.

tcauchois (commented):
Hey all,

I'm going to try to land this momentarily but we had a few requested changes to the CMake.  If you're ok with the changes but would like me to make them I'm happy to do so as I pull it down.  Alternatively, if you want me to hold on while you rev the PR let me know.

Thanks!

build_scripts/build_usd.py
dgovil: Sure, would you like me to document somewhere that they should ideally not be used together? Or just let Hio pick one itself at runtime?

dgovil:
Either way works for me! I can take a pass in the morning so you don't have to. I think your changes all sound good to me and simplify things

build_scripts/build_usd.py
dgovil: I ended up going with your suggestion below to make ImageIO take precedence.

dgovil:
Okay, I've rebased on latest dev + addressed your changes.
I also realized that the original PR had `zfile` and `tx` listed, which `sips --formats` does not list as supported, so I removed them. It's now just the two tif extensions.

pxr/imaging/plugin/hioImageIO/imageIOImage.mm
```diff
@@ -0,0 +1,891 @@
...
+            return false;
+        }
+    }
+    
+    CGColorSpaceRef colorSpace;
+    CGBitmapInfo bitmapInfo;
+    if (IsColorSpaceSRGB()) {
+        // if this is recognized as sRGB then set the colorSpace and bitmapInfo accordingly
+        // this is for cases in which the input format may be CMYK,
+        // which OpenImageIO converts to rgb when ImageSpec's attribute oiio:RawColor is false (default)
```
tcauchois:
```suggestion
```

Delete OIIO comment that I assume doesn't apply here.

pxr/imaging/plugin/hioImageIO/imageIOImage.mm
```diff
@@ -0,0 +1,891 @@
...
+        case HioFormatInt32Vec4:
+        case HioFormatFloat32Vec4:
+            return 4;
+        default:
+            TF_CODING_ERROR("Unsupported type");
+            return 4;
+    }
+}
+
+// For compatability with Ice/Imr we transmogrify some matrix metadata
```
tcauchois:
```suggestion
// For compatability with legacy metadata keys we transmogrify some matrix metadata
```

Ice/Imr are Pixar-internal libraries, though these metadata keys also show up in OIIO and OpenEXR.  Anyway, hopefully makes things less confusing down the road.

tcauchois:
Two more documentation suggestions, which I can make inline or you can approve here.

dgovil:
Thanks @tcauchois , I just accepted your changes.

PR #3152, opened 2024-07-03
Autodesk: Delay-load vulkan dlls on Windows

erikaharrison-adsk:
### Description of Change(s)

Remove the dependency of hgiVulkan in hgiInterop on Windows, so that the vulkan dlls can be delay loaded.
On Windows, it is possible to have both hgiVulkan and hgiDX. Delay loading allow the user to select their preferred hgi at run time.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9832

erikaharrison-adsk:
Closing. We're going in a different direction - see related: https://github.com/PixarAnimationStudios/OpenUSD/pull/3603

PR #3154, opened 2024-07-05
feat: add MaterialBindingAPIChecker

beersandrew:
### Description of Change(s)
- Added all necessary validation pieces including plugInfo.json, valdator tokens, tests, and validator code itself for the MaterialBindingAPIChecker

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65425035

### DIFF
--- a/pxr/usd/usdShade/plugInfo.json
+++ b/pxr/usd/usdShade/plugInfo.json
@@ -133,7 +133,10 @@
                         "schemaTypes": [
                             "UsdShadeShader"
                         ]
-                    }, 
+                    },
+                    "MaterialBindingApiChecker": {
+                        "doc": "Verify a prim has the MaterialBindingAPI applied if it has a material relationship binding."
+                    },
                     "keywords": [
                         "UsdShadeValidators"
                     ]
--- a/pxr/usd/usdShade/plugInfo.json
+++ b/pxr/usd/usdShade/plugInfo.json
@@ -134,8 +134,8 @@
                             "UsdShadeShader"
                         ]
                     },
-                    "MaterialBindingApiChecker": {
-                        "doc": "Verify a prim has the MaterialBindingAPI applied if it has a material relationship binding."
+                    "MaterialBindingApiAppliedValidator": {
+                        "doc": "Verify a prim has the MaterialBindingAPI applied if it has a material binding relationship."
                     },
                     "keywords": [
                         "UsdShadeValidators"
--- a/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
+++ b/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
@@ -14,6 +14,10 @@
 #include "pxr/usd/usdShade/shader.h"
 #include "pxr/usd/usdShade/shaderDefUtils.h"
 #include "pxr/usd/usdShade/validatorTokens.h"
+#include <pxr/usd/usdShade/material.h>
+#include <pxr/usd/usdShade/materialBindingAPI.h>
+#include <pxr/usd/usd/relationship.h>
+#include <pxr/usd/usd/prim.h>
 
 #include <algorithm>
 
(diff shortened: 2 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9833

pxr/usd/usdShade/validators.cpp
```diff
@@ -192,13 +193,45 @@ _ShaderPropertyTypeConformance(const UsdPrim &usdPrim)
...
 }
 
+
+static
+UsdValidationErrorVector
+_MaterialBindingApiChecker(const UsdPrim &usdPrim)
+{
+    UsdValidationErrorVector errors;
+
+    if (!usdPrim.CanApplyAPI<UsdShadeMaterialBindingAPI>()){
```
dgovil: Question for the group: I know the current check doesn't check for it, but should the test flag a warning (not an error) when the MaterialBindingAPI is applied to a prim that it can't be applied to?

dgovil:
Also LGTM! Great work :-)

pxr/usd/usdShade/validators.cpp
tallytalwar: What is an example of such an application @dgovil? I thought MaterialBindingAPI can be applied to any prim which wants to provide a material:binding relationship.

pxr/usd/usdShade/plugInfo.json
```diff
@@ -133,7 +133,10 @@
                         "schemaTypes": [
                             "UsdShadeShader"
                         ]
-                    }, 
+                    },
+                    "MaterialBindingApiChecker": {
```
tallytalwar: Checker or not to Checker?
MaterialBindingAPIApplicationValidator, MaterialBindingAPIApplicationCompliance?

pxr/usd/usdShade/plugInfo.json
```diff
@@ -133,7 +133,10 @@
...
+                        "doc": "Verify a prim has the MaterialBindingAPI applied if it has a material relationship binding."
```
tallytalwar: material binding relationship.

pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
```diff
@@ -118,11 +122,50 @@ TestUsdShadeShaderPropertyCompliance()
...
+    errors = validator->Validate(usdPrim);
+
+    // Verify the errors are fixed
+    TF_AXIOM(errors.size() == 0);
+}
+
 int
 main()
 {
     TestUsdShadeValidators();
```
tallytalwar, line 166: Can you also include this validator to the list above for TestUsdShadeValidators?

pxr/usd/usdShade/validators.cpp
```diff
@@ -192,13 +193,45 @@ _ShaderPropertyTypeConformance(const UsdPrim &usdPrim)
...
+        return errors;
+    }
+
+    UsdShadeMaterialBindingAPI materialBindingAPI(usdPrim);
+
+    if (materialBindingAPI.GetDirectBinding().IsBound() && !usdPrim.HasAPI<UsdShadeMaterialBindingAPI>()) {
```
tallytalwar: Hmm.. this is only checking for direct binding, but a prim could also provide a collection based binding rel, which I think it will not cover.

Might be simpler to do the exact same thing which is being done in the python code.. that is, get all relationships in "materialBinding" namespace / startswith, if exists and API not applied report error.

tallytalwar (commented):
Left some notes for you to consider @beersandrew. Thanks

pxr/usd/usdShade/validators.cpp
dgovil: Oh is MaterialBindingAPI unbounded? I thought it required at least being a gprim?

tallytalwar: I think a prim's ancestor also particate in determining what material is bound: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdShade/materialBindingAPI.cpp#L718, so there is no restriction on a type of prim which can't have MaterialBindingAPI applied. Also if there was such a restriction, the schema should be tightened by having "CanOnlyApplyTo" in schema.usda.

pxr/usd/usdShade/validators.cpp
```diff
@@ -192,13 +192,48 @@ _ShaderPropertyTypeConformance(const UsdPrim &usdPrim)
     return errors;
 }
 
+static bool _HasMaterialBindingRelationship(const UsdPrim& usdPrim) {
```
tallytalwar: Might be better to make this a lambda within `_MaterialBindingApiAppliedValidator` method?

And also make the `materialBindingString` a `static const`, or its totally fine if you do `UsdShadeTokens->materialBinding.GetString()` in place.

tallytalwar (approved):
Just 1 cleanup comment. Otherwise lgtm. Thanks @beersandrew

pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
```diff
@@ -118,11 +123,50 @@ TestUsdShadeShaderPropertyCompliance()
...
+    const UsdPrim usdPrim = usdStage->DefinePrim(SdfPath("/Test"));
+    UsdShadeMaterial material = UsdShadeMaterial::Define(usdStage, SdfPath("/Test/Material"));
+
+    // Create the material binding relationship manually
+    UsdRelationship materialBinding = usdPrim.CreateRelationship(TfToken("material:binding"));
+    materialBinding.AddTarget(material.GetPath());
+
+    UsdValidationErrorVector errors = validator->Validate(usdPrim);
+
+    auto site = errors[0].GetSites()[0];
```
pablode: (leftover line / before size check)

beersandrew: fixed, thanks!

PR #3157, opened 2024-07-09
Apple : UsdMtlx MaterialXConfigAPI schema and MaterialX versioning

ld-kerley:
### Description of Change(s)

Initial pass at adding MaterialXInfoAPI applied schema, and then using this in UsdMtlx file format reader to append the version to the UsdShadeMaterial prims created.

HdMtlx also updated to explicitly write the MaterialX version in to the document that is reconstructed.

Outstanding: I need some input on the best way to get to the MaterialX version attribute on the UsdShadeMaterial prim from inside HdMtlx.  I've left a hardcoded version string in the code so far at the place I think I need it.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9835

pxr/usd/usdMtlx/materialXInfoAPI.h
```diff
@@ -0,0 +1,209 @@
...
+    /// MaterialX library version that the data has been authored against.
+    /// Defaults to 1.38 to allow correct verisoning of old files.
+    ///
+    /// | ||
+    /// | -- | -- |
+    /// | Declaration | `string mtlxinfo:version = "1.38"` |
+    /// | C++ Type | std::string |
+    /// | \ref Usd_Datatypes "Usd Type" | SdfValueTypeNames->String |
+    USDMTLX_API
+    UsdAttribute GetMtlxinfoVersionAttr() const;
```
crydalch: Style question: should "info" and "x" be capitalized in the function @ld-kerley ? (i.e. `GetMtlXInfoVersionAttr()`)

pxr/imaging/hdMtlx/hdMtlx.cpp
```diff
@@ -562,6 +562,12 @@ HdMtlxCreateMtlxDocumentFromHdMaterialNetworkInterface(
 
     // Initialize a MaterialX Document
     mx::DocumentPtr mxDoc = mx::createDocument();
+
+    // Need to figure out how to read the actual value from the netInterface
+    std::string materialXVersionString = "1.37";
```
crydalch: @ld-kerley , should this default value be 1.38 like the schema has? Or maybe it's just a temporary/filler value that doesn't matter?

pxr/usd/usdMtlx/materialXInfoAPI.h
ld-kerley: Following other MaterialX use in USD I think it should be `Mtlx`, so lower case x.  But I think you're right about the `Info` part - thanks - I'll make sure that gets changed next time I push something here.

pxr/imaging/hdMtlx/hdMtlx.cpp
ld-kerley: Yes - this is place holder - I'm hoping that someone with a bit more intimate hydra understanding can guide me as to the right way to get the real MaterialX version string from USD here.

pxr/usd/usdMtlx/materialXInfoAPI.h
crydalch: Cool that's great. Thank you for getting this PR up and going!

pxr/usd/usdMtlx/api.h
```diff
@@ -1,11 +1,28 @@
...
+// Section 6. Trademarks. is deleted and replaced with:
 //
-#ifndef PXR_USD_USDMTLX_API_H
-#define PXR_USD_USDMTLX_API_H
+// 6. Trademarks. This License does not grant permission to use the trade
+//    names, trademarks, service marks, or product names of the Licensor
+//    and its affiliates, except as required to comply with Section 4(c) of
+//    the License and to reproduce the content of the NOTICE file.
+//
+// You may obtain a copy of the Apache License at
```
tcauchois, line 14: Note that we recently updated our license text, so you'll want to go with the top-of-tree license text.  If you're running usdSchemaGen with an old USD version (< 24.08) that might put the wrong text in.

pxr/usd/usdMtlx/reader.cpp
```diff
@@ -1458,9 +1460,14 @@ UsdShadeMaterial
 _Context::BeginMaterial(const mx::ConstNodePtr& mtlxMaterial)
 {
     if (TF_VERIFY(!_usdMaterial)) {
+        auto mtlxVersionValue = VtValue(mtlxMaterial->getDocument()->getVersionString());
```
tcauchois: I'm guessing this will return one of "" and "1.39" at the moment?  Do we want to handle the fallback case and put 1.38 in instead here?  Or does the document function do that already?

pxr/imaging/hdMtlx/hdMtlx.cpp
```diff
@@ -585,6 +591,23 @@ HdMtlxCreateMtlxDocumentFromHdMaterialNetworkInterface(
         mxType,
         mxShaderNode);
 
+    // just for testing...
+    bool localDebug = false;
```
tcauchois: There's a system in hydra called debug codes that would be perfect here.  The idea is that you create a local "debugCodes.h/.cpp", like the ones in imaging/hd (for example); add a new debug code, such as HDMTLX_VERSION_UPGRADE; and here, replace "localDebug" with TfDebug::IsEnabled(HDMTLX_VERSION_UPGRADE).

You can trigger this either with the debug codes editor in usdview, or by setting the environment variable "TF_DEBUG" to "HDMTLX_VERSION_UPGRADE".

Having this as a debug code seems like a fantastic idea!

tcauchois (commented):
Thanks for putting this together!  Left some thoughts on the rest, but for the question of how to get the version string through I think we need to discuss our options :).  But we're thinking about it, and will get back to you with ideas!

pxr/imaging/hdMtlx/hdMtlx.cpp
```diff
@@ -562,6 +562,12 @@ HdMtlxCreateMtlxDocumentFromHdMaterialNetworkInterface(
 
     // Initialize a MaterialX Document
     mx::DocumentPtr mxDoc = mx::createDocument();
+
+    // Need to figure out how to read the actual value from the netInterface
```
tcauchois: A thought that @spiffmon had:

1.) Rename the property info:mtlx:version or something (such that it starts with "info:").
2.) Update materialAdapter.cpp (a) and dataSourceMaterial.cpp (b) to grab all attributes with the "info:" namespace prefix, and insert them into a new VtDictionary "info" on HdMaterialNetwork (a)/a new HdContainerDataSource named "info" on the material data source (b).
3.) Update dataSourceLegacyPrim.cpp and sceneIndexAdapterSceneDelegate.cpp to translate from the VtDictionary to the ContainerDataSource and back.
4.) Finally grab the version string off the material network here.

How does that feel, design wise/difficulty wise?

ld-kerley:
@tcauchois - I took a stab at what you suggested.  I think it's functional, but likely there are style things that could be tidied up, but before going too far down that road I wanted to share where I'm at for feedback, and make sure this is along the right lines.

pxr/usdImaging/usdImaging/materialAdapter.cpp
```diff
@@ -379,6 +379,19 @@ UsdImagingMaterialAdapter::GetMaterialResource(UsdPrim const &prim,
             time);
     }
 
+    // finally collect any 'info' on the Material prim
+    VtDictionary infoDict;
```
tcauchois: (1) Do we only want to include authored attributes, or are we expecting the fallback of "1.38" to ever be useful?  This won't pull fallbacks, but GetAttributes would().  Maybe moot since usdMtlx/reader.cpp looks to be manually creating the attribute.

(2) We may want to use prim.GetPropertiesInNamespace("info") (or GetAuthoredPropertiesInNamespace) instead here?  And then auto attr = prop.As<UsdAttribute>(), which returns null if it's not an attribute (e.g. if it's a relationship).  See also: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdGeom/primvarsAPI.cpp#L215.

pxr/imaging/hd/utils.cpp
```diff
@@ -274,21 +276,37 @@ ConvertHdMaterialNetworkToHdMaterialNetworkSchema(
                 .Build());
     }
 
+    infoNames.reserve(hdNetworkMap.info.size());
+    infoValues.reserve(hdNetworkMap.info.size());
+    for (const auto& infoEntry : hdNetworkMap.info)
+    {
+      // from _dataSourceLegacyPrim.cpp - _ToContainerDS(VtDictionary)
```
tcauchois: Here & elsewhere, no need to update this but if it's annoying you I think we could add VtDictionary <-> ContainerDS translators somewhere accessible (maybe retainedDataSource.h?).

pxr/imaging/hdMtlx/hdMtlx.cpp
```diff
@@ -585,6 +607,21 @@ HdMtlxCreateMtlxDocumentFromHdMaterialNetworkInterface(
         mxType,
         mxShaderNode);
 
+    if (TfDebug::IsEnabled(HDMTLX_VERSION_UPGRADE)) {
+      writeToXmlFile(
+          mxDoc, mx::FilePath(mxMaterial->getName()+"_before.mtlx"));
```
tcauchois: Possibly want to add a printout of which files you're writing to stdout?

pxr/usdImaging/usdImaging/materialAdapter.cpp
ld-kerley: I can't see that it would hurt any.

pxr/imaging/hd/utils.cpp
ld-kerley: There were other places in the code where these conversions were happening, but I couldn't see a clean way to share that code.  I don't want to bite off more than I can chew here, so maybe I'll leave that for a later refacor - or if anyone wants to jump in and make that change, thats also fine.

pxr/imaging/hdMtlx/hdMtlx.cpp
ld-kerley: make sense - added.

spiffmon:
Hi @ld-kerley , we've been discussing the latest changes, and realizing that since the new data we're adding will potentially affect the network that gets built, it will, for _instanced_ materials, be "prototype-affecting" data, rather than "per-instance data". With that in mind, we were thinking that the term "Config" captures the "weight" of the data better than "Info", which kind of implies it's just advisory (admitting it's unfortunate "info" is what's used in Shader/NodeDefAPI, a legacy Pixar (or maybe Katana?) thing).

Would you be willing to change the schema name to MaterialXConfigAPI and the new datasource member to config?

ld-kerley:
I'm personally open to any naming, and can see why "Info" implies advisory.  Is there precedent for "Config" elsewhere in USD?  I'd like to raise this with the wider MaterialX/USD community and ensure the name change doesn't raise any flags for anyone else.  If there's no existing precedent for "Config" then I might open the discussion for alternatives, but if the "Config" suggestion fits in to some already established USD nomenclature, we should certainly stick with that.

spiffmon:
There is "internal pixar" precedent in the Presto scenegraph that "config attributes" imply structural changes to a prim's definition; while we are continuing discussions on whether that mechanism will need to be reflected in OpenUSD, it's not a sure thing.

So sure, please do verify with the wider group and let us know if there are alt sugs!

nvmkuruc:
Perhaps of note, `config` is currently a keyword in `usda` (but I don't think that limits its usage in property names and namespace elements)

https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/sdf/textFileFormatParser.h#L99

spiffmon:
That's correct, @nvmkuruc .  In fact, we've stopped authoring it in Presto/menva, but don't disallow it, for old assets.  And yes, it doesn't restrict uses in property names or namespaces.

PR #3159, opened 2024-07-10
Apple : Compilation support for MaterialX 1.38.X and 1.39.X

ld-kerley:
### Description of Change(s)

MaterialX 1.39.X introduced API changes, this PR allows OpenUSD to be built against either MaterialX 1.38.X or 1.39.X.

Most of the changed arise from an API change around `MaterialX::TypeDesc` being passed by value instead of pointer. The other change is the complete removal of the `channels` attribute.

NOTE : The MaterialX versioning PR (#3157) should be merged before this. 

### Fixes Issue(s)

Fixes compile issues when using MaterialX 1.39.X

-

### DIFF
--- a/pxr/imaging/hdSt/materialXFilter.cpp
+++ b/pxr/imaging/hdSt/materialXFilter.cpp
@@ -637,10 +637,15 @@ _GetGlTFSurfaceMaterialTag(HdMaterialNode2 const& terminal)
 static const mx::TypeDesc
 _GetMxTypeDescription(std::string const& typeName)
 {
-#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION==38
-  // Add whatever is necessary for current codebase:
-  static const auto _typeLibrary =
-      std::map<std::string, const mx::TypeDesc*>{
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION <= 38
+    using MxTypeDesc = const mx::TypeDesc*;
+#else
+    using MxTypeDesc = const mx::TypeDesc;
+#endif
+
+    // Add whatever is necessary for current codebase:
+    static const auto _typeLibrary =
+      std::map<std::string, MxTypeDesc>{
           {"float", mx::Type::FLOAT},
           {"color3", mx::Type::COLOR3},
           {"color4", mx::Type::COLOR4},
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9837

pxr/imaging/hdSt/materialXShaderGen.cpp
```diff
@@ -285,12 +289,15 @@ HdStMaterialXShaderGen<Base>::_EmitMxSurfaceShader(
         if (outputConnection) {
 
             std::string finalOutput = outputConnection->getVariable();
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION==38
```
tcauchois: Worth doing "MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION < 39"?

pxr/imaging/hdSt/materialXShaderGen.cpp
```diff
@@ -156,13 +156,17 @@ HdStMaterialXShaderGen<Base>::_EmitGlslfxHeader(mx::ShaderStage& mxStage) const
...
-            if (mxType == nullptr) {
+            const mx::TypeDesc mxType = getMxTypeDesc(primvarPair.second);
+            if (mxTypeIsNone(mxType)) {
                 TF_WARN("MaterialX geomprop '%s' has unknown type '%s'",
                         primvarPair.first.c_str(), primvarPair.second.c_str());
             }
-            const std::string type = mxType 
-                ? Base::_syntax->getTypeName(mxType) : "vec2";
+
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION==38
```
tcauchois: I appreciated how most of the other ifdefs are hidden away inside helper functions; is there a way to do that here?

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -634,26 +634,48 @@ _GetGlTFSurfaceMaterialTag(HdMaterialNode2 const& terminal)
     return materialToken.GetString();
 }
 
-static const mx::TypeDesc*
+static const mx::TypeDesc
 _GetMxTypeDescription(std::string const& typeName)
 {
+#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION==38
+  // Add whatever is necessary for current codebase:
+  static const auto _typeLibrary =
```
tcauchois: Minor thing, but I wonder if you could have an ifdef around
"using typename MxTypeDesc = const mx::TypeDesc"
or something and then only define the map once, since aside from the map type everything is character-for-character identical.  You'd still need another ifdef around the find/return code below, but that's less likely to be edited than this type map table.

pxr/imaging/hdSt/materialXShaderGen.cpp
ld-kerley: Sounds good - I'll look to see if this is possibly cleanly.

pxr/imaging/hdSt/materialXShaderGen.cpp
ld-kerley: I don't think any version of MaterialX prior to 1.38.something was ever used or would compile against the USD codebase, so this seemed more direct, but if you prefer the `<` I have no objections to changing it. 

Maybe `#if MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION<=38` is better overall?

pxr/imaging/hdSt/materialXFilter.cpp
ld-kerley: yep - seems like a good improvement.

zwillikon:
May you please rebase _materialXFilter.cpp_ ?

nyabinary:
bump  :p

PR #3160, opened 2024-07-11
Apple: Enable iOS Imaging support 

dgovil:
### Description of Change(s)

This PR ports Thor's imaging changes against the latest dev. This should enable imaging on all iOS derivatives, including visionOS.
I **think** I got everything, and it does build/run for me, but as mentioned I haven't done significant testing just yet.

I think we can use this branch to iterate the changes for imaging as needed.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9841

pxr/imaging/hgiGL/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(PXR_PREFIX pxr/imaging)
 set(PXR_PACKAGE hgiGL)
 
-if (NOT ${PXR_ENABLE_GL_SUPPORT})
+if (NOT ${PXR_ENABLE_GL_SUPPORT} AND NOT PXR_APPLE_EMBEDDED)
```
meshula, line 4: What's the difference between `AND NOT PXR_APPLE_EMBEDDED` and `&& !defined(ARCH_OS_IPHONE)` ? Should probably use `&&` and `!` in general, but I'm wondering why the different macro here & there?

pxr/imaging/hgiMetal/texture.mm
```diff
@@ -95,7 +103,11 @@
         // to our original, private texture.
 
         // Modify texture descriptor to describe the temp texture.
+#if defined(ARCH_OS_OSX)
```
meshula, line 110: could we have a comment here about why the Mode is not the same on desktop and embedded? I realize Managed isn't available on iOS but why aren't both Shared?

pxr/imaging/hgiGL/CMakeLists.txt
dgovil: PXR_APPLE_EMBEDDED is what we tagged it in the cmake files as, whereas the ARCH_OS_IPHONE was for the CPP files. The EMBEDDED bit was to be clearer in CMake land where configuration is more common, whereas the OS_IPHONE was to keep in line with the TARGET_OS_IPHONE styling that the compilers already provide.

Regarding the style change to symbols, afaik CMake only supports the words not the symbol? at any rate, none of the existing Pixar cmake uses symbols.

pxr/imaging/hgiMetal/texture.mm
dgovil: I'll double check with Thor if there's any other reason, but afaik it's because Managed is required for older Macs with a dGPU (AMD/NVIDIA but not Intel).

A better check would probably be to use [hasUnifiedMemory](https://developer.apple.com/documentation/metal/mtldevice/3229025-hasunifiedmemory), instead of the ifdef. 

However, I am torn on whether its best to leave Mac behaviour alone for now and address it in a future PR, so that we can keep the scope of this PR tight? Let me know how you feel, and I can put it in this PR if you prefer

dgovil: Checked with Thor, and it is the reason above. Anyway, I'll wait to hear back from you before I add a comment or just swap to the unifiedMemory check which should then be self explanatory.

pxr/imaging/hgiGL/CMakeLists.txt
meshula: Gotcha, it was kinda confusing then, and it's kinda confusing now, but doesn't need an update.

pxr/imaging/hgiMetal/texture.mm
meshula: tbh I'd prefer we switch to the check because I worry that such a minor detail might just slip away for more months. @davidgyu any thought on this?

dgovil: Fair enough. I made the change after discussing with a few folks who seem confident it shouldn't cause issues.
Unfortunately I do not have a Mac with an intel gpu to test on, but it seems to be working fine on Apple Silicon Macs, so I think we should be good.

I don't have a good benchmark scene, but I suspect it'll be slightly faster too with this change now.

dgovil:
I need to fix up some issues the arose from rebasing. Fixing it up now, and should have it good to go soon.

dgovil:
With some of the changes in 24.8, I'm just going to close this PR and re-open it later once I can resolve some things. Seems like its a little more involved than I thought.

pxr/imaging/hgiMetal/texture.mm
davidgyu: I like that suggestion @meshula. I might be able to check for correct behavior on my stash of old macOS systems with Intel and AMD GPUs.

dgovil:
Okay that actually was less involved than I thought.
It required a patch to MaterialX (thanks Maddy, and I'll look into mainlining it) and some linker changes.

I'll resubmit this PR because it won't let me re-open this for some reason.

dgovil:
Reopening

dgovil:
Re-opened here https://github.com/PixarAnimationStudios/OpenUSD/pull/3215

Issue #3156, opened 2024-07-08
hdf5 version in build_usd.py doesn't build with newer clang versions


jonlm:
### Description of Issue

Building on macOS Sonoma with XCode 15.3 and Alembic + HDF5 enabled fails with this error:

```
src/hdf5-1.10.0-patch1/src/H5Dint.c:2193:6: error: incompatible pointer to integer conversion assigning to 'herr_t' (aka 'int') from 'void *' [-Wint-conversion]
            HGOTO_ERROR(H5E_DATASET, H5E_CANTRESET, NULL, "unable to reset layout info")
            ^                                       ~~~~
```

I found some discussion of this same problem [here](https://gitlab.gnome.org/GNOME/gimp/-/issues/8649):

_"The -Wint-conversion warning diagnostic for implicit int <-> pointer conversions now defaults to an error in all C language modes. It may be downgraded to a warning with -Wno-error=int-conversion, or disabled entirely with -Wno-int-conversion."_
https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html

The hdf5 version ([1.10.0 patch 1](https://github.com/HDFGroup/hdf5/releases/tag/hdf5-1_10_0-patch1)) is from 2016, so this isn't too surprising. I tried replacing it with [the latest version of 1.10](https://github.com/HDFGroup/hdf5/releases/tag/hdf5-1_10_11), but in order for it to build I had to remove this chunk of code from build_usd.py since neither patch cleanly applies anymore, nor seems to be necessary:

```
    with CurrentWorkingDirectory(DownloadURL(HDF5_URL, context, force)):
        if MacOS():
            PatchFile("config/cmake_ext_mod/ConfigureChecks.cmake",
                    [("if (ARCH_LENGTH GREATER 1)", "if (FALSE)")])
            if context.targetUniversal:
                PatchFile("config/cmake/H5pubconf.h.in",
                        [(" #define H5_SIZEOF_LONG_LONG 8",
                        " #define H5_SIZEOF_LONG_LONG 8\n" +
                        " #define H5_SIZEOF_LONG_DOUBLE 16")])
```

At that point it did build fine, however I didn't test any hdf5 alembic files so I can't speak to the correctness of the build.

Related commits: 
https://github.com/PixarAnimationStudios/OpenUSD/commit/5e280f6ffb37e350a7011ed9227667cbf4b881ec
[...]

PR #3165, opened 2024-07-15
Apple: Remove HDF5 support

dgovil:
### Description of Change(s)

This PR consists of two commits. One that removes HDF5 support from the build_usd.py, and the other that removes it from the USD codebase as a whole.

I split it into two commits so it may ease the ability to pick the level of support you want to leave in. However, I would perhaps suggest removing HDF5 altogether if nobody has objections, as I worry the code will just eventually bit rot, if it hasn't already. 

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3156


asluk:
@marktucker , @crydalch -- does Houdini still need HDF5 Alembic support? Thanks! https://www.sidefx.com/docs/houdini/nodes/out/alembic.html

jesschimein:
Filed as internal issue #USD-9848

marktucker:
We are reluctant to drop support for reading HDF5 Alembics without a strong reason. We don't want to make any unnecessary assumptions about what kind of legacy data or tools our customers may be using.

So SideFX will continue to support HDF5 Alembic loading in USD as long as we support it in Houdini proper. And although this discussion has triggered us to think more about our position moving forward, we need to support reading of HDF5 Alembics for at least the next version or two of Houdini (to allow for a period of safe/orderly/officially announced deprecation).

So from our perspective it would be great if at least the C++ code in alembicReader.cpp could be left alone, with whatever compile guards the community may consider appropriate. We don't use the build_usd.py script, and already build our own Alembic and HDF5 libraries, so we have no problem with the removal of HDF5 as an option there.

The benefit of leaving the C++ code in place is simply to minimize differences from our fork of USD to the Pixar standard fork. If HDF5 support is officially removed from the Pixar fork, we'll just add the required code back into our own fork and accept responsibility for maintaining this bit of the code for our customers.

PR #3166, opened 2024-07-15
feat: add validator SkelBindingAPIAppliedChecker

beersandrew:
### Description of Change(s)
add validator SkelBindingAPIAppliedChecker
- added a validation check for when a skel binding property is present but the skel binding api is not applied
- added a validation check for when a skel binding api is applied to a prim that is not parented by a skel root
- added a unit test

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65425045
- 

### DIFF
--- a/pxr/usd/usdSkel/CMakeLists.txt
+++ b/pxr/usd/usdSkel/CMakeLists.txt
@@ -99,13 +99,12 @@ pxr_library(usdSkel
 )
 
 pxr_build_test(testUsdSkelValidators
-        LIBRARIES
-        sdr
-        tf
-        usd
-        usdSkel
-        CPPFILES
-        testenv/testUsdSkelValidators.cpp
+    LIBRARIES
+    tf
+    usd
+    usdSkel
+    CPPFILES
+    testenv/testUsdSkelValidators.cpp
 )
    
 pxr_test_scripts(
@@ -130,6 +129,11 @@ pxr_install_test_dir(
 pxr_install_test_dir(
     SRC testenv/testUsdSkelCache
     DEST testUsdSkelCache
+)
+
+pxr_install_test_dir(
+    SRC testenv/testUsdSkelRoot
+    DEST testUsdSkelRoot
 )
 
 pxr_install_test_dir(
--- a/pxr/usd/usdSkel/CMakeLists.txt
+++ b/pxr/usd/usdSkel/CMakeLists.txt
@@ -37,6 +37,10 @@ pxr_library(usdSkel
         tokens
         topology
         utils
+        validatorTokens
+
+    CPPFILES
+        validators.cpp
 
     PUBLIC_HEADERS
         api.h
@@ -93,6 +97,16 @@ pxr_library(usdSkel
         images/influencesPrimvarLayout.svg
         images/unboundedInterpolationExample.svg
 )
+
+pxr_build_test(testUsdSkelValidators
+        LIBRARIES
+        sdr
+        tf
+        usd
+        usdSkel
+        CPPFILES
+        testenv/testUsdSkelValidators.cpp
+)
    
 pxr_test_scripts(
     testenv/testUsdSkelAnimMapper.py
@@ -116,11 +130,6 @@ pxr_install_test_dir(
 pxr_install_test_dir(
     SRC testenv/testUsdSkelCache
     DEST testUsdSkelCache
-)
-
-pxr_install_test_dir(
-    SRC testenv/testUsdSkelRoot
-    DEST testUsdSkelRoot
 )
 
 pxr_install_test_dir(
@@ -189,3 +198,9 @@ pxr_register_test(testUsdSkelUtils
     COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testUsdSkelUtils"
     EXPECTED_RETURN_CODE 0
 )
+
+pxr_register_test(testUsdSkelValidators
+    COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testUsdSkelValidators"
+    EXPECTED_RETURN_CODE 0
+)
+
--- a/pxr/usd/usdSkel/plugInfo.json
+++ b/pxr/usd/usdSkel/plugInfo.json
@@ -71,8 +71,8 @@
                     }
                 },
                 "Validators": {
-                    "SkelBindingApiAppliedChecker": {
-                        "doc": "Verify a prim has the SkelBindingAPI applied if it has a UsdSkelBinding property. If a prim has the SkelBindingAPI applied, make sure it's of type SkelRoot or under a SkelRoot."
+                    "SkelBindingApiAppliedValidator": {
+                        "doc": "Verify a prim has the SkelBindingAPI applied if it has a UsdSkelBinding property. If a prim has the SkelBindingAPI applied, make sure it's of type SkelRoot or parented by a SkelRoot."
                     },
                     "keywords": [
                         "UsdSkelValidators"
--- a/pxr/usd/usdSkel/plugInfo.json
+++ b/pxr/usd/usdSkel/plugInfo.json
@@ -69,6 +69,14 @@
                         "implementsComputeExtent": true, 
                         "schemaKind": "concreteTyped"
                     }
+                },
+                "Validators": {
+                    "SkelBindingApiAppliedChecker": {
+                        "doc": "Verify a prim has the SkelBindingAPI applied if it has a UsdSkelBinding property. If a prim has the SkelBindingAPI applied, make sure it's of type SkelRoot or under a SkelRoot."
+                    },
+                    "keywords": [
+                        "UsdSkelValidators"
+                    ]
                 }
             }, 
             "LibraryPath": "@PLUG_INFO_LIBRARY_PATH@", 
--- a/pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
+++ b/pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
@@ -24,14 +24,14 @@ PXR_NAMESPACE_USING_DIRECTIVE
 void
 TestUsdSkelBindingAPIChecker()
 {
-    // Verify validator exists
+    // Verify that the validator exists.
     UsdValidationRegistry &registry = UsdValidationRegistry::GetInstance();
     const UsdValidator *validator = registry.GetOrLoadValidatorByName(
             UsdSkelValidatorNameTokens->skelBindingApiAppliedChecker);
     TF_AXIOM(validator);
 
     // Create Stage and mesh with a skel binding property
-    UsdStageRefPtr usdStage = UsdStage::CreateNew("/Users/andrewbeers/test.usda");
+    UsdStageRefPtr usdStage = UsdStage::CreateInMemory();
     UsdGeomMesh mesh = UsdGeomMesh::Define(usdStage, SdfPath("/SkelRoot/Mesh"));
     UsdGeomPrimvarsAPI primvarsApi(mesh);
     UsdGeomPrimvar jointIndicesPrimvar = primvarsApi.CreatePrimvar(TfToken("skel:jointIndices"), SdfValueTypeNames->IntArray, UsdGeomTokens->vertex);
@@ -39,7 +39,7 @@ TestUsdSkelBindingAPIChecker()
 
     UsdValidationErrorVector errors = validator->Validate(mesh.GetPrim());
 
-    // test and verify error for not having a skel binding api applied
+    // Verify the error for not having the SkelBindingAPI schema applied is present.
     TF_AXIOM(errors.size() == 1);
     TF_AXIOM(errors[0].GetType() == UsdValidationErrorType::Error);
     TF_AXIOM(errors[0].GetSites().size() == 1);
@@ -47,17 +47,17 @@ TestUsdSkelBindingAPIChecker()
     TF_AXIOM(errors[0].GetSites()[0].IsPrim());
     TF_AXIOM(errors[0].GetSites()[0].GetPrim().GetPath() ==
              SdfPath("/SkelRoot/Mesh"));
-    std::string expectedErrorMsg = "Found a UsdSkelBinding property (primvars:skel:jointIndices)" \
-                                    ", but no SkelBindingAPI applied on the prim </SkelRoot/Mesh>." \
-                                    "(fails 'SkelBindingAPIAppliedChecker')";
+    std::string expectedErrorMsg = ("Found a UsdSkelBinding property (primvars:skel:jointIndices)"
+                                    ", but no SkelBindingAPI applied on the prim </SkelRoot/Mesh>."
+                                    "(fails 'SkelBindingAPIAppliedChecker')");
     TF_AXIOM(errors[0].GetMessage() == expectedErrorMsg);
 
-    // apply skel binding api
+    // Apply the SkelBindingAPI.
     UsdSkelBindingAPI skelBindingApi = UsdSkelBindingAPI::Apply(mesh.GetPrim());
 
     errors = validator->Validate(mesh.GetPrim());
 
-    // test and verify error for not having a skel root parenting the mesh
+    // Verify the error for not having a SkelRoot parenting a prim with the SkelBindingAPI applied.
     TF_AXIOM(errors.size() == 1);
     TF_AXIOM(errors[0].GetType() == UsdValidationErrorType::Error);
     TF_AXIOM(errors[0].GetSites().size() == 1);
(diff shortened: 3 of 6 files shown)


jesschimein:
Filed as internal issue #USD-9849

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
...
+TestUsdSkelBindingAPIChecker()
+{
+    // Verify validator exists
+    UsdValidationRegistry &registry = UsdValidationRegistry::GetInstance();
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdSkelValidatorNameTokens->skelBindingApiAppliedChecker);
+    TF_AXIOM(validator);
+
+    // Create Stage and mesh with a skel binding property
+    UsdStageRefPtr usdStage = UsdStage::CreateNew("/Users/andrewbeers/test.usda");
```
dgovil: Nit: I'm sure you'll remove it but good to just write it to a temp test location instead before it's merged in, and/or just keep it in memory if you don't need to write it out.

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
...
+    UsdGeomMesh mesh = UsdGeomMesh::Define(usdStage, SdfPath("/SkelRoot/Mesh"));
+    UsdGeomPrimvarsAPI primvarsApi(mesh);
+    UsdGeomPrimvar jointIndicesPrimvar = primvarsApi.CreatePrimvar(TfToken("skel:jointIndices"), SdfValueTypeNames->IntArray, UsdGeomTokens->vertex);
+    jointIndicesPrimvar.Set(VtIntArray{0, 1, 2});
+
+    UsdValidationErrorVector errors = validator->Validate(mesh.GetPrim());
+
+    // test and verify error for not having a skel binding api applied
```
dgovil: Smallest of nitpicks: Comment style should be sentence case just to match the rest of the comments.

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // test and verify error for not having a skel root parenting the mesh
+    TF_AXIOM(errors.size() == 1);
+    TF_AXIOM(errors[0].GetType() == UsdValidationErrorType::Error);
+    TF_AXIOM(errors[0].GetSites().size() == 1);
+    TF_AXIOM(errors[0].GetSites()[0].IsValid());
+    TF_AXIOM(errors[0].GetSites()[0].IsPrim());
+    TF_AXIOM(errors[0].GetSites()[0].GetPrim().GetPath() ==
+             SdfPath("/SkelRoot/Mesh"));
+    expectedErrorMsg = "UsdSkelBindingAPI applied on prim: </SkelRoot/Mesh>, " \
```
dgovil: Style nit: You can put the string inside parenthesis and avoid escaping the line break. It's a little cleaner and less fragile if someone forgets in the future.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,90 @@
...
+        for (const TfToken &primToken : primPropertyNames){
+            for (const TfToken &skelToken : skelPropertyNames){
+                if (skelToken == primToken){
+                    errors.emplace_back(
+                            UsdValidationErrorType::Error,
+                            UsdValidationErrorSites{
+                                    UsdValidationErrorSite(usdPrim.GetStage(),
+                                                           usdPrim.GetPath())
+                            },
+                            TfStringPrintf("Found a UsdSkelBinding property (%s), " \
```
dgovil, line 47: Perhaps a question for @tallytalwar , but are we always printing out the warning messages during the validator, or do we collect it and then print it at the end?

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (!usdPrim.HasAPI<UsdSkelBindingAPI>()){
+        UsdSchemaRegistry& usdSchemaRegistry = UsdSchemaRegistry::GetInstance();
+        std::unique_ptr<UsdPrimDefinition> primDef = usdSchemaRegistry.BuildComposedPrimDefinition(TfToken(), {TfToken("SkelBindingAPI")});
+
+        std::vector<TfToken> skelPropertyNames = primDef->GetPropertyNames();
+        std::vector<TfToken> primPropertyNames = usdPrim.GetPropertyNames();
+
+        for (const TfToken &primToken : primPropertyNames){
+            for (const TfToken &skelToken : skelPropertyNames){
+                if (skelToken == primToken){
```
dgovil: Nit: If you invert this comparison + use a continue,  you can unnest the contents of the check. Ever so slightly improves readability.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,90 @@
...
+                            "but no SkelBindingAPI applied on the prim <%s>.(fails 'SkelBindingAPIAppliedChecker')",
+                                           skelToken.GetText(), usdPrim.GetPath().GetText()));
+                }
+            }
+        }
+    }
+    else {
+        if (usdPrim.GetTypeName() != UsdSkelTokens->SkelRoot) {
+            UsdPrim parentPrim = usdPrim.GetParent();
+            while (!parentPrim.IsPseudoRoot()) {
```
dgovil: I think you should also check if the parentPrim isn't null, just in case you get a bad prim here.
`while (parentPrim && !parentPrim.IsPseudoRoot())`

dgovil:
LGTM! other than a couple small notes, most of which are just minor style ones.

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
beersandrew: didn't know about this! thanks!

beersandrew:
> LGTM! other than a couple small notes, most of which are just minor style ones.

Made the suggested changes. Thanks so much for the review & ideas.

tallytalwar:
Just curious why is this a draft PR?

pxr/usd/usdSkel/CMakeLists.txt
```diff
@@ -118,11 +132,6 @@ pxr_install_test_dir(
     DEST testUsdSkelCache
 )
 
-pxr_install_test_dir(
```
tallytalwar: Did you intend to remove this test?

pxr/usd/usdSkel/CMakeLists.txt
```diff
@@ -93,6 +97,16 @@ pxr_library(usdSkel
         images/influencesPrimvarLayout.svg
         images/unboundedInterpolationExample.svg
 )
+
+pxr_build_test(testUsdSkelValidators
+        LIBRARIES
+        sdr
```
tallytalwar: I am not sure if sdr is needed as a dependency here. Can you remove this please? Thanks

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the terms set forth in the LICENSE.txt file available at
+// https://openusd.org/license.
+//
+
+#include "pxr/pxr.h"
+#include "pxr/base/tf/token.h"
+#include "pxr/usd/sdr/registry.h"
```
tallytalwar: Can you remove these extra headers? Thanks

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,91 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the terms set forth in the LICENSE.txt file available at
+// https://openusd.org/license.
+//
+
+#include "pxr/usd/sdr/shaderProperty.h"
```
tallytalwar: Similarly sdr related headers shouldn't be required in usdSkel.

So this one and the sdr/registry.h below.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,91 @@
...
+UsdValidationErrorVector
+_SkelBindingApiChecker(const UsdPrim &usdPrim)
+{
+    UsdValidationErrorVector errors;
+
+    if (!usdPrim.HasAPI<UsdSkelBindingAPI>()){
+        UsdSchemaRegistry& usdSchemaRegistry = UsdSchemaRegistry::GetInstance();
+        std::unique_ptr<UsdPrimDefinition> primDef = usdSchemaRegistry.BuildComposedPrimDefinition(TfToken(), {TfToken("SkelBindingAPI")});
+
+        std::vector<TfToken> skelPropertyNames = primDef->GetPropertyNames();
```
tallytalwar: You can clean this further.

I would recommend something along these lines:

```cpp
static const std::vector<TfToken> skelPropertyNames = []() {
    UsdSchemaRegistry& usdSchemaRegistry = UsdSchemaRegistry::GetInstance();
    std::unique_ptr<UsdPrimDefinition> primDef = usdSchemaRegistry.BuildComposedPrimDefinition(TfToken(), {TfToken("SkelBindingAPI")});
        return primDef->GetPropertyNames();
    }();
```

This skelProperties are not repeatedly computed and can be saved in the static variable here.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,91 @@
...
+        std::vector<TfToken> primPropertyNames = usdPrim.GetPropertyNames();
+
+        for (const TfToken &primToken : primPropertyNames){
+            for (const TfToken &skelToken : skelPropertyNames){
+                if (skelToken != primToken)
+                    continue;
```
tallytalwar: Most of openUSD code follows a convention of surrounding if blocks with {}, even for single line blocks.

tallytalwar: Also maybe make skelPropertyNames in unordered_set of Tokens, then you can only iterate on primProperties and do a find of the primProperty in the set, if any is found in the set, then you can report the validation error below and break the for loop! No need to report multiple errors in this case I think.

I think this way it will be more performant. Let me know what you think?

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
...
+    errors = validator->Validate(mesh.GetPrim());
+
+    // Verify all errors are gone
+    TF_AXIOM(errors.size() == 0);
+}
+
+int
+main()
+{
+    TestUsdSkelBindingAPIChecker();
```
tallytalwar: Can you also add that subset UsdSkelValidator keywords test here (similar to what you did for UsdGeom / UsdShade)?

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+                errors.emplace_back(
+                        UsdValidationErrorType::Error,
+                        UsdValidationErrorSites{
+                                UsdValidationErrorSite(usdPrim.GetStage(),
+                                                       usdPrim.GetPath())
+                        },
+                        TfStringPrintf(("Found a UsdSkelBinding property (%s), "
+                            "but no SkelBindingAPI applied on the prim <%s>.(fails 'SkelBindingAPIAppliedChecker')"),
```
tallytalwar: I will also recommend to remove the test name from the error string. Validator which resulted in this error will automatically be embedded in the Error (https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usd/validator.cpp#L61).

pxr/usd/usdSkel/plugInfo.json
```diff
@@ -69,6 +69,14 @@
                         "implementsComputeExtent": true, 
                         "schemaKind": "concreteTyped"
                     }
+                },
+                "Validators": {
+                    "SkelBindingApiAppliedChecker": {
```
tallytalwar: I think in the 2 validators I introduced in the dev branch, I was trying not to use "Checker" in the name. So

`SkelBindingAPIApplicationValidator` or `SkelBindingAPIApplicationCompliance` ?

cc. @dgovil if you have opinions on this?

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
```diff
@@ -0,0 +1,88 @@
...
+                "of type SkelRoot, as required by the UsdSkel schema.(fails 'SkelBindingAPIAppliedChecker')");
+    const std::string message = errors[0].GetMessage();
+    TF_AXIOM(errors[0].GetMessage() == expectedErrorMsg);
+
+    // Add SkelRoot as a parent to the mesh.
+    UsdSkelRoot skelRoot = UsdSkelRoot::Define(usdStage, SdfPath("/SkelRoot"));
+    errors = validator->Validate(mesh.GetPrim());
+
+    // Verify all errors are gone
+    TF_AXIOM(errors.size() == 0);
```
tallytalwar, line 97: Very nice test!!!

tallytalwar (commented):
Hey @beersandrew 

Thanks for doing this. Left a few suggestions for you to consider. Let me know if you want to talk through any of these or more here or on slack.

pxr/usd/usdSkel/CMakeLists.txt
beersandrew: removed, thanks

pxr/usd/usdSkel/CMakeLists.txt
beersandrew: this was not intentional, re-added.

pxr/usd/usdSkel/plugInfo.json
beersandrew: I did not change this yet, as I'll wait for some agreement (same issue in the MaterialBindingAPI PR). 

My suggestions: (or any combinations of:
 `<APISchema>` + `<Existence | Presence>` + `<Validator | Verifier | Checker>`
 )
- `SkelBindingAPIExistenceVerifier`
- `SkelBindingAPIPresenceValidator`
- ...

`Application` here makes me think of a program / app, rather than being applied, but I'm happy to move forward with any of the options.

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
beersandrew: updated

pxr/usd/usdSkel/testenv/testUsdSkelValidators.cpp
beersandrew: added

pxr/usd/usdSkel/validators.cpp
beersandrew: removed

pxr/usd/usdSkel/validators.cpp
beersandrew: great suggestion, updated

pxr/usd/usdSkel/validators.cpp
beersandrew: fixed

beersandrew: i think this is a great idea, implemented.

pxr/usd/usdSkel/validators.cpp
beersandrew: removed

pxr/usd/usdSkel/plugInfo.json
dgovil: Since the schemas are an appliedAPI schema, I kind of like  `SkelBindingAPIAppliedValidator`

tallytalwar: I am fine with that. Will also be renaming the StageMetadataChecker ones I have already synced in to be called StageMetadataValidator. Thanks both.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,89 @@
...
+PXR_NAMESPACE_OPEN_SCOPE
+
+static
+UsdValidationErrorVector
+_SkelBindingApiAppliedValidator(const UsdPrim &usdPrim)
+{
+    UsdValidationErrorVector errors;
+
+    if (!usdPrim.HasAPI<UsdSkelBindingAPI>()){
+        std::vector<TfToken> primPropertyNames = usdPrim.GetPropertyNames();
```
tallytalwar: nitpick: Its a good practice to declare variables just before use. So I would move  `primPropertyNames` below `skelPropertyNames`.

Also you can make this a `const vector`.

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,89 @@
...
+        static const std::unordered_set<TfToken, TfHash> skelPropertyNames = []() {
+            UsdSchemaRegistry& usdSchemaRegistry = UsdSchemaRegistry::GetInstance();
+            std::unique_ptr<UsdPrimDefinition> primDef = usdSchemaRegistry.BuildComposedPrimDefinition(TfToken(), {TfToken("SkelBindingAPI")});
```
tallytalwar: Ahh apologies, I should have noticed this before, you can use `UsdSkelTokens->SkelBindingAPI`, instead of `TfToken("SkelBindingAPI")`. Such tokens are often generated along with the schema.

https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdSkel/tokens.cpp#L55

pxr/usd/usdSkel/validators.cpp
```diff
@@ -0,0 +1,89 @@
...
+            std::vector<TfToken> vectorOfNames = primDef->GetPropertyNames();
```
tallytalwar: `const vector`

tallytalwar (approved):
Thanks a bunch @beersandrew. Will sync this in.

PR #3167, opened 2024-07-15
Apple: Include the USD patch version in the namespace

dgovil:
### Description of Change(s)

USD has set its internal namespace to include the major (always 0), and minor (year) versions, but hasn't thus far included the patch version.

Since USD doesn't follow semver, the patch version is actually quite important, as there may be ABI changes between USD versions within the same year.

Including the patch version will improve the intended nature of the namespacing IMHO. For example, 24.3 and 24.11 may be quite dramatically different in terms of ABI and API, but include the same namespace


jesschimein:
Filed as internal issue #USD-9850

nvmkuruc:
This seems like a good idea.  👍

PR #3170, opened 2024-07-17
Autodesk: Fix HdSt tests failing under HgiVulkan with Lavapipe

erikaharrison-adsk:
### Description of Change(s)

⚠ Note: This PR is for feature-hgi-vulkan branch.

## Introduction
The goal of this PR is to fix various issue with HgiVulkan to enable Lavapipe (software rasterizer) compatibility. This was done by fixing bugs found when running the HdSt tests. Lavapipe is a conformant Vulkan 1.3 implementation: all issues encountered are actually Vulkan API usage problems, and could very well be encountered on actual hardware device. No special code is required for supporting Lavapipe as long as HgiVulkan uses the Vulkan API correctly.

Most fixes fall into two categories:
- Device limits are ignored, mostly max memory size and alignment requirements
- Tests are OpenGL specific and should explicitly disable Vulkan

## Breakdown of changes
- testHgiVulkan
    - Problem: assumes Vulkan debugging is enabled
        - Fix: add HGI_ENABLE_VULKAN=1 and HGIVULKAN_DEBUG=1 to pxr_register_test ENV argument
- testHdStBufferAggregation
    - Problem: assumes that VkPhysicalDeviceLimits::minUniformBufferOffsetAlignment is 256
        - Fix: use HgiCapabilities::GetUniformBufferOffsetAlignment()
    - Problem: can generate more than 2GB of log files on failure
        - Fix: limit log to first 100 incorrect values in the huge buffer test
- testHdStBufferArray
    - Problem: output is device dependent (uboSize depends on VkPhysicalDeviceLimits::minUniformBufferOffsetAlignment) and diff'd against a constant baseline
        - Fix: ignore uboSize in diff. ssboSize is also ignored for the same reason as uboSize (depends on minStorageBufferOffsetAlignment)
- testHdStBarAllocationLimit
    - Problem: assumes that VkPhysicalDeviceLimits::maxStorageBufferRange is 2^30 (default value of setting HD_MAX_VBO_SIZE)
        - Fix: use the smallest of HgiCapabilities::GetMaxShaderStorageBlockSize() and HD_MAX_VBO_SIZE
- testHdStBasicDrawing
    - Problem: geometry shaders with more than VkPhysicalDeviceLimits::maxGeometryInputComponents input locations are being generated.
        - Fix: Inputs are passed as arrays of int, consuming one location per int. Instead pack them into ivec4 (for array sizes smaller or equal to 4, which is the max right now for quads), to pass up to 4 values per location.
- testHdStBufferAggregation
    - Problem: assumes that VkPhysicalDeviceLimits::maxStorageBufferRange is 2^30 (default value of setting HD_MAX_VBO_SIZE) when checking the vboRelocated performance counter (expects it to be 9)
        - Fix: use the smallest of HgiCapabilities::GetMaxShaderStorageBlockSize() and HD_MAX_VBO_SIZE to compute the correct expected value of vboRelocated.
- testHdStCodeGen
    - Problem: generated Vulkan shader results are different from the OpenGL baseline.
        - NOTE: this problem was fixed by another commit before this PR could be submitted. The PR retains some changes since they remove a lot of CMake script duplication.
        - Fix: Create GL and Vulkan variants of the tests so they can use a different baseline each.
- testHdStDrawItemsCache
    - Problem: is consuming command buffers faster than Lavapipe can execute them, so it reaches the maximum allowed inflight of 64. Normally buffers would be reset by _EndFrameSync after submission (or by EndFrame if it gets called), but since this sync is non-blocking, and Lavapipe is slow, it's called too early and does nothing (because all the buffers are still executing). And since all these calls must be on the main thread, it deadlocks: main thread is waiting for more command buffers, but those require calling _EndFrameSync to be made available, which it can't do because it's blocked...
        - Fix: Be smarter about allocating in-flight bits: search for the next available bit. If no in-flight bits are available, then update the existing command buffers in-flight status to release ones no longer in use, instead of just waiting (which is a deadlock).
- testHdStGL
    - Problem: OpenGL specific test
        - Fix: Disable Vulkan
- testHdStGLSL
    - Problem: OpenGL specific test
        - Fix: Disable Vulkan
- testHdStHWFaceCulling
    - Problem: output is device dependent (uboSize depends on VkPhysicalDeviceLimits::minUniformBufferOffsetAlignment) and diff'd against a constant baseline
        - Fix: ignore uboSize in diff. ssboSize is also ignored for the same reason as uboSize (depends on minStorageBufferOffsetAlignment)
- testHdStPrimvars
    - Problem: output is device dependent (uboSize depends on VkPhysicalDeviceLimits::minUniformBufferOffsetAlignment) and diff'd against a constant baseline
        - Fix: ignore uboSize in diff. ssboSize is also ignored for the same reason as uboSize (depends on minStorageBufferOffsetAlignment)

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9861

pxr/imaging/hdSt/codeGen.cpp
```diff
@@ -4611,6 +4718,8 @@ HdSt_CodeGen::_GenerateDrawingCoord(
                                   /*name=*/name,
                                   /*dataType=*/_tokens->_int);
         }
+
+        _PlumbInterstageElements();
```
DDoS: Turns out this `if (!_hasCS) { ... _PlumbInterstageElements(); }` block should be moved after the `static const std::vector<std::string> drawingCoordParams { ... }` block near the top of the function. Otherwise the code at line 4433 doesn't find the mapping. It's also missing the "vs_" prefix.

pxr/imaging/hgiVulkan/device.cpp
```diff
@@ -69,7 +69,8 @@ _SupportsPresentation(
                 vkGetInstanceProcAddr(
                     vkInstance,
                     "vkGetPhysicalDeviceXlibPresentationSupportKHR");
-        return vkGetPhysicalDeviceXlibPresentationSupportKHR(
+        return vkGetPhysicalDeviceXlibPresentationSupportKHR &&
```
clach, line 72: Just FYI, this gives me a compilation error: `the address of 'VkBool32 vkGetPhysicalDeviceXlibPresentationSupportKHR(VkPhysicalDevice, uint32_t, Display*, VisualID)' will never be NULL`

DDoS: That's interesting, I actually observed this being null, when I used Lavapipe headless (didn't build it with X support). Is it a warning as an error or an actual error?

clach: It's a warning as an error. Do you know if there's some define we can use to branch this?

DDoS: ~Maybe the WSI macros? https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/WSIheaders.html~
Nevermind, we're already using those. I might need to retest this to understand what's going wrong better.

DDoS: `vkGetInstanceProcAddr` can return null, so I don't understand why the compiler would warn here.

pxr/imaging/hdSt/codeGen.cpp
```diff
@@ -4104,58 +4196,72 @@ static void _EmitFVarAccessor(
...
-           << " = " << "dc.instanceIndex[" << index << "]" << ";\n";
+        const auto [groupName, component] = _GetDrawingCoordMapping("instanceIndexI" + std::to_string(i));
+        ss << "  " << outputPrefix << groupName << outArraySize << component
+           << " = " << "dc.instanceIndex[" << i << "]" << ";\n";
     }
     for(int i = 0; i < instanceIndexWidth-1; ++i) {
-        std::string const index = std::to_string(i);
-        ss << "  " << outputPrefix << "instanceCoordsI" << index << outArraySize
-           << " = " << "dc.instanceCoords[" << index << "]" << ";\n";
+        const auto [groupName, component] = _GetDrawingCoordMapping("instanceIndexI" + std::to_string(i));
```
DDoS: Bug: `"instanceIndexI"` should be `"instanceCoordsI"`

PR #3171, opened 2024-07-17
Autodesk: Update multisample state

erikaharrison-adsk:
### Description of Change(s)

[hdx] Update multisample state
- Update render pass multisample state based on parameters
- Update color correction task multisample state based on the sample count

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9862

PR #3182, opened 2024-07-31
[UsdLux] Add explanation of UsdLux quantities and behavior

pmolodo:
### Description of Change(s)

Adds more detailed and precise explanations of expected behavior for UsdLux lights to UsdLux docs.

---------------------------------------------------------------

### Related PRs:

:chains: **hdEmbree Reference Implementation:**
This PR is documentation changes only; see here for the PR chain that provides a reference implementation as part of hdEmbree:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3199

:eight_spoked_asterisk: **All UsdLux update related PRs:**
To see ALL UsdLux update related PRs (documentation AND reference implementation) in one place, see:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3200

:hourglass: **Original PR:**
These PRs are an update of this original PR:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/2758

---------------------------------------------------------------

### But... why??

**Why make these PRs in the first place?**

The current specifications of the various UsdLux prims + attributes are imprecise or vague in many places, and as a result, actual implementations of them by various renderers have diverged, sometimes quite significantly.  For instance, here is Intel's [4004 Moore Lane](https://dpel.aswf.io/4004-moore-lane/) scene, with the same UsdLux lights defined, in 3 different renderers:

Karma:
4004 Moore Lane, rendered in Karma

Arnold:
4004 Moore Lane, rendered in Arnold

Omniverse RTX:
4004 Moore Lane, rendered in Omniverse RTX

For a full descpription of the problem, see here:
- [Problem Statement](https://github.com/anderslanglands/light_comparison/blob/main/README.md#problem-statement)

**Why so many PRs?**
This was my attempt to break up a rather large change into smaller, more easily reviewable changes, that can be merged in incrementally, to help ease the burden for code reviewers.

If you find this confusing, and would rather just one big PR (ie, just this one), or have them organized in some other way, please let me know!

**Why are the documentation changes in their own PR?**
In some ways, the [documentation changes](https://github.com/PixarAnimationStudios/OpenUSD/pull/3182) are the heart of this effort - we wish to specify more exactly what the various UsdLux prims and attributes represent.  However, building consensus on this may take time - so we expect some dialogue on the exact language or formulas.

Because that may take time, the [hdEmbree reference implementation changes](https://github.com/PixarAnimationStudios/OpenUSD/pull/3199) are separate, and broken up, so we can hopefully start integrating portions of them even before final consensus has been reached on the final form of the specification.

**Why are the documentation changes not broken up into smaller pieces, like the hdEmbree reference implementation changes?**
Because I wasn't sure if that would be desirable or not!  If people think that would be helpful, I can do so - perhaps breaking out by schema or individual attribute?

---------------------------------------------------------------

### DIFF
--- a/pxr/usd/usdLux/lightAPI.h
+++ b/pxr/usd/usdLux/lightAPI.h
@@ -46,6 +46,41 @@ class SdfAssetPath;
 /// type (e.g. RectLight or DistantLight) or is applied directly to a Gprim 
 /// that should be treated as a light.
 /// 
+/// <b>Quantities and Units</b>
+/// 
+/// Most renderers consuming OpenUSD today are RGB renderers, rather than
+/// spectral. Units in RGB renderers are tricky to define as each of the red,
+/// green and blue channels transported by the renderer represents the
+/// convolution of a spectral exposure distribution, e.g. CIE Illuminant D65,
+/// with a sensor response function, e.g. CIE 1931 𝓍̅. Thus the main quantity
+/// in an RGB renderer is neither radiance nor luminance, but "integrated
+/// radiance" or "tristimulus weight".
+/// 
+/// The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is
+/// an Illuminant D spectral distribution with chromaticity matching the
+/// rendering color space white point, normalized such that a ray normally
+/// incident upon the sensor with EV0 exposure settings will generate a pixel
+/// value of [1, 1, 1] in the rendering color space.
+/// 
+/// Given the above definition, that means that the luminance of said default
+/// light will be 1 *nit (cd∕m²)* and its emission spectral radiance
+/// distribution is easily computed by appropriate normalization.
+/// 
+/// For brevity, the term *emission* will be used in the documentation to mean
+/// "emitted spectral radiance" or "emitted integrated radiance/tristimulus
+/// weight", as appropriate.
+/// 
+/// The method of "uplifting" an RGB color to a spectral distribution is
+/// unspecified other than that it should round-trip under the rendering
+/// illuminant to the limits of numerical accuracy.
+/// 
+/// Note that some color spaces, most notably ACES, define their white points
+/// by chromaticity coordinates that do not exactly line up to any value of a
+/// standard illuminant. Because we do not define the method of uplift beyond
+/// the round-tripping requirement, we discourage the use of such color spaces
+/// as the rendering color space, and instead encourage the use of color spaces
+/// whose white point has a well-defined spectral representation, such as D65.
+/// 
 /// <b>Linking</b>
 /// 
 /// Lights can be linked to geometry.  Linking controls which geometry
@@ -268,7 +303,22 @@ public:
     // --------------------------------------------------------------------- //
     // INTENSITY 
     // --------------------------------------------------------------------- //
-    /// Scales the power of the light linearly.
+    /// Scales the brightness of the light linearly.
+    /// 
+    /// Expresses the "base", unmultiplied luminance emitted (L) of the light,
+    /// in nits (cd∕m²):
+    /// 
+    /// <center><b>
+    /// L<sub>Scalar</sub> = intensity
+    /// </b></center>
+    /// 
+    /// Normatively, the lights' emission is in units of spectral radiance
+    /// normalized such that a directly visible light with `intensity` 1 and
+    /// `exposure` 0 normally incident upon the sensor plane will generate a
+    /// pixel value of [1, 1, 1] in an RGB renderer, and thus have a luminance
+    /// of 1 nit. A light with `intensity` 2 and `exposure` 0 would therefore
+    /// have a luminance of 2 nits.
+    /// 
     ///
     /// | ||
     /// | -- | -- |
@@ -290,9 +340,21 @@ public:
     // --------------------------------------------------------------------- //
     // EXPOSURE 
     // --------------------------------------------------------------------- //
-    /// Scales the power of the light exponentially as a power
+    /// Scales the brightness of the light exponentially as a power
     /// of 2 (similar to an F-stop control over exposure).  The result
-    /// is multiplied against the intensity.
+    /// is multiplied against the intensity:
+    /// 
+    /// <center><b>
+    /// L<sub>Scalar</sub> = L<sub>Scalar</sub> ⋅ 2<sup>exposure</sup>
+    /// </b></center>
+    /// 
+    /// Normatively, the lights' emission is in units of spectral radiance
+    /// normalized such that a directly visible light with `intensity` 1 and
+    /// `exposure` 0 normally incident upon the sensor plane will generate a
+    /// pixel value of [1, 1, 1] in an RGB renderer, and thus have a luminance
+    /// of 1 nit (cd∕m²). A light with `intensity` 1 and `exposure` 2 would
+    /// therefore have a luminance of 4 nits.
+    /// 
     ///
     /// | ||
     /// | -- | -- |
@@ -360,10 +422,119 @@ public:
     // --------------------------------------------------------------------- //
     // NORMALIZE 
     // --------------------------------------------------------------------- //
-    /// Normalizes power by the surface area of the light.
-    /// This makes it easier to independently adjust the power and shape
-    /// of the light, by causing the power to not vary with the area or
-    /// angular size of the light.
+    /// Normalizes the emission such that the power of the light
+    /// remains constant while altering the size of the light, by dividing the
+    /// luminance by the world-space surface area of the light.
+    /// 
+    /// This makes it easier to independently adjust the brightness and size
+    /// of the light, by causing the total illumination provided by a light to
+    /// not vary with the area or angular size of the light.
+    /// 
+    /// Mathematically, this means that the luminance of the light will be
+    /// divided by a factor representing the "size" of the light:
+    /// 
+    /// <center><b>
+    /// L<sub>Scalar</sub> = L<sub>Scalar</sub> / sizeFactor
+    /// </b></center>
+    /// 
+    /// ...where `sizeFactor` = 1 if `normalize` is off, and is calculated
+    /// depending on the family of the light as described below if `normalize`
+    /// is on.
+    /// 
+    /// ### DomeLight / PortalLight:
+    /// 
+    /// For a dome light (and its henchman, the PortalLight), this attribute is
+    /// ignored:
+    /// 
+    /// <center><b>
+    /// sizeFactor<sub>dome</sub> = 1
+    /// </b></center>
+    /// 
+    /// ### Area Lights:
+    /// 
+    /// For an area light, the `sizeFactor` is the surface area (in world
+    /// space) of the shape of the light, including any scaling applied to the
+    /// light by its transform stack. This includes the boundable light types
+    /// which have a calculable surface area:
+    /// 
+    /// - MeshLightAPI
+    /// - DiskLight
+    /// - RectLight
+    /// - SphereLight
+    /// - CylinderLight
+    /// - (deprecated) GeometryLight
+    /// 
+    /// <center><b>
+    /// sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light)
+    /// </b></center>
+    /// 
+    /// ### DistantLight:
+    /// 
+    /// For distant lights, we first define 𝛳<sub>max</sub> as:
+    /// 
+    /// <center><b>
+    /// 𝛳<sub>max</sub> = clamp(toRadians(distantLightAngle) / 2, 0, 𝜋)
+    /// </b></center>
+    /// 
+    /// Then we use the following formula:
+    /// 
+    /// * <i>if 𝛳<sub>max</sub> = 0:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> = 1
+    /// </b></center>
+    /// 
+    /// * <i>if 0 < 𝛳<sub>max</sub> ≤ 𝜋 / 2:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> = sin²𝛳<sub>max</sub> ⋅ 𝜋
+    /// </b></center>
+    /// 
+    /// * <i>if 𝜋 / 2 < 𝛳<sub>max</sub> ≤ 𝜋:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> =
+    /// (2 - sin²𝛳<sub>max</sub>) ⋅ 𝜋
+    /// </b></center>
+    /// 
+    /// This formula is used because it satisfies the following two properties:
+    /// 
+    /// 1. When normalize is enabled, the received illuminance from this light
+    /// on a surface normal to the light's primary direction is held constant
+    /// when angle changes, and the "intensity" property becomes a measure of
+    /// the illuminance, expressed in lux, for a light with 0 exposure.
+    /// 
+    /// 2. If we assume that our distant light is an approximation for a "very
+    /// far" sphere light (like the sun), then (for
+    /// *0 < 𝛳<sub>max</sub> ≤ 𝜋/2*) this definition agrees with the
+    /// definition used for area lights - ie, the total power of this distant
+    /// sphere light is constant when the "size" (ie, angle) changes, and our
+    /// sizeFactor is proportional to the total surface area of this sphere.
+    /// 
+    /// ### Other Lights
+    /// 
+    /// The above taxonomy describes behavior for all built-in light types.
+    /// (Note that the above is based on schema *family* - ie, `DomeLight_1`
+    /// follows the rules for a `DomeLight`, and ignores `normalize`.)
+    /// 
+    /// Lights from other third-party plugins / schemas must document their
+    /// expected behavior with regards to normalize.  However, some general
+    /// guidelines are:
+    /// 
+    /// - Lights that either inherit from or are strongly associated with one of
+    /// the built-in types should follow the behavior of the built-in type
+    /// they inherit/resemble; ie, a renderer-specific "MyRendererRectLight"
+    /// should have its size factor be its world-space surface area
+    /// - Lights that are boundable and have a calcuable surface area should
+    /// follow the rules for an Area Light, and have their sizeFactor be their
+    /// world-space surface area
+    /// - Lights that are non-boundable and/or have no way to concretely or even
+    /// "intuitively" associate them with a "size" will ignore this attribe
+    /// (and always set sizeFactor = 1)
+    /// 
+    /// Lights that don't clearly meet any of the above criteria may either
+    /// ignore the normalize attribute or try to implement support using
+    /// whatever hueristic seems to make sense - for instance,
+    /// MyMandelbulbLight might use a sizeFactor equal to the world-space
+    /// surface area of a sphere which "roughly" bounds it.
+    /// 
     ///
     /// | ||
     /// | -- | -- |
@@ -385,7 +556,20 @@ public:
     // --------------------------------------------------------------------- //
     // COLOR 
     // --------------------------------------------------------------------- //
-    /// The color of emitted light, in energy-linear terms.
+    /// The color of emitted light, in the rendering color space.
+    /// 
+    /// This color is just multiplied with the emission:
+    /// 
+    /// <center><b>
+    /// L<sub>Color</sub> = L<sub>Scalar</sub> ⋅ color
+    /// </b></center>
+    /// 
+    /// In the case of a spectral renderer, this color should be uplifted such
+    /// that it round-trips to within the limit of numerical accuracy under the
+    /// rendering illuminant.  We recommend the use of a rendering color space
+    /// well defined in terms of a Illuminant D illuminant, to avoid unspecified
+    /// uplift.  See: \ref usdLux_quantities
+    /// 
     ///
     /// | ||
     /// | -- | -- |
@@ -436,6 +620,17 @@ public:
     /// enableColorTemperature is set to true.  When active, the
     /// computed result multiplies against the color attribute.
     /// See UsdLuxBlackbodyTemperatureAsRgb().
+    /// 
+    /// This is always calculated as an RGB color using a D65 white point,
+    /// regardless of the rendering color space, normalized such that the
+    /// default value of 6500 will always result in white, and then should be
+    /// transformed to the rendering color space.
+    /// 
+    /// Spectral renderers should do the same and then uplift the resulting
+    /// color after multiplying with the `color` attribute.  We recommend the
+    /// use of a rendering color space well defined in terms of a Illuminant D
+    /// illuminant, to avoid unspecified uplift.  See: \ref usdLux_quantities
+    /// 
     ///
     /// | ||
     /// | -- | -- |
--- a/pxr/usd/usdLux/overview.dox
+++ b/pxr/usd/usdLux/overview.dox
@@ -101,6 +101,55 @@ constraints, such as to attach lights to moving geometry.  This is
 consistent with USD's policy of storing the computed result of rigging,
 not the rigging itself.
 
+\subsection usdLux_quantities Quantities and Units
+
+Most renderers consuming OpenUSD today are RGB renderers, rather than spectral.
+Units in RGB renderers are tricky to define as each of the red, green and blue
+channels transported by the renderer represents the convolution of a spectral
+exposure distribution, e.g. CIE Illuminant D65, with a sensor response function,
+e.g. CIE 1931 𝓍̅. Thus the main quantity in an RGB renderer is neither radiance
+nor luminance, but "integrated radiance" or "tristimulus weight".
+
+The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is an
+Illuminant D spectral distribution with chromaticity matching the rendering
+color space white point, normalized such that a ray normally incident upon the
+sensor with EV0 exposure settings will generate a pixel value of [1, 1, 1] in
+the rendering color space.
+
+Given the above definition, that means that the luminance of said default light
+will be 1 *nit (cd∕m²)* and its emission spectral radiance distribution is
+easily computed by appropriate normalization.
+
+For brevity, the term *emission* will be used in the documentation to mean
+"emitted spectral radiance" or "emitted integrated radiance/tristimulus weight",
+as appropriate.
+
+The method of "uplifting" an RGB color to a spectral distribution is unspecified
+other than that it should round-trip under the rendering illuminant to the
+limits of numerical accuracy.
+
+Note that some color spaces, most notably ACES, define their white points by
+chromaticity coordinates that do not exactly line up to any value of a standard
+illuminant. Because we do not define the method of uplift beyond the round-
+tripping requirement, we discourage the use of such color spaces as the
+rendering color space, and instead encourage the use of color spaces whose white
+point has a well-defined spectral representation, such as D65.
+
+\subsubsection usdLux_quantities_exposure Exposure
+
+In order to generate an image, a camera effectively integrates irradiance at a patch
+on the sensor over time to get exposure, which is converted to an electrical signal.
+The default behaviour of many renderers is to ignore this, which as we have seen means
+that 1 nit incident on the sensor results in an output signal of 1.0.
+
+1 nit is a tiny amount of light, equivalent by definition to a single candle viewed at a distance. Thus
+using physically plausible values for light brightnesses will result in blown out images
+unless exposure is taken into account. A full physical camera model would be a good
+future addition to OpenUSD, but there is already a means of communicating sensor
+exposure via `UsdGeomCamera::GetExposureAttr()` and it is expected that all renderers
+implementing UsdLux must respect the value of that attribute by scaling the rendered
+image by `pow(2, exposure)`.
+
 \subsection usdLux_Behavior Properties & Behavior
 
 Colors specified in attributes are in energy-linear terms,
@@ -112,6 +161,14 @@ distance is a consequence of reduced visible solid angle. Environments
 that do not measure the visible solid angle are expected to provide an
 approximation, such as inverse-square falloff. Further artistic control
 over attenuation can be modelled as light filters.
+
+The behavior of all attributes is in UsdLuxLightAPI is documented on their `Get()`
+methods. All attributes are expected to be supported by all light types, with
+the effect of each attribute applying multiplicatively in sequence.
+
+The example hdEmbree delegate provides a reference implementation, as well as a
+series of unit tests and reference images for renderers to check their own
+implementation against.
 
 More complex behaviors are provided via a set of API classes.
 These behaviors are common and well-defined but may not be supported
--- a/pxr/usd/usdLux/schema.usda
+++ b/pxr/usd/usdLux/schema.usda
@@ -53,6 +53,41 @@ class "LightAPI" (
     type (e.g. RectLight or DistantLight) or is applied directly to a Gprim 
     that should be treated as a light.
 
+    <b>Quantities and Units</b>
+
+    Most renderers consuming OpenUSD today are RGB renderers, rather than
+    spectral. Units in RGB renderers are tricky to define as each of the red,
+    green and blue channels transported by the renderer represents the
+    convolution of a spectral exposure distribution, e.g. CIE Illuminant D65,
+    with a sensor response function, e.g. CIE 1931 𝓍̅. Thus the main quantity
+    in an RGB renderer is neither radiance nor luminance, but "integrated
+    radiance" or "tristimulus weight".
+
+    The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is
+    an Illuminant D spectral distribution with chromaticity matching the
+    rendering color space white point, normalized such that a ray normally
+    incident upon the sensor with EV0 exposure settings will generate a pixel
+    value of [1, 1, 1] in the rendering color space.
+
+    Given the above definition, that means that the luminance of said default
+    light will be 1 *nit (cd∕m²)* and its emission spectral radiance
+    distribution is easily computed by appropriate normalization.
+
+    For brevity, the term *emission* will be used in the documentation to mean
+    "emitted spectral radiance" or "emitted integrated radiance/tristimulus
+    weight", as appropriate.
+
+    The method of "uplifting" an RGB color to a spectral distribution is
+    unspecified other than that it should round-trip under the rendering
+    illuminant to the limits of numerical accuracy.
+
+    Note that some color spaces, most notably ACES, define their white points
+    by chromaticity coordinates that do not exactly line up to any value of a
+    standard illuminant. Because we do not define the method of uplift beyond
+    the round-tripping requirement, we discourage the use of such color spaces
+    as the rendering color space, and instead encourage the use of color spaces
+    whose white point has a well-defined spectral representation, such as D65.
+
     <b>Linking</b>
 
     Lights can be linked to geometry.  Linking controls which geometry
@@ -152,7 +187,22 @@ class "LightAPI" (
     float inputs:intensity = 1 (
         displayGroup = "Basic"
         displayName = "Intensity"
-        doc = """Scales the power of the light linearly."""
+        doc = """Scales the brightness of the light linearly.
+
+        Expresses the "base", unmultiplied luminance emitted (L) of the light,
+        in nits (cd∕m²):
+
+        <center><b>
+                        L<sub>Scalar</sub> = intensity
+        </b></center>
+
+        Normatively, the lights' emission is in units of spectral radiance
+        normalized such that a directly visible light with `intensity` 1 and
+        `exposure` 0 normally incident upon the sensor plane will generate a
+        pixel value of [1, 1, 1] in an RGB renderer, and thus have a luminance
+        of 1 nit. A light with `intensity` 2 and `exposure` 0 would therefore
+        have a luminance of 2 nits.
+        """
         customData = {
             token apiName = "intensity"
         }
@@ -160,9 +210,21 @@ class "LightAPI" (
     float inputs:exposure = 0 (
         displayGroup = "Basic"
         displayName = "Exposure"
-        doc = """Scales the power of the light exponentially as a power
+        doc = """Scales the brightness of the light exponentially as a power
         of 2 (similar to an F-stop control over exposure).  The result
-        is multiplied against the intensity."""
+        is multiplied against the intensity:
+
+        <center><b>
+                L<sub>Scalar</sub> = L<sub>Scalar</sub> ⋅ 2<sup>exposure</sup>
+        </b></center>
+
+        Normatively, the lights' emission is in units of spectral radiance
+        normalized such that a directly visible light with `intensity` 1 and
+        `exposure` 0 normally incident upon the sensor plane will generate a
+        pixel value of [1, 1, 1] in an RGB renderer, and thus have a luminance
+        of 1 nit (cd∕m²). A light with `intensity` 1 and `exposure` 2 would
+        therefore have a luminance of 4 nits.
+        """
         customData = {
             token apiName = "exposure"
         }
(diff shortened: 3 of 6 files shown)


jesschimein:
Filed as internal issue #USD-9900

pxr/usd/usdLux/lightAPI.h
```diff
@@ -46,6 +46,40 @@ class SdfAssetPath;
...
+/// 
+/// Given the above definition, that means that the luminance of said default
+/// light will be 1 *nit (cd∕m²)* and its emission spectral radiance
+/// distribution is easily computed by appropriate normalization.
+/// 
+/// For brevity, the term *emission* will be used in the documentation to mean
+/// "emitted spectral radiance" or "emitted integrated radiance/tristimulus
+/// weight", as appropriate.
+/// 
+/// Note that some color spaces, most notably ACES, define their white points
```
meshula, line 77: We should reword this section about ACES.We should convey this information, but emphasize that because a standard uplift is not specified beyond the requirement of a round-trip, we recommend using color spaces specified with respect to a standard illuminant.

The intent of this suggestion is to encourage the use of color spaces that are naturally conducive to physical colorimetry. If we word this section as it is now, it is an invitation to endless troubleshooting and a continuation of work arounds and hacks to accommodate adaptation and "uplift" which leaves us outside the realm of strictly measurable science.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -46,6 +46,40 @@ class SdfAssetPath;
...
+/// 
+/// Most renderers consuming OpenUSD today are RGB renderers, rather than
+/// spectral. Units in RGB renderers are tricky to define as each of the red,
+/// green and blue channels transported by the renderer represents the
+/// convolution of a spectral exposure distribution, e.g. CIE Illuminant D65,
+/// with a sensor response function, e.g. CIE 1931 𝓍̅. Thus the main quantity
+/// in an RGB renderer is neither radiance nor luminance, but "integrated
+/// radiance" or "tristimulus weight".
+/// 
+/// The emision of a default light with `intensity` 1 and `color` [1, 1, 1] is
```
meshula: typo, "emission"

pxr/usd/usdLux/lightAPI.h
```diff
@@ -267,7 +301,22 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// Scales the brightness of the light linearly.
+    /// 
+    /// Expresses the "base", unmultiplied luminance emitted (L) of the light,
+    /// in nits (cd∕m²):
+    /// 
+    /// <center><b>
+    /// L<sub>Scalar</sub> = intensity
+    /// </b></center>
+    /// 
+    /// More precisely, the lights' emission is in units of spectral radiance
```
meshula: "More precisely" as used throughout this document slightly carries the meaning that there might be "even more precise" ways to describe things, but that's not the case. Can we use a wording such as "Normatively" or "Mathematically" instead? I don't want people to wonder if there is another shoe to drop in the future.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +420,120 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// </b></center>
+    /// 
+    /// 
+    /// ...where `sizeFactor` = 1 if `normalize` is off, and is calculated
+    /// depending on the family of the light as described below if `normalize`
+    /// is on.
+    /// 
+    /// ### DomeLight / PortalLight:
+    /// 
+    /// For a dome light (and it's henchman, the PortalLight), this attribute is
```
meshula: typo "its"

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +420,120 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// <center><b>
+    /// sizeFactor<sub>dome</sub> = 1
+    /// </b></center>
+    /// 
+    /// ### Area Lights:
+    /// 
+    /// For an area light, the `sizeFactor` is the surface area (in world
+    /// space) of the shape of the light, including any scaling applied to the
+    /// light by its transform stack. This includes the boundable light types
+    /// which can have a calculable surface area, such as:
```
meshula: we should be more normative. Instead of "can have" we should say "does have". If there is an exception we should call it out.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +420,120 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// (2 - sin²𝛳<sub>max</sub>) ⋅ 𝜋
+    /// </b></center>
+    /// 
+    /// This formula is used because it satisfies the following two properties:
+    /// 
+    /// 1. When normalize is enabled, the received illuminance from this light
+    /// on a surface normal to the light's primary direction is held constant
+    /// when angle changes, and the "intensity" property becomes a measure of
+    /// the illuminance, expressed in lux, for a light with 0 exposure.
+    /// 2. If we assume that our distant light is an approximation for a "very
```
meshula, line 504: could we insert a blank line here preceding the "2." bullet

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +420,120 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// 
+    /// Lights from other third-party plugins / schemas must document their
+    /// expected behavior with regards to normalize.  However, some general
+    /// guidelines are:
+    /// 
+    /// - Lights that either inherit from or are strongly associated with one of
+    /// the built-in types should follow the behavior of the built-in type
+    /// they inherit/resemble; ie, a renderer-specific "MyRendererRectLight"
+    /// should have it's size factor be it's world-space surface area
+    /// - Lights that are boundable and have a calcuable surface area should
```
meshula, line 525: typo "calculable"

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +420,120 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// generally follow the rules for an Area Light, and have their
+    /// sizeFactor be their world-space surface area
+    /// - Lights that are non-boundable and/or have no way to "intuitively"
```
meshula: typo + suggestion

> Lights that are non-boundable or otherwise have no meaningful notion of a size should ignore this attribute.

Logic: if they do not ignore size, then by definition size is meaningful.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -384,7 +555,17 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
     // --------------------------------------------------------------------- //
     // COLOR 
     // --------------------------------------------------------------------- //
-    /// The color of emitted light, in energy-linear terms.
+    /// The color of emitted light, in the rendering color space.
+    /// 
+    /// This color is just multiplied with the emission. In the case of a
```
meshula: Again, the suggestion here is to phrase this section to bias towards avoiding the need to uplift in the first place. We are better served to gently steer people away from wobbly-wobbly bits and towards methods based on measurable colorimetry.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -435,6 +616,15 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
     /// enableColorTemperature is set to true.  When active, the
     /// computed result multiplies against the color attribute.
     /// See UsdLuxBlackbodyTemperatureAsRgb().
+    /// 
+    /// This is always calculated as an RGB color using a D65 white point,
+    /// regardless of the rendering color space, normalized such that the
+    /// default value of 6500 will always result in white, and then should be
+    /// transformed to the rendering color space.
+    /// 
+    /// Spectral renderers should do the same and then uplift the resulting
```
meshula, line 629: "uplift" please imagine my tears.

with respect to my preceding comment, a spectral render, and any other renderer !!!!! should utilize the observed CIEXY triple as noted above for maximum goodness.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -435,6 +616,15 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
     /// enableColorTemperature is set to true.  When active, the
     /// computed result multiplies against the color attribute.
     /// See UsdLuxBlackbodyTemperatureAsRgb().
+    /// 
+    /// This is always calculated as an RGB color using a D65 white point,
```
meshula, line 624: The property should have an associated color space, explicitly restricted to one of the canonical color spaces described in the GfColorSpace enumeration. It is then the responsibility of the user to use the facilities in GfColor/ColorSpace to resolve the value into an rgb triple.  [Edit] the following has no real bearing on this discussion.  ~The Color Interop standard, to which GfColorSpace adheres, has introduced a CIEXY space which would be a good choice for the default color space associated with the color temperature.~

pxr/usd/usdLux/overview.dox
```diff
@@ -101,11 +101,54 @@ constraints, such as to attach lights to moving geometry.  This is
...
+\subsection usdLux_quantities Quantities and Units
+
+Most renderers consuming OpenUSD today are RGB renderers, rather than spectral.
+Units in RGB renderers are tricky to define as each of the red, green and blue
+channels transported by the renderer represents the convolution of a spectral
+exposure distribution, e.g. CIE Illuminant D65, with a sensor response function, e.g.
+CIE 1931 𝓍̅. Thus the main quantity in an RGB renderer is neither radiance nor
+luminance, but "integrated radiance" or "tristimulus weight".
+
+The emision of a default light with `intensity` 1 and `color` [1, 1, 1] is an
```
meshula: since this section is copy/paste of things I already noted above, eg, "emision" typo and subsequent notes, this comment is merely a note to propagate fixes here as well.

pxr/usd/usdLux/overview.dox
```diff
@@ -101,11 +101,54 @@ constraints, such as to attach lights to moving geometry.  This is
...
+using physically plausible values for light brightnesses will result in blown out images
+unless exposure is taken into account. A full physical camera model would be a good
+future addition to OpenUSD, but there is already a means of communicating sensor
+exposure via `UsdGeomCamera::GetExposureAttr()` and it is expected that all renderers
+implementing UsdLux must respect the value of that attribute by scaling the rendered
+image by `pow(2, exposure)`.
+
 \subsection usdLux_Behavior Properties & Behavior
 
 Colors specified in attributes are in energy-linear terms,
```
meshula, line 155: This section is now obsolete. The colorspace metadata must conform to the available colorimetry defined by GfColor/ColorSpace. Further the UsdStage::SetColorConfiguration() **does not** have any bearing on rendering. Within the rendering domain, colorimetry is strictly reduced to that specified by ocio/nanocolor as implemented by gfColor*. The mention of SetColorConfiguration should be removed in this context, as it is used to set color configurations governed by OCIO, such as the viewing transform. Color configurations, explicitly can not influence the colorimetry defined normatively by the ocio/nanocolor.

This section should reflect that by pointing to gfcolor*, ocio/nano, and so on. Perhaps we could workshop it a bit here.

meshula:
Just to add a general point, thank you so much for taking such a detailed approach to the documentation, and for bringing the concerns of spectral rendering to a clear focus for readers of this documentation.

pxr/usd/usdLux/overview.dox
pmolodo: Are you saying that the attribute colorSpace metadata should apply, but the stage color configuration should not?  If so, what should the default colorSpace be when an attribute doesn't explicitly specify?

Also - this implies that we're now creating two classes of color attributes - those to which the stage level SetColorConfiguration applies, and those to which it doesn't.  I feel that there should be some programmatic way to query this, rather than relying on an end user knowing which attributes it applies to and those to which it doesn't. 
 (Though I don't necessarily think we should gate these changes until the presence of such an API - just that they should be on the near-term roadmap.)
 
 Also - just wanted to make clear that I'm not trying to argue against this approach, just pointing out some resulting considerations, and trying to get additional clarification!

pmolodo:
> Just to add a general point, thank you so much for taking such a detailed approach to the documentation, and for bringing the concerns of spectral rendering to a clear focus for readers of this documentation.

And just to make sure that credit is directed appropriately - I'm just the caretaker/window-dresser for this PR; the true meaty content is almost all thanks to @anderslanglands's  original PR: https://github.com/PixarAnimationStudios/OpenUSD/pull/2758...

And that goes doubly true for all the spectral rendering content, where my knowledge is fairly limited!.  🙏 @anderslanglands

pxr/usd/usdLux/lightAPI.h
pmolodo: Oops, fixed: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R59

pxr/usd/usdLux/lightAPI.h
pmolodo: @anderslanglands @meshula - does this seem acceptable?

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R73-R82

pxr/usd/usdLux/lightAPI.h
pmolodo: Updated all these to "Normatively" or "Mathematically" as appropriate:
https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R314

pxr/usd/usdLux/lightAPI.h
pmolodo: Oops, fixed all these: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R445

pxr/usd/usdLux/lightAPI.h
pmolodo: @anderslanglands @meshula  - suggested change here:

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R560-R569

pxr/usd/usdLux/lightAPI.h
pmolodo: @meshula @anderslanglands - I was assuming you meant that this sentence could just be removed?

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4L625-L627

pxr/usd/usdLux/lightAPI.h
pmolodo: Fixed: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R445

pxr/usd/usdLux/lightAPI.h
pmolodo: Changed to "light types which have":

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R457

pxr/usd/usdLux/lightAPI.h
pmolodo: Done: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R502

Note that I checked, and this would only affect the in-source comment - the doxygen rendered html stayed the same:

image

If we want to give extra space there, I'll have to... I dunno, add another blank line, or `<p>`, or something...

pxr/usd/usdLux/lightAPI.h
pmolodo: Fixed: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-11634bb5438749e3c529c9f44011d464e79bdb1352c0b7fc7c9408bece4415e4R524

pxr/usd/usdLux/lightAPI.h
pmolodo: I assume the typo was the "calcuable" you also noted [here](https://github.com/PixarAnimationStudios/OpenUSD/pull/3182#discussion_r1702859881)?

As for the logic note - I agree!  ...but I'm not clear on what change you wanted to make as a result.  Did you mean that we should just eliminate the sentence because it's obviously true? Or just eliminate the use of the word "meaningful"?

Suggested change: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-bdeeccf2a8491e83067bc3c1e20e5a696612da9b31d1e3b7e6e7fc33b3592ec7R355-R357

pxr/usd/usdLux/lightAPI.h
pmolodo: See my comments / questions here: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182#discussion_r1703195632

pxr/usd/usdLux/overview.dox
pmolodo: Yup, done!

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/a1d7d8328149bf6067832d65c79f003fcfc262fa#diff-3042b10f5eed7d1dc28467437ef7f7c3cad9ccb30eb8dd93820575402328c11eR113

pxr/usd/usdLux/lightAPI.h
anderslanglands: If we’re going to do this then do we want to say something like “encourage the use of colour spaces whose white point has a well-defined spectral representation, such as D65”?

pxr/usd/usdLux/lightAPI.h
anderslanglands: In this case I think the original wording was good. You _have_ to uplift an RGB in order to render spectrally.

pxr/usd/usdLux/lightAPI.h
anderslanglands: I worded it this way so as to maintain compatibility between RGB and spectral. Later, when we add illuminants, then we would do this properly. 

The alternative is we just go the whole hog and add in the illuminant spectra now, making one of the options a blackbody, and change the meaning of this to only apply when that illuminant spectrum is selected.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +421,119 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
     // --------------------------------------------------------------------- //
     // NORMALIZE 
     // --------------------------------------------------------------------- //
-    /// Normalizes power by the surface area of the light.
+    /// Normalizes the emission such that the power of the light
+    /// remains constant while altering the size of the light, by dividing the
+    /// luminance by the world-space surface area of the light.
+    /// 
     /// This makes it easier to independently adjust the power and shape
```
meshula: should be `adjust the brightness and size` per @anderslanglands

pxr/usd/usdLux/overview.dox
```diff
@@ -101,11 +101,60 @@ constraints, such as to attach lights to moving geometry.  This is
...
+
+Most renderers consuming OpenUSD today are RGB renderers, rather than spectral.
+Units in RGB renderers are tricky to define as each of the red, green and blue
+channels transported by the renderer represents the convolution of a spectral
+exposure distribution, e.g. CIE Illuminant D65, with a sensor response function,
+e.g. CIE 1931 𝓍̅. Thus the main quantity in an RGB renderer is neither radiance
+nor luminance, but "integrated radiance" or "tristimulus weight".
+
+The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is an
+Illuminant D spectral power distribution with chromaticity matching the
```
meshula: `spectral distribution` ~ the word power isn't helping clarity

pxr/usd/usdLux/lightAPI.h
pmolodo: Tweaked: 

@anderslanglands 
https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/e493051f2d48654482586cbc32d46eb3a84b930d#diff-ec32e12ccc5be018c3a056d787ed4acdddd0653b33d1cc95593a8e5c26a1a75fR51

pxr/usd/usdLux/lightAPI.h
pmolodo: Restored the instructions about uplifting, but also kept the recommendation to use a "well defined" color space:

@anderslanglands 

https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/e493051f2d48654482586cbc32d46eb3a84b930d#diff-ec32e12ccc5be018c3a056d787ed4acdddd0653b33d1cc95593a8e5c26a1a75fR85-R89

pxr/usd/usdLux/lightAPI.h
pmolodo: @anderslanglands @meshula - 

Hmm - I restored the sentence* (just so it's more visible for reference in discussion) for now, but it seems we need to reach some more consensus on how to handle this.

*(Sorry if this causes more :eye: :droplet:, @meshula !)

pxr/usd/usdLux/lightAPI.h
pmolodo: @anderslanglands 
Done: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/2533cb0a9bbfed73a5e2c970c67d8b2f91a27b52#diff-bdeeccf2a8491e83067bc3c1e20e5a696612da9b31d1e3b7e6e7fc33b3592ec7R256

pxr/usd/usdLux/overview.dox
pmolodo: @anderslanglands 
Done: https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/2533cb0a9bbfed73a5e2c970c67d8b2f91a27b52#diff-bdeeccf2a8491e83067bc3c1e20e5a696612da9b31d1e3b7e6e7fc33b3592ec7R67

pxr/usd/usdLux/lightAPI.h
meshula: This wording is much more specific about avoiding unspecified uplift, so I like the revision.

pmolodo: We actually hadn't added any new language to this specific section (`GetColorTemperatureAttr()`) about avoiding unspecified uplift, so I added some (and mirrored the changes to the docs for `GetColorAttr()`).


- [specifically mention desire to avoid unspecified uplight - e12942c](https://github.com/PixarAnimationStudios/OpenUSD/pull/3182/commits/e12942c9740de80af0ed285e4f05527d9de75464)

pxr/usd/usdLux/lightAPI.h
meshula: it's fine, just the code

pxr/usd/usdLux/lightAPI.h
```diff
@@ -46,6 +46,41 @@ class SdfAssetPath;
...
+/// with a sensor response function, e.g. CIE 1931 𝓍̅. Thus the main quantity
+/// in an RGB renderer is neither radiance nor luminance, but "integrated
+/// radiance" or "tristimulus weight".
+/// 
+/// The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is
+/// an Illuminant D spectral distribution with chromaticity matching the
+/// rendering color space white point, normalized such that a ray normally
+/// incident upon the sensor with EV0 exposure settings will generate a pixel
+/// value of [1, 1, 1] in the rendering color space.
+///
```
meshula, line 64: Is this possible with radiometric units? I'd prefer that we move away from standard observers and photometry where possible, and more directly to the realm of radiometry so that we can just deal with measurements of energy, not measurements mapped to human vision.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -46,6 +46,41 @@ class SdfAssetPath;
...
+/// 
+/// The emission of a default light with `intensity` 1 and `color` [1, 1, 1] is
+/// an Illuminant D spectral distribution with chromaticity matching the
+/// rendering color space white point, normalized such that a ray normally
+/// incident upon the sensor with EV0 exposure settings will generate a pixel
+/// value of [1, 1, 1] in the rendering color space.
+/// 
+/// Given the above definition, that means that the luminance of said default
+/// light will be 1 *nit (cd∕m²)* and its emission spectral radiance
+/// distribution is easily computed by appropriate normalization.
```
meshula, line 67: Watts and Joules >>> lumens, nits and candelas

pxr/usd/usdLux/lightAPI.h
meshula: I like this latest version the best. I just added new notes about photometry above, and the new formulation works with radiometry.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -289,9 +339,21 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// <center><b>
+    /// L<sub>Scalar</sub> = L<sub>Scalar</sub> ⋅ 2<sup>exposure</sup>
+    /// </b></center>
+    /// 
+    /// Normatively, the lights' emission is in units of spectral radiance
+    /// normalized such that a directly visible light with `intensity` 1 and
+    /// `exposure` 0 normally incident upon the sensor plane will generate a
+    /// pixel value of [1, 1, 1] in an RGB renderer, and thus have a luminance
+    /// of 1 nit (cd∕m²). A light with `intensity` 1 and `exposure` 2 would
+    /// therefore have a luminance of 4 nits.
```
meshula, line 356: Ditto, can we have a think about radiometry here. Like to get rid of the standard observer concept.

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +421,119 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
-    /// of the light, by causing the power to not vary with the area or
-    /// angular size of the light.
+    /// Normalizes the emission such that the power of the light
+    /// remains constant while altering the size of the light, by dividing the
+    /// luminance by the world-space surface area of the light.
+    /// 
+    /// This makes it easier to independently adjust the brightness and size
+    /// of the light, by causing the total illumination provided by a light to
+    /// not vary with the area or angular size of the light.
+    ///
```
meshula, line 432: this is easy with radiometric units :]

pxr/usd/usdLux/lightAPI.h
```diff
@@ -384,7 +555,20 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// This color is just multiplied with the emission:
+    /// 
+    /// <center><b>
+    /// L<sub>Color</sub> = L<sub>Scalar</sub> ⋅ color
+    /// </b></center>
+    /// 
+    /// In the case of a spectral renderer, this color should be uplifted such
+    /// that it round-trips to within the limit of numerical accuracy under the
+    /// rendering illuminant.  We recommend the use of a rendering color space
+    /// well defined in terms of a Illuminant D illuminant, to avoid unspecified
```
meshula, line 570: do we wanna go further and just say D50 and D65 to be in line with common practice? The canonicals include D55 and D75 as well, but I'm not aware that they're used anywhere that might also use UsdLux. The canonical illumininants are magical because the use the newer value of Planck's constant. If we just say a "D" illuminant, it makes me think we might recommend 1931 D illuminants, in addition to the four "magic" illuminants.

pxr/usd/usdLux/overview.dox
meshula: UsdStage::SetColorConfiguration doesn't apply to the rendering color space at all, it refers to the viewing transform that is applied to the final framebuffer. 

So we're not creating two classes of color attributes at all.

As to the default color space to be used for scene referred colors, that's a topic for a new color space schema, not yet published. In the mean time, the default is what it is today, which is linear with the primaries and white point of rec bt.709. When the new schema is published, the default, in absence of any other indication, will still be that. So there won't be an issue with backwards compatibility.

pxr/usd/usdLux/shapingAPI.h
```diff
@@ -249,6 +322,37 @@ class UsdLuxShapingAPI : public UsdAPISchemaBase
...
+    /// 
+    /// https://store.ies.org/product/lm-63-19-approved-method-ies-standard-file-format-for-the-electronic-transfer-of-photometric-data-and-related-information/
+    /// 
+    /// The luminous intensity values in the ies profile are sampled using
+    /// the emission direction in the light's local space (after a possible
+    /// transformtion by a non-zero shaping:ies:angleScale, see below). The
+    /// sampled value is then potentially normalized by the overal power of the
+    /// profile if shaping:ies:normalize is enabled, and then used as a scaling
+    /// factor on the returned luminance:
+    ///
```
meshula, line 337: We should probably also mention that IES profile's photometry may or may not be in relative to a benchmark "bare bulb" (that they had on hand! and that we cannot characterize!), and so we expect that IES profiles should be considered to require scaling of some sort in any case, whether or not we are also imposing normalization.

pxr/usd/usdLux/lightAPI.h
anderslanglands: Yeah that seems reasonable

pxr/usd/usdLux/lightAPI.h
anderslanglands: It's just as easy with photometric units :)

pxr/usd/usdLux/lightAPI.h
anderslanglands: It's much easier to talk about expected values in an RGB renderer using photometric units, which is of course 99% of renderers out there.

pxr/usd/usdLux/lightAPI.h
anderslanglands: In this case, not really. If you want to talk about the radiance here, what wavelength should we talk about it at? How do we tie that to an RGB pixel colour? Or should we talk about the integrated radiance over some range of wavelengths?

pxr/usd/usdLux/shapingAPI.h
anderslanglands: We're actually going to need to revisit this. We're getting into the weeds with this internally here and the correct handling for converting from intensities in the IES to exitant luminance from an area source is more involved than what's being done in any renderer currently.

pxr/usd/usdLux/lightAPI.h
meshula: Yes, here, and with all the radiometric comments, I was thinking about integrated radiance over some range. I will follow up.

pxr/usd/usdLux/shapingAPI.h
meshula: When I wrote that, I was trying to correlate the IES specification to what we're saying here. I had been thinking of the intensity pattern as if it is a modulating the emittance from a small spherical radiator. Since the pattern is measured from a goniophotometer, the measurement itself pretends that the luminaire is a sphere; so basically a sphere large enough to have a surface area that can be measured, but small enough that it can treated like a point light in terms of how the light looks like from two different viewpoints, ie, not considering a difference in irradiance across the span of the luminaire... When you say "from an area source", that's very interesting. Whenever I look at Annex D, I always wonder if the height, width, and length they talk about are angular or linear. Since no units are mentioned (that I can find), I've always assumed angular, and assumed that there's no way to guess how big a luminaire is from an IES file. I'm very curious to learn how you're thinking about this, and if there is a clear explanation in LM-63-19 and I just haven't spotted it.

anderslanglands: Yeah so basically units in the IES are candela by definition. To compute the correct exitant luminance one needs to divide that by area of the source along the exitant direction. For a sphere this is always $\pi r^2$ since the sphere projects to a disk along any direction. For a disk light this would be the area of the disk, multiplied by the cosine of the angle between the exitant direction and the disk surface normal. For rectangles and cylinders it gets more complicated. 

I’m currently pondering what the best approach is here. IES files are only really valid for point lights by their nature. For spheres and disks we can do something reasonable and simple as noted above. For other shapes do we try and be exact (cylinders seem like they’d be particularly hard/expensive to handle)? Or do we just make some simple approximation and note that the output is not expected to be correct for those shapes. The latter seems reasonable to me but haven’t thought it through fully yet.

anderslanglands: Would be keen to know your thoughts

meshula: Ah, I see where you are going with that, I think. I wonder if it might be easiest to think about by taking a step beyond a sphere, by considering one face of a 10cm x 10cm x 10cm cube, with a point emitter at 0,0,0. A disc projected from the emitter to the face would have an area of 100 cm^2. We could then just treat the face as a disc light using `L = I / area`.  To finish that thought, we could treat the whole ideal cube luminaire as six of those.

That wouldn't be a bad approximation, I bet.

If we thought about an icosahedron, with 20 disc lights, the approximation is better.

Following that train of thought, does this reduce to a differential form, that we could plug the small set of regular IES shapes into? I think there's sphere, ellipsoid, box, cylinder, and not much else?

For a cylinder it might be $dA = _frac{ dI } { r d\theta dz cos(\theta ') $ on the walls, and a plain old disk for the top and bottom.

pxr/usd/usdLux/lightAPI.h
meshula: Apologies, I was being bone headed here. You specifically said the luminaire has an Illuminant D spectral distribution with chromaticity matching the rendering color space white point. The observer is the sensor response function. I have no idea what I was on about, this is already the way I was hoping it would be.

pxr/usd/usdLux/lightAPI.h
meshula: I think the answer to my question is "you are doubilng down on the boneheaded misinterpretation in your previous question. please recall, where we specify a D series illuminant, and the sensor response function from which radiometric values may be recovered"

pxr/usd/usdLux/lightAPI.h
meshula: same answer, D series illuminant plus sensor exposure function, yes?

pxr/usd/usdLux/shapingAPI.h
anderslanglands: To be clear I'm talking about the shapes of the usdlux lights, so sphere, disk, rect, cylinder

pxr/usd/usdLux/lightAPI.h
anderslanglands: :D

jesschimein:
Internal issue number for reference: USD-9900

pxr/usd/usdLux/schema.usda
```diff
@@ -187,18 +249,140 @@ class "LightAPI" (
...
+        * <i>if 0 < 𝛳<sub>max</sub> ≤ 𝜋 / 2:</i>
+        <center><b>
+            sizeFactor<sub>distant</sub> = sin²𝛳<sub>max</sub> ⋅ 𝜋
+        </b></center>
+
+        * <i>if 𝜋 / 2 < 𝛳<sub>max</sub> ≤ 𝜋:</i>
+        <center><b>
+                    sizeFactor<sub>distant</sub> =
+                        (2 - sin²𝛳<sub>max</sub>) ⋅ 𝜋
+        </b></center>
```
curtisblack, line 323: Curious if you have a link to the derivation of this?

If I compute a normalization factor as a fraction of solid angle coverage I get a different answer: https://www.desmos.com/calculator/coa7luomx3

anderslanglands: It's the integral over _projected_ solid angle. See for example page 11 here: https://www.cs.cornell.edu/courses/cs5625/2022sp/notes/radiometry-notes.pdf

One way of looking at this is it expresses the light in units of irradiance, which is useful. A simpler way to look at it is "the light doesn't get dimmer when you make the angle bigger".

pxr/usd/usdLux/lightAPI.h
```diff
@@ -359,10 +421,119 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// Then we use the following formula:
+    /// 
+    /// * <i>if 𝛳<sub>max</sub> = 0:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> = 1
+    /// </b></center>
+    /// 
+    /// * <i>if 0 < 𝛳<sub>max</sub> ≤ 𝜋 / 2:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> = sin²𝛳<sub>max</sub> ⋅ 𝜋
```
tcauchois, line 488: Apologies for the dumb questions, but how does this formulation relate to e.g. if you pretend a distant light is a cone section of a sphere?  If you normalize by computing the solid angle of the light vs the solid angle of a light with angle 0.53, the solid angle is 2 * pi * (1 - cos(theta)), or equivalently, 4 * pi * sqr(sin(0.5 * theta)).
https://en.wikipedia.org/wiki/Solid_angle#Cone,_spherical_cap,_hemisphere

... the math looks similar but I'm having trouble drawing the correspondences.  Can you explain what's going on here geometrically?

pxr/usd/usdLux/shapingAPI.h
```diff
@@ -201,8 +229,29 @@ class UsdLuxShapingAPI : public UsdAPISchemaBase
     // --------------------------------------------------------------------- //
     // SHAPING:CONE:ANGLE 
     // --------------------------------------------------------------------- //
-    /// Angular limit off the primary axis to restrict the
-    /// light spread.
+    /// Angular limit off the primary axis to restrict the light
+    /// spread, in degrees.
```
tcauchois, line 233: It looks like this has been rewritten since I last took a look at the docs/test images.  Do you know where this stands w.r.t. Renderman conformance at the moment?

pxr/usd/usdLux/lightAPI.h
anderslanglands: See [the link I posted in response to Curtis, above (page 11)](https://www.cs.cornell.edu/courses/cs5625/2022sp/notes/radiometry-notes.pdf). This is defining it in terms of *projected* solid angle instead of solid angle. This has two nice properties:
1. The light doesn't change brightness when you make the angle bigger (which is the whole point of normalize), see [renders 21-30 here](https://molodowitch.com/luxtest-diffs/2024-07-31.4k-sample-embree.luxtest-1a88c027.usd-6187b7a3.renders-e6fc4200/luxtest.html#distant) comparing first and second columns. First is this normalization, second is Karma, which does solid angle normalization.
2. This makes the light's brightness be defined in units of illuminance, which means we now have a way to ascribe a SI unit to the light.

pxr/usd/usdLux/schema.usda
curtisblack: Thanks for the explanation @anderslanglands, maybe this should be in the documentation for future reference?

Just a few follow up questions if I may. This normalization is defined for surfaces, does this mean a volume lit with a normalized distant light will not stay the same brightness for different sizes? Should there be a different normalization technique for volumes?

pmolodo: Heh, at one point I wrote this explanation up, before axing it because it was too wordy:

https://github.com/pmolodo/OpenUSD/blob/pr/usd-lux-update-doc-normalizingDistantLights.pdf/pxr/usd/usdLux/doxygen/normalizingDistantLights.pdf

(There's also a [.md version](https://github.com/pmolodo/OpenUSD/blob/pr/usd-lux-update-doc-normalizingDistantLights.pdf/pxr/usd/usdLux/doxygen/normalizingDistantLights.md), but it's formatted for Doxygen's markdown flavor, not github's, so not all formating works 100%.)

pxr/usd/usdLux/lightAPI.h
pmolodo: For a geometrical derivation of this formula, see "Property 2" in this explanation I wrote up at one point (because I had the same question):

https://github.com/pmolodo/OpenUSD/blob/pr/usd-lux-update-doc-normalizingDistantLights.pdf/pxr/usd/usdLux/doxygen/normalizingDistantLights.pdf

(There's also a [.md version](https://github.com/pmolodo/OpenUSD/blob/pr/usd-lux-update-doc-normalizingDistantLights.pdf/pxr/usd/usdLux/doxygen/normalizingDistantLights.md), but it's formatted for Doxygen's markdown flavor, not github's, so not all formating works 100%.)

pxr/usd/usdLux/shapingAPI.h
pmolodo: It's the same for coneAngle < 90 degrees.  Renderman seems to cap the value at 90 degrees - if coneAngle >= 90 degrees, it behaves as though comeAngle = 180 / coneAngle is disabled.

pmolodo:
I've updated the description of `inputs:ies:angleScale` to match the "bimodal" formula I proposed here:

https://github.com/pmolodo/OpenUSD/blob/ies-angleScale-findings-questions/pxr/usd/usdLux/iesAngleScale/UsdLux-IESAngleScale-FindingsAndQuestions.md#a-bimodal---scaling-origin-at-0-for-anglescale--0-180-for-anglescale--0

pxr/usd/usdLux/lightAPI.h
```diff
@@ -360,10 +422,119 @@ class UsdLuxLightAPI : public UsdAPISchemaBase
...
+    /// </b></center>
+    /// 
+    /// * <i>if 𝜋 / 2 < 𝛳<sub>max</sub> ≤ 𝜋:</i>
+    /// <center><b>
+    /// sizeFactor<sub>distant</sub> =
+    /// (2 - sin²𝛳<sub>max</sub>) ⋅ 𝜋
```
tcauchois, line 494: So I've convinced myself that normalizing on irradiance makes more sense than normalizing on power, but the Renderman flag specifically normalizes on angle; e.g. it scales intensity by (sizefactor(theta) / sizefactor(0.53)).  This lets you configure intensity (to get the correct amount of illumination) and angle (to get the right surface response) separately.

With this formulation, of not scaling by the ratio, the value of "intensity" takes on a radically different meaning between normalize being off and on.  A distant light of intensity 1 lumen, angle 0.53, will provide no illumination if normalize is off but provide twilight-levels of illumination with normalize on, in HoudiniGL/Karma.  This seems less correct if our intent is to keep numbers inspired by radiometry?

Can you explain the theory of dividing by (2 - sin^2 (theta) * pi), rather than (2 - sin^2(theta) * pi)/(2 - sin^2(0.53)*pi)?

tcauchois: (Note that the scaling behavior as you change intensity/angle is the same between the PR version, and the ratio version; the only difference is that turning on the normalize checkbox effectively multiplies intensity by the constant (2-sin^2(0.53)*pi) in the PR version).

anderslanglands: Well if you're convinced that illuminance/irradiance is the right unit then we're good :) the sin^2(theta) term accounts for projected solid angle, so puts the brightness of the light in units of illuminance/irradiance. In other words, with this factor, when normalize is on, then "intensity" is actually specifying illuminance in lux.

This is subtly different from what Karma and most everyone else does in that it keeps brightness constant as the angle of the light is changed, while Karma et al will get slightly darker the larger the angle because they're normalizing by solid angle rather than projected solid angle.

AFAICT The RenderMan method is basically the same as Karma et al, except it's normalized differently. Similarly your proposed factor in the end of your first comment would be in units of lux scaled by some arbitrary factor.

tcauchois: Renderman scales intensity by (sizeFactor(0.53) / sizeFactor(theta)), rather than just scaling it by 1/sizeFactor(theta).  This means that for a light with a default angle of 0.53, clicking "Normalize" doesn't change the intensity at all, which is a nice property.  With the PR proposal, since you don't normalize to 0.53 I think you end up brightening the intensity by a constant amount every time you click normalize (except maybe for angle = 0).

tcauchois: I think Paul's arguments for both the irradiance normalization and the analogy to distant sphere lights both work whether you normalize to 0 or 0.53, fwiw, since the distant sphere light argument involves a magic constant and mostly concerns itself with the theta-dependent scaling term.

pmolodo: > With this formulation, of not scaling by the ratio, the value of "intensity" takes on a radically different meaning between normalize being off and on. A distant light of intensity 1 lumen, angle 0.53, will provide no illumination if normalize is off but provide twilight-levels of illumination with normalize on, in HoudiniGL/Karma.

So - I'm not sure I would say "radically different" - as you noted, the only difference is a constant factor.

However - you're right that, without the $sin^2(\dfrac{0.53\degree}{2})*\pi$ term, the render of a default .53 distant light WILL change when you toggle normalize on / off.

> With the PR proposal, since you don't normalize to 0.53 I think you end up brightening the intensity by a constant amount every time you click normalize (except maybe for angle = 0).

With EITHER formulation, how a given light's render will change, when normalization is toggled on and off, varies depending on the `distantLightAngle`.  It's just that the "normalization reference point" - ie, the `distantLightAngle` at which there is NO change when normalize is toggled - differs.  With the renderman formulation (with the $sin^2(\dfrac{0.53\degree}{2})*\pi$ term), the "normalization reference point" is .53° - ie, it equals the default value for `distantLightAngle`, which **IS** a handy property.

Without the $sin^2(\dfrac{0.53\degree}{2})*\pi$, the normalization reference point is ~68.7° degrees, ie:

$$
2 * degrees(arcsin(\sqrt{1 / \pi}))
$$

This is a much less useful "normalization reference point" - it is not only different than the default value of .53°, but also much larger than the typical range for which a distant light is typically used.  So, in a practical sense, you can say that without your extra constant, the typical distant lights will "change a lot" when you toggle normalize, while with the $sin^2(\dfrac{0.53\degree}{2})*\pi$ term, typical distant lights will "change only a little" when you toggle normalize.

> I think Paul's arguments for both the irradiance normalization and the analogy to distant sphere lights both work whether you normalize to 0 or 0.53, fwiw, since the distant sphere light argument involves a magic constant and mostly concerns itself with the theta-dependent scaling term.

Definitely true for **Property 2: Proportional to surface area of a distant sphere** - as you note, there's already a magic multiplier constant, so multiplying by any other constant won't change the proportionality!

However, there is no magic constant for **Property 1: Intensity = Illuminance** - so that will no longer hold if the extra 0.53 balancing factor is added.  That is, without it you can say:

- With `normalize` OFF, the `intensity` value represents **luminance** in nits
- With `normalize` ON, the `intensity` represents **illuminance** in lux, if surface area is in meter^2

...whereas, if you use the .53 balancing factor, then it "represents *illuminance* times the .53 balance factor".

So - which you prefer is a question of whether:

- you prioritize having the "normalization reference point" == the default `distantLightAngle` (so, for a default distant light, toggling normalize on and off won't change it)
- you want to have `intensity` represent a direct (ie, not corrected by an arbitrary constant) physical measurement of **illuminance** when normalize is on

anderslanglands: Great summary, thanks @pmolodo 

So yeah, it comes down to do you want the light to not change brightness when toggling normalize on and all other values at default (note this is not currently true for all other lights), or do you want to have normalize on mean the light is in a well-defined unit?

I'd _much_ prefer the latter, but given that it's a constant scale factor it's not a hill I want to die on.

PR #3186, opened 2024-07-31
[hdEmbree] Initial UsdLux reference implementation (hdEmbree-UsdLux-PR06)

pmolodo:
### Description of Change(s)

This modifies the hdEmbree example plugin to do direct lighting so as to provide a reference implementation of the expected behaviour for UsdLux.

If no lights are present in the stage, the old ambient occlusion path will be used.

---------------------------------------------------------------

### Related PRs:

This PR is part of a chain of PRs that provide a reference implementation of UsdLux as part of hdEmbree.

| :arrow_backward: **Previous PR in chain:**  |  :arrow_down_small: **This PR changes only:** | :arrow_forward: **Next PR in chain:** |
|-----------------------------|------------------------------|------------------------|
|  https://github.com/PixarAnimationStudios/OpenUSD/pull/3198| [Diff vs Previous](https://github.com/NVIDIA-Omniverse/OpenUSD/pull/4) <p> Due to a github limitation, this PR must include all previous PRs in the chain.  To see JUST the changes in this PR vs the previous in the chain, click this link  | https://github.com/PixarAnimationStudios/OpenUSD/pull/3196|

:chains: **All PRs in this chain:**
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3199

:page_with_curl: **Documentation PR:**
This chain of PRs (implementing UsdLux support in hdEmbree), was made separate from the PR documenting expected UsdLux behavior:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3182

:eight_spoked_asterisk: **All UsdLux update related PRs:**
To see ALL UsdLux update related PRs (documentation AND reference implementation) in one place, see:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/3200

:hourglass: **Original PR:**
These PRs are an update of this original PR:
- https://github.com/PixarAnimationStudios/OpenUSD/pull/2758

---------------------------------------------------------------

### But... why??

**Why make these PRs in the first place?**

The current specifications of the various UsdLux prims + attributes are imprecise or vague in many places, and as a result, actual implementations of them by various renderers have diverged, sometimes quite significantly.  For instance, here is Intel's [4004 Moore Lane](https://dpel.aswf.io/4004-moore-lane/) scene, with the same UsdLux lights defined, in 3 different renderers:

Karma:
4004 Moore Lane, rendered in Karma

Arnold:
4004 Moore Lane, rendered in Arnold

Omniverse RTX:
4004 Moore Lane, rendered in Omniverse RTX

For a full descpription of the problem, see here:
- [Problem Statement](https://github.com/anderslanglands/light_comparison/blob/main/README.md#problem-statement)

**Why so many PRs?**
This was my attempt to break up a rather large change into smaller, more easily reviewable changes, that can be merged in incrementally, to help ease the burden for code reviewers.

If you find this confusing, and would rather just one big PR (ie, just this one), or have them organized in some other way, please let me know!

**Why are the documentation changes in their own PR?**
In some ways, the [documentation changes](https://github.com/PixarAnimationStudios/OpenUSD/pull/3182) are the heart of this effort - we wish to specify more exactly what the various UsdLux prims and attributes represent.  However, building consensus on this may take time - so we expect some dialogue on the exact language or formulas.

Because that may take time, the [hdEmbree reference implementation changes](https://github.com/PixarAnimationStudios/OpenUSD/pull/3199) are separate, and broken up, so we can hopefully start integrating portions of them even before final consensus has been reached on the final form of the specification.

**Why are the documentation changes not broken up into smaller pieces, like the hdEmbree reference implementation changes?**
Because I wasn't sure if that would be desirable or not!  If people think that would be helpful, I can do so - perhaps breaking out by schema or individual attribute?

---------------------------------------------------------------

### DIFF
--- a/pxr/imaging/plugin/hdEmbree/light.cpp
+++ b/pxr/imaging/plugin/hdEmbree/light.cpp
@@ -1,0 +1,171 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the terms set forth in the LICENSE.txt file available at
+// https://openusd.org/license.
+//
+#include "pxr/imaging/plugin/hdEmbree/light.h"
+
+#include "light.h"
+#include "pxr/imaging/plugin/hdEmbree/debugCodes.h"
+#include "pxr/imaging/plugin/hdEmbree/renderParam.h"
+#include "pxr/imaging/plugin/hdEmbree/renderer.h"
+
+#include "pxr/imaging/hd/sceneDelegate.h"
+#include "pxr/imaging/hio/image.h"
+
+#include <embree3/rtcore_buffer.h>
+#include <embree3/rtcore_scene.h>
+
+#include <fstream>
+#include <sstream>
+#include <vector>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+HdEmbree_Light::HdEmbree_Light(SdfPath const& id, TfToken const& lightType)
+    : HdLight(id) {
+    if (id.IsEmpty()) {
+        return;
+    }
+
+    // Set the variant to the right type - Sync will fill rest of data
+    if (lightType == HdSprimTypeTokens->cylinderLight) {
+        _lightData.lightVariant = HdEmbree_Cylinder();
+    } else if (lightType == HdSprimTypeTokens->diskLight) {
+        _lightData.lightVariant = HdEmbree_Disk();
+    } else if (lightType == HdSprimTypeTokens->rectLight) {
+        // Get shape parameters
+        _lightData.lightVariant = HdEmbree_Rect();
+    } else if (lightType == HdSprimTypeTokens->sphereLight) {
+        _lightData.lightVariant = HdEmbree_Sphere();
+    } else {
+        TF_WARN("HdEmbree - Unrecognized light type: %s", lightType.GetText());
+        _lightData.lightVariant = HdEmbree_UnknownLight();
+    }
+}
+
+HdEmbree_Light::~HdEmbree_Light() = default;
+
+void
+HdEmbree_Light::Sync(HdSceneDelegate *sceneDelegate,
+                         HdRenderParam *renderParam, HdDirtyBits *dirtyBits)
+{
+    HD_TRACE_FUNCTION();
+    HF_MALLOC_TAG_FUNCTION();
+
+    HdEmbreeRenderParam *embreeRenderParam =
+        static_cast<HdEmbreeRenderParam*>(renderParam);
+
+    // calling this bumps the scene version and causes a re-render
+    embreeRenderParam->AcquireSceneForEdit();
+
+    SdfPath const& id = GetId();
+
+    // Get _lightData's transform. We'll only consider the first time sample for now
+    HdTimeSampleArray<GfMatrix4d, 1> xformSamples;
+    sceneDelegate->SampleTransform(id, &xformSamples);
+    _lightData.xformLightToWorld = GfMatrix4f(xformSamples.values[0]);
+    _lightData.xformWorldToLight = _lightData.xformLightToWorld.GetInverse();
+    _lightData.normalXformLightToWorld =
+        _lightData.xformWorldToLight.ExtractRotationMatrix().GetTranspose();
+
+    // Store luminance parameters
+    _lightData.intensity = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->intensity).GetWithDefault(1.0f);
+    _lightData.exposure = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->exposure).GetWithDefault(0.0f);
+    _lightData.color = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->color).GetWithDefault(GfVec3f{1.0f, 1.0f, 1.0f});
+    _lightData.normalize = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->normalize).GetWithDefault(false);
+    _lightData.colorTemperature = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->colorTemperature).GetWithDefault(6500.0f);
+    _lightData.enableColorTemperature = sceneDelegate->GetLightParamValue(
+        id, HdLightTokens->enableColorTemperature).GetWithDefault(false);
+
+    // Get visibility
+    _lightData.visible = sceneDelegate->GetVisible(id);
+
+    // Switch on the _lightData type and pull the relevant attributes from the scene
+    // delegate
+    std::visit([this, &id, &sceneDelegate](auto& typedLight) {
+        using T = std::decay_t<decltype(typedLight)>;
+        if constexpr (std::is_same_v<T, HdEmbree_UnknownLight>) {
+            // Do nothing
+        } else if constexpr (std::is_same_v<T, HdEmbree_Cylinder>) {
+            typedLight = HdEmbree_Cylinder{
+                sceneDelegate->GetLightParamValue(id, HdLightTokens->radius)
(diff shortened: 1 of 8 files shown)


jesschimein:
Filed as internal issue #USD-9904

pxr/imaging/plugin/hdEmbree/light.cpp
```diff
@@ -0,0 +1,171 @@
...
+                sceneDelegate->GetLightParamValue(id, HdLightTokens->height)
+                    .Get<float>(),
+            };
+        } else if constexpr (std::is_same_v<T, HdEmbree_Sphere>) {
+            typedLight = HdEmbree_Sphere{
+                sceneDelegate->GetLightParamValue(id, HdLightTokens->radius)
+                    .GetWithDefault(0.5f),
+            };
+        } else {
+            static_assert(false, "non-exhaustive _LightVariant visitor");
```
tcauchois, line 121: At least in my compile, this static assert always fires; I replaced this locally with a typedLight = HdEmbree_UnknownLight{} (& combined with the first case).  It might be possible to do something better?  But this might give us less build trouble.

pxr/imaging/plugin/hdEmbree/light.h
```diff
@@ -0,0 +1,182 @@
...
+/// - No support for motion blur (currently, if motion blur is enabled, all
+///   samples taken at the first time sample, ie, when the shutter opens).
+/// - No support for instanced lights
+
+#include "pxr/base/gf/vec3f.h"
+#include "pxr/base/gf/matrix3f.h"
+#include "pxr/base/gf/matrix4f.h"
+#include "pxr/imaging/hd/light.h"
+
+#include <embree3/rtcore_common.h>
```
tcauchois, line 88: Pulled in embree 4 already and fixed up the places where I needed to (it was pretty easy).

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -58,11 +69,118 @@ class _ScopedThreadScheduler {
 
 #endif  // TBB_INTERFACE_VERSION_MAJOR < 12
 
-
 namespace {
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+// -------------------------------------------------------------------------
```
tcauchois, line 76: I feel like it might be a kindness to future us if I pull all of the math and sampling code into a separate file?  lightSamplingUtils.h?  Any objections?

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -883,60 +1355,86 @@ HdEmbreeRenderer::_ComputePrimvar(RTCRayHit const& rayHit,
...
+    // occluded sets tfar < 0 if the ray hit anything
+    return shadow.tfar > 0.0f;
+}
+
 GfVec4f
 HdEmbreeRenderer::_ComputeColor(RTCRayHit const& rayHit,
                                 std::default_random_engine &random,
                                 GfVec4f const& clearColor)
 {
-    if (rayHit.hit.geomID == RTC_INVALID_GEOMETRY_ID) {
```
tcauchois, line 919: Don't understand how you could delete this; the idea here is that if your ray misses, you return the clear color, which I'd expect to happen in many scenes (if the geometry doesn't fill the viewport).  If rayHit.hit.geomID is invalid, the unguarded accesses to user data pointers there are a quick trip to sigsegv, and I didn't see any compensating code in TraceRay to not call _ComputeColor when the ray misses?  Reinstating this block locally, but if there's something I'm missing or a bit of code that should have been here let me know.

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -960,38 +1458,63 @@ HdEmbreeRenderer::_ComputeColor(RTCRayHit const& rayHit,
...
-    // Lighting model: (camera dot normal), i.e. diffuse-only point light
-    // centered on the camera.
-    GfVec3f dir = GfVec3f(rayHit.ray.dir_x, rayHit.ray.dir_y, rayHit.ray.dir_z);
-    float diffuseLight = fabs(GfDot(-dir, normal)) *
-        HdEmbreeConfig::GetInstance().cameraLightIntensity;
+    GfVec3f lightingColor(0.0f);
+
+    // If there are no lights, then keep the existing camera light + AO path to
+    // be able to inspect the scene
+    if (_lightMap.empty())
```
tcauchois, line 1465: The main change I've made locally is to add a render setting "enableLighting" and a corresponding env setting HDEMBREE_USE_LIGHTING to turn on the light computation, as opposed to AO.  Happy to rename those if you'd like.  I also set that setting to false by default, though it should be easy to enable in usdview and then it'll get stored in user state forever.  If "enableLighting" is on, it overrides AO settings; if it's off, we fall back to AO.

Having it as an explicit mode seems preferable, and also makes it easier to merge what with the unit tests/etc.  But let me know your thoughts on this.

tcauchois (commented):
Lots of really great improvements here! :).  Had to make a few changes locally and it's running through testing for (hopefully) the final time!  Let me know your take on my comments so we can land this.

pxr/imaging/plugin/hdEmbree/renderer.cpp
anderslanglands: Hi Tom, we're not sure about this one either. If the renders looks the same after reinstating this, then that seems good.

pxr/imaging/plugin/hdEmbree/renderer.cpp
anderslanglands: Seems good!

pxr/imaging/plugin/hdEmbree/renderer.cpp
anderslanglands: No objections here, seems like a good idea

pxr/imaging/plugin/hdEmbree/light.h
anderslanglands: Great

pxr/imaging/plugin/hdEmbree/light.cpp
pmolodo: Sorry - this was a msvc / gcc difference (and I was building / testing on windows, because I couldn't get RenderMan and Arnold installed on Ubuntu...).

We could replace with this, which gives us a valid check on both linux and windows:

```
        } else {
            // We should never get to this branch, as all possible variants
            // should be handled above. If we do, sizeof(T) will always be 0;
            // having, having the expression rely on T ensures the assert is
            // not compiled unless such a T exists, on both msvc and gcc.
            static_assert(sizeof(T) == 0, "non-exhaustive _LightVariant visitor");
        }
```

pxr/imaging/plugin/hdEmbree/renderer.cpp
pmolodo: Apologies - I think this was a casualty of the slicing / dicing I did to split this into multiple PRs.  It gets added back in when we add dome light support, here:

https://github.com/PixarAnimationStudios/OpenUSD/pull/3188/commits/ecdcbbc95aa1a4e1eede6b483993fb4e559bd70e#diff-3af68bc81d28e73ceb1b6f3584442a93021c53709fbf95cb796e3ca42dd7f911R1488-R1491

...with an additional check to see if there are any dome lights.  However, you're totally right that it shouldn't have been removed in the first place.

pxr/imaging/plugin/hdEmbree/light.cpp
tcauchois: Elegant and tricky! :)

PR #3120, opened 2024-06-12
Autodesk: Fix MSVC build with pch

erikaharrison-adsk:
### Description of Change(s)

Fix Windows build failure when PCH is ON when using ninja.

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3408 (Updated 2024-11-22)

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #3141, opened 2024-06-25
cmake: ensure directory holding pch.pch file is created

pmolodo:
### Description of Change(s)

When specifying a precompiled header output via `cl /Fp${folder}/${filename}`, cl won't automatically create `${folder}`.  Ensure we do this, or `cl` will error.

This may be an issue only with VS2022 or Windows 11, as I've never encountered this before, and I those are the main new variables.  (Note that I was also build with Ninja, though I've been doing that for a while, and haven't changed the version in a while.)

### Versions

- USD: latest dev (5332d627cb28991e812acd67b8971e87367d9191)
- CMake:  3.27.9
- OS: Windows 11, 23H2, Build 22631.3737
- Visual Studio: 2022 LTSC 17.8.11+34931.61
- .NET Framework: 4.8.09032
- Visual C++: 2022   00476-80000-00000-AA016
- Ninja: 1.10.2

### Fixes Issue(s)
- Without this, I got the following error when compiling with precompiled headers enabled:

```
[479/4528] Building CXX object pxr\base\arch\CMakeFiles\arch.dir\pch.cpp.obj
FAILED: pxr/base/arch/CMakeFiles/arch.dir/pch.cpp.obj
[...]

PR #3204, opened 2024-08-03
Build: fix pre-compiled header files for MSVC/Ninja

mistafunk:
### Description of Change(s)
- Fix use of pre-compiled header when using MSVC with Ninja
- Fix potential compilation errors with Ninja caused by concurrent writes to PDB files when building with multiple cores.

### DIFF
--- a/cmake/defaults/msvcdefaults.cmake
+++ b/cmake/defaults/msvcdefaults.cmake
@@ -134,3 +134,9 @@ set(_PXR_CXX_FLAGS "${_PXR_CXX_FLAGS} /Gm-")
 # with no symbols in it.  We do this a lot because of a pattern of having
 # a C++ source file for many header-only facilities, e.g. tf/bitUtils.cpp.
 set(CMAKE_STATIC_LINKER_FLAGS "${CMAKE_STATIC_LINKER_FLAGS} /IGNORE:4221")
+
+# Enforce synchronous PDB writes when using Ninja
+# (this prevents "permission denied" compile errors on program databases)
+if("${CMAKE_GENERATOR}" STREQUAL "Ninja")
+    set(_PXR_CXX_FLAGS "${_PXR_CXX_FLAGS} /FS")
+endif()
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -523,7 +523,8 @@ function(_pxr_enable_precompiled_header TARGET_NAME)
     # Headers live in subdirectories.
     set(rel_output_header_path "${PXR_PREFIX}/${TARGET_NAME}/${output_header_name}")
     set(abs_output_header_path "${PROJECT_BINARY_DIR}/include/${rel_output_header_path}")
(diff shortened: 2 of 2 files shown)


cmake/defaults/Options.cmake
```diff
@@ -72,7 +72,7 @@ option(PXR_ENABLE_GL_SUPPORT "Enable OpenGL based components" ON)
 
 # Precompiled headers are a win on Windows, not on gcc.
```
meshula: can we also revise the comment to not say Windows, but rather "msvc with msbuild"

meshula:
thanks for noticing this

mistafunk:
> thanks for noticing this

You are welcome. I took another look and found the actual bugs. There was a missing directory and wrong cmake command order when PCH was enabled. I reused this PR and force pushed.

mistafunk:
A note for CLion users, do not get confused (like I did) by output like this in the CMake window:
```
Problems were encountered while collecting compiler information:
	C:\Users\<xxx>\AppData\Local\Temp\compiler-file7019753171835842146: fatal error C1083: Cannot open include file: 'pxr/usd/_ar/pypch.h': No such file or directory
```

This is an artifact of the CLion "intellisense" and not related to the USD build system...

cmake/defaults/Options.cmake
mistafunk: Maybe just remove this comment - or do the homework and measure this properly :-)

I did a quick stopwatch comarison... msvc/ninja with PCH took about 5min and without PCH about 6min... so we could also just say "... are a win with MSVC ..."

meshula:
Thanks, this looks ready for an internal review.

mistafunk:
Btw, I tried to run the tests, there are over 100 failing out of 600 - is this exepected on `dev`?

jesschimein:
Filed as internal issue #USD-9931

sunyab:
@mistafunk Our internal CI is currently showing no failures, so I would not expect anything like that. I also can't imagine anything specific to this change that would be causing that either.

Do you have an example of a failing test you could share here? I'd bet they're all failing for the same reason.

pmolodo:
I'll close my [PR3141](https://github.com/PixarAnimationStudios/OpenUSD/pull/3141), since it looks like this fixes that and does a bit more as well!

cmake/defaults/Options.cmake
pmolodo: On another note - I've been using sccache with windows msvc USD builds recently, and been pretty happy with it so far.  Haven't done any time comparisons, but rebuilds feel faster - particularly if you switch branches and come back, so a lot of stuff has updated timestamps but hasn't actually changed.

mistafunk:
I cannot reproduce the linux build failure locally... is the pipeline timeout too low?

DDoS:
There is also #3120 that fixes this same issue.

adskWangl:
> There is also #3120 that fixes this same issue.

Thank you and we'll close #3120 as that is for the same issue.

PR #3205, opened 2024-08-04
Improve pxr CMake Config support by using `CMakeFindDependencyMacro` CMake module instead of raw `find_package`

theblackunknown:
### Description of Change(s)

This PR improves the `pxr` CMake Config supports by using `find_dependency` from `CMakeFindDependencyMacro` CMake module instead of raw `find_package`.

Doing so handling of `REQUIRED` is not needing anymore as it will be transively forwarded from the `find_pckage(pxr ...)` calls:
- `find_package(pxr REQUIRED ...)` results in dependencies being required too
- `find_package(pxr ...)` results in dependencies being optional too

More information on [`CMakeFindDependencyMacro` CMake module](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) documentation page.

### DIFF
--- a/pxr/pxrConfig.cmake.in
+++ b/pxr/pxrConfig.cmake.in
@@ -16,6 +16,8 @@ set(PXR_MINOR_VERSION "@PXR_MINOR_VERSION@")
 set(PXR_PATCH_VERSION "@PXR_PATCH_VERSION@")
 set(PXR_VERSION "@PXR_VERSION@")
 
+include(CMakeFindDependencyMacro)
+
 # If Python support was enabled for this USD build, find the import
 # targets by invoking the appropriate FindPython module. Use the same
 # LIBRARY and INCLUDE_DIR settings from the original build if they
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-9932

theblackunknown:
It looks like the Linux build went through but the CI was cancelled, any idea what went wrong ?

sunyab:
@theblackunknown It looks like the build succeeded but there was some kind of hiccup on the Azure side of things. I requeued the job to double-check.

theblackunknown:
Hi @sunyab what are my chances to integrate my change for the next release ?

parkerlreed:
Maybe https://github.com/PixarAnimationStudios/OpenUSD/issues/3310 can be included in this?

Currently unable to build f3d as it depends on usd which then depends on materialx. Fails at OpenGL since nothing includes it in the cmake.

Manually adding `find_package(OpenGL REQUIRED)` below the pxr worked to get a build going in the mean time.

theblackunknown:
> Maybe #3310 can be included in this?
> 
> Currently unable to build f3d as it depends on usd which then depends on materialx. Fails at OpenGL since nothing includes it in the cmake.
> 
> Manually adding `find_package(OpenGL REQUIRED)` below the pxr worked to get a build going in the mean time.

Looking at your issue @parkerlreed, I think I know what is going on and this requires further improvements which are currently not in this PR. 
I would be happy to follow up on those but I would like to get this PR merged first.

PR #3206, opened 2024-08-05
usdUtils: add USDZ Package Encapsulation Validator

beersandrew:
### Description of Change(s)
- Adding all necessary validator files to support a validator in usdUtils
- Added Initial code for USDZ package encapsulation validation

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65424899


pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // Verify the validator exists
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->packageEncapsulationValidator);
+
+    TF_AXIOM(validator);
+
+    const std::string testAssetDirectory = ArchGetCwd() + "/ctest/testUsdUtilsPackageEncapsulation/";
+
+    // Load the pre-created usdz stage with fully qualified paths to both a reference and an asset
+    const UsdStageRefPtr& stage = UsdStage::Open(testAssetDirectory + "fail.usdz");
```
beersandrew: @tallytalwar I'm looking for feedback on this section. I'd like to do this in a way that does not depend on so many test files. I tried using [UsdUtilsCreateNewUsdzPackage](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/usdzPackage.cpp#L81) however when using that on my `usda` file with fully qualified paths, the function corrects the paths internally so I don't get the failure condition I want. 

My next idea was to manually set the paths in code:

1. Create `subLayer.usda` (which is empty)
2. Create `image.jpg` (or some other asset file that's mostly empty)
3. Create `root.usda` that references the fully qualified paths of `subLayer.usda` & `image.jpg`
4. "Manually" turn `root.usda` into `test.usdz` (not sure what this function is, but not the above function since that corrects my paths, `usdzip` seems to work for this so maybe I can do the equivalent of that)
5. Verify errors
6. Update reference and asset paths in root.usda to be relative paths
7. "Manually" turn `root.usda` into `test.usdz` again (I guess here I could run `UsdUtilsCreateNewUsdzPackage` however to me it would be more clear to create the usdz the same way for someone reading the function
8. Verify the errors are gone
9. Delete `subLayer.usda`, `image.jpg`, 'root.usda', and `test.usdz`

This path requires no tests assets which seems better to me but I wanted to get your thoughts before going forward incase you like the committed solution better or had some other idea

jesschimein:
Filed as internal issue #USD-9933

tallytalwar:
Will get to reviewing this soon @beersandrew. It will be good to also get @dgovil 's eyes on any usdz related validator, so including him. Thanks both.

pxr/usd/usd/plugInfo.json
```diff
@@ -148,7 +148,7 @@
                 "Validators": {
                     "CompositionErrorTest": {
                         "doc": "Validator aims at providing all composition errors, which were generated while composing the stage."
-                    },
```
tallytalwar: We have some code internally which checks if any generated code differs.. unfortunately this change will make that fail, can you revert this file?

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <iostream>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+static
+void
+TestUsdUsdzValidators()
+{
+    // This should be updated with every new validator added with
+    // UsdUsdzValidators keyword.
```
tallytalwar: I think you meant UsdUtilsValidators here in the comment?

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    UsdValidationRegistry& registry = UsdValidationRegistry::GetInstance();
+
+    // Verify the validator exists
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->packageEncapsulationValidator);
+
+    TF_AXIOM(validator);
+
+    const std::string testAssetDirectory = ArchGetCwd() + "/ctest/testUsdUtilsPackageEncapsulation/";
```
tallytalwar: So I don't think you need to do this. Refer this example: `testUsdTimeValueAuthoring`,
cmake install dir: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/CMakeLists.txt#L681-L684
test usage of the installed usda files:
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/testenv/testUsdTimeValueAuthoring.cpp#L81
And the files are directly used as in the test, that is because thats where ctest will place the test binary and the installed file. 

And doing this way of hard coding ctest path will make this fail internally, as we do not have cmake based build system internally.

tallytalwar: In this code you should be able to use "testUsdUtilsPackageEncapsulation/fail.usdz" directly with UsdStage::Open.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
tallytalwar: I am not sure I am groking fully the aim here? Lets talk on slack what you want to achieve here.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,94 @@
...
+    for (size_t i = 0; i < errors.size(); ++i)
+    {
+        TF_AXIOM(errors[i].GetType() == UsdValidationErrorType::Error);
+        TF_AXIOM(errors[i].GetSites().size() == 1);
+        TF_AXIOM(!errors[i].GetSites()[0].GetLayer().IsInvalid());
+        TF_AXIOM(errors[i].GetMessage() == expectedErrorMessages[i]);
+    }
+
+    // Load the pre-created usdz stage with relative paths to both a reference and an asset.
+    const UsdStageRefPtr &passStage = UsdStage::Open(testAssetDirectory + "pass.usdz");
```
tallytalwar: Similar to fail.usdz, you should be able to provide the path: "testUsdUtilsPackageEncapsulation/pass.usdz" here.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,94 @@
...
+    TF_AXIOM(errors.size() == 0);
+}
+
+int
+main()
+{
+    TestUsdUsdzValidators();
+    TestPackageEncapsulationValidator();
+
+    std::cout << "OK\n";
```
tallytalwar: I have been removing these cout statements internally.

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "pxr/usd/usd/validationError.h"
+#include "pxr/usd/usd/validationRegistry.h"
+#include "pxr/usd/usdUtils/validatorTokens.h"
+#include "pxr/usd/ar/packageUtils.h"
+#include "pxr/usd/usdUtils/dependencies.h"
+#include "pxr/usd/usdUtils/userProcessingFunc.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+    static
```
tallytalwar: fix indentation.

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,88 @@
...
+                                       &unresolvedPaths, nullptr);
+
+        const std::string& realPath = rootLayer->GetRealPath();
+        const std::string& packagePath = ArIsPackageRelativePath(rootLayer->GetIdentifier()) ?
+                                        ArSplitPackageRelativePathOuter(realPath).first :
+                                        realPath;
+
+        if (packagePath.length() > 0) {
+            for (const SdfLayerRefPtr& subLayer : layers) {
+                const std::string& realPath = subLayer->GetRealPath();
```
tallytalwar: I was comparing this implementation with the ComplianceChecker implementation, and I saw a note about skipping session layers, which won't have a realPath, seems thats missing here.
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/complianceChecker.py#L753-L755

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        UsdUtilsComputeAllDependencies(path, &layers, &assets,
+                                       &unresolvedPaths, nullptr);
+
+        const std::string& realPath = rootLayer->GetRealPath();
+        const std::string& packagePath = ArIsPackageRelativePath(rootLayer->GetIdentifier()) ?
+                                        ArSplitPackageRelativePathOuter(realPath).first :
+                                        realPath;
+
+        if (packagePath.length() > 0) {
```
tallytalwar: `if (!packagePath.empty())`

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,88 @@
...
+            for (const SdfLayerRefPtr& subLayer : layers) {
+                const std::string& realPath = subLayer->GetRealPath();
+                if (!TfStringStartsWith(realPath, packagePath)) {
+                    errors.emplace_back(
+                            UsdValidationErrorType::Error,
+                            UsdValidationErrorSites{
+                                    UsdValidationErrorSite(rootLayer,
+                                                           subLayer->GetDefaultPrimAsPath())
+                            },
+                            TfStringPrintf(("Found loaded layer '%s' that "
```
tallytalwar: I know the error message is coming from the current compliance check code. But why "loaded" layer? Maybe @dgovil or @spiffmon can clarify?

tallytalwar (commented):
Thanks @beersandrew. Left a few notes and feel free to ping on slack with the test question you had.

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,88 @@
...
+    const SdfLayerRefPtr& rootLayer = usdStage->GetRootLayer();
+    const bool isPackage = [](const SdfLayerHandle& layer){
+        return layer->GetFileFormat()->IsPackage() || ArIsPackageRelativePath(layer->GetIdentifier());
+    }(rootLayer);
+
+    if (!isPackage){
+        return errors;
+    }
+
+    std::vector<TfRefPtr<SdfLayer>> layers;
```
tallytalwar: You should be able to use SdfLayerRefPtrVector, defined here: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/sdf/declareHandles.h#L335

beersandrew:
Thanks for the notes @tallytalwar I think I've addressed all of them.

pxr/usd/usdUtils/validators.cpp
beersandrew: I'm second guessing my decision to remove 'loaded' in the error message, also interested about it's meaning in this context.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -47,24 +47,25 @@ TestPackageEncapsulationValidator()
 
     TF_AXIOM(validator);
 
-    const std::string testAssetDirectory = ArchGetCwd() + "/ctest/testUsdUtilsPackageEncapsulation/";
-
-    // Load the pre-created usdz stage with fully qualified paths to both a reference and an asset
-    const UsdStageRefPtr& stage = UsdStage::Open(testAssetDirectory + "fail.usdz");
+    // Load the pre-created usdz stage with paths to a layer and asset that are not included in the package, but exist
```
tallytalwar: 80 column width limit please.

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const std::string& packagePath = ArIsPackageRelativePath(rootLayer->GetIdentifier()) ?
+                                    ArSplitPackageRelativePathOuter(realPath).first :
+                                    realPath;
+
+    if (!packagePath.empty()) {
+        for (const SdfLayerRefPtr& subLayer : layers) {
+            const std::string& realPath = subLayer->GetRealPath();
+
+            // Make sure the realPath exists, as the validator should not consider in-memory layers
+            if (!realPath.empty() && !TfStringStartsWith(realPath, packagePath)) {
```
tallytalwar: nit:

I would have done:
```cpp
if (realPath.empty()) {
    continue;
}
if (!TfStringStartsWith(realPath, packagePath)) {
....
}
```

tallytalwar (approved):
LGTM @beersandrew Thanks

pxr/usd/usdUtils/validators.cpp
beersandrew: thanks! I agree that's clearer, updated

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,102 @@
...
+    UsdValidationErrorVector errors = validator->Validate(stage);
+
+    // Verify both the layer & asset errors are present
+    TF_AXIOM(errors.size() == 2);
+
+    const std::string& realRootPath = stage->GetRootLayer()->GetRealPath();
+    const std::string& rootDirectory = realRootPath
+            .substr(0, realRootPath.find_last_of("/") + 1);
+
+    const std::string& errorLayer = rootDirectory + "excludedDirectory"
```
tallytalwar: In the validator code you have the error string use identifier, which is correct. But here you are using realPath, which is platform specific. So for windows this will fail as the expected error will be having a "\\" and ValidationError string will be using the identifier path which always use the PathDelimiter ("/").

pxr/usd/usdUtils/CMakeLists.txt
```diff
@@ -290,6 +309,11 @@ pxr_install_test_dir(
     DEST testUsdUtilsUserProcessingFunc
 )
 
+pxr_install_test_dir(
+    SRC testenv/testUsdUtilsPackageEncapsulation
```
tallytalwar: nit: We generally keep the test asset directory same name as the test. So "testUsdUtilsValidators", instead of "testUtilsPackageEncapsulation".

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,102 @@
...
+    // Verify the validator exists
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->packageEncapsulationValidator);
+
+    TF_AXIOM(validator);
+
+    // Load the pre-created usdz stage with paths to a layer and asset
+    // that are not included in the package, but exist
+    const UsdStageRefPtr& stage = UsdStage::Open(
+            "testUsdUtilsPackageEncapsulation/fail.usdz");
```
tallytalwar: The assets are being "installed" in the test dir, so the file here shouldn't include the test dir. Just "fail.usdz" should pick the asset. This was the reason for test failures on all platforms internally.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
tallytalwar: And as mentioned ^, this does fail on windows in internal test runs.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,102 @@
...
+        TF_AXIOM(errors[i].GetType() == UsdValidationErrorType::Error);
+        TF_AXIOM(errors[i].GetSites().size() == 1);
+        TF_AXIOM(!errors[i].GetSites()[0].GetLayer().IsInvalid());
+        TF_AXIOM(errors[i].GetMessage() == expectedErrorMessages[i]);
+    }
+
+    // Load the pre-created usdz stage with relative paths to both a reference
+    // and an asset that are included in the package.
+    const UsdStageRefPtr& passStage = UsdStage::Open(
+            "testUsdUtilsPackageEncapsulation/pass.usdz");
```
tallytalwar: As mentioned above for fail.usda, this need to be updated to just "pass.usdz". 

> The assets are being "installed" in the test dir, so the file here shouldn't include the test dir. Just "fail.usdz" should pick the asset. This is the reason for test failures.

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            if (!TfStringStartsWith(realPath, packagePath)) {
+                errors.emplace_back(
+                        UsdValidationErrorType::Error,
+                        UsdValidationErrorSites{
+                                UsdValidationErrorSite(rootLayer,
+                                                       subLayer->GetDefaultPrimAsPath())
+                        },
+                        TfStringPrintf(("Found layer '%s' that "
```
tallytalwar: Had a chat with Spiff internally. "Loaded" is not correct, but maybe it should be "used"/"referenced". Lets do "Found referenced layer...."

pxr/usd/usdUtils/validators.cpp
```diff
@@ -0,0 +1,95 @@
...
+    UsdUtilsComputeAllDependencies(path, &layers, &assets,
+                                   &unresolvedPaths, nullptr);
+
+    const std::string& realPath = rootLayer->GetRealPath();
+    const std::string& packagePath = ArIsPackageRelativePath(rootLayer->GetIdentifier()) ?
+                                    ArSplitPackageRelativePathOuter(realPath).first :
+                                    realPath;
+
+    if (!packagePath.empty()) {
+        for (const SdfLayerRefPtr& subLayer : layers) {
```
tallytalwar: Spiff also correctly pointed and I agree that the use of variable name is incorrect, "subLayer" doesn't signify the concept of used/referenced "dependencies" of the usdz package. Lets rename this to be "referencedLayer".

tallytalwar (commented):
I think 1 last set of notes @beersandrew. Can you please address these?

Thanks

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
beersandrew: I've updated this to use std::filesystem to build the paths properly, let me know if there's a better way. It seems when I just try to use the identifier of the root stage, it uses the relative path "pass.usdz" or "fail.usdz" but the error message from the code ends up including the full actual path in the error message, using the realPath of the root usdz file is the best way I can think of to get that root, however since it's known the file will be in the current working directory, maybe it makes sense to use something like `ArchGetCwd()` but I still think i need `std::filesystem` in order to append `excludedDirectory` and get the correct "/" or "\" afterwards.

let me know what you think

tallytalwar: Hi @beersandrew 

Do you think https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/tf/pathUtils.h#L80-L88 might be helpful?

beersandrew: appreciate your patience on this one @tallytalwar, I've updated the pathing to use Tf functions where it seemed appropriate, let me know what you think

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -55,24 +55,21 @@ TestPackageEncapsulationValidator()
...
-    const std::string& errorAsset = rootPath.parent_path()
-            .append("excludedDirectory")
-            .append("image.jpg");
+    const std::string& absoluteUsdzPath = TfAbsPath(
+            stage->GetRootLayer()->GetIdentifier());
+    const std::string& usdzRootDirectory = TfGetPathName(absoluteUsdzPath);
+    const std::string& errorLayer = TfStringCatPaths(
+            usdzRootDirectory, "excludedDirectory/layer.usda");
+    const std::string& errorAsset = TfStringCatPaths(
+            usdzRootDirectory, "excludedDirectory/image.jpg");
```
tallytalwar, line 66: For Identifier its returning the const reference of the saved layer's identifier, but for all other calls I won't do a reference. Just a `const string` should be fine.

The use of reference for TfGetPathName, TfStringCatPaths is actually incorrect, as it will to get a reference of a temporary string.

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -0,0 +1,100 @@
...
+            stage->GetRootLayer()->GetIdentifier());
+    const std::string usdzRootDirectory = TfGetPathName(absoluteUsdzPath);
+    const std::string errorLayer = TfStringCatPaths(
+            usdzRootDirectory, "excludedDirectory/layer.usda");
+    const std::string errorAsset = TfStringCatPaths(
+            usdzRootDirectory, "excludedDirectory/image.jpg");
+
+    std::vector<std::string> expectedErrorMessages = {
+            TfStringPrintf(("Found referenced layer '%s' that does not "
+                            "belong to the package '%s'."),
```
tallytalwar, line 70: Now the layer's path is good, but notice that you are using real path in the validator code for packagePath in the error string, I think which is correct so this tests should also not normalize the packagePath (rootLayer's realPath).

beersandrew: thanks, I've updated this to be the same as the validator code. I wasn't sure if I should bring over: 

`const std::string& packagePath = ArIsPackageRelativePath(rootLayer->GetIdentifier()) ?
                                    ArSplitPackageRelativePathOuter(realPath).first :
                                    realPath;`
                                    
 but since it will always be the realPath for these test files, I took it out. Let me know if I should add it back in or not.

PR #3208, opened 2024-08-06
[usdview] fix for windows users with PYTHONSTARTUP files

pmolodo:
### Description of Change(s)

fix for issue where backslashes in paths weren't correctly escaped when trying to run PYTHONSTARTUP file

fix ensures proper formatting of string literals by using `repr(str(path))` - the `repr` ensures proper backslash escaping AND proper quoting (and the `str` is there in case someone passes in, ie, a `pathlib.Path`).

### Fixes Issue(s)
- errors on python interpreter startup for windows users with a PYTHONSTARTUP set


jesschimein:
Filed as internal issue #USD-9939

PR #3215, opened 2024-08-08
Apple: Enable Imaging support for Embedded Platforms

dgovil:
### Description of Change(s)

This PR ports Thor's Imaging work onto the latest USD dev branch and enables imaging for iOS and visionOS. It also includes some patches from Maddy Adams and me to suit the current state of USD and its dependencies.

There is a single patch to MaterialX to enable support with USD not using OpenGL builds.

Additionally, as requested by Nick Porcino, I have added support for shared memory texture handling on the Mac as well where applicable.

I've tested this on macOS with Apple Silicon but do not have Intel macs to test with. I've also tested building this for both iOS and visionOS.

(Jess I believe you'd tracked this as #USD-9841)

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-9955

pxr/imaging/hgiMetal/blitCmds.mm
```diff
@@ -203,7 +203,11 @@
     
     mtlDesc.mipmapLevelCount = dstTexDesc.mipLevels;
     mtlDesc.arrayLength = dstTexDesc.layerCount;
+#if defined(ARCH_OS_OSX)
     mtlDesc.resourceOptions = MTLResourceStorageModeManaged;
+#else
+    mtlDesc.resourceOptions = MTLResourceStorageModeShared;
```
tcauchois: This line seems like it comes up a lot in this PR; it would be nice if we could encapsulate the #if #else in a function, on hgi.mm or capabilities.mm or something.

dgovil: yeah good callout. I'll take a pass at notes next week and hit that up.

dgovil:
~~I added a commit that moves all the various shader gen backends behind their respective ifdefs~~

pxr/imaging/hgiMetal/blitCmds.mm
dgovil: Should be addressed in the latest commit. I moved it into capabilities as preferredStorageMode, to mirror the existing defaultStorageMode.

dgovil:
@davidgyu actually never mind, @ld_kerley caught some issues so I backed out my shader gen changes. I think we need to revisit it on the MaterialX side and then follow up in OpenUSD later. I'll leave the MaterialX build patch in for that reason.

dgovil:
Just a heads up that I added some changes from @furby-tm to fix a few minor issues they found.
One thing too though is that it guards the releases behind an arc check, and if we can merge in https://github.com/PixarAnimationStudios/OpenUSD/pull/3226 then we can potentially remove those as well.

PR #3216, opened 2024-08-08
Autodesk: Configurable anisotropic filtering

### LINKED ISSUE #3180: All Hgi implementations force maximum anisotropic filtering for all textures regardless of usage
### Description of Issue
In all Hgi implementations, texture sampling is set to the maximum anisotropic filtering supported by the implementation, without any way to override it. See `pxr/imaging/hgiGL/sampler.cpp`, `pxr/imaging/hgiGL/texture.cpp`, `pxr/imaging/hgiMetal/sampler.mm`, `pxr/imaging/hgiVulkan/sampler.cpp`. This can be a significant performance issue, as it increases the texture memory bandwidth usage. 
This is also done for every sampler, regardless of its usage. For example when enabling dome lighting, every sample on the pre-filtered light maps is done with anisotropic filtering, even if it's not a UV mapped texture (not applied to geometry). This can cause accuracy issues (because the sample is not on a fragment), but it's also completely wasted bandwidth.
Performance of anisotropic filtering is highly hardware dependent. Each vendor does it differently, and there are no standards (graphics API specs are very vague about how it should work). It's typically cheap on modern high performance hardware, but no so much on low spec hardware devices or software rasterizers.

A few simple changes could help fix most problems:
- Expose `maxAnisotropy` in `HgiSamplerDesc`
- Define an environment setting `HGI_MAX_ANISOTROPY` as an override for the implementation max anisotropy
- Update current users of textures/samples to disable anisotropy when not needed (like dome lighting texture)

### Steps to Reproduce
N/A

N/A

### Package Versions
N/A

### Build Flags
N/A

AdamFelt:
### Description of Change(s)

Make anisotropic filtering configurable. See issue #3180 for rationale.

Expose maxAnisotropy in HgiSamplerDesc
Define an environment setting HGI_MAX_ANISOTROPY as an override for the implementation max anisotropy
Update current users of textures/samplers to disable anisotropy when not needed (like dome lighting texture)

### Fixes Issue(s)

Mostly fixes #3180, but more work should be done to configure anisotropic filtering from the materials (can be a separate issue).
-

### DIFF
--- a/pxr/imaging/hd/material.cpp
+++ b/pxr/imaging/hd/material.cpp
@@ -270,6 +270,7 @@ _GetSamplerParameters(
                  nodeTypeId, parameters, sdrNode, nodePath),
              _ResolveMagSamplerParameter(
                  nodeTypeId, parameters, sdrNode, nodePath),
+             /*maxAnisotropy*/16,
              HdBorderColorTransparentBlack, 
              /*enableCompare*/false, 
              HdCmpFuncNever };
--- a/pxr/imaging/hd/types.cpp
+++ b/pxr/imaging/hd/types.cpp
@@ -109,7 +109,8 @@ HdSamplerParameters::HdSamplerParameters()
 
 HdSamplerParameters::HdSamplerParameters(
     HdWrap wrapS, HdWrap wrapT, HdWrap wrapR, 
-    HdMinFilter minFilter, HdMagFilter magFilter, 
+    HdMinFilter minFilter, HdMagFilter magFilter,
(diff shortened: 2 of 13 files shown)


jesschimein:
Filed as internal issue #USD-9957

AdamFelt:
> Filed as internal issue #USD-9957
> 

Can you please verify that I am on the Autodesk CLA.

jesschimein:
Sorry for the confusion!  Yes, you are on the Autodesk CLA!

PR #3217, opened 2024-08-10
Conform SdfLayer and UsdStage dtors to C++ Standard.

furby-tm:
* Without these `noexcept` exception specifications, **SdfLayer** and **UsdStage** fail to compile when compiling with for example, **Swift/Clang** with the following error:
  ```
  error: exception specification is more lax than base version
  ```
* Please see [this reference](https://timsong-cpp.github.io/cppwp/n4659/except.spec#4) where this is mentioned in the standard:

  > 18.4 Exception speciﬁcations [except.spec]
  > ...
  > 4. If a virtual function has a non-throwing exception speciﬁcation, all declarations,
  > including the deﬁnition, of any function that overrides that virtual function in any
  > derived class shall have a non-throwing exception speciﬁcation, unless the overriding
  > function is deﬁned as deleted. 

### Description of Change(s)

- Add `noexcept` exception specifications to the virtual dtors of **SdfLayer** and **UsdStage**.

### DIFF
--- a/pxr/usd/sdf/layer.cpp
+++ b/pxr/usd/sdf/layer.cpp
@@ -198,7 +198,7 @@ SdfLayer::SdfLayer(
     _MarkCurrentStateAsClean();
 }
 
-SdfLayer::~SdfLayer()
+SdfLayer::~SdfLayer() noexcept
 {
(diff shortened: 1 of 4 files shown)


furby-tm:
Closed in favor of **#3219**.

PR #3218, opened 2024-08-10
Clang Modules: Remove includes within pxr namespace.

furby-tm:
* When compiling with **Swift/Clang** (**`-fmodule`**), it is an error to include modules within namespaces:
  ```
  error: import of module appears within namespace
  ```

### Description of Change(s)

- **Arch.FileSystem**: Moved the namespace block: **`PXR_NAMESPACE_OPEN_SCOPE`** below the includes (**imports**) of **`Darwin.C.limits`** and **`Glibc.limits`**, so those modules are not imported inside of the **`pxr`** namespace.

- **Sdf.MapEditor**: ~~Closed the namespace block: **`PXR_NAMESPACE_CLOSE_SCOPE`** before the includes, and then re-opened the namespace block: **`PXR_NAMESPACE_OPEN_SCOPE`** immediately after the includes.~~ Per @nvmkuruc's [suggestion](https://github.com/PixarAnimationStudios/OpenUSD/pull/3218#discussion_r1713937224), it was deemed cleaner to simply move the includes to the top of the file.

<hr/>

### DIFF
--- a/pxr/usd/sdf/mapEditor.cpp
+++ b/pxr/usd/sdf/mapEditor.cpp
@@ -9,7 +9,9 @@
 #include "pxr/usd/sdf/mapEditor.h"
 #include "pxr/usd/sdf/path.h"
 #include "pxr/usd/sdf/schema.h"
+#include "pxr/usd/sdf/types.h"
 
+#include "pxr/base/vt/dictionary.h"
 #include "pxr/base/vt/value.h"
 #include "pxr/base/tf/diagnostic.h"
 #include "pxr/base/tf/mallocTag.h"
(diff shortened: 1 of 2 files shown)


pxr/usd/sdf/mapEditor.cpp
```diff
@@ -178,8 +178,13 @@ Sdf_CreateMapEditor(const SdfSpecHandle& owner, const TfToken& field)
     template std::unique_ptr<Sdf_MapEditor<MapType> >                \
         Sdf_CreateMapEditor(const SdfSpecHandle&, const TfToken&);   \
 
+PXR_NAMESPACE_CLOSE_SCOPE
+
 #include "pxr/base/vt/dictionary.h"
 #include "pxr/usd/sdf/types.h"
+
```
nvmkuruc: Is there a reason that the includes need to be down here?  Could they just be relocated to the top of the header instead of closing/reopening the namespace?

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf4-include-header-files-before-other-declarations-in-a-file

furby-tm: Thanks for reviewing this, I believe your suggestion should work out just fine, and I will modify this PR per your suggestion, my original goal was to attempt to keep the changes to this file minimal, but I agree that moving the includes to the top is a much cleaner solution.

Thanks!

furby-tm: Modified this PR per your suggestion, and all checks still passed on my end. 👍🏻

jesschimein:
Filed as internal issue #USD-9958

furby-tm:
Sorry, I had realized I had made these PRs against `release` and they should be PRs made out to the `dev` branch. Closing in favor of https://github.com/PixarAnimationStudios/OpenUSD/pull/3224

Issue #3214, opened 2024-08-08
USDA Writes Sensitive to User's Locale


erslavin:
### Description of Issue

The USDA writer appears to be sensitive to the current locale setting when writing integer content.  Indeed digging down it looks to be using `TfStringify`, with no specialization for integers, which results in the underlying `ostream` being initialized with whatever the current global locale is set to.  I isolated this by adding the following sequence into the test for `TfStringUtils`:

```
// verify that TfStringify is agnostic to locale for
// numerical values
std::locale originalLocale;
std::locale::global(std::locale(""));
TF_AXIOM(TfStringify(1000.56) == "1000.56");
TF_AXIOM(TfStringify(1000) == "1000");
std::locale::global(originalLocale);
TF_AXIOM(TfStringify(1000) == "1000");
TF_AXIOM(TfStringify(1000.56) == "1000.56");
```

Indeed, this fails on the line you would expect:

`TF_AXIOM(TfStringify(1000) == "1000");

Either this behavior should be guarded against in `TfStringify` by potentially specializing for integer values, or if not globally desirable for `TfStringify` the USDA writer needs to be aware that this is a possibility and initialize the writes with the standard C++ locale prior to using the streams (and restore it after write is complete).

### Steps to Reproduce
1.  Add the above block to the `TfStringUtils` tests
2.  Run the `TfStringUtils` test to note the failure

A similar issue was noted here (likely with the same underlying cause), but closed: https://github.com/PixarAnimationStudios/OpenUSD/issues/1350

### System Information (OS, Hardware)

Windows

### Package Versions

Current `dev` branch

jesschimein:
Filed as internal issue #USD-9954

meshula:
From a discussion, a plausible solution is to specialize using `std::to_chars` since it is optimized for performance, and does not use locale.

eg
```
std::string result(32, '\0'); // big enough
auto [ptr, ec] = std::to_chars(&result[0], &result[0] + result.size(), value);
```

nvmkuruc:
@jesschimein #3222 should have fixed this issue.  I think we can close this.

jesschimein:
Thank you!  Closing :)

PR #3222, opened 2024-08-12
Specialized TfStringify for integral types to prevent locale dependent writes

erslavin:
### Description of Change(s)

Specializes TfStringify for common integral types to prevent locale dependent writes of numeric data.

### Fixes Issue(s)
- #3214 

### DIFF
--- a/pxr/base/tf/stringUtils.cpp
+++ b/pxr/base/tf/stringUtils.cpp
@@ -913,65 +913,6 @@ TfStringify(std::string const& s)
     return s;
 }
 
-template <typename T>
-std::string
-_TfStringifyIntegralImpl(const T value)
-{
-    std::ostringstream stream;
-    stream.imbue(std::locale::classic());
-    stream << value;
-
-    return stream.str();
-}
-
-std::string
-TfStringify(short val)
-{
-    return _TfStringifyIntegralImpl<short>(val);
-}
-
-std::string
-TfStringify(unsigned short val)
-{
-    return _TfStringifyIntegralImpl<unsigned short>(val);
-}
-
-std::string
-TfStringify(int val)
-{
-    return _TfStringifyIntegralImpl<int>(val);
-}
-
-std::string
-TfStringify(unsigned int val)
-{
-    return _TfStringifyIntegralImpl<unsigned int>(val);
-}
-
-std::string
-TfStringify(long val)
-{
-    return _TfStringifyIntegralImpl<long>(val);
-}
-
-std::string
-TfStringify(unsigned long val)
-{
-    return _TfStringifyIntegralImpl<unsigned long>(val);
-}
-
-std::string
-TfStringify(long long val)
-{
-    return _TfStringifyIntegralImpl<long long>(val);
-}
-
-std::string
-TfStringify(unsigned long long val)
-{
-    return _TfStringifyIntegralImpl<unsigned long long>(val);
-}
-
 static
 const
 pxr_double_conversion::DoubleToStringConverter& 
--- a/pxr/base/tf/stringUtils.cpp
+++ b/pxr/base/tf/stringUtils.cpp
@@ -917,18 +917,11 @@ template <typename T>
 std::string
 _TfStringifyIntegralImpl(const T value)
 {
-    // plus one because for signed values, digits10 will give one less
-    // than what is actually needed in cases where the amount of characters
-    // could represent an overflow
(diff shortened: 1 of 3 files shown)


pxr/base/tf/stringUtils.cpp
```diff
@@ -912,6 +913,72 @@ TfStringify(std::string const& s)
     return s;
 }
 
+template <typename T>
+std::string
+_TfStringifyIntegralImpl(const T& value)
```
nvmkuruc: Should this be just `const T` or `T` in the spirit of this--

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

jesschimein:
Filed as internal issue #USD-9967

pxr/base/tf/stringUtils.h
```diff
@@ -572,6 +572,22 @@ TF_API std::string TfStringify(std::string const&);
...
+/// \overload
+TF_API std::string TfStringify(unsigned int);
+/// \overload
+TF_API std::string TfStringify(long);
+/// \overload
+TF_API std::string TfStringify(unsigned long);
+/// \overload
+TF_API std::string TfStringify(long long);
+/// \overload
+TF_API std::string TfStringify(unsigned long long);
```
sunyab: Instead of using std::to_chars, could we just make the `ostringstream` in the base `TfStringify` implementation always use the classic locale?

```
std::ostringstream stream;
stream.imbue(std::locale::classic());
...
```

This is simpler and ensures other types going through this code path will also pick up this behavior. Also, our performance tests for layer writing (surprisingly?) showed negligible effects using `std::to_chars` but noticeable improvements using the classic locale.

pxr/base/tf/testenv/stringUtils.cpp
```diff
@@ -386,6 +395,35 @@ TestStrings()
...
+    
+    // verify that TfStringify is agnostic to locale for
+    // numerical values
+    std::locale originalLocale;
+    std::locale::global(std::locale(""));
+    try
+    {
+        TF_AXIOM(TfStringify(1000.56) == "1000.56");
+        TF_AXIOM(TfStringify(1000) == "1000");
+        std::locale::global(originalLocale);
```
sunyab: This new test did not fail for me when I ran it without the `TfStringify` fix. I'm guessing that your environment might have a locale set that would cause this to fail?

We could install a facet that would force formatting for integral types into the locale to make this independent of environment. With the below setup, the test fails without the `TfStringify` fix and passes with it.

```
struct separate_thousands : std::numpunct<char>
{
    std::string do_grouping() const override { return "\003"; }
    char do_thousands_sep() const override { return ','; }
};
                                                                                                                                                                                                                                     
std::locale originalLocale;
std::locale::global(std::locale(std::locale(""), new separate_thousands));
...
```

pxr/base/tf/testenv/stringUtils.cpp
```diff
@@ -264,6 +265,14 @@ DoPrintfStr(const char *fmt, ...)
     return ret;
 }
 
+template <typename T>
+bool
+_RoundtripStringifyLimits()
+{
+    return (TfUnstringify<T>(TfStringify(std::numeric_limits<T>::min())) == std::numeric_limits<T>::min()) &&
+        (TfUnstringify<T>(TfStringify(std::numeric_limits<T>::max())) == std::numeric_limits<T>::max());
```
sunyab: Could you format this to adhere to the 80 char line limit?

sunyab (changes requested):

pxr/base/tf/testenv/stringUtils.cpp
erslavin: My environment has no specific locale set, it's the default; not sure why it would be different for you.

erslavin: Probably a difference between Windows and Linux.  I suspect Linux the default native locale on Linux is probably "C" (which would be the same as the locale the program is started with)

spitzak:
`std::to_chars` prints the shortest possible sequence of digits that accurately represents the floating-point number being encoded. You can't do that with `printf` the best you can do is `%.18g` which prints a lot of misleading extra digits instead of rounding to the nearest representative value. It would be a huge advantage to assume all conversions in USD are done using to_chars, it is not only being used to print '.' instead of ','.

PR #3224, opened 2024-08-13
Clang Modules: Remove includes within pxr namespace.

furby-tm:
* When compiling with **Swift/Clang** (**`-fmodule`**), it is an error to include modules within namespaces:
  ```
  error: import of module appears within namespace
  ```

### Description of Change(s)

- **Arch.FileSystem**: Moved the namespace block: **`PXR_NAMESPACE_OPEN_SCOPE`** below the includes (**imports**) of **`Darwin.C.limits`** and **`Glibc.limits`**, so those modules are not imported inside of the **`pxr`** namespace.

- **Sdf.MapEditor**: ~~Closed the namespace block: **`PXR_NAMESPACE_CLOSE_SCOPE`** before the includes, and then re-opened the namespace block: **`PXR_NAMESPACE_OPEN_SCOPE`** immediately after the includes.~~ Per @nvmkuruc's [suggestion](https://github.com/PixarAnimationStudios/OpenUSD/pull/3218#discussion_r1713937224), it was deemed cleaner to simply move the includes to the top of the file.

##### Update 8.29.24
- **Arch.Attributes**: I had just noticed another include in the pxr namespace block that had creeped up on me when compiling for Windows, moved the include for **api.h** up to the top of the file, above the **`PXR_NAMESPACE_OPEN_SCOPE`** block.

<hr/>

### DIFF
--- a/pxr/base/arch/attributes.h
+++ b/pxr/base/arch/attributes.h
@@ -15,6 +15,7 @@
 
 #include "pxr/pxr.h"
 #include "pxr/base/arch/export.h"
+#include "pxr/base/arch/api.h"
 
 PXR_NAMESPACE_OPEN_SCOPE
 
@@ -259,8 +260,6 @@ struct Arch_ConstructorEntry {
         static void _name(__VA_ARGS__)
 
 #elif defined(ARCH_OS_WINDOWS)
-    
-#    include "pxr/base/arch/api.h"
     
 // Entry for a constructor/destructor in the custom section.
     __declspec(align(16))
--- a/pxr/base/arch/fileSystem.h
+++ b/pxr/base/arch/fileSystem.h
@@ -38,8 +38,6 @@
 #include <stringapiset.h>
 #endif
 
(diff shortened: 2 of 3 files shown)


jesschimein:
Filed as internal issue #USD-9969

dgovil:
Thanks for putting this up. I think this would be a good change in general to cleanup the codebase. It aligns with our changes as well so would be great to see integrated if Pixar don't see any side effects from it.

furby-tm:
@dgovil in hindsight, I don't believe this include needs to be guarded for Windows, does it? Whatever you think might be the cleanest option [here](https://github.com/PixarAnimationStudios/OpenUSD/pull/3224/commits/d181cbd0146768279234806b579ae1ef8ac51321#diff-b40dbe788e027f40571f10933a12de6ddbb133d54a8499ac41995e7a89facceaR19-R21).

dgovil:
Yeah I believe you’re right.

PR #3225, opened 2024-08-13
Conform SdfLayer and UsdStage dtors to C++ Standard.

furby-tm:
* Without these `noexcept` exception specifications, **SdfLayer** and **UsdStage** fail to compile when compiling with for example, **Swift/Clang** with the following error:
  ```
  error: exception specification is more lax than base version
  ```
* Please see [this reference](https://timsong-cpp.github.io/cppwp/n4659/except.spec#4) where this is mentioned in the standard:

  > 18.4 Exception speciﬁcations [except.spec]
  > ...
  > 4. If a virtual function has a non-throwing exception speciﬁcation, all declarations,
  > including the deﬁnition, of any function that overrides that virtual function in any
  > derived class shall have a non-throwing exception speciﬁcation, unless the overriding
  > function is deﬁned as deleted. 

### Description of Change(s)

- Add `noexcept` exception specifications to the virtual dtors of **SdfLayer** and **UsdStage**.

<hr/>

### DIFF
--- a/pxr/usd/sdf/layer.cpp
+++ b/pxr/usd/sdf/layer.cpp
@@ -198,7 +198,7 @@ SdfLayer::SdfLayer(
     _MarkCurrentStateAsClean();
 }
 
-SdfLayer::~SdfLayer()
+SdfLayer::~SdfLayer() noexcept
 {
     TF_PY_ALLOW_THREADS_IN_SCOPE();
 
--- a/pxr/usd/sdf/layer.h
+++ b/pxr/usd/sdf/layer.h
@@ -84,7 +84,7 @@ class SdfLayer
 public:
     /// Destructor
     SDF_API
-    virtual ~SdfLayer(); 
+    virtual ~SdfLayer() noexcept; 
 
     /// Noncopyable
     SdfLayer(const SdfLayer&) = delete;
--- a/pxr/usd/usd/stage.cpp
+++ b/pxr/usd/usd/stage.cpp
@@ -683,7 +683,7 @@ UsdStage::UsdStage(const SdfLayerRefPtr& rootLayer,
     _cache->SetVariantFallbacks(GetGlobalVariantFallbacks());
 }
 
-UsdStage::~UsdStage()
+UsdStage::~UsdStage() noexcept
 {
     TF_DEBUG(USD_STAGE_LIFETIMES).Msg(
         "UsdStage::~UsdStage(rootLayer=@%s@, sessionLayer=@%s@)\n",
--- a/pxr/usd/usd/stage.h
+++ b/pxr/usd/usd/stage.h
@@ -399,7 +399,7 @@ public:
                InitialLoadSet load=LoadAll);
     
     USD_API
-    virtual ~UsdStage();
+    virtual ~UsdStage() noexcept;
 
     /// Calls SdfLayer::Reload on all layers contributing to this stage,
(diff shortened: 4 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9970

dgovil:
Thanks for putting this up. I think it's a good change and mirrors what we have ourselves. If Pixar don't see any side effects, I think its a good generalized cleanup to the codebase.

nvmkuruc:
I agree with @dgovil 👍

gitamohr:
I don't understand this one -- destructors are `noexcept` by default unless the class has a member with a throwing destructor.  None of the base classes of `SdfLayer` and `UsdStage` have exception specifications, and they don't have members with throwing destructors either, best I can tell.  If I add `static_assert(std::is_nothrow_destructible_v<SdfLayer>)` and similarly for `UsdStage`, those pass.  Internally we do not see this error in our Clang builds of USD.

furby-tm:
When compiling for Swift/Clang, out of the entire USD codebase, these are the only two classes that fail to compile, with the compiler complaining that the destructor exception specification must not be more lax than it's base class.

I'm not sure what exactly is causing this, but have not yet experienced any issues by adding `noexcept` to each of the destructors.

gitamohr:
It wouldn't cause any problems because they happen to be `noexcept` today.  I feel leery about stamping that out, though, since if someone _did_ introduce a member with a throwing dtor, we then wouldn't hear about it.  I.e. if we added the `static_assert`s like I mentioned above, then if we stamped `noexcept` on these dtors, those `static_assert`s could never fail.  What does "compiling for Swift/Clang" mean?  Is that just compiling with Clang or is it some special thing?  Could it be a bug with that?  Does it give any diagnostic information about _why_ it believes those classes don't have `noexcept` dtors?

furby-tm:
> What does "compiling for Swift/Clang" mean?

It means building the OpenUSD project from the `swift build` command, utilizing the embedded clang compiler with the Swift programming language with the Swift Package Manager (not cmake).

> Could it be a bug with that? Does it give any diagnostic information about why it believes those classes don't have noexcept dtors?

It could be, but I would find it strange as these are the only two classes that experience this compiler error not only across the USD codebase, but also across all OpenUSD's dependencies and plugin libraries (like MaterialX, OIIO, OCIO, and co.), having successfully built all those other libraries with `swift build`.

I don't see anything obvious using swift's verbose mode as to why the `noexcept` specifications are necessary, but I'd love to find a path forward to fixing this for swift.

gitamohr:
Does it complain whenever those header files are included in any .cpp file, or just when compiling certain .cpp files?  Could you try boiling it down?  I.e. copy/paste `SdfLayer` or `UsdStage` to make a duplicate, then strip away as much as you can until either whatever triggers the error is removed, or we have a minimal repro?

furby-tm:
Sure thing, what I can go ahead and do is copy over a super slimmed down UsdStage class in a header and cpp implementation, as well as a basic class that inherits from TfRefBase / TfWeakBase in a super minimal isolated swiftpm package to see if we can reproduce this in a minimal setting.

I think this should allow us to get a better feel whether this is something that is better addressed on the Swift end or if there's a minimal change that can instead be addressed in the OpenUSD code to workaround this error.

Just give me a couple weeks to get to this, but I think that should give us a clearer path forward.

PR #3227, opened 2024-08-13
usdShade: Include connectable validation from usdchecker PrimEncapsulationRule

beersandrew:
### Description of Change(s)
- adding necessary validation files for connectable validation (no nested shaders, no shaders inside scopes)

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65424985

### DIFF
--- a/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
+++ b/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
@@ -424,7 +424,7 @@ TestUsdShadeConnectableValidator()
 
     // Create a Material > Shader > Shader hierarchy
     UsdShadeMaterial::Define(usdStage, SdfPath("/RootMaterial"));
-    UsdShadeShader::Define(usdStage, SdfPath("/RootMaterial/Shader"));
+    const UsdShadeShader& topShader = UsdShadeShader::Define(usdStage, SdfPath("/RootMaterial/Shader"));
     const UsdShadeShader& insideShader = UsdShadeShader::Define(usdStage, SdfPath("/RootMaterial/Shader/InsideShader"));
 
     // Verify error that does not allow a connectable to be parented by non-container connectable
@@ -439,6 +439,10 @@ TestUsdShadeConnectableValidator()
     std::string expectedErrorMsg =
             "Connectable Shader </RootMaterial/Shader/InsideShader> cannot reside under a non-Container Connectable Shader";
     TF_AXIOM(errors[0].GetMessage() == expectedErrorMsg);
+
+    // Verify the first Shader is valid
+    errors = validator->Validate(topShader.GetPrim());
+    TF_AXIOM(errors.size() == 0);
 
     // Create a Material > Scope > Shader hierarchy
     usdStage->RemovePrim(SdfPath("/RootMaterial/Shader/InsideShader"));
--- a/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
+++ b/pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
@@ -16,6 +16,7 @@
 #include "pxr/usd/usd/validationRegistry.h"
 #include "pxr/usd/usd/validator.h"
 #include "pxr/usd/usdGeom/validatorTokens.h"
+#include "pxr/usd/usdGeom/scope.h"
 #include "pxr/usd/usdShade/shader.h"
 #include "pxr/usd/usdShade/shaderDefUtils.h"
 #include "pxr/usd/usdShade/tokens.h"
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-9979

pxr/usd/usdShade/validators.cpp
```diff
@@ -30,6 +30,70 @@
...
 
+static
+UsdValidationErrorVector
+_ConnectableValidator(const UsdPrim& usdPrim)
+{
+    UsdValidationErrorVector errors;
+
+    const UsdShadeConnectableAPI& connectable = UsdShadeConnectableAPI(usdPrim);
+
+    if (!usdPrim.GetTypeName().IsEmpty() && connectable) {
```
tallytalwar: nit: you can negate this and do an early return. And then you can move this entire logic outside the if block.

pxr/usd/usdShade/validators.cpp
```diff
@@ -30,6 +30,70 @@
...
+        const UsdPrim& parentPrim = usdPrim.GetParent();
+
+        if (parentPrim && !parentPrim.IsPseudoRoot()) {
+            UsdShadeConnectableAPI parentConnectable = UsdShadeConnectableAPI(parentPrim);
+            if (parentPrim.GetTypeName().IsEmpty()){
```
tallytalwar: How about you use ternary operator here:

```
UsdShadeConnectableAPI parentConnectable = (parentPrim.GetTypeName().IsEmpty()) ? UsdShadeConnectableAPI() : UsdShadeConnectableAPI(parentPrim)
```

But why do you want to have this check to begin with? parentConnectable for prim with no typename should return false for IsContainer, so I think you can simply remove this parentPrim typename is empty check.
```

Comment (tallytalwar) [reply pxr/usd/usdShade/validators.cpp]: I see you are doing this because of the comment in complianceChecker to work around a bug! I think that comment is outdated, the bug its talking about is all fixed now.

File: pxr/usd/usdShade/validators.cpp (hunk 1 above)
Comment (tallytalwar) [inline pxr/usd/usdShade/validators.cpp]: Also no need to do check for primType here as well.

File: pxr/usd/usdShade/validators.cpp [hunk 3]
```diff
@@ -30,6 +30,70 @@
...
+                parentConnectable = UsdShadeConnectableAPI();
+            }
+
+            if (parentConnectable && !parentConnectable.IsContainer()) {
```
tallytalwar: I will retain the original code comments. Code comments are great as it makes sure the intent being described in the code comment matches in the code. Specially for a reviewer or a future developer to work with.

pxr/usd/usdShade/validators.cpp
```diff
@@ -30,6 +30,70 @@
...
+                        },
+                        TfStringPrintf("Connectable %s <%s> cannot reside "
+                                       "under a non-Container Connectable %s",
+                                       usdPrim.GetTypeName().GetText(),
+                                       usdPrim.GetPath().GetText(),
+                                       parentPrim.GetTypeName().GetText()));
+            }
+            else if (!parentConnectable) {
+                UsdPrim currentAncestor = parentPrim.GetParent();
+                while(currentAncestor && !currentAncestor.IsPseudoRoot()) {
```
tallytalwar: You can maybe break this while loop to a recursive lambda?

pxr/usd/usdShade/validators.cpp
```diff
@@ -30,6 +30,70 @@
...
+                    if (!currentAncestor.GetTypeName().IsEmpty()) {
+                        const UsdShadeConnectableAPI& ancestorConnectable = UsdShadeConnectableAPI(currentAncestor);
+                        if (ancestorConnectable) {
+                            errors.emplace_back(
+                                    UsdValidationErrorType::Error,
+                                    UsdValidationErrorSites {
+                                            UsdValidationErrorSite(usdPrim.GetStage(),
+                                                                   usdPrim.GetPath())
+                                    },
+                                    TfStringPrintf("Connectable %s <%s> can only have"
```
tallytalwar: Will be good to retain the code comment from UsdUtilsComplianceChecker here.

pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
```diff
@@ -410,6 +412,55 @@ TestUsdShadeMaterialBindingAPIAppliedValidator()
...
+
+    UsdStageRefPtr usdStage = UsdStage::CreateInMemory();
+
+    // Create a Material > Shader > Shader hierarchy
+    UsdShadeMaterial::Define(usdStage, SdfPath("/RootMaterial"));
+    UsdShadeShader::Define(usdStage, SdfPath("/RootMaterial/Shader"));
+    const UsdShadeShader& insideShader = UsdShadeShader::Define(usdStage, SdfPath("/RootMaterial/Shader/InsideShader"));
+
+    // Verify error that does not allow a connectable to be parented by non-container connectable
+    UsdValidationErrorVector errors = validator->Validate(insideShader.GetPrim());
```
tallytalwar, line 431: Might also make sense to have a Validate for "/RootMaterial/Shader" and show no errors are reported for the same.

tallytalwar (commented):
Thanks @beersandrew . Left a couple of notes to be addressed.

beersandrew:
Thanks for the notes here @tallytalwar, I think I've addressed them all

PR #3228, opened 2024-08-14
Autodesk: [build_usd.py, hioAvif] Add option to build with AVIF

erikaharrison-adsk:
### Description of Change(s)

- Add `--avif` and `--no-avif` options to build_usd.py

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-9980

spiffmon:
Hi @erikaharrison-adsk , we appreciate the bugfix in CMakeLists.txt, which we will make, but we don't see any compelling reason for the additional build_usd options, given the almost trivial cost of building our interned AVIF, and that other hio plugins are free to override it in plugin priority?

kohakukun:
Hello @spiffmon . The change was in the context of WASM, since it wasn't supported and there was no way to programatically disable it. We might want to close this PR, since now we are in the process of discussing all wasm changes directly with some other folks at PIXAR

spiffmon:
OK, thanks, @erikaharrison-adsk - closing for now!

PR #3223, opened 2024-08-13
Add `TfEnum` for validation error

roggiezhang-nv:
### Description of Change(s)

Register `UsdValidationErrorType` as `TfEnum` and simplify function `UsdValidationError::GetErrorAsString`.

### Fixes Issue(s)
-

- [x] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement
****

PR #3232, opened 2024-08-15
Add Python binding for `UsdValidatorMetadata`

roggiezhang-nv:
### Description of Change(s)

Add Python binding for `UsdValidatorMetadata`.

It is stacked on https://github.com/PixarAnimationStudios/OpenUSD/pull/3223 for adding Python bindings for validator framework.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/usd/wrapValidator.cpp
+++ b/pxr/usd/usd/wrapValidator.cpp
@@ -1,0 +1,106 @@
+//
+// Copyright 2024 Pixar
+//
+// Licensed under the terms set forth in the LICENSE.txt file available at
+// https://openusd.org/license.
+//
+#include "pxr/pxr.h"
+#include "pxr/usd/usd/validator.h"
+
+#include "pxr/base/tf/pyContainerConversions.h"
+#include "pxr/base/tf/pyPtrHelpers.h"
+#include "pxr/base/tf/pyResultConversions.h"
+
+#include <boost/python/class.hpp>
+#include <boost/python/make_constructor.hpp>
+#include <boost/python/operators.hpp>
+#include <boost/python/object.hpp>
+
+using namespace boost::python;
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace
+{
+
+    UsdValidatorMetadata *
+    _NewMetadata(
+        const TfToken &name,
+        const PlugPluginPtr &plugin,
+        const TfTokenVector &keywords,
+        const TfToken &doc,
+        const TfTokenVector &schemaTypes,
+        bool isSuite)
+    {
+        return new UsdValidatorMetadata{name, plugin, keywords, doc, schemaTypes, isSuite};
+    }
+
+    TfToken
+    _GetMetadataName(const UsdValidatorMetadata &metadata)
+    {
+        return metadata.name;
+    }
+
+    void
+    _SetMetadataName(UsdValidatorMetadata &metadata, const TfToken &name)
+    {
+        metadata.name = name;
+    }
+
+    TfTokenVector
+    _GetMetadataKeywords(const UsdValidatorMetadata &metadata)
+    {
+        return metadata.keywords;
+    }
+
+    void
+    _SetMetadataKeywords(UsdValidatorMetadata &metadata, const TfTokenVector &keywords)
+    {
+        metadata.keywords = keywords;
+    }
+
+    TfTokenVector
+    _GetMetadataSchemaTypes(const UsdValidatorMetadata &metadata)
+    {
+        return metadata.schemaTypes;
+    }
+
+    void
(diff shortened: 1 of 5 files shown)


roggiezhang-nv:
@nvmkuruc for vis.

jesschimein:
Filed as internal issue #USD-9985

pxr/usd/usd/wrapValidator.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                          (arg("name") = TfToken(),
+                                           arg("plugin") = PlugPluginPtr(),
+                                           arg("keywords") = TfTokenVector(),
+                                           arg("doc") = TfToken(),
+                                           arg("schemaTypes") = TfTokenVector(),
+                                           arg("isSuite") = false)))
+        .add_property("name", &_GetMetadataName, &_SetMetadataName)
+        .add_property("plugin", &_GetMetadataPlugin, &_SetMetadataPlugin)
+        .add_property("keywords", make_function(
+            &_GetMetadataKeywords, return_value_policy<TfPySequenceToList>()), _SetMetadataKeywords)
```
tallytalwar: I was having some good discussion with Alex here at Pixar about python APIs and how it might be confusing for a python developer to not be able to do something like: `metadata.keywords[0] = "foo"`. We also discussed and have many examples of this in our code base where we wrap something as a python property if value of the type is immutable in python.

So that gets me to, will it make more sense to provide a GetKeywords/SetKeywords (and similarly for schemaTypes below) which returns a list and takes in a list respectively? (We have example of this here: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/wrapAttribute.cpp#L123-L124

But now this results in inconsistency with the cpp API, which at present doesn't provide any member functions on the UsdValidationMetadata struct. Do you think it will be possible for you to add these methods also for consistency?

Thanks

pxr/usd/usd/wrapValidator.cpp
```diff
@@ -0,0 +1,106 @@
...
+        .def_readwrite("doc", &UsdValidatorMetadata::doc)
+        .add_property("schemaTypes", make_function(
```
tallytalwar: Same note about schemaTypes also.

tallytalwar (commented):
Left a note about modifying the usage of add_property for metadata data members which work off a token vector. Thanks

pxr/usd/usd/wrapValidator.cpp
roggiezhang-nv: That makes sense to me. I'll change it to Set/Get function instead. Thanks.

roggiezhang-nv: @tallytalwar I have changed it to use getter/setter to access keywords and schemaTypes, but not for cpp struct. One question is that should I keep `keywords` and `schemaTypes` as private or add public getter/setter for them only? That adds another inconsistency that it has two set of methods to access the two members if I keep them public. While it breaks back compatibility if I moved them into private.

nvmkuruc: What about using a `vector_indexing_suite` to preserve consistency with the CPP API?

nvmkuruc: Rather than use `vector_indexing_suite` or adding explicit get/set methods, I think I'd actually argue in favor of just making `UsdVaidatorMetadata` immutable in python.  In general, the usage for `UsdValidatorMetadata` is to just reflect and inspect metadata described in C++ validator plugins, not editing in python.

There's already a keyword argument interface for easy construction for testing.  `keywords` and `schemaTypes` could use `TfPySequenceToTuple`.

roggiezhang-nv: Yes, that makes sense to me if we don't need to register validator in python.

tallytalwar: > UsdValidatorMetadata is to just reflect and inspect metadata described in C++ validator plugins, not editing in python.

Hmm, I think we should allow registering of validators from python also. So I can see we providing API for this https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/validationRegistry.h#L226 in python.

For this usecase, we will have to have immutable metadata constructs in python, no? @nvmkuruc

nvmkuruc: Even when registering validators in python, the keyword arg constructor @roggiezhang-nv provides for validator metadata is probably the most pythonic way to initialize and somewhat symmetrical to aggregate initialization of structs with designated initializers in C++.

tallytalwar: Yup I missed the constructor! Agreed with your statements @nvmkuruc

roggiezhang-nv: I'll make those properties immutable then.

tallytalwar:
Thanks @roggiezhang-nv , syncing this in now.

PR #3245, opened 2024-08-22
Ios imaging

gvanush:
### Fixes Issue(s)
-

- [ ] I have submitted a signed Contributor License Agreement


dgovil:
Hi, @gvanush ,
Thanks for putting up this PR.
We have another PR that adds imaging support here: https://github.com/PixarAnimationStudios/OpenUSD/pull/3215 and covers a little different ground.

It would be useful if you could provide feedback on that PR so we can track the support in one place. I think it captures most of your changes here with the exception of support for building in simulators. 

I've stubbed in simulator build support here in the frameworks PR though that is based on top of my Imaging PR above https://github.com/PixarAnimationStudios/OpenUSD/pull/2969 in case that helps. However I'm currently refactoring this PR a bit to meet some feedback.

I'd appreciate any feedback of yours however on either PR.

jesschimein:
Filed as internal issue #USD-10011

gvanush:
> Hi, @gvanush , Thanks for putting up this PR. We have another PR that adds imaging support here: #3215 and covers a little different ground.
> 
> It would be useful if you could provide feedback on that PR so we can track the support in one place. I think it captures most of your changes here with the exception of support for building in simulators.
> 
> I've stubbed in simulator build support here in the frameworks PR though that is based on top of my Imaging PR above #2969 in case that helps. However I'm currently refactoring this PR a bit to meet some feedback.
> 
> I'd appreciate any feedback of yours however on either PR.

Great to hear @dgovil ! I just started with the iOS port but can use yours as it seems to work already. 
One thing I needed and it works already is building for iOS simulator Arm 64. Also have a script (that maybe worth incorporating to the build usd script) to build xcframework from all Apple platform variations. 

Is there an ongoing work for simulator support?

dgovil:
I'm adding simulator support in the frameworks PR so it discovers the right sdk for both. 

For xcframeworks, I haven't tried setting that up within OpenUSD directly yet because it would require doing multiple builds before the combining, but I do think it would make sense to have a script that does the multi build + combining

dgovil:
I just wanted to reply here with an update:
Imaging support has been merged in and will be in USD 24.11 unless something goes wrong.
The [Framework PR](https://github.com/PixarAnimationStudios/OpenUSD/pull/2969) has updated support for Simulators and building as an XCFramework. It's now rebased on top of latest dev.

Issue #3233, opened 2024-08-15
`UsdLuxCylinderLight` documentation and bounds computation are inconsistent


nvmkuruc:
### Description of Issue

OpenUSD's documentation describes a cylinder light as being "centered at the origin and has its major axis on the X axis".
https://openusd.org/dev/api/class_usd_lux_cylinder_light.html#details

The bounds computation implementation however centers the axis along the Z axis.
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdLux/cylinderLight.cpp#L193
`(*extent)[1] = GfVec3f(radius, radius, length * 0.5f);`

Should extents computation be updated to match the documentation?

### Steps to Reproduce
1.

jesschimein:
Filed as internal issue #USD-9986

spiffmon:
Yep, sounds right, Matt, let’s just double check that HdPrman is X axis… I
reckon the bounds comp got copied from UsdGeomCylinder, which is Z…

pmolodo:
> Yep, sounds right, Matt, let’s just double check that HdPrman is X axis… 

Yup, it looks like HdPrman is indeed using X axis:

https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.08/third_party/renderman-26/plugin/hdPrman/light.cpp#L753-L766

```c++
        } else if (_lightShaderType == us_PxrCylinderLight) {
            // radius (YZ only, default 0.5)
            VtValue radius = sceneDelegate->GetLightParamValue(id,
                HdLightTokens->radius);
            if (radius.IsHolding<float>()) {
                geomScale[1] *= radius.UncheckedGet<float>() / 0.5;
                geomScale[2] *= radius.UncheckedGet<float>() / 0.5;
            }
            // length (X-axis)
            VtValue length = sceneDelegate->GetLightParamValue(id,
                HdLightTokens->length);
            if (length.IsHolding<float>()) {
                geomScale[0] *= length.UncheckedGet<float>();
            }
```

nvmkuruc:
Thanks @pmolodo!

Do you want a fix from us @spiffmon?

spiffmon:
Yeah, that'd be great, Matt!

--spiff

pmolodo:
> Yeah, that'd be great, Matt!
> --spiff

Done: https://github.com/PixarAnimationStudios/OpenUSD/pull/3246

nvmkuruc:
Fixed by #3246

PR #3246, opened 2024-08-22
[usdLux] UsdLuxCylinderLight - bounds computation puts length along x axis

pmolodo:
### Description of Change(s)
UsdLuxCylinderLight - bounds computation puts length along x axis, to match documentation

### Fixes Issue(s)
- #3233 


jesschimein:
Filed as internal issue #USD-10018

Issue #3249, opened 2024-08-24
Crate write out gets exponentially slower with high time sample count


dgovil:
### Description of Issue

We're seeing a reproducible issue across USD 24.2-24.8 (didn't test earlier) where writing crate becomes exponentially slower as you add more samples.

- With 100 samples, crate takes 2.5x longer than a USDA to write on average
- With 10,000 samples, crate takes 15x longer. (0.105 vs 0.006)
- With 20,0000 samples, it takes about 30x longer (0.4s vs 0.014s)
- With a single attribute at 30,000 samples , it takes 36x longer to write out than a text file. (0.75s vs 0.02s)
- When this jumps to 35,000 samples, it does not finish writing the crate file on my M2 Ultra after even 10 minutes

We've reproduced this fairly reliably on multiple machines and builds of USD. The ratios are always roughly the same and it gives up at about the same count.

Additionally it completely pegs a single core at 100%.
It has roughly 48 threads active at this time on my system (I have 24 cores, so it seems to double the cores?) but most(all but one?) seem to be inactive during this. 
It also doesn't seem to use much memory at all during this time, at roughly 77MB.

The chart here shows how much slower than USDA crate becomes as the samples get larger.
Granted my sample sizes start jumping after a bit, but it shows somewhat linear growth but then at about 35k samples it just won't complete at all.
Bear in mind this is a multiplier on the usda document write speed which is roughly O(N). This makes crate something like O(n^2). 

<img width="2190" alt="image" src="https://github.com/user-attachments/assets/b57206ca-ef73-44ec-8598-a06c94b5ed61">

Some observations:
1. This doesn't seem to change depending on the data type of the attribute. Only the number of samples seems to change it.
2. Timesamples are stored with their key as a DoubleVector. So in theory there should be no compression feeding into this either. Unless I'm mistaken, that should be faster than usda because it should just be able to dump things linearly as it writes out, without needing to format things...so its odd that its slower even at a low sample count.

### Steps to Reproduce
1. Run this script
4. Get a few coffees

```python
from pxr import Usd, Sdf
import tempfile, os, time

stage = Usd.Stage.CreateInMemory()
prim = stage.DefinePrim("/test")
attr = prim.CreateAttribute("foo", Sdf.ValueTypeNames.Float)

count = (1, 100, 1_000, 10_000, 20_000, 30_000, 35_000, 40_000)

for c in count:
    print(f"Running with {c} samples")

    for i in range(c):
[...]

Issue #2662, opened 2023-09-07
Sporadic crash during recomposition on M1 Mac builds


marktucker:
### Description of Issue
There seems to be a recently (between 23.02 and 23.08) introduced race condition to USD composition on M1 builds. See https://forum.aousd.org/t/anyone-seeing-issues-with-usd-on-m1-macs/599 for some additional information.

### Steps to Reproduce
1. Run the following python code (or something similar) many many times on an M1 Mac build of USD:
```
s = Usd.Stage.CreateInMemory()
prim = s.DefinePrim('/asset')
prim.GetReferences().AddReference(asset_file_path)
```
2. Eventually get either a crash (in 23.08) or a runtime error (23.05)

### System Information (OS, Hardware)
MacOS 13, clang 14, M1 Mac

### Package Versions
23.05, 23.08

### Build Flags
M1 arm64 builds

jesschimein:
Filed as internal issue #USD-8668

dgovil:
Just an FYI this is still an issue with 24.8 as well..

Sample snippet to run it in a loop


Running this `arch -arm64 python3 test.py` but not with `arch -x86_64 python3`. Substitute the path to the gramophone with https://developer.apple.com/augmented-reality/quick-look/models/gramophone/gramophone.usdz

I suspect there's some subtle memory ordering bug somewhere

```
import platform, sys
print(sys.version_info)
print(platform.processor())

from pxr import Usd
print(Usd.GetVersion())

for x in range(40_000):
    s = Usd.Stage.CreateInMemory()
    prim = s.DefinePrim('/asset')

    prim.GetReferences().AddReference("/Users/dhruvgovil/Downloads/gramophone.usdz")

print("Works")
```

dgovil:
Actually, I notice that at 40k iterations, newer versions of USD deadlock. 
I went down the iteration count and I think its related to the lock here https://github.com/PixarAnimationStudios/OpenUSD/issues/3249

Under 32k iterations, you still see the issue Mark describes, but over 32k iterations I think the hashing structure enters the infinite loop I also describe in that other issue.

jesschimein:
Hi @marktucker @dgovil, do you know if this is still happening with more recent OS and compilers? 🙏🏻

dgovil:
It looks to be fixed for me. I had closed my diagnostic issue but I think we can close this one too unless mark objects?

marktucker:
I'm not seeing this on recent Houdini builds either.

[...]

PR #3253, opened 2024-08-28
Apple: Change robin_hash.h distance_type to int64_t to prevent overflow loop

dgovil:
### Description of Change(s)

Robin hash has a distance type that is declared as an int16_t. This causes anything that calls into it with 32k or higher entries to wrap around and form an infinite loop, locking up the current process.

Alternatively, the latest robin_hash code has checks for this, so this might be a stopgap till the project can upgrade the version used. 

This appears to mostly happen on Clang, but I assume that's because GCC might have other behaviour on wrap, and we haven't tested MSVC.

I don't notice a performance regression with this patch. Timing of common benchmarks stays largely the same, and memory usage seems very similar too. I don't believe this type is used for longer term storage outside of the functions it's used in.

Thanks to Damien for finding this as half the cause to the issues below. Both issues below still have their respective performance issues elsewhere in the code, but at least no longer loop infinitely.

### Fixes Issue(s)
Fixes the issue of an infinite loop in both these issues, but not the remaining performance issue that both have.
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2662
- https://github.com/PixarAnimationStudios/OpenUSD/issues/3249

### DIFF
--- a/pxr/base/tf/pxrTslRobinMap/robin_hash.h
+++ b/pxr/base/tf/pxrTslRobinMap/robin_hash.h
@@ -174,7 +174,7 @@ class bucket_entry : public bucket_entry_hash<StoreHash> {
 
  public:
   using value_type = ValueType;
-  using distance_type = std::int16_t;
+  using distance_type = std::int64_t;
 
   bucket_entry() noexcept
       : bucket_hash(),


e4lam:
Why not cherry-pick https://github.com/Tessil/robin-map/commit/68ff7325b3898fca267a103bad5c509e8861144d instead? Reading the [referenced issue](https://github.com/Tessil/robin-map/issues/52#issuecomment-1367498868), the cause of this is when there are too many hash collisions, which may also be symptomatic of a poorly chosen hash function.

spitzak:
I agree with e4lam's comment. If the problem was an infinite loop for more than 2^16 entries it would have been obvious long ago. It also seems likely that a poor hash function is the explanation for the different performance between compilers.

dgovil:
Oh I’d missed that specific commit around it.
If that fixes the issue, then yeah that seems wise. Alternatively, perhaps
if the issue is that old , it might be time to just upgrade the version
whole sale in case there are other endemic issues from the version used.

jesschimein:
Filed as internal issue #USD-10037

gitamohr:
I've updated this to version 1.3.0.

dgovil:
Awesome. I'll close this out and test with your latest changes.

PR #3254, opened 2024-08-29
chore: delete useless residual member function declaration for SdfPath

sssooonnnggg:


jesschimein:
Filed as internal issue #USD-10044

PR #3255, opened 2024-08-29
perf: cherry pick of robin-map's optimization from upstream, use er…

sssooonnnggg:
…ase_fast instead of erase when iter is not needed as a return value to improve the performance of erase operation

### Description of Change(s)
robin_hash_map's erase has performace issues, see: https://github.com/Tessil/robin-map?tab=readme-ov-file#performance-pitfalls
Cherry-pick https://github.com/Tessil/robin-map/pull/76 and replace erase with erase_fast

### Fixes Issue(s)


jesschimein:
Filed as internal issue #USD-10045

gitamohr:
Thank you!  I'm updating the interned copy of `robin_map` to version 1.3.0 separately, and I will merge your other changes to make use of `erase_fast` from this PR.

PR #3271, opened 2024-09-04
Apple: Accessibility  API schemas

dgovil:
This PR adds an Accessibility API schemas to UsdUI as an attempt to make it a Universally Accessible Scene Description. Previously it also included Localization, but after discussion, that has been moved to https://github.com/PixarAnimationStudios/OpenUSD/pull/3475

Disclaimer: This addition does not represent any future feature support in our products. These schemas are borne out of discussing the needs for the ecosystem to accommodate a wider range of users.

## Overview
A brief overview of the schema:

[Proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/accessibility)

```
def "Foo"(
    apiSchemas = ["AccessibilityAPI:default"]
) {
    string accessibility:default:label = "Short Label"
    string accessibility:default:description = "A longer description of this prim"
    uniform token accessibility:default:priority = "standard"
}
```

UsdUIAccessibilityAPI is a multiple apply schema.
Each instance has three attributes:

label: A short description of the prim
description: An extended description of the prim
priority: A hint to a runtime on how this instance should be valued

## Feedback and Remaining Work

It would be greatly appreciated if feedback could be provided with a focus on the schema first so we can nail down how to represent the data in the USD.


nvmkuruc:
Do you have any thoughts on how localization should work across schema domains? For example--  consider the following scene description.

```
def Xform "prim_with_primvar" (apiSchemas = ["LocalizationAPI:default", "LocalizationAPI:fr_CA"]) {
    string primvars:greeting = "hello"
    string primvars:greeting:lang:fr_CA = "bonjour"
}
```

I could see some clients may interact with `primvars:greeting` as a localized attribute with the `UsdUILocalizationAPI` and others may use the `UsdGeomPrimvarsAPI` and treat `greeting` and `greeting:lang:fr_CA` as two distinct primvars.

Localization and `UsdShade` is an interesting case to consider as well.

dgovil:
@nvmkuruc That's a good question and something I forgot to mention in my PR docs (I had a line in the proposal discussion somewhere)

But yes, I think I wanted to have language on restricting the set of attributes one should localize. I wasn't sure how best to phrase that or whether the API should do validation itself around that.

I think longer term, my thought  would be that localization should only be used for things like strings (localized text), asset paths (localized textures?) and rels (localized animation bindings for lip-syncs?).  

However, even there, I think at the start my recommendation phrasing is that it should be limited to only strings at the start, and that it's only for strings that will be presented to a user as part of content consumption.

I think that recommendation would perhaps side-step the issue you mention since I don't know of a scenario where a primvar would fall under that. If that kind of wording would help, I think we could brainstorm how to phrase it here.

nvmkuruc:
I'm not sure that it sidesteps it because `string` and `asset` paths are commonly used as primvars and shader inputs for specifying or generating texture paths.

I'd expect the current implementation of `UsdGeomPrimarsAPI` and `UsdShadeShader` would return a unique input or primvar that had localized `string` or `asset` properties.
https://openusd.org/dev/api/class_usd_geom_primvars_a_p_i.html#a2336d8d750d52f1a8950abe5cfafb75a
https://openusd.org/dev/api/class_usd_shade_shader.html#aef8e1bf11a790edd7ffa7a90de4c5cb7

dgovil:
Hmm, I think that would be fine if they get returned in the list of properties.
Unless a UI is showing every single property on a prim, I don't believe it would cause any issues since the use would be targeted, so the existence of extra properties would be benign? Would you have a specific scenario where it might cause issues?

But also I think the initial recommendation of only limiting it to strings that are presented to users for consumption might avoid that for the time being

jesschimein:
Filed as internal issue #USD-10083

spiffmon:
Apologies for joining the party so late, but my _hope_ would be that the schemas' scoping to UsdUI would (self and explicit) document that we should only be using it for localization/accessibility of UI-facing data, and such localization should **not** have any cross-schema interaction (I can't decide whether UsdUI is actually "lower level" than, e.g. UsdGeom, but it doesn't feel right for UI concerns to affect scene evaluation for rendering).

@nvmkuruc , is it reasonable to say that asset rendering localization is a different problem than Accessibility, or are we really trying to get both of those birds in the same sack?

dgovil:
@spiffmon I believe I'm advocating for the same thing as you. I think (at least to start) it should only be used for strings that are presented to a user in a consumable way. Ie not generic attributes that are shown in an editor UI, but strings that themselves are shown as UI (such as the Autodesk text proposal) or expressed to the UI like accessibility over. 

I do think that people might use it to express other localizations in the future, which is why I mention the other possibilities. However I have language in the doc to say that any uses of localization are dependent on knowing what the endpoints support. 

I can strengthen the wording to emphasize the recommendation to only use it for user facing strings etc...

spiffmon:
From my POV, it would be good to keep the scoping tighter for now... there's just some meaty problems to talk through and solve if we start allowing it to be used for localizing texture names, _especially_ in a procedural way, and/or through the primvar-propagation mechanism.

dgovil:
Yep, agreed. I'll try and come up with suitable language for the doc to strengthen that recommendation

nvmkuruc:
> Yep, agreed. I'll try and come up with suitable language for the doc to strengthen that recommendation

I think one of the things that has been throwing me is that the example in the PR description shows `string foo` as an attribute that can be localized.  That implied to me that "any attribute" (or "any string attribute") can be localized and why I started to wonder about the implications for primvars, shader inputs, etc.

spiffmon:
I agree that's a concern, @nvmkuruc , and the only way in USD-of-today to make the limited scope obvious in the data itself instead of documentation would be to require a namespace for all localizable strings.

In the future, if we are copacetic with core USD behaviors relying on OpenExec, I _believe_ it should be possible for the "localization behavior" (which means tracking dependencies between attributes to get proper dirtying, as well as evaluation behavior for which attribute should be consulted) to be injected where it is needed, given an applied schema to impart it.  I am not 100% sure of that and haven't had time to talk to exec folks about it yet.  But if we prescribe a namespace now, it cuts off that possibility.  If we leave it in the not-completely-satisfying-or-validatable state of "limitations by fiat/documentation", then we leave the door open. 

Would it be worth taking some time to see how far we _can_ go with validation without generating false negatives?

dgovil:
I'm thinking a reasonable check right now might be:


- Is a string attribute
- Is not a primvar or an input/output

I think that would cover most of the concerns?

dgovil:
Just pinging this thread again, do the changes in my previous message sound reasonable? I can update the PR accordingly?

dgovil:
@nvmkuruc and @spiffmon I added validation functions and parameters in the code to check if something can be localized. 

CanLocalize(attribute) -> Checks if its a string attribute and if its not a primvar
CanLocalize(relationship) -> Returns false always but I figured it was nice to just have the API symmetry

All the creation functions now have a `validate` method that checks first if they can actually localize this, but I allow this to be skipped with a documentation note that this requires caution when doing so.

Let me know if that sounds better to you?

spiffmon:
Hi @dgovil , `Can*()` and `Validate`methods sound great.  

But also, upon another read, I did not think it was possible, but in any case not desirable, to have  a multiple-apply schema that elides the instance-name from its builtin properties, and I think we **do** want to canonize the Accessibility properties.  So I'd like to push back on eliding the "default" instance name from the "base" prim properties.  We could still have a helper `ApplyDefaultAPI()` method to canonize the default without programs needing to supply the literal?

dgovil:
Okay that's fair. I updated it so it keeps the `default` schema name in the attribute name.
Regarding making the default easier to make, I just set the default token name in the header to `default_`. So you can create the API without naming it.

One downside is that the UsdGenSchema overwrites that default when run. I wonder if maybe having it understand a default instance name would be a good addition to the UsdGenSchema as well?

dgovil:
Also changed the localization attribute for default language to include the instance name too, but I disallow creation from other instance names than `default`, and fetching the attribute will always get it from the `default` instance.

spiffmon:
> One downside is that the UsdGenSchema overwrites that default when run. I wonder if maybe having it understand a default instance name would be a good addition to the UsdGenSchema as well?

That could be something to consider for the future, but for now, you just can't, and the change will fail our internal automation if usdGenSchema gets a different result.  So, similarly to [UsdShade::CreateMaterialBindSubset()|https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdShade/materialBindingAPI.h#L924], can we stick to a custom method that we can document as the primary entrypoint?

dgovil:
@spiffmon if you're re-running usdGenSchema in your automation, how are you getting around needing extra headers in the cpp file? e.g I need `primvars.h`to check if its a primvar, but if I include it up top where all the other headers belong, it gets overwritten when running it again.

Are you putting the headers lower down in the file after the point where Jinja doesn't replace content in that case?

dgovil:
Anyway, made the change by adding `CreateDefaultAPI` and `ApplyDefaultAPI` methods on both schemas.
So now re-running the usdGenSchema doesn't cause any changes.

I haven't added any Python bindings for the custom methods yet. I can do that once I get a 👍 on the C++ API additions.

spiffmon:
See [Customizing Per-Class Properties](https://openusd.org/dev/api/_usd__page__generating_schemas.html#Usd_PerClassProperties) for exactly that use-case (which appears in UsdGeom schema.usda) - spolier, it's `customData[extraIncludes]`

dgovil:
Ah that's good to know. I changed that over.
I'll probably make a separate PR down the road for the schema gen to allow for a section in the `.cpp` file where includes can be preserved since the extraIncludes puts it in the header itself (which is fine for this, but I imagine other cases where that might be undesirable)

spiffmon:
Are there cases where it's insufficient to just put includes in **custom** section of the cpp - that's what we generally do.  Unless the header actually needed types defined in an external header, I don't think it's possible for the boilerplate section of the cpp to incur any other dependencies.

dgovil:
Not so much insufficient as it feels odd to have includes not at the top of
the file? I know USD doesn’t have a style guide but it feels like the
convention elsewhere is included at the top.

Anyway I digress from the subject of the PR

dgovil:
@nvmkuruc and @spiffmon , I just wanted to check and see if you had any more thoughts on the schema and API as presented thus far? I can rebase on dev, and add the Python bindings if you're okay with what's in here thus far

pxr/usd/usdUI/schema.usda
```diff
@@ -183,3 +188,155 @@ class Backdrop "Backdrop" (
...
+        in it its place. A description is an optional attribute, and some accessibility systems may only use the label.
+
+        Descriptions may be time varying for runtimes that support it. For example, you may describe what a character
+        is doing at a given time.
+        """
+    )
+
+    uniform token priority = "standard" (
+        doc = """A hint to the accessibility runtime of how to prioritize this instances label and description,
+        relative to others. This attribute is optional. Accessibility runtimes may choose to apply their own
```
nvmkuruc: Hi @dgovil.  I wanted to get some clarity on what you mean by runtimes may choose to apply their own priority systems that override this.

Are you saying that other systems can choose their own set of tokens instead of `low` / `standard` / `high`?  Or that there may be additional runtime specific attributes or configurations that may further refine the baseline priority?

If it's the former, I think I'd discourage that as it complicates validation.

pxr/usd/usdUI/schema.usda
```diff
@@ -183,3 +188,83 @@ class Backdrop "Backdrop" (
...
+            }
+            dictionary priority = {
+                string value = "priority"
+                string doc = """Name of the priority attribute"""
+            }
+        }
+    }
+) {
+    string label (
+        doc = """A short label to concisely describe the prim.
```
nvmkuruc, line 247: Any guidance on "shortness" or other constraints (like avoiding new lines) that you'd recommend for label?

Asking from the perspective of validation.

dgovil:
Per the request, I've split up the PR so this one focuses just on Accessibility. I'll put up the Localization schema later today.

nvmkuruc (commented):
Thanks Dhruv!

pxr/usd/usdUI/schema.usda
dgovil: I mean that its simply a hint, and runtimes don't have to strictly stick with everything being `high` being higher priority than everything `low` if some other runtime context would otherwise prioritize the `low` state.

So they can't add their own tokens. I'll add some verbiage accordingly.

pxr/usd/usdUI/schema.usda
dgovil: Unfortunately I don't believe there's any standard length.
It's just meant to be shorter than the description.

dgovil:
Would you guys have an example of wrapping method overloads for Python? I have two overloads for `CreateDefaultAPI`. One for each initializer: 

1. `static UsdUIAccessibilityAPI CreateDefaultAPI(const UsdPrim& prim);`
2. `static UsdUIAccessibilityAPI CreateDefaultAPI(const UsdSchemaBase& schemaObj);`

 Boost Python says `RuntimeError: Boost.Python - All overloads must be exported before calling 'class_<...>("AccessibilityAPI").staticmethod("CreateDefaultAPI")'`

In the meantime, I've only exposed the prim version of it to Python, because I couldn't find an incantation that allowed for both with the same name, and wasn't sure what the USDism is here

nvmkuruc:
> Would you guys have an example of wrapping method overloads for Python? I have two overloads for `CreateDefaultAPI`. One for each initializer:
> 
> 1. `static UsdUIAccessibilityAPI CreateDefaultAPI(const UsdPrim& prim);`
> 2. `static UsdUIAccessibilityAPI CreateDefaultAPI(const UsdSchemaBase& schemaObj);`
> 
> Boost Python says `RuntimeError: Boost.Python - All overloads must be exported before calling 'class_<...>("AccessibilityAPI").staticmethod("CreateDefaultAPI")'`
> 
> In the meantime, I've only exposed the prim version of it to Python, because I couldn't find an incantation that allowed for both with the same name, and wasn't sure what the USDism is here

Here's an example from `UsdSemantics` in case it helps.  https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdSemantics/wrapLabelsAPI.cpp#L176

dgovil:
Split out the Localization PR here https://github.com/PixarAnimationStudios/OpenUSD/pull/3475

dgovil:
@nvmkuruc I don't see an overload of either custom method. Apologies if I'm missing it though.

nvmkuruc:
Apologies, I misunderstood the issue.  How many times are you calling `.staticmethod`?  Once or once per overload?

dgovil:
No worries! I tried once per overload, and just the once for one of the overloads, but both seem to make it unhappy. :-/

dgovil:
Awesome, thank you for merging this. It is and will be much appreciated!

PR #3278, opened 2024-09-06
cmake: add some missed source files

rpoyner-tri:
### Description of Change(s)

Add some previously missing source files to the explicit lists of source files in various CMakeLists.txt.

### Fixes Issue(s)

This probably doesn't fix a visible issue in OpenUSD itself. It does help with downstream libraries that integrate portions of OpenUSD as source. In the case of [Drake](https://drake.mit.edu), its build system relies on file lists from OpenUSD's CMakeLists.txt files.


jesschimein:
Filed as internal issue #USD-10093

Issue #3279, opened 2024-09-06
Improved control of symbol visibility 


rpoyner-tri:
### Description of Issue

[Drake](https://drake.mit.edu) has begun incorporating portions of OpenUSD, via building from source code. One Drake-specific requirement (that other downstream library developers may have) is to minimize the symbols that are visible from its shared libraries.

Drake uses  two techniques to control symbols:
1. compiler symbol visibility attributes (https://gcc.gnu.org/wiki/Visibility)
2. rewriting namespace names to contain the string 'drake_vendor'

Technique 1 has the limitation that it sometimes interferes with c++ in surprising ways (e.g https://github.com/PixarAnimationStudios/OpenUSD/issues/1475). It can work well for plain C code.

Happily, OpenUSD has a very nice system to let downstream builders rewrite its namespaces. However, the Drake integration experience so far has uncovered a few holes in the namespace scheme. I hope to help address this by offering a few small PRs, with reference back to this issue.

### Steps to Reproduce

It's easiest to give an example from OpenUSD's own shared library output. These instructions are for Ubuntu Jammy.

1. `$cd path-to/OpenUSD`
1. `$ git checkout dev`
1. `$ python3 ./build_scripts/build_usd.py --no-python ~/tmp/usd/dist`
2. `$ cd ~/tmp/usd/dist`
3.  `$ readelf --wide --symbols --demangle ./lib/libusd_work.so | grep 'GLOBAL *DEFAULT' |egrep -v 'pxrInternal|GLIBC|GCC|CXXABI|tbb::'`
4. In the output from above, the data section symbol `PXR_WORK_THREAD_LIMIT` is GLOBAL and not namespaced.
5. Repeat the above on libusd_gf.so, libusd_pegtl.so, etc., to see other cases.

### System Information (OS, Hardware)

Ubuntu LTS versions for x84_64, macOS recent versions for arm64. See the Supported Configurations table at https://drake.mit.edu/installation.html .

### Package Versions

The problems vary over time, but several were seen as of v24.08. Current dev has fewer problems already.

jesschimein:
Filed as internal issue #USD-10094

sunyab:
The PRs associated with this issue have all been merged, so I'm going to close this out. Please file a new issue if this problem is still present even after these changes. Thank you!

PR #3280, opened 2024-09-06
work: move work thread limit within namespace

rpoyner-tri:
### Description of Change(s)

Puts the PXR_WORK_THREAD_LIMIT symbol within the  PXR_NAMESPACE.  

Other TF_DEFINE_ENV_SETTING() invocations are within the namespace; this is the only one I've found that is not.

### Fixes Issue(s)
Partial fix for #3279.

### DIFF
--- a/pxr/base/work/threadLimits.cpp
+++ b/pxr/base/work/threadLimits.cpp
@@ -29,6 +29,8 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+PXR_NAMESPACE_OPEN_SCOPE
+
 // The environment variable used to limit the number of threads the application
 // may spawn:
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10095

nvmkuruc:
This seems like simple and helpful fix.

@pmolodo FYI-- if this gets accepted, it may affect #3370

pmolodo:
> This seems like simple and helpful fix.
> 
> @pmolodo FYI-- if this gets accepted, it may affect #3370

Actually, released some changes to #3370 so that it won't be affected - PXR_WORK_THREAD_LIMIT  isn't exported as part of the interface, so I had to retrieve the env var by name (via `Tf_GetEnvSettingByName`, which despite the `Tf_` prefix, is exported).

PR #3282, opened 2024-09-06
pegtl: fix namespace problems

rpoyner-tri:
### Description of Change(s)

- Fix preprocessor-time composition of PXR_PEGTL_NAMESPACE
-- the prior solution failed to do name expansion, always yielding a not-configurable `PXR_INTERNAL_NS_pegtl`
- Put the "workaround" symbol into the namespace

### Fixes Issue(s)
Partial fix for #3279.

### DIFF
--- a/pxr/base/pegtl/README.md
+++ b/pxr/base/pegtl/README.md
@@ -39,7 +39,9 @@ This file defines the namespace name that PEGTL uses:
 #include "pxr/pxr.h"
 
 #if PXR_USE_NAMESPACES
-#define PXR_PEGTL_NAMESPACE PXR_INTERNAL_NS ## _pegtl
+#define PXR_PEGTL_impl_PASTE2(x, y) x ## y
+#define PXR_PEGTL_impl_PASTE(x, y) PXR_PEGTL_impl_PASTE2(x, y)
+#define PXR_PEGTL_NAMESPACE PXR_PEGTL_impl_PASTE(PXR_INTERNAL_NS, _pegtl)
 #else
 #define PXR_PEGTL_NAMESPACE pxr_pegtl
 #endif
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-10097

pxr/base/pegtl/README.md
```diff
@@ -39,7 +39,9 @@ This file defines the namespace name that PEGTL uses:
 #include "pxr/pxr.h"
 
 #if PXR_USE_NAMESPACES
-#define PXR_PEGTL_NAMESPACE PXR_INTERNAL_NS ## _pegtl
+#define PXR_PEGTL_impl_PASTE2(x, y) x ## y
```
nvmkuruc, line 42: Could `TF_PP_CAT` be used instead?

cpichard:
Hi, we found that this incorrect namespacing causes symbols duplication and lead to crashes in a configuration using different USD versions static and shared with the symbols namespaced. 
It would be great to have this bug fixed soon.
@nvmkuruc should I create a new PR with TF_PP_CAT ? 
Looking at pxr/base/tf/preprocessorUtilsLite.h, it seems to bring a lot of defines additional to TF_PP_CAT, should we be worried about the extra compile time including this file in a header if we just use TF_PP_CAT ? 
Thanks,
Cyril

nvmkuruc:
> Hi, we found that this incorrect namespacing causes symbols duplication and lead to crashes in a configuration using different USD versions static and shared with the symbols namespaced. It would be great to have this bug fixed soon. @nvmkuruc should I create a new PR with TF_PP_CAT ? Looking at pxr/base/tf/preprocessorUtilsLite.h, it seems to bring a lot of defines additional to TF_PP_CAT, should we be worried about the extra compile time including this file in a header if we just use TF_PP_CAT ? Thanks, Cyril

Not wanting to make `pxr/base/pegtl` dependent on `pxr/base/tf` makes sense to me. I wouldn't let that question hold anything up.

cpichard:
Sounds good, thanks !

PR #3287, opened 2024-09-11
Added AVIF to allowed USDZ image file types

AlexSchwank:
### Description of Change(s)

This change is adding AV1 Image File Format (AVIF) to the list of allowed image file types in USDZ.

The AV1 Image Format is a royalty-free open-source picture format, with modern compression, flexible color specification, high dynamic range values, depth images and alpha channels, and support for layered and sequential images. 

HioAvif was introduced in OpenUSD 24.08 ([commit](https://github.com/PixarAnimationStudios/OpenUSD/commit/1cba8a9a1ce170eb61400175d7d3a3b7a0641a2b)) which enabled reading of `.avif` image files in Hio. Thanks to @meshula for his work.

### Fixes Issue(s)
- 

### DIFF
--- a/docs/user_guides/render_user_guide.rst
+++ b/docs/user_guides/render_user_guide.rst
@@ -1138,11 +1138,29 @@ compression is applied. It is expected that more complex treatment of OpenEXR
 files including the construction of multilayer files will be completed by 
 pipeline tools.
 
+AV1 Image File Format (AVIF)
+============================
+
+The AV1 Image Format is a royalty-free open-source picture format, with modern 
+compression, flexible color specification, high dynamic range values, depth 
+images and alpha channels, and support for layered and sequential images. 
+
+The supported feature set in Hydra's builtin texture manager is currently 
+restricted to single frame images, which are decoded to linear Rec709 RGB or 
+RGBA if an alpha channel is present.
+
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-10109

docs/user_guides/render_user_guide.rst
```diff
@@ -1138,11 +1138,29 @@ compression is applied. It is expected that more complex treatment of OpenEXR
...
+AV1 Image File Format (AVIF)
+============================
+
+The AV1 Image Format is a royalty-free open-source picture format, with modern 
+compression, flexible color specification, high dynamic range values, depth 
+images and alpha channels, and support for layered and sequential images. 
+
+The supported feature set in Hio is currently restricted to single frame 
+images, which are decoded to linear Rec709 RGB or RGBA if an alpha channel is 
+present.
```
spiffmon: This is end-user doc, so rather than call out Hio, can we say "Hydra's builtin texture manager" or something like that?

spiffmon (commented):
Just one note from me, @AlexSchwank - thank you!

docs/user_guides/render_user_guide.rst
AlexSchwank: Done.

PR #3288, opened 2024-09-11
Ensure HdSceneIndexAdapterSceneDelegate::GetMaterialResource considers all contexts

### LINKED ISSUE #3286: HdSceneIndexAdapterSceneDelegate::GetMaterialResource() returns wrong network in certain situations
_The specific alignment of the stars is as follows:_

* `MyRenderDelegate::GetMaterialRenderContexts()` returns `["foo", "bar"]`
* USD Material has `outputs:bar:surface` and `outputs:surface`

_The issue:_

I would expect that `HdSceneIndexAdapterSceneDelegate::GetMaterialResource()` would return the `"bar"` network but, instead, it returns the `""` network.

_The cause (as best I can tell):_

When we call `GetMaterialResource()`, it iterates through our delegate's list of supported contexts, starting with `"foo"`, and calls `HdMaterialSchema::GetMaterialNetwork("foo")`. This context doesn't exist in the USD file, so `GetMaterialNetwork()` instead attempts to return a network for the `""` context which, in this case, *does* exits. `GetMaterialResource()` recognises that a valid network has been returned and immediately sends it back to us.

robpieke:
### Description of Change(s)

Moving the responsibility for iterating over the render delegate's material contexts out of the scene delegate and into the material schema, to ensure the fallback network is only returned after *all* the delegate's contexts have been considered (and failed).

### Fixes Issue(s)

Fixes #3286 

### DIFF
--- a/pxr/imaging/hd/materialSchema.cpp
+++ b/pxr/imaging/hd/materialSchema.cpp
@@ -41,10 +41,12 @@ HdMaterialSchema::GetMaterialNetwork()
 }
 
 HdMaterialNetworkSchema
-HdMaterialSchema::GetMaterialNetwork(TfToken const &context)
+HdMaterialSchema::GetMaterialNetwork(TfTokenVector const &contexts)
 {
-    if (auto b = _GetTypedDataSource<HdContainerDataSource>(context)) {
-        return HdMaterialNetworkSchema(b);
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-10110

PR #3291, opened 2024-09-14
Autodesk: Fix Security Issue on Linux build for LZ4

erikaharrison-adsk:
### Description of Change(s)

Upgrades LZ4 version to 1.10.0 to fix security issue on lz4-1.9.2

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10126

PR #3292, opened 2024-09-15
Fix TfMallocTag.Tls.Find static method on Darwin.

furby-tm:
The original reasoning for disabling this was because originally **`thread_local`** was not available in Apple's (now ancient) Clang **pre-Xcode 8**, however now all modern versions of Apple's Clang do have support for **`thread_local`**.

### Description of Change(s)

- The static method for **`TfMallocTag::Tls::Find()`** now returns the per-thread data for **`TfMallocTag`** on all **Darwin** platforms, which before this change, has always returned nothing (and was silently failing).
- Remove the **`ARCH_HAS_THREAD_LOCAL`** preprocessor.
- Remove any/all conditional compilation checks for **`ARCH_HAS_THREAD_LOCAL`**.

<hr/>

### Note

I had noticed the following two **Tf** test cases are failing, but this is unrelated to this revision, as they were failing before this revision was introduced.

```swift
99% tests passed, 2 tests failed out of 791

Total Test time (real) = 1495.16 sec

The following tests FAILED:
	 15 - TfAtomicOfstreamWrapper (Failed)
	 56 - TfSafeOutputFile (Failed)
```

<hr/>

### Fixes Issue(s)

- #3323

### DIFF
--- a/pxr/base/arch/defines.h
+++ b/pxr/base/arch/defines.h
@@ -79,12 +79,6 @@
 #define ARCH_HAS_GNU_STL_EXTENSIONS
 #endif
 
-// The current version of Apple clang does not support the thread_local
-// keyword.
-#if !(defined(ARCH_OS_DARWIN) && defined(ARCH_COMPILER_CLANG))
-#define ARCH_HAS_THREAD_LOCAL
-#endif
-
 // The MAP_POPULATE flag for mmap calls only exists on Linux platforms.
 #if defined(ARCH_OS_LINUX)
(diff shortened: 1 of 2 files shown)


furby-tm:
(cc. @dgovil, @AlexSchwank), I was not able to locate Nicholas Blasingame's GitHub username, but feel free to ping him here to sync up, thanks!

jesschimein:
Filed as internal issue #USD-10127

dgovil:
Another one that we talked about on slack but I forgot to respond here. This looks like a good change to me. Thanks for solving that.

PR #3293, opened 2024-09-16
Simplify grammar when parsing large strings

nvmkuruc:
### Description of Change(s)

A performance regression was observed when parsing large strings after the merging of #3005.

Parsing time for the schema registry layers (which has large documentation strings) went from ~4-5ms to over 7ms.

This was traced back to the `String` parsing grammar.  One notable difference from the PEGTL implementation and the original implementation was that even though it permissively let any character be "escaped" by `\`, it was explicit in the grammar about those that would be considered valid for replacement by `TfEscapeStringReplaceChar`. This was unnecessary and appeared to slow down the parser. (Were the grammar ever to become less permissive, it also had a bug-- as `OctDigit` was incorrectly specified.)  These unnecessary checks and rules have now been removed.

This change includes some additional optimizations and simplifications to the grammar.
* Preferring `PEGTL_NS::string` (and the wrappers `PEGTL_NS::two` and `PEGTL_NS::three`) over sequences of individual `PEGTL_NS::one` characters.
* Replacing `opt_must` with `opt` in the number parsing grammar.  An early version of the PEGTL grammar had required `plus<digit>.` to be followed by another digit (likely via `plus<digit>`).  Once it became optional (ie. `plus` became `star`), the `must` was spurious.
* Some complexity was originally added to the comment parsing to minimize backtracking when matching comments, but we suspect the `PEGTL:_NS::string` matching will be efficient and simpler.

```
env PXR_ENABLE_GLOBAL_TRACE=1 PXR_WORK_THREAD_LIMIT=1 python3 -c "from pxr import Usd; Usd.SchemaRegistry()"
```
```
    4.410 ms     4.410 ms      14 samples    |   | pxrInternal_v0_24_11__pxrReserved__::Sdf_ParseLayer
```

### Fixes Issue(s)
-

### DIFF
--- a/pxr/usd/sdf/textFileFormatParser.h
+++ b/pxr/usd/sdf/textFileFormatParser.h
@@ -70,26 +70,10 @@ struct Ampersand : PEGTL_NS::one<'&'> {};
 
 // character classes
 struct Digit : PEGTL_NS::digit {};
-struct HexDigit : PEGTL_NS::xdigit {};
-struct OctDigit : PEGTL_NS::range<0, 7> {};
 struct Eol : PEGTL_NS::one<'\r', '\n'> {};
 struct Eolf : PEGTL_NS::eolf {};
 struct Utf8 : PEGTL_NS::utf8::any {};
 struct Utf8NoEolf : PEGTL_NS::minus<Utf8, Eol> {};
-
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10128

PR #3294, opened 2024-09-16
Reduce copies of string values during text parsing

nvmkuruc:
### Description of Change(s)

While investigating the parser performance for #3293, we also looked at reducing the number of copies of large documentation strings observed in the schema registry.

With this change, `string_view` is used in many cases instead of creating a copy and `std::move` is used instead of copying the evaluated string.

### Fixes Issue(s)
-


jesschimein:
Filed as internal issue #USD-10129

PR #3296, opened 2024-09-17
Autodesk: Add per RenderPass shadow control

erikaharrison-adsk:
### Description of Change(s)

Add a per RenderPass level shadowing control so that we can decide which RenderPasses receive shadows and which do not.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10142

PR #3299, opened 2024-09-18
Adding color4f(array) support to the _FactoryMap

### LINKED ISSUE #3298: UsdImagingDataSourceAttributeNew doesn't support color4
In `dataSourceAttribute.cpp` the `_FactoryMap` has entries for `Color3f` and `Color3fArray` but not `Color4f` or `Color4fArray`. The result of this is that any shading node which has a `color4f` parameter will *not* have a `value` generated.

For example:
```
        def Shader "mtlxover1"
        {
            uniform token info:id = "ND_over_color4"
            color4f inputs:bg = (0.9, 0.1, 0.6, 1)
            color4f inputs:fg.connect = </materials/collect1/mtlxtiledimage1.outputs:out>
            color3f outputs:out
        }
```

In this case, the `bg` parameter ends up generating no `value`.

robpieke:
### Description of Change(s)

Copy-paste to add Color4f/Color4fArray to the _FactoryMap of convertible types to support shader nodes with color4f parameters.

### Fixes Issue(s)
Fixes #3298 


jesschimein:
Filed as internal issue #USD-10150

PR #3300, opened 2024-09-18
UsdView: Display resolved labels in attribute editor

nvaszabo:
### Description of Change(s)
This PR uses the `usdSemantics` schema and API to compute the taxonomies and the associated labels and display them in `UsdView's` attribute editor under a new row named `Resolved Labels`.

The `Resolved Labels` for a selected prim include the labels on the prim itself and those on all its ancestors (inherited labels).
For each taxonomy, the unique set of labels is computed and sorted lexicographically by the 'UsdSemantics.LabelsQuery.ComputeUniqueInheritedLabels()` API.

#### Performance considerations
The computation of direct and indirect taxonomies and their labels is instantaneous for prim hierarchies that are ~30 items deep with ~10 taxonomies and ~10 labels each. The same seems to lag just very slightly with hierarchies that are ~30 items deep and with a random number of maximum of 100 taxonomies, each with a random number of maximum of of 100 labels.

image

### Fixes Issue(s)
-


pxr/usdImaging/usdviewq/customAttributes.py
```diff
@@ -5,7 +5,8 @@
 # https://openusd.org/license.
 #
 
-from pxr import Usd, UsdGeom, UsdShade
+from collections import defaultdict
```
nvmkuruc: This seems unused.

jesschimein:
Filed as internal issue #USD-10151

nvaszabo:
@jesschimein

nvaszabo:
[bookshelf.zip](https://github.com/user-attachments/files/17603496/bookshelf.zip)
An example of using UsdSemantics schema.

jmk:
> [bookshelf.zip](https://github.com/user-attachments/files/17603496/bookshelf.zip) An example of using UsdSemantics schema.

Thanks for the example file, it was helpful to validate the change on our end. We've completed our review and this should be merged shortly.

Issue #2961, opened 2024-02-20
[usdAbc] UsdAbcAlembicFileFormat does not use ArAsset API for reading assets


nvidia-jomiller:
### Description of Issue

We are unable to load any Alembic `SdfAssetPath`s that resolve to a non file-based path such as a URL / URI. The internal implementation of [UsdAbcAlembicFileFormat::Read](https://github.com/PixarAnimationStudios/OpenUSD/blob/v23.11/pxr/usd/plugin/usdAbc/alembicFileFormat.cpp#L99) always assumes that the provided `resolvedPath` argument is a normal file path that can be fed directly to Alembic's `IArchive`. If `resolvedPath` is a URL Alembic will fail with an [Unsupported format: 'Unknown'](https://github.com/PixarAnimationStudios/OpenUSD/blob/v23.11/pxr/usd/plugin/usdAbc/alembicReader.cpp#L946) error since the `UsdAbcAlembicFileFormat` has not converted, or mapped, `resolvedPath` to something that Alembic's `IArchive` can load. Ideally, `UsdAbcAlembicFileFormat` should use `ArGetResolver().OpenAsset()` with the `ArAsset` API to make sure that the underlying ArAssetResolver plugin is invoked when trying to load Alembic assets. 

### Steps to Reproduce

1. Download [repro.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14380660/repro.zip)
2. Extract [repro.zip](https://github.com/PixarAnimationStudios/OpenUSD/files/14380660/repro.zip) to `/some/path/repro`
3. Navigate to `/some/path/repro`
4. Run `usdview Scene.usdz`
5. The following warning / error should appear:
```
Warning: in _ReportErrors at line 3170 of /home/jomiller/projects/usd-builds/dev/USD/pxr/usd/usd/stage.cpp -- In </World/Scene>: Could not open asset @/home/jomiller/repros/Scene.usdz[Assets/testAbcCamera.abc]@ for reference introduced by @Scene.usdz@</World/Scene> -- Failed to open Alembic archive "/home/jomiller/repros/Scene.usdz[Assets/testAbcCamera.abc]": Unsupported format: 'Unknown'. (instantiating stage on stage @Scene.usdz@ <0x3244120>)
```

### System Information (OS, Hardware)
Windows 11 Enterprise
Processor	AMD Ryzen Threadripper PRO 3975WX 32-Cores 3.50 GHz
[...]

PR #3302, opened 2024-09-19
Improve Alembic plugin to read assets with ArResolver to support other sources except local files

roggiezhang-nv:
### Description of Change(s)

Alembic plugin only accepts assets that are from local paths. In order to fix this, we need to use ArResolver/ArAsset to read/write contents so customized resolver can read assets from other locations. However, interfaces of Alembic library have limitations that can only accept local urls or `std::istream`. This PR uses the following solution:

  Open assets with `ArResolver`. Instead of reading them into the memory, we get the mapped file path with [ArchGetFileName](https://github.com/PixarAnimationStudios/OpenUSD/blob/59992d2178afcebd89273759f2bddfe730e59aa8/pxr/base/arch/fileSystem.h#L177) from `ArAsset::GetFileUnsafe`. 

### Fixes Issue(s)
- https://github.com/PixarAnimationStudios/OpenUSD/issues/2961

### DIFF
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -860,6 +860,8 @@ private:
     _PrimMap _prims;
     Prim* _pseudoRoot;
     UsdAbc_TimeSamples _allTimeSamples;
+
+    std::vector<std::shared_ptr<ArAsset>> _assetHolders;
 };
 
 static
@@ -895,6 +897,7 @@ _ReaderContext::_OpenAndGetMappedFilePath(const std::string& filePath)
             {
                 // If file handle is presented, use mapped file path instead of original one.
                 const std::string mappedFilePath = ArchGetFileName(fileHandle);
+                _assetHolders.emplace_back(std::move(asset));
 
                 return mappedFilePath;
             }
@@ -1540,6 +1543,7 @@ _ReaderContext::_Clear()
     _allTimeSamples.clear();
     _instanceSources.clear();
     _instances.clear();
+    _assetHolders.clear();
 }
 
 const _ReaderContext::Prim*
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -23,6 +23,7 @@
 #include "pxr/base/tf/staticData.h"
 #include "pxr/base/tf/staticTokens.h"
 #include "pxr/base/tf/ostreamMethods.h"
+#include "pxr/base/tf/fileUtils.h"
 #include <Alembic/Abc/ArchiveInfo.h>
 #include <Alembic/Abc/IArchive.h>
 #include <Alembic/Abc/IObject.h>
@@ -881,18 +882,22 @@ _ReaderContext::_ReaderContext() :
 std::string
 _ReaderContext::_OpenAndGetMappedFilePath(const std::string& filePath)
 {
-    // Open asset with Ar to support URLs other than local paths.
-    std::shared_ptr<ArAsset> asset =
-        ArGetResolver().OpenAsset(ArResolvedPath(filePath));
-    if (asset)
+    // Return as it is if it's a local path and not a symlink.
+    if (!TfIsFile(filePath))
     {
-        FILE* fileHandle = asset->GetFileUnsafe().first;
-        if (fileHandle)
+        // Open asset with Ar to support URLs other than local paths.
+        std::shared_ptr<ArAsset> asset =
+            ArGetResolver().OpenAsset(ArResolvedPath(filePath));
+        if (asset)
         {
-            // If file handle is presented, use mapped file path instead of original one.
-            const std::string mappedFilePath = ArchGetFileName(fileHandle);
+            FILE* fileHandle = asset->GetFileUnsafe().first;
+            if (fileHandle)
+            {
+                // If file handle is presented, use mapped file path instead of original one.
+                const std::string mappedFilePath = ArchGetFileName(fileHandle);
 
-            return mappedFilePath;
+                return mappedFilePath;
+            }
         }
     }
 
--- a/pxr/usd/plugin/usdAbc/alembicReader.cpp
+++ b/pxr/usd/plugin/usdAbc/alembicReader.cpp
@@ -861,6 +861,8 @@ private:
     Prim* _pseudoRoot;
     UsdAbc_TimeSamples _allTimeSamples;
 
+    // Asset holders to keep a reference alive so that resolver doesn't 
+    // attempt to cleanup asset until the asset is no longer in use.
     std::vector<std::shared_ptr<ArAsset>> _assetHolders;
 };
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10155

anwang2009:
Hi @roggiezhang-nv, thanks for opening the discussion with an implementation! We took a step back and looked at the current design and came up with the following concerns:
- _ReaderContext::_OpenAndGetMappedFilePath adds two file handle opens per path, even if the file paths are already supported in the original implementation. This adds bloat in a commonly used code path.
- > However, interfaces of Alembic library have limitations that only accept local paths or std::istreams ... Instead of reading them into the memory, we get the mapped file path with [ArchGetFileName](https://github.com/PixarAnimationStudios/OpenUSD/blob/59992d2178afcebd89273759f2bddfe730e59aa8/pxr/base/arch/fileSystem.h#L177)

  I want to clarify what is meant by "mapped file path". Could you give an example where having a mapped file path is necessary in linux, macos, and Windows? Is the mapping a separate process to the code path that has been changed here?
- Related to the above points I'm wondering if there's an easy way at runtime to distinguish between paths that "already work" vs paths that need "special treatment"

We're considering two other designs as well: 
1. For your immediate needs, if there is a custom resolver you are using it's the least invasive to make the changes you need in your custom plugins. But if the desire is to solve this problem more generally:
2. We can consider producing istreams in the filepath cases that aren't covered by the original implementation. I see that you looked into this in #3361, and I'm curious what specific istream flaws you ran into? We know that istream is less performant and requires deeper code changes, so we would hope to invoke the istream case only as necessary. The primary upside would be implementation clarity.

:thinking: What do you think?

roggiezhang-nv:
Hi, An. Thanks for your reply. Here are my thoughts about your questions:

> I want to clarify what is meant by "mapped file path". Could you give an example where having a mapped file path is necessary in linux, macos, and Windows? Is the mapping a separate process to the code path that has been changed here?

I may not describe it clearly. So Alembic library only accepts local URLs or override std::istream. In order to read assets that are not local, we'd have to do that in two ways:
1. Cache remote assets to local drive, and get the local path (that's what we called "mapped file path", and that's what we have done in _OpenAndGetMappedFilePath). For default resolver in OpenUSD, it simply returns the original path since it's already local. We still need to mention that not all customized resolvers implement this kind of cache behavior. For those resolvers, we simply fail it as before in this PR. So the behavior in this PR can be summarized as:
   * For local files, it's the same behavior as before.
   * For customized resolver that caches file and ArAsset::GetFileUnsafe returns valid FILE descriptor, it can work now with this PR. (This is the main point of this PR)
   * For other resolvers, it simply fails as before.
3. Override std::istream to read contents into a istream. **However, this solution will break support for multi-layered ABC files.** That's why we chose to use std::istream implementation in the beginning and gave up finally.

> Related to the above points I'm wondering if there's an easy way at runtime to distinguish between paths that "already work" vs paths that need "special treatment"

We can simply check if the given urls exist so we can isolate the current work from the current PR. The only overhead for the current PR was in the `_OpenAndGetMappedFilePath` that opens the file and gets its mapped path, which still has no harm to performance. This step is not necessary for files that already exist in the local drive. This is why we thought it's the best solution currently.

anwang2009:
hi @roggiezhang-nv ! Are you planning any further changes before I start some performance testing and reviews?

roggiezhang-nv:
> hi @roggiezhang-nv ! Are you planning any further changes before I start some performance testing and reviews?

@anwang2009 I added one strict checker to avoid open and release local asset twice so we treat local files the same as before without any side-effects. You can start evaluating the perf and review now.

roggiezhang-nv:
@anwang2009 I added one more change per request to hold assets within the lifetime of _ReaderContext.

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -872,6 +881,33 @@ _ReaderContext::_ReaderContext() :
...
+        std::shared_ptr<ArAsset> asset =
+            ArGetResolver().OpenAsset(ArResolvedPath(filePath));
+        if (asset)
+        {
+            FILE* fileHandle = asset->GetFileUnsafe().first;
+            if (fileHandle)
+            {
+                // If file handle is presented, use mapped file path instead of original one.
+                const std::string mappedFilePath = ArchGetFileName(fileHandle);
+                _assetHolders.emplace_back(std::move(asset));
```
nvmkuruc, line 905: It might be helpful to explain that you're holding onto the assets to keep a reference alive so that resolver doesn't attempt to cleanup until the layer is no longer in use.

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -872,6 +881,33 @@ _ReaderContext::_ReaderContext() :
...
+    // Return as it is if it's a local path and not a symlink.
+    if (!TfIsFile(filePath))
+    {
+        // Open asset with Ar to support URLs other than local paths.
+        std::shared_ptr<ArAsset> asset =
+            ArGetResolver().OpenAsset(ArResolvedPath(filePath));
+        if (asset)
+        {
+            FILE* fileHandle = asset->GetFileUnsafe().first;
+            if (fileHandle)
```
anwang2009: Maybe also add the condition that the offset from GetFileUnsafe is not 0?

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -872,6 +881,33 @@ _ReaderContext::_ReaderContext() :
     // Do nothing
 }
 
+std::string
+_ReaderContext::_OpenAndGetMappedFilePath(const std::string& filePath)
+{
+    // Return as it is if it's a local path and not a symlink.
+    if (!TfIsFile(filePath))
```
anwang2009: Let's resolve symlinks in this TfIsFile call

anwang2009 (commented):
Thanks @roggiezhang-nv , lgtm besides some minor things!

roggiezhang-nv:
@anwang2009 All comments addressed. Thanks.

pxr/usd/plugin/usdAbc/alembicReader.cpp
```diff
@@ -872,6 +883,33 @@ _ReaderContext::_ReaderContext() :
...
+    // Return as it is if it's a local path or symlink.
+    if (!TfIsFile(filePath, true))
+    {
+        // Open asset with Ar to support URLs other than local paths.
+        std::shared_ptr<ArAsset> asset =
+            ArGetResolver().OpenAsset(ArResolvedPath(filePath));
+        if (asset)
+        {
+            FILE* fileHandle = asset->GetFileUnsafe().first;
+            if (fileHandle && ftell(fileHandle) == 0)
```
anwang2009: Can we use the offset that's returned by GetFileUnsafe().second instead of calling ftell?

It would be good to explain why the check against the file position is needed. I'm guessing it's to ensure that the .abc file we're looking at isn't embedded in a .usdz or other package, since the Alembic code itself won't know what to do with that.

- sunya

roggiezhang-nv: Done.

PR #3304, opened 2024-09-19
Autodesk: Fix RemoveInputScene

erikaharrison-adsk:
### Description of Change(s)

Move the initialization of variable removalTestQueue before deleting the iterator it was using.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10160

PR #3309, opened 2024-09-21
Adding python bindings for VtArray<SdfPathExpression> objects.

marktucker:
### Description of Change(s)
Adding python bindings for VtArray<SdfPathExpression> objects.

Without these additions, it is not possible to use python to get or set attributes or metadata of type Sdf.ValueTypeNames.PathExpressionArray. There is obviously no pressing need for this as there are no schemas (that I am aware of) that use this data type, but since this is a valid data type, it seemed to me that it should be possible to use it in python.

### DIFF
--- a/pxr/usd/sdf/pathExpression.cpp
+++ b/pxr/usd/sdf/pathExpression.cpp
@@ -39,6 +39,11 @@ TF_REGISTRY_FUNCTION(TfType)
     TfType::Define<VtArray<SdfPathExpression>>();
 }
 
+TF_REGISTRY_FUNCTION(VtValue)
+{
+    VtValue::RegisterSimpleCast<std::string, SdfPathExpression>();
+}
+
 SdfPathExpression::ExpressionReference const &
 SdfPathExpression::ExpressionReference::Weaker()
 {
--- a/pxr/usd/sdf/pathExpression.cpp
+++ b/pxr/usd/sdf/pathExpression.cpp
@@ -39,11 +39,6 @@ TF_REGISTRY_FUNCTION(TfType)
     TfType::Define<VtArray<SdfPathExpression>>();
 }
 
-TF_REGISTRY_FUNCTION(VtValue)
-{
-    VtValue::RegisterSimpleCast<std::string, SdfPathExpression>();
-}
-
 SdfPathExpression::ExpressionReference const &
 SdfPathExpression::ExpressionReference::Weaker()
 {
--- a/pxr/usd/sdf/testenv/testSdfPathExpressionArray.py
+++ b/pxr/usd/sdf/testenv/testSdfPathExpressionArray.py
@@ -1,0 +1,32 @@
+#!/pxrpythonsubst
+#
+# Copyright 2024 Pixar
+#
+# Licensed under the terms set forth in the LICENSE.txt file available at
+# https://openusd.org/license.
+
+from pxr import Sdf, Tf
+import sys, unittest
+
+MatchEval = Sdf._MakeBasicMatchEval
+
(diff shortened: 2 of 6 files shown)


jesschimein:
Filed as internal issue #USD-10177

pxr/usd/sdf/testenv/testSdfPathExpressionArray.py
```diff
@@ -0,0 +1,32 @@
+#!/pxrpythonsubst
+#
+# Copyright 2024 Pixar
```
sunyab: 2024 -> 2025

pxr/usd/sdf/wrapArrayPathExpression.cpp
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright 2016 Pixar
```
sunyab: 2016 -> 2025

pxr/usd/sdf/pathExpression.cpp
```diff
@@ -39,6 +39,11 @@ TF_REGISTRY_FUNCTION(TfType)
     TfType::Define<VtArray<SdfPathExpression>>();
 }
 
+TF_REGISTRY_FUNCTION(VtValue)
+{
+    VtValue::RegisterSimpleCast<std::string, SdfPathExpression>();
+}
+
```
sunyab: We'd prefer to disallow implicit casts from std::string to SdfPathExpression. Construction of an SdfPathExpression may raise a TfError if the input string doesn't parse, so we'd prefer the user be explicit about when they want to do the conversion and be aware of the potential error.

I believe that means we can remove this and the implicit_convertible registration I'll call out below.

pxr/usd/sdf/wrapPathExpression.cpp
```diff
@@ -199,6 +205,9 @@ void wrapPathExpression()
         .def(self == self)
         .def(self != self)
         ;
+
+    implicitly_convertible<std::string, SdfPathExpression>();
```
sunyab: See comment above. An additional reason to disallow the implicit conversion is that it can lead to surprising scene description being authored if the input string is invalid. For example:

```
>>> from pixar import Sdf
>>> l = Sdf.Layer.CreateAnonymous('.usda')
>>> p = Sdf.CreatePrimInLayer(l, '/foo')
>>> a = Sdf.AttributeSpec(p, 'a', Sdf.ValueTypeNames.PathExpression)
>>> a.default = '%#$@'
pixar.Tf.ErrorException: 
        Error in 'SdfPathExpression::SdfPathExpression' at line 69 in file pxr/usd/sdf/pathExpression.cpp : '<input>:1:1: parse error matching pxr_pegtl::pad<{anonymous}::PathExpr, pxr_pegtl::ascii::blank, pxr_pegtl::ascii::blank> -- <input>:1:1'

>>> print(l.ExportToString())
#usda 1.0

over "foo"
{
    pathExpression a = ""
}
```

sunyab (changes requested):
Hey @marktucker , left some notes for you, thanks!

pxr/usd/sdf/testenv/testSdfPathExpressionArray.py
marktucker: To be fair, it was 2024 when I submitted this PR :)

nvmkuruc:
Hi @sunyab and @marktucker, was this merged as 33a027070c197d68a5a52336a97261bc66c7c577?

jesschimein:
It was!  Thanks for catching that!  Closing this issue as fixed.

PR #3312, opened 2024-09-23
Optimize Sdf_PathNode::GetPathToken() by avoiding unnecessary memory allocation

petrihakkinen:
### Description of Change(s)

GetPathToken no longer does a memory allocation when the token already exists. A benchmark that calls SdfPath::GetString repeatedly on a large amount of SdfPaths shows 2.3x improvement on Windows.

Note that the following unit tests fail on Windows with or without the changes:

158 - testWorkThreadLimitsRawTBBMax (Failed)
767 - testUsdviewNavigationKeys (Failed)
783 - testUsdResolverExample (Failed)

### Fixes Issue(s)

None.

### DIFF
--- a/pxr/usd/sdf/pathNode.cpp
+++ b/pxr/usd/sdf/pathNode.cpp
@@ -438,6 +438,10 @@ namespace {
 // nullptr property node pointer) plus all the properties that hang off it.
 struct _PropToTokenTable
 {
+    void Init() {
+        _data = std::make_shared<_Data>();
+    }
+
     // Note that this function returns a TfToken lvalue reference that needs to
     // live/be valid as long as this object is around.
     template <class Fn>
@@ -471,7 +475,7 @@ private:
         tbb::spin_mutex mutex;
     };
 
-    std::shared_ptr<_Data> _data { new _Data };
+    std::shared_ptr<_Data> _data;
 };
 
 } // anon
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10180

pxr/usd/sdf/pathNode.cpp
```diff
@@ -494,9 +498,12 @@ Sdf_PathNode::GetPathToken(Sdf_PathNode const *primPart,
     TfAutoMallocTag tag2("Sdf_PathNode::GetPathToken");
 
     _PrimToPropTokenTables::accessor primAccessor;
-    _pathTokenTable->insert(
+    bool newValue = _pathTokenTable->insert(
```
nvmkuruc: I'm curious you considered this form of `insert` (https://github.com/uxlfoundation/oneTBB/blob/master/include/oneapi/tbb/concurrent_hash_map.h#L1141) as an alternative.  I haven't looked too deep, but it seems like it would defer construction of the value.  That would avoid the need to leave the `_PathToTokenTable` in a temporarily uninitialized state.
```
_pathTokenTable->insert(primAccessor, primPart);
```

petrihakkinen: I need to refresh my memory (it's been a while...). I'll get back to this next week.

petrihakkinen:
@nvmkuruc Sorry it took a bit longer than a week, but I've now pushed a new version with the suggested changes.

PR #3313, opened 2024-09-23
Autodesk: Fix leaking Vulkan buffer from HdSt_PipelineDrawBatch

erikaharrison-adsk:
### Description of Change(s)

This went undetected by the validation layer because the tests were also leaking the Vulkan instance and device, so they were never destroyed, and the validation check never kicked in. So fix the leaking tests too.

This also fixes an occasional crash from a driver trying to use a resource after the tests had exited main(), because the HgiVulkan leak also meant no synchronization.

Added some more buffer naming to help debug the leak and included it in the PR as it's useful for debugging.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10181

PR #3314, opened 2024-09-23
[docs] update VERSIONS.md to note that OpenUSD requires Jinja2 >= 3.0

pmolodo:
### Description of Change(s)

OpenUSD makes use of `jinja2.pass_context`:

- https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.08/pxr/imaging/hd/hdGenSchema.py#L9
- https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.08/third_party/renderman-25/plugin/hdPrman/rileyGenPrim.py#L10
- https://github.com/PixarAnimationStudios/OpenUSD/blob/v24.08/third_party/renderman-26/plugin/hdPrman/rileyGenPrim.py#L10

... which did not exist before 3.0
(was previously named `contextfunction`):

- https://github.com/pallets/jinja/commit/788d8bc1728490196f75ba4f8b7b1794a1974f0a

### Fixes Issue(s)
- Incompatible jinja2 version listed in VERSIONS.md

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -31,7 +31,7 @@ Our test machines have the following software versions installed.
 | Alembic       | 1.8.5                | 1.8.5                        | 1.8.5                          |
 | OpenEXR       | 3.1.11               | 3.1.11                       | 3.1.11                         |
 | MaterialX     | 1.38.10              | 1.38.10                      | 1.38.10                        |
-| Jinja2        | 2.0                  |                              |                                |
+| Jinja2        | >=3.0                |                              |                                |
 | Flex          | 2.5.39               |                              |                                |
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10188

dsyu-pixar:
@pmolodo thank you for this update. Just FYI I've checked our test/build machines and confirmed we are using jinja2 v3.1.2, so I'm going to use that specific version in VERSIONS.md to be consistent.

pmolodo:
> @pmolodo thank you for this update. Just FYI I've checked our test/build machines and confirmed we are using jinja2 v3.1.2, so I'm going to use that specific version in VERSIONS.md to be consistent.

Sounds great!

PR #3316, opened 2024-09-24
Apple: Correct re-declaration of structs as classes

dgovil:
### Description of Change(s)
This change fixes two areas where something was defined a structure, but was referenced as a class elsewhere. While this is harmless, it does cause some spurious compile time warnings.

I'm trying to reduce the number of warnings I get when building USD (which is a lot) to make it less noisy when developing on the USD codebase. So I'll put up PRs as I have downtime.

This is low priority and low/no impact to the codebase, but the reduced compile time noise is a win.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/imaging/hd/sortedIds.h
+++ b/pxr/imaging/hd/sortedIds.h
@@ -72,7 +72,7 @@ public:
     void Clear();
 
 private:
-    class _UpdateImpl;
+    struct _UpdateImpl;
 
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10190

nvmkuruc:
Thanks @dgovil. 👍

Issue #3093, opened 2024-05-23
Open usdview without specifying a file


jakeanq:
### Description of Issue

It would be useful to be able to launch `usdview` without being required to specify a file path, in which case it would open to an empty stage.

Use-cases for this:
 * Users who would prefer to use the File→Open menu option to find their USD files.
 * Writing a .desktop file (or equivalent for other OSs) so that `usdview` can be launched from system menus.
 * Writing a usdview plugin which determines what to open (perhaps selecting from an asset browser or dynamically generating USD content).

### Steps to Reproduce
1. Run `usdview` with no arguments; it will print `error: the following arguments are required: usdFile`

### System Information (OS, Hardware)

Tested on Arch Linux, hardware not applicable.

### Package Versions

`usd 24.03`

jesschimein:
Filed as internal issue #USD-9701

spiffmon:
Those are great points, @jakeanq !  Given our other priorities and roadmap, this is something we might not get to for awhile, but we'd be very interested to look at a PR for this, if anyone is willing!

MishDivyAmzn:
I would like to work on this issue for ASWF Dev Days, Sep. 26-27, 2024. I have experience working with OpenUSD python and C++ API and have used `usdview`. Please let me know if there are any concerns.

jesschimein:
Hi!  We'd love your contribution!  We're not hooked into the full Dev Days apparatus, but here are [contributing guidelines](https://openusd.org/release/contributing_to_usd.html) for OpenUSD!  The first thing we'd need is a signed CLA in order to accept your PR.  You can find links and instructions at the very top of the contributing guidelines page :)

ghypz:
Hello hello! I am also interested in working on this issue for Dev Days. I was informed that a CLA was just signed on behalf of Netflix Animation for some of us to be able to contribute to OpenUSD.

Would you be willing to collaborate on this @MishDivyAmzn? :)

MishDivyAmzn:
Hey,

I have requested a signed CLA from my employer. I'll send it as soon as it is ready.

I have this feature working on my machine. I am just writing some tests, and then I will test on other platforms (Windows, Linux and MacOS). If you are still interested in collaborating then I am down. But I also understand if you would prefer to tackle some other issue. :)

ghypz:
Sounds like you've already got this one covered! I'll have a browse at the other issues. Happy dev!

PR #3321, opened 2024-09-27
Open usdview without specifying a file

MishDivyAmzn:
Running `usdview` without `usdFile` positional argument will now open an empty stage. If `usdFile` is passed, then there will be no behavior change.   

### Description of Change(s)
- Made `usdFile` positional argement Optional
- Updated the `GetResolverContext` method to return a default context when `usdFile` is not passed.
- Updated the `appController` `_openStage` method to create an empty stage when `usdFilePath` is a falsy value.
- Updated the `_saveFlattendedAs`, `_saveOverridesAs` and `_saveViewerImage` methods to use "new_scene" prefix for recommended file name when `usdFile` is a falsy value.

**Note:** Prior to this change, running `usdview` without `usdFile` argument would print `error: the following arguments are required: usdFile`.

### Fixes Issue(s)
#3093

### DIFF
--- a/pxr/usdImaging/usdviewq/appController.py
+++ b/pxr/usdImaging/usdviewq/appController.py
@@ -403,7 +403,7 @@ class AppController(QtCore.QObject):
             self._ui = Ui_MainWindow()
             self._ui.setupUi(self._mainWindow)
 
-            self._mainWindow.setWindowTitle(parserData.usdFile)
+            self._mainWindow.setWindowTitle(parserData.usdFile or "Empty Stage")
             self._statusBar = QtWidgets.QStatusBar(self._mainWindow)
             self._mainWindow.setStatusBar(self._statusBar)
 
@@ -1203,7 +1203,16 @@ class AppController(QtCore.QObject):
             # layers populated after loading
             _MuteMatchingLayers()
             
-    def _openStage(self, usdFilePath, sessionFilePath,
+    def _getEmptyStage(self):
+
+        with self._makeTimer('create empty stage'):
+            stage = Usd.Stage.CreateInMemory()
+
+        if not stage:
+            sys.stderr.write('Error: Unable to create empty stage\n')
+        return stage
+            
+    def _getStageForFile(self, usdFilePath, sessionFilePath,
                    populationMaskPaths, muteLayersRe):
 
         def _GetFormattedError(reasons=None):
@@ -1274,6 +1283,25 @@ class AppController(QtCore.QObject):
         if not stage:
             sys.stderr.write(_GetFormattedError())
         else:
+            stage.SetEditTarget(stage.GetSessionLayer())
+
+        if self._mallocTags == 'stage':
+            DumpMallocTags(stage, "stage-loading")
+
+        return stage
+
+    def _openStage(self, usdFilePath, sessionFilePath,
+                   populationMaskPaths, muteLayersRe):
+
+        if self._mallocTags != 'none':
+            Tf.MallocTag.Initialize()
+
+        if not usdFilePath:
+            stage = self._getEmptyStage()
+        
+        else:
+            stage = self._getStageForFile(usdFilePath, sessionFilePath,
+                                          populationMaskPaths, muteLayersRe)
             stage.SetEditTarget(stage.GetSessionLayer())
 
         if self._mallocTags == 'stage':
@@ -2795,6 +2823,12 @@ class AppController(QtCore.QObject):
 
         # Start timer to measure Qt shutdown time
         self._startQtShutdownTimer()
+        
+    
+    def _getRecommendedFilenamePrefix(self):
+        return (self._parserData.usdFile.rsplit('.', 1)[0]
+                if self._parserData.usdFile
+                else 'new_file')
 
     def _openFile(self):
         extensions = Sdf.FileFormat.FindAllFileFormatExtensions()
@@ -2836,14 +2870,13 @@ class AppController(QtCore.QObject):
         return saveName
 
     def _saveOverridesAs(self):
-        recommendedFilename = self._parserData.usdFile.rsplit('.', 1)[0]
-        recommendedFilename += '_overrides.usd'
+        recommendedFilename = self._getRecommendedFilenamePrefix() + '_overrides.usd'
 
         saveName = self._getSaveFileName(
             'Save Overrides As', recommendedFilename)
         if len(saveName) == 0:
             return
-        elif (os.path.isfile(saveName) and
+        elif (os.path.isfile(saveName) and self._parserData.usdFile and
             os.path.samefile(saveName, self._parserData.usdFile)):
             msg = QtWidgets.QMessageBox()
             msg.setIcon(QtWidgets.QMessageBox.Critical)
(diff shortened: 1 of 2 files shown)


MishDivyAmzn:
Requested a signed CLA from my employer. Will send the CLA to Pixar and update this PR as soon as I receive the CLA.

jesschimein:
Filed as internal issue #USD-10199

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2804,9 +2826,13 @@ def _getSaveFileName(self, caption, recommendedFilename):
         
         return saveName
 
+    def _getRecommendedFilenamePrefix(self):
+        return (self._parserData.usdFile.rsplit('.', 1)[0]
+                if self._parserData.usdFile
+                else 'new_file')
```
spiffmon: Should this be something unique, like when we are creating temp files?  It's probably unlikely, but if a bunch of "file-less" usdview's are launched and you do live-editing and want to save out the overrides from more than one, if you go with the recommendation you'll stomp your results.  OTOH, you will see that the suggestion is the same for both, and have the ability to change it.  Just a thought to consider.

spiffmon: ... if you do go that route, given that there's multiple uses of this recommended name, and only one "empty/new stage" per usdview, Seems like we would want just one unique name per-process, right?

spiffmon (commented):
Just one suggestion, though others may have feedback also - thanks!

pxr/usdImaging/usdviewq/appController.py
MishDivyAmzn: Good point! The unique names usually look a bit ugly to people, so I wasn't sure about it :). I can replace this function with the following if people think this would be better.

```
def _getRecommendedFilenamePrefix(self):
    return (self._parserData.usdFile.rsplit('.', 1)[0]
            if self._parserData.usdFile
            else f'new_file_{os.getpid()}_{datetime.now().strftime("%Y%m%d_%H%M%S_%f")[:18]}')
```
The above function would make a unique name suggestion based on process ID and current datetime (millisecond precision), so almost unique everytime. But could be a bit jarring to look at. Open to suggestions :)

Here's an example of a potential suggestion from the above function:
`new_file_12345_20240927_154510_12`

We can potentially remove the `datetime` portion if we just need a unique name per process ID.

Thanks for the feedback, lmk if anything else stands out.

anwang2009:
Hi @MishDivyAmzn, thanks for the PR! Any updates on the CLA?

MishDivyAmzn:
The CLA is still in process by the legal. It should be ready very soon. I'll update this as soon as I hear back.

MishDivyAmzn:
Update: Pixar should have received the CLA now :)

pxr/usdImaging/usdviewq/appController.py
anwang2009: Let's go ahead with the current naming, since file saving does ask for a confirmation if you try to override an existing file

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -2804,9 +2826,13 @@ def _getSaveFileName(self, caption, recommendedFilename):
...
+
     def _saveOverridesAs(self):
-        recommendedFilename = self._parserData.usdFile.rsplit('.', 1)[0]
-        recommendedFilename += '_overrides.usd'
+        recommendedFilename = self._getRecommendedFilenamePrefix() + '_overrides.usd'
```
anwang2009, line 2873: Below at `os.path.samefile(saveName, self._parserData.usdFile)`: 

With an empty stage, if we want to save overrides to an existing file, `self._parserData.usdFile` will be None and an error occurs.

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -1262,8 +1270,22 @@ def _GetFormattedError(reasons=None):
...
+
+        if not usdFilePath:
+            stage = self._getEmptyStage()
+        
         else:
-            stage.SetEditTarget(stage.GetSessionLayer())
+            stage = self._getStageForFile(usdFilePath, sessionFilePath,
+                                          populationMaskPaths, muteLayersRe)
+
+        stage.SetEditTarget(stage.GetSessionLayer())
```
anwang2009: It looks like saving overrides saves the root layer only. If someone opens an empty stage, makes some edits via the Interpreter, then does "Save Overrides As", nothing will get written out. We should only create a session layer and set it as the edit target when we have a real stage, and avoid creating the session layer for the empty stage.

anwang2009 (commented):
@MishDivyAmzn thanks for your patience! We've added a few comments, when you get the chance to take a look.

MishDivy:
Thank you for the feedback, I'll update this PR asap. :)

pmolodo:
@MishDivy - any chance you could find some time to update this PR? We'd also like to be able to open UsdView with no args...

MishDivy:
Apologies for the delay, this slipped out of my radar. Let me rebase this PR to latest, and fix any issues pointed out.

MishDivy:
### Update Summary
I have updated this PR to address previous review comments:

- When **`usdview`** is launched **without a `usdFile` argument**, it no longer creates a Session layer.  
  All overrides are now authored to the **anonymous root layer** instead.
- Fixed the issue where attempting to **save overrides with the same filename** would raise an error.

### Testing
I verified the changes by:
- Launching `usdview` with no arguments (empty stage).  
- Authoring scene description using the Python interpreter.  
- Using **Save Overrides As…** and **Save Flattened As…** to confirm that data exports correctly from `usdview`.  
- Reopening the saved files to ensure they load successfully and edits persist as expected.

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -1281,6 +1290,25 @@ def _GetFormattedError(reasons=None):
...
+        else:
+            stage = self._getStageForFile(usdFilePath, sessionFilePath,
+                                          populationMaskPaths, muteLayersRe)
+            stage.SetEditTarget(stage.GetSessionLayer())
+
+        if self._mallocTags == 'stage':
+            DumpMallocTags(stage, "stage-loading")
+
+        return stage
+
```
sdobbs-blz, line 1311: Looks like this PR is missing a lot of logic from the `_openStage` method, perhaps a bad merge? - https://github.com/PixarAnimationStudios/OpenUSD/blob/43b8c713427fac6eb1230d8f4933d1358b9746b2/pxr/usdImaging/usdviewq/appController.py#L1206

MishDivy: Thanks for the comment!

I renamed the old `_openStage` function to `_getStageForFile` and now the `_openStage` calls `_getStageForFile` or `_getEmptyStage` based on whether `usdFile` was passed as an argument. This way the callers of `_openStage` are unaffected. 

Please let me know if there are any other concerns.

sdobbs-blz: ahh sorry, missed that. thanks!

PR #3332, opened 2024-10-02
Autodesk: Fix Loading Large Size Image Crash with hioOpenEXR

erikaharrison-adsk:
### Description of Change(s)

In USD 24.08, the crash when loading an extreme-sized texture returns. The root cause is that now USD uses Hio_OpenEXRImage instead of HioOiio to load the exr images, and there are issues in OpenEXRCore & HioOpenEXR.

This solution is to fix loading large-size image crashes mainly resulting from int overflows. 

Related: https://github.com/PixarAnimationStudios/OpenUSD/pull/3119

### Fixes Issue(s)
- N/A

### DIFF
--- a/pxr/imaging/hio/OpenEXR/OpenEXRCore/encoding.c
+++ b/pxr/imaging/hio/OpenEXR/OpenEXRCore/encoding.c
@@ -373,7 +373,7 @@ exr_encoding_run (
     {
         priv_from_native32 (
             encode->sample_count_table,
-            encode->chunk.width * encode->chunk.height);
+            (uint64_t) encode->chunk.width * (uint64_t) encode->chunk.height);
     }
 
     if (rv == EXR_ERR_SUCCESS)
@@ -420,7 +420,7 @@ exr_encoding_run (
     {
         priv_to_native32 (
             encode->sample_count_table,
-            encode->chunk.width * encode->chunk.height);
+            (uint64_t) encode->chunk.width * (uint64_t) encode->chunk.height);
     }
 
     return rv;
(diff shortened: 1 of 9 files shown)


jesschimein:
Filed as internal issue #USD-10227

meshula:
Thank you for these fixes. The changes to OpenEXRCore do not exist upstream, but the fixes are important enough that they should be upstreamed ~ we don't want the interned copy of OpenEXRCore to get ahead of the originating code with substantial changes like this. Could I ask that you also upstream the OpenEXRCore fixes to OpenEXR in parallel please? The OpenEXR test suite is supposed to be robust enough to catch very large files and so on, so could you also post at OpenEXR how to repro the failure? This should have been caught OpenEXR side :)

Thanks!

adskWangl:
> Thank you for these fixes. The changes to OpenEXRCore do not exist upstream, but the fixes are important enough that they should be upstreamed ~ we don't want the interned copy of OpenEXRCore to get ahead of the originating code with substantial changes like this. Could I ask that you also upstream the OpenEXRCore fixes to OpenEXR in parallel please? The OpenEXR test suite is supposed to be robust enough to catch very large files and so on, so could you also post at OpenEXR how to repro the failure? This should have been caught OpenEXR side :)
> 
> Thanks!

Thank you for your response! I'll try to reproduce this failure with an independent OpenEXR case and upstream the OpenEXRCore fixes to OpenEXR as this fix is also important to us.

adskWangl:
@meshula sorry for the delay. The fix in OpenEXRCore has been submitted into OpenEXR and I assume you'll update OpenEXR sometime. So for this PR, I now limit the scope of changes to the USD code itself.

meshula:
Thanks @adskWangl !

PR #3333, opened 2024-10-02
Autodesk: Added TF_VERIFY_VK_RESULT for Vulkan

erikaharrison-adsk:
### Description of Change(s)

Added TF_VERIFY_VK_RESULT to verify the result of the Vulkan call and print the result string if it fails.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10228

Issue #3114, opened 2024-06-07
Importing pxr in python on Windows changes PATH which breaks Usd in subprocesses


tbttfox:
### Description of Issue
If you `import pxr` in a python interpreter then launch a subprocess that has its own USD libraries (eg. Maya), the extra entry added into PATH loads the incorrect DLL's and prevents the subprocess from loading USD.

This is something that *can* be worked around on our end. But I really don't like that importing a module changes PATH.
It looks like this only happens on Windows, and is only added to the PyPi upload to make anaconda work

The PATH change happens here: [build_scripts/pypi/package_files/setup.py](https://github.com/PixarAnimationStudios/OpenUSD/blob/2864f3d04f396432f22ec5d6928fc37d34bb4c90/build_scripts/pypi/package_files/setup.py#L98)

### Steps to Reproduce
1. Open a python interpreter
2. `import pxr, subprocess`
3. `subprocess.Popen(r'C:\Program Files\Autodesk\Maya2024\bin\maya.exe')`
4. Load mayaUsdPlugin.mll
5. Note the error. It will be something like this:
```
// Error: file: C:/Program Files/Autodesk/Maya2024/scripts/others/pluginWin.mel line 316: Unable to dynamically load : C:/Program Files/Autodesk/MayaUSD/Maya2024/0.25.0/mayausd/MayaUSD/plugin/adsk/plugin/mayaUsdPlugin.mll
// The specified procedure could not be found.
// Error: file: C:/Program Files/Autodesk/Maya2024/scripts/others/pluginWin.mel line 316: The specified procedure could not be found.
//  (mayaUsdPlugin)
```
### System Information (OS, Hardware)
Windows 10

### Package Versions
System Python USD version: 0.24.5
Maya Python USD version: 0.22.11
MayaUSD version 0.25.0
Python 3.9.12

jesschimein:
Filed as internal issue #USD-9744

PR #3353, opened 2024-10-08
Move all the Windows dll path logic into the WindowsImportWrapper

tbttfox:
### Description of Change(s)
There's already a python context in `pxr/tf/__init__.py` that handles adding and removing dll paths for windows, so I moved the Windows specific import path logic out of the `pxr/__init__.py` to there.  This way, we don't have to permanently change the `PATH`

I couldn't get the tests to pass locally from the dev build, even *without* my changes.
And I couldn't get the release branch to build because boost breaks.
Any suggestions how to deal with this?

However, I was able to test that my changes load the correct DLL's on python 3.7, python 3.11, and anaconda 3.12 both with and without the `PXR_USD_WINDOWS_DLL_PATH` variable set. (I just copied my changes directly into different pip installs)

### Fixes Issue(s)
https://github.com/PixarAnimationStudios/OpenUSD/issues/3114

### DIFF
--- a/build_scripts/pypi/package_files/setup.py
+++ b/build_scripts/pypi/package_files/setup.py
@@ -59,28 +59,6 @@ if windows():
     for f in dll_files:
         shutil.move(f, os.path.join(BUILD_DIR, "lib/python/pxr"))
 
-    # Because there are no RPATHS, patch __init__.py
-    # See this thread and related conversations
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10276

✅  Signed CLA on file

dgovil:
@sunyab does USD still support Python 3.7 ? If not then perhaps the version check here could be simplified and moved to a build time check for 3.8+?

Issue #3335, opened 2024-10-02
Compiler Warnings: Overrides a member function but is not marked override


dgovil:
### Description of Issue

There are a range of methods that do not declare themselves as overrides. This should be a fairly straightforward change to make, and would clean up a lot of issues/compiler ambiguity. I think this is a good first time issue for someone getting started with the USD codebase.

I am building with Clang 16 from Xcode 16 with `python3 ./build_scripts/build_usd.py ~/Downloads/usd  --build-variant debug --no-imaging  -vvv` but this should reproduce with other compilers. The warnings are reproduced below:

```
[1833/4632] Building CXX object pxr/usd/usdUtils/CMakeFiles/usdUtils.dir/assetLocalizationPackage.cpp.o
In file included from /Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationPackage.cpp:18:
In file included from /Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalization.h:17:
/Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationDelegate.h:274:38: warning: 'ProcessPayloads' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  274 |     virtual std::vector<std::string> ProcessPayloads(
      |                                      ^
/Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationDelegate.h:50:38: note: overridden virtual function is here
   50 |     virtual std::vector<std::string> ProcessPayloads(
      |                                      ^
1 warning generated.
In file included from /Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationPackage.cpp:18:
In file included from /Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalization.h:17:
/Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationDelegate.h:274:38: warning: 'ProcessPayloads' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  274 |     virtual std::vector<std::string> ProcessPayloads(
      |                                      ^
/Users/dhruvgovil/Projects/usd/pxr/usd/usdUtils/assetLocalizationDelegate.h:50:38: note: overridden virtual function is here
   50 |     virtual std::vector<std::string> ProcessPayloads(
      |                                      ^
1 warning generated.
[...]

PR #3355, opened 2024-10-09
Compiler Warnings: Overrides a member function but is not marked override

furby-tm:
### Description of Change(s)

Add missing `override` declarations to member methods which override virtual methods of their base classes.

### Fixes Issue(s)
- #3335


furby-tm:
cc. @dgovil

jesschimein:
Filed as internal issue #USD-10278

dgovil:
Forgot to leave a review comment on here when you put this up but this change looks good!

Issue #3324, opened 2024-09-28
Compiler Warnings: Static Functions that should be Static Inline


dgovil:
### Description of Issue

There are several static functions that the compiler believes should be declared as static inline. These are reported from Apple Clang 16 bundled with Xcode 16, but should reproduce in other compilers with slight variations.

Most are found in `pxr/base/vt/wrapArray.h` with a few in `usd/pxr/imaging/hdSt/unitTestHelper.h`.
I think this would be a good first issue to get folks used to building USD and make contributions that will help provide cleaner outputs.

The warning is reproduced below:

```
[1065/4632] Building CXX object pxr/base/vt/CMakeFiles/vt.dir/wrapArray.cpp.o
In file included from /Users/dhruvgovil/Projects/usd/pxr/base/vt/wrapArray.cpp:10:
/Users/dhruvgovil/Projects/usd/pxr/base/vt/wrapArray.h:307:1: warning: 'static' function 'Vt_ComputeEffectiveRankAndLastDimSize' declared in header file should be declared 'static inline' [-Wunneeded-internal-declaration]
  307 | Vt_ComputeEffectiveRankAndLastDimSize(
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
In file included from /Users/dhruvgovil/Projects/usd/pxr/base/vt/wrapArray.cpp:10:
/Users/dhruvgovil/Projects/usd/pxr/base/vt/wrapArray.h:307:1: warning: 'static' function 'Vt_ComputeEffectiveRankAndLastDimSize' declared in header file should be declared 'static inline' [-Wunneeded-internal-declaration]
  307 | Vt_ComputeEffectiveRankAndLastDimSize(
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.

[1337/4632] Building CXX object pxr/base/vt/CMakeFiles/vt.dir/arrayPyBuffer.cpp.o
In file included from /Users/dhruvgovil/Projects/usd/pxr/base/vt/arrayPyBuffer.cpp:15:
/Users/dhruvgovil/Projects/usd/pxr/base/vt/wrapArray.h:307:1: warning: 'static' function 'Vt_ComputeEffectiveRankAndLastDimSize' declared in header file should be declared 'static inline' [-Wunneeded-internal-declaration]
  307 | Vt_ComputeEffectiveRankAndLastDimSize(
[...]

PR #3356, opened 2024-10-09
Compiler Warnings: Static Functions that should be Static Inline

furby-tm:
### Description of Change(s)

Add `inline` to the static methods the clang compiler thinks should be declared as `static inline`.

### Fixes Issue(s)
- #3324


furby-tm:
cc. @dgovil

jesschimein:
Filed as internal issue #USD-10279

dgovil:
Also forgot to review this one here (my bad, I know we talked about it on slack though), but it looks good to me.

PR #3359, opened 2024-10-10
usdShade: Adding NormalMapTexture validator

beersandrew:
### Description of Change(s)
Added an implementation of NormalMapTexture validation rules 

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65425006


pxr/usd/usdShade/validators.cpp
```diff
@@ -590,6 +589,244 @@ _SubsetsMaterialBindFamily(const UsdPrim& usdPrim)
...
+static
+UsdValidationErrorVector
+_NormalMapTextureValidator(const UsdPrim& usdPrim) {
+
+    if (!usdPrim.IsA<UsdShadeShader>()) {
+        return {};
+    }
+
+    const UsdShadeShader shader(usdPrim);
+    if (!shader) {
```
beersandrew: @tallytalwar I see this check here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L468 however I'm not sure what USD structure would trigger this failure, after it's already passed the `IsA` check.

pxr/usd/usdShade/validators.cpp
```diff
@@ -590,6 +589,244 @@ _SubsetsMaterialBindFamily(const UsdPrim& usdPrim)
...
+    // effect of normal perturbations.
+    // don't really care about fourth components...
+    bool nonCompliantScaleValues = scale[0] != 2 ||
+        scale[1] != 2 || scale[2] != 2;
+
+    if (nonCompliantScaleValues)
+    {
+        errors.emplace_back(
+            UsdShadeValidationErrorNameTokens->nonCompliantScale,
+            UsdValidationErrorType::Warn,
```
beersandrew, line 777: @tallytalwar in the compliance checker, this is a warning, however the bias check below is an error. Is that correct? I wanted to change the bias error below to be a warning, let me know what you think

pxr/usd/usdShade/testenv/testUsdShadeValidators.cpp
```diff
@@ -557,6 +559,213 @@ TestUsdShadeEncapsulationRulesValidator()
...
+    TF_AXIOM(errors[0].GetSites().size() == 1);
+    TF_AXIOM(errors[0].GetSites()[0].IsValid());
+    TF_AXIOM(errors[0].GetSites()[0].IsPrim());
+    TF_AXIOM(errors[0].GetSites()[0].GetPrim().GetPath() ==
+             expectedPrimPath);
+    TF_AXIOM(errors[0].GetMessage() == expectedErrorMsg);
+}
+
+void
+TestUsdShadeNormalMapTextureValidator()
```
beersandrew, line 580: @tallytalwar There are quite a few failure conditions in this test and validator, let me know if you think it would be better to break this up in to multiple validators or not

pxr/usd/usdShade/testenv/testUsdShadeValidators/normalMap.jpg
beersandrew, line 1: @tallytalwar We had the issue in the past with the encapsulation validator where me adding external files did not work for you, just calling this out here as this may happen again as I don't know what the difference was here in the past.

jesschimein:
Filed as internal issue #USD-10296

pxr/usd/usdShade/validators.cpp
tallytalwar: Yeah I don't think so also.

pxr/usd/usdShade/validators.cpp
tallytalwar: Referring the original commit for bias and scale adjustments to compliance checker: https://github.com/PixarAnimationStudios/OpenUSD/commit/b718721744f72c310f9b08aeac101c9fecd9d89a

When complying with UsdPreviewTexture and UsdUVTexture (that is when nonCompliantScaleValues is false) we need to make sure to report error for bias values. Refer these here: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usdImaging/plugin/usdShaders/shaders/shaderDefs.usda#L100-L112

Regarding non-conforming scale values being reported as warnings: There is a reasoning here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L550-L552 (I will request to keep all these code comments as-is in the cpp so as to save the reasoning). I believe Blender is the application mentioned here (cc. @dgovil is that still the case?)

beersandrew: made sure to include all comments from the complianceChecker code (I did miss one). Also rebased this branch

beersandrew:
Closing this PR, this work is now done on https://github.com/PixarAnimationStudios/OpenUSD/pull/3443

PR #3360, opened 2024-10-10
Autodesk: Task controller custom tasks

erikaharrison-adsk:
### Description of Change(s)

Enables the addition of custom render tasks in the task controller. The user can select the insertion point in the standard stack and edit the render task parameters.

### Fixes Issue(s)
- N/A


jesschimein:
Filed as internal issue #USD-10301

PR #3361, opened 2024-10-11
Improve `ArchGetFileName` for windows to return full path

roggiezhang-nv:
### Description of Change(s)

In a fix to improve Alembic plugin to read assets with ArResolver to support other sources except local files (https://github.com/PixarAnimationStudios/OpenUSD/pull/3302), Alembic library provides 4 APIs: one of them supports to read from local paths and the other 3 are using istream interfaces which has flaws to support layered ABC files. Therefore, we can only use the API that supports to read from local paths. While it involves to get the mirroed local path from ArAsset, and that's `ArchGetFileName` supposed to work for. 

However, `ArchGetFileName` only returns paths without drive letter, which is relative path related to the current drive the application is running from. This PR is to improve `ArchGetFileName` to return full path from `FILE*`. Here is an example before and after this PR:

Assuming the file path behind the FILE* handle is `C:/path/includes/drive/letter.ext`, and the application is running from `D` drive.
// Before --
ArchGetFileName(file_handle) => "/path/includes/drive/letter.ext"
// After --
ArchGetFileName(file_handle) => "C:/path/includes/drive/letter.ext"

The path returned from old API is `/path/includes/drive/letter.ext`, which is corresponding to `D:/path/includes/drive/letter.ext` in full path, which is wrong.

### Fixes Issue(s)
-

### DIFF
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -532,24 +532,34 @@ ArchGetFileName(FILE *file)
     }
     return result;
 #elif defined (ARCH_OS_WINDOWS)
-    static constexpr DWORD bufSize =
-        sizeof(FILE_NAME_INFO) + sizeof(WCHAR) * 4096;
-    HANDLE hfile = _FileToWinHANDLE(file);
-    auto fileNameInfo = reinterpret_cast<PFILE_NAME_INFO>(malloc(bufSize));
     string result;
-    if (GetFileInformationByHandleEx(
-            hfile, FileNameInfo, static_cast<void *>(fileNameInfo), bufSize)) {
+    WCHAR filePath[MAX_PATH];
+    HANDLE hfile = _FileToWinHANDLE(file);
+    if (GetFinalPathNameByHandleW(hfile, filePath, MAX_PATH, VOLUME_NAME_DOS)) {
         size_t outSize = WideCharToMultiByte(
-            CP_UTF8, 0, fileNameInfo->FileName,
-            fileNameInfo->FileNameLength/sizeof(WCHAR),
+            CP_UTF8, 0, filePath,
+            wcslen(filePath),
             NULL, 0, NULL, NULL);
         result.resize(outSize);
         WideCharToMultiByte(
-            CP_UTF8, 0, fileNameInfo->FileName,
-            fileNameInfo->FileNameLength/sizeof(WCHAR),
+            CP_UTF8, 0, filePath,
+            -1,
             &result.front(), outSize, NULL, NULL);
+
+        if (result.length() > 4)
+        {
+            // It needs to strip the path prefix as
+            // the path returned is DOS device path, and the
+            // syntax is one of:
+            //   \\.\C:\Test\Foo.txt
+            //   \\?\C:\Test\Foo.txt
+            if (result.compare(0, 4, "\\\\?\\") == 0 ||
+                result.compare(0, 4, "\\\\.\\") == 0)
+            {
+                result.erase(0, 4);
+            }
+        }
     }
-    free(fileNameInfo);
     return result;                                        
 #else
 #error Unknown system architecture
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -21,6 +21,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cerrno>
+#include <filesystem>
 #include <memory>
 #include <utility>
 
@@ -555,21 +556,11 @@ ArchGetFileName(FILE *file)
             -1,
             &result.front(), outSize, NULL, NULL);
 
+        // Strip path prefix if necessary.
         // See https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
-        // for format of DOS device paths. For UNC paths, it's returned as untouched.
-        if (result.length() > 4 &&
-            result.compare(0, 8, "\\\\?\\UNC\\") != 0)
-        {
-            // Otherwise, strip prefix from paths returned by GetFinalPathNameByHandleW,
-            // which is one of:
-            //   \\.\C:\Test\Foo.txt
-            //   \\?\C:\Test\Foo.txt
-            if (result.compare(0, 4, "\\\\?\\") == 0 ||
-                result.compare(0, 4, "\\\\.\\") == 0)
-            {
-                result.erase(0, 4);
-            }
-        }
+        // for format of DOS device paths.
+        auto canonicalPath = std::filesystem::canonical(result);
+        result = canonicalPath.string();
     }
     return result;                                        
 #else
--- a/pxr/base/arch/fileSystem.cpp
+++ b/pxr/base/arch/fileSystem.cpp
@@ -533,24 +533,35 @@ ArchGetFileName(FILE *file)
     return result;
 #elif defined (ARCH_OS_WINDOWS)
     string result;
-    WCHAR filePath[MAX_PATH];
+    std::vector<WCHAR> filePath(MAX_PATH);
     HANDLE hfile = _FileToWinHANDLE(file);
-    if (GetFinalPathNameByHandleW(hfile, filePath, MAX_PATH, VOLUME_NAME_DOS)) {
+    DWORD dwSize = GetFinalPathNameByHandleW(hfile, filePath.data(), MAX_PATH, VOLUME_NAME_DOS);
+    // * dwSize == 0. Fail.
+    // * dwSize < MAX_PATH. Success, and dwSize returns the size without null terminator.
+    // * dwSize >= MAX_PATH. Buffer is too small, and dwSize returns the size with null terminator.
+    if (dwSize >= MAX_PATH) {
+        filePath.resize(dwSize);
+        dwSize = GetFinalPathNameByHandleW(hfile, filePath.data(), dwSize, VOLUME_NAME_DOS);
+    }
+
+    if (dwSize != 0) {
         size_t outSize = WideCharToMultiByte(
-            CP_UTF8, 0, filePath,
-            wcslen(filePath),
+            CP_UTF8, 0, filePath.data(),
+            dwSize,
             NULL, 0, NULL, NULL);
         result.resize(outSize);
         WideCharToMultiByte(
-            CP_UTF8, 0, filePath,
+            CP_UTF8, 0, filePath.data(),
             -1,
             &result.front(), outSize, NULL, NULL);
 
-        if (result.length() > 4)
+        // See https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
+        // for format of DOS device paths. For UNC paths, it's returned as untouched.
+        if (result.length() > 4 &&
+            result.compare(0, 8, "\\\\?\\UNC\\") != 0)
         {
-            // It needs to strip the path prefix as
-            // the path returned is DOS device path, and the
-            // syntax is one of:
+            // Otherwise, strip prefix from paths returned by GetFinalPathNameByHandleW,
+            // which is one of:
             //   \\.\C:\Test\Foo.txt
             //   \\?\C:\Test\Foo.txt
             if (result.compare(0, 4, "\\\\?\\") == 0 ||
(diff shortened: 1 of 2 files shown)


roggiezhang-nv:
@nvmkuruc @nvidia-jomiller  for vis.

pxr/base/arch/testenv/testFileSystem.cpp
```diff
@@ -92,6 +114,12 @@ int main()
     fclose(firstFile);
     ARCH_AXIOM(ArchGetFileLength(firstName.c_str()) == strlen(testContent));
 
+    // Open a file, check that the file path from FILE* handle is matched.
+    ARCH_AXIOM((firstFile = ArchOpenFile(firstName.c_str(), "rb")) != NULL);
+    std::string filePath = ArchGetFileName(firstFile);
+    ARCH_AXIOM(_DosDevicePathFilter(filePath) == _DosDevicePathFilter(firstName));
```
nvmkuruc: Would comparing `ArchAbsPath` be sufficient instead of `_DosDevicePathFilter`?

roggiezhang-nv: It doesn't work to me.

jesschimein:
Filed as internal issue #USD-10302

anwang2009:
Thanks for the PR @roggiezhang-nv !

`testFileSystem` with these changes is failing on both macos and windows - could you address this? Here's the offending failed test and context on windows.:
```
72: first name C:\Windows\TEMP/archFS.12176
72: filePath C:\Windows\Temp\archFS.12176
72: ArchNormPath(first name) C:/Windows/TEMP/archFS.12176
72: ArchNormPath(filePath) C:/Windows/Temp/archFS.12176
72:  ArchError: [ArchNormPath(filePath) == ArchNormPath(firstName)] axiom failed
 ```

**Comment by nvmkuruc** [comment]
@roggiezhang-nv Maybe the test should compare `ArchAbsPath` to get rid of any symlink?

https://openusd.org/dev/api/group__group__arch___system_functions.html#gab146e5fb7a6c2f805936960bc58e3911 suggests it should be returning the canonical path.

**Comment by roggiezhang-nv** [comment]
Seems so. I'll fix it.

**Comment by roggiezhang-nv** [comment]
Replaced. Let's see how it goes.

**Comment by anwang2009** [comment]
By the way @roggiezhang-nv - the test is still failing on Windows with similar output as I've commented above. Is it working on your Windows build?

**Comment by roggiezhang-nv** [comment]
> By the way @roggiezhang-nv - the test is still failing on Windows with similar output as I've commented above. Is it working on your Windows build?

@anwang2009 Yes, it's working for my machine. The issue is that we don't have a way to return a uniform path on windows to check the equality of paths. Is it ok to use the uppercase or lowercase of paths on windows to compare?

**Comment by roggiezhang-nv** [comment]
Btw, @anwang2009, @nvmkuruc wants me to compare the perf between `GetFinalPathNameByHandle` (new) and `GetFileInformationByHandleEx` (old) to give more confidence. 

I called each of them 100000 times, and `GetFinalPathNameByHandle` is indeed twice slower than `GetFileInformationByHandleEx`. However, it costs only 6us for `GetFinalPathNameByHandle` compared to 3us for `GetFileInformationByHandleEx` in each call, so the perf influence can be ignored. 

Let me if I can simply the uppercase or lowercase to compare the paths for windows. Or you have any new suggestions.

**Comment by roggiezhang-nv** [comment]
Btw, I fixed the path compare issue with `std::filesystem::equivalent`, which can handle cases.

**Comment by roggiezhang-nv** [comment]
@anwang2009 I also need to correct my number as I passed a wrong param which is not similar to the real implementation to GetFinalPathNameByHandle. In reality, it's 5-6 times slower than old implementation (windows only, but not Linux and Mac) because of GetFinalPathNameByHandle. So it's 15us per call for ArchGetFileName this time on my machine. Not sure this is an acceptable cost.

**Comment by meshula** [comment]
I haven't spotted it being discussed elsewhere, so I'll note here, forgive the dry tone, and ignore as you wish :)

GetFinalPathNameByHandle resolves the final path of a file, including symbolic links, mount points, or junctions in addition to the work GetFileInformationByHandleEx might perform.

Given the reported benchmark result of resolving in microseconds, I have a suspicion the benchmark is of cached performance not unwarmed performance. Also, I think there might be a more "interesting" difference (ie, tens of milliseconds or more) if a the path needs resolving. but, it's an unavoidable penalty if we want correct operation.

A significant call point is in Crate, where it's going to be called once per asset so there'll be a general linear overhead. If it's really 16us, and you have a thousand files, it'll be 16ms, time lost in the noise.

If it's resolving across a network mount, right now it'll fail, and after a fix, the overhead will add up. But as I said, that's unavoidable.

Where the overhead would likely be significant is if someone wrote a loop to get info on all the files in a massive directory but that doesn't happen in the USD codebase itself.

**Comment by roggiezhang-nv** [comment]
> I haven't spotted it being discussed elsewhere, so I'll note here, forgive the dry tone, and ignore as you wish :)
> 
> GetFinalPathNameByHandle resolves the final path of a file, including symbolic links, mount points, or junctions in addition to the work GetFileInformationByHandleEx might perform.
> 
> Given the reported benchmark result of resolving in microseconds, I have a suspicion the benchmark is of cached performance not unwarmed performance. Also, I think there might be a more "interesting" difference (ie, tens of milliseconds or more) if a the path needs resolving. but, it's an unavoidable penalty if we want correct operation.
> 
> A significant call point is in Crate, where it's going to be called once per asset so there'll be a general linear overhead. If it's really 16us, and you have a thousand files, it'll be 16ms, time lost in the noise.
> 
> If it's resolving across a network mount, right now it'll fail, and after a fix, the overhead will add up. But as I said, that's unavoidable.
> 
> Where the overhead would likely be significant is if someone wrote a loop to get info on all the files in a massive directory but that doesn't happen in the USD codebase itself.

@meshula Thanks for your insights. Yes, it's cached result. For non-cached result, that's 75us versus 20us for both functions on my machine. And I agree it should also depend on whether it's symbolic links, etc. 

>> > A significant call point is in Crate, where it's going to be called once per asset so there'll be a general linear overhead. If it's really 16us, and you have a thousand files, it'll be 16ms, time lost in the noise.

That's one of the issue I saw from Crate implementation as [the filename returned by the call](https://github.com/PixarAnimationStudios/OpenUSD/blob/9b0c13b2efa6233c8a4a4af411833628c5435bde/pxr/usd/usd/crateFile.cpp#L2270) is not used anywhere, which means we can reduce that cost from USD codebase. @nvmkuruc for this issue.

File: pxr/base/arch/fileSystem.cpp [hunk 2]
```diff
@@ -532,24 +532,34 @@ ArchGetFileName(FILE *file)
...
     return result;
 #elif defined (ARCH_OS_WINDOWS)
-    static constexpr DWORD bufSize =
-        sizeof(FILE_NAME_INFO) + sizeof(WCHAR) * 4096;
-    HANDLE hfile = _FileToWinHANDLE(file);
-    auto fileNameInfo = reinterpret_cast<PFILE_NAME_INFO>(malloc(bufSize));
     string result;
-    if (GetFileInformationByHandleEx(
-            hfile, FileNameInfo, static_cast<void *>(fileNameInfo), bufSize)) {
+    WCHAR filePath[MAX_PATH];
```
anwang2009: The docs aren't super clear to me, but I think using MAX_PATH for the buffer size seems like it could be insufficient for Windows configurations where long paths are enabled.

GetFilePathNameByHandleW is supposed to return the required buffer size in the case where the filename would overflow the given buffer – it seems like we should be handling that somehow.

from @sunyab

pxr/base/arch/fileSystem.cpp
```diff
@@ -532,24 +532,34 @@ ArchGetFileName(FILE *file)
...
+    HANDLE hfile = _FileToWinHANDLE(file);
+    if (GetFinalPathNameByHandleW(hfile, filePath, MAX_PATH, VOLUME_NAME_DOS)) {
         size_t outSize = WideCharToMultiByte(
-            CP_UTF8, 0, fileNameInfo->FileName,
-            fileNameInfo->FileNameLength/sizeof(WCHAR),
+            CP_UTF8, 0, filePath,
+            wcslen(filePath),
```
anwang2009: Can we avoid calling wcslen by using the return value of GetFilePathNameByHandleW?

from @sunyab

pxr/base/arch/fileSystem.cpp
```diff
@@ -532,24 +532,34 @@ ArchGetFileName(FILE *file)
...
+            -1,
             &result.front(), outSize, NULL, NULL);
+
+        if (result.length() > 4)
+        {
+            // It needs to strip the path prefix as
+            // the path returned is DOS device path, and the
+            // syntax is one of:
+            //   \\.\C:\Test\Foo.txt
+            //   \\?\C:\Test\Foo.txt
```
anwang2009: This comment might be easier to understand as:

```
// Strip prefix from paths returned by GetFinalPathNameByHandleW,
// which is one of:
//   \\.\C:\Test\Foo.txt
//   \\?\C:\Test\Foo.txt
```

from @sunyab

pxr/base/arch/fileSystem.cpp
```diff
@@ -532,24 +532,34 @@ ArchGetFileName(FILE *file)
...
+            if (result.compare(0, 4, "\\\\?\\") == 0 ||
+                result.compare(0, 4, "\\\\.\\") == 0)
+            {
+                result.erase(0, 4);
+            }
```
anwang2009: While doing some research, I saw that network paths on Windows will look like "\\?\UNC\...". I think we've run into issues with network file paths in the past, so we need to figure out whether we need to deal with this specially.

from @sunyab

anwang2009 (commented):
Thanks @roggiezhang-nv , the tests pass now. Here are a few notes:

pxr/base/arch/fileSystem.cpp
roggiezhang-nv: Addressed.

pxr/base/arch/fileSystem.cpp
roggiezhang-nv: Addressed.

pxr/base/arch/fileSystem.cpp
roggiezhang-nv: Improved.

pxr/base/arch/fileSystem.cpp
roggiezhang-nv: I treated it specially without stripping the prefix if it's UNC path.

pxr/base/arch/fileSystem.cpp
```diff
@@ -532,24 +533,35 @@ ArchGetFileName(FILE *file)
...
-            fileNameInfo->FileNameLength/sizeof(WCHAR),
+            CP_UTF8, 0, filePath.data(),
+            -1,
             &result.front(), outSize, NULL, NULL);
+
+        // Strip path prefix if necessary.
+        // See https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
+        // for format of DOS device paths.
+        auto canonicalPath = std::filesystem::canonical(result);
+        result = canonicalPath.string();
```
anwang2009, line 563: We should actually consider removing this entire block. [According to this post from three years ago](https://developercommunity.visualstudio.com/t/stdfilesystemcanonical-behaves-different-for-netwo/1619001#T-N1624131) `std::filesystem::canonical` is identical to `GetFinalPathNameByHandleW`. Have you found differently? If not, we should just be able to return `result` directly instead of trying to canonicalize it again.

One thing to also note is that `GetFinalPathNameByHandleW` by default (with `VOLUME_NAME_DOS`) includes prefixes by default, and therefore so does `std::filesystem::canonical`, if the above post is correct. So the prefixes are not stripped in this current implementation, which seems fine as long as downstream use cases are handled correctly. Why did you want prefixes stripped originally?

roggiezhang-nv: @anwang2009 We just thought we should return a canonical path. 

Not sure what you mean that `std::filesystem::canonical` is identical to `GetFinalPathNameByHandleW`. `std::filesystem::canonical` will strip prefixes if necessary to return a canonical path. For linux paths, they are the same, but not for windows paths.

\\?\C:\test_files\main.usd   <---- before `std::filesystem::canonical` is called.
C:\test_files\main.usd         <---- after `std::filesystem::canonical` is called.

anwang2009: I see, the post must be out of date. Thanks @roggiezhang-nv !

PR #3364, opened 2024-10-16
usdUtils: Add Validation Framework rule ARKitLayerChecker

beersandrew:
### Description of Change(s)
- Added validation logic and a test for ARKitLayerChecker 

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65424926

### DIFF
--- a/pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
+++ b/pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
@@ -30,7 +30,7 @@ TestUsdUsdzValidators()
     UsdValidationRegistry& registry = UsdValidationRegistry::GetInstance();
     UsdValidatorMetadataVector metadata =
             registry.GetValidatorMetadataForPlugin(_tokens->usdUtilsPlugin);
-    TF_AXIOM(metadata.size() == 1);
+    TF_AXIOM(metadata.size() == 2);
     // Since other validators can be registered with a UsdUtilsValidators
     // keyword, our validators registered in usd are a subset of the entire
     // set.
@@ -40,7 +40,8 @@ TestUsdUsdzValidators()
     }
 
     const std::set<TfToken> expectedValidatorNames =
-            {UsdUtilsValidatorNameTokens->packageEncapsulationValidator};
+            {UsdUtilsValidatorNameTokens->packageEncapsulationValidator,
+            UsdUtilsValidatorNameTokens->layerFileFormatValidator};
 
     TF_AXIOM(validatorMetadataNameSet == expectedValidatorNames);
 }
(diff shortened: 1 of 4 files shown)


pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
```diff
@@ -114,11 +115,33 @@ TestPackageEncapsulationValidator()
...
+    UsdValidationRegistry& registry = UsdValidationRegistry::GetInstance();
+
+    // Verify the validator exists
+    const UsdValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->layerFileFormatValidator);
+
+    TF_AXIOM(validator);
+
+    // Create an empty layer with a core extension
+    const SdfLayerRefPtr passLayer = SdfLayer::CreateNew("pass.usda");
```
beersandrew, line 131: @dgovil @tallytalwar I wanted your opinions on what a reasonable failing test would be for this validation rule. It seems to me that I would need to install some other file format plugin in order for Sdf to allow me to use some non core usd extension for my layer, however then shipping and installing that plugin just to make this test better seems unreasonable. Any ideas?

dgovil: You could just rename a USDC file with an arbitrary extension, and then use the zipfile writer to author a file with it in it. I don't think your test needs to actually read the file within the zip though, just needs to use the zipfile reader to check the file names

beersandrew: That seems to work for the `ARKitFileExtensionChecker` which is helpful for when I get to that one however now I'm a little confused on the practical difference between `ARKitFileExtensionChecker`([complianceChecker code](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/complianceChecker.py#L925)) and `ArKitLayerChecker` ([complianceChecker code](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/complianceChecker.py#L766)).

[Here](https://github.com/beersandrew/usd-test-assets/tree/main/Validation/examples/ARLayerChecker) is my hello world example that triggers `ARKitFileExtensionChecker`, but I'm struggling to do something similar for specifically the `ArKitLayerChecker` since it seems to need to get an `SdfLayer` first.

dgovil: Weird. Those look identical to me. 

I guess taking a step back: 
- The files must have extensions that match the usdz spec
- The first file must be a valid usd type

I think they could be the same test to be honest?

beersandrew: I agree and think only `ARKitFileExtensionChecker` is necessary here, will add that PR tomorrow. 

I'll leave this PR open in case @tallytalwar or anyone else has a strong opinion to keep this / modify it in some way. 

Thanks!

beersandrew: https://github.com/PixarAnimationStudios/OpenUSD/pull/3368

jesschimein:
Filed as internal issue #USD-10318

pxr/usd/usdUtils/testenv/testUsdUtilsValidators.cpp
tallytalwar: Isn't the `ArKitFileExtensionChecker` also checking for image file extensions? (So seems like `ArKitLayerChecker` is a subset of `ArKitFileExtensionChecker`)? Maybe we can keep just the FileExtensionChecker and it return separate UsdValidationError to cater to both the layer (usda, usdc, usd, usdz) and bundled image extensions ("exr", "jpg", "jpeg", "png")?

tallytalwar: Right, looks like you are doing just that in #3368, just split the ValidationError to cater to layer extensions and image extensions?

I am fine with closing this PR (will leave it to you @beersandrew :D)

beersandrew: Yep this makes sense to me. I will:

1. close out this PR
2. adjust https://github.com/PixarAnimationStudios/OpenUSD/pull/3368 to have 2 errors 
3. ignore porting over the [TextureChecker](https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65424815)

Thank you both!

beersandrew:
Closing: this work will be included as part of https://github.com/PixarAnimationStudios/OpenUSD/pull/3368

PR #3365, opened 2024-10-16
Support OpenImageIO 3.x inside the `hioOiio` plugin

jessey-git:
### Description of Change(s)
The soon to be released OpenImageIO 3.x series drops a variety of old, deprecated, methods and types.

In the context of the imaging `hioOiio` plugin, this simply means to adjust the following [1]:
- The `ImageIOParameter` type alias is removed and just `ParamValue` should be used
- The `read_image` method must take in subimage, miplevel, and channel values explicitly

These changes are compatible with the OpenImageIO 2.x series and shouldn't present a problem for those not updating just yet.

[1] https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/Deprecations-3.0.md#imageioh

### Fixes Issue(s)
- Building against OIIO 3


jesschimein:
Filed as internal issue #USD-10319

jessey-git:
Closing since PR https://github.com/PixarAnimationStudios/OpenUSD/pull/3418 was done in its place.

PR #3369, opened 2024-10-17
Ensure HdDependenciesSchema::GetEntries is linkable

robpieke:
Adding `HD_API` to `HdDependenciesSchema::GetEntries` so that the symbol will be found on Windows.

### Description of Change(s)

Without this change, Windows-based users of `HdDependenciesSchema` are unable to use the `GetEntries` method as the symbol isn't "exposed" (I don't know the strictly proper terminology)


jesschimein:
Filed as internal issue #USD-10327

PR #3370, opened 2024-10-21
[hdEmbree] ensure we respect PXR_WORK_THREAD_LIMIT (standalone)

pmolodo:
> [!IMPORTANT] 
> This is a duplicate of this PR:
>
> - https://github.com/PixarAnimationStudios/OpenUSD/pull/3198
>
> ...except pulled out of the UsdLux-HdEmbree PR "chain", to make for easier standalone integration, as it has no other PR dependencies, and is a simple/small change.
>
> I will close this PR if the other is merged first, and vice-versa.

### Description of Change(s)

Fixes an issue where pre-oneTBB, threads spawned from the non-root thread (such as the render thread) would not respect PXR_WORK_THREAD_LIMIT

### Fixes Issue(s)
- renderThread-spawned processes don't respect PXR_WORK_THREAD_LIMIT

### DIFF
--- a/pxr/imaging/plugin/hdEmbree/renderer.cpp
+++ b/pxr/imaging/plugin/hdEmbree/renderer.cpp
@@ -22,6 +22,87 @@
 #include <chrono>
 #include <thread>
 
+// -------------------------------------------------------------------------
+// Old TBB workaround - can remove once OneTBB is mandatory
+// -------------------------------------------------------------------------
+#include <tbb/tbb_stddef.h>
+
+#if TBB_INTERFACE_VERSION_MAJOR < 12
+
+#include <memory>
+
+#include <tbb/task_scheduler_init.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+// PXR_WORK_THREAD_LIMIT isn't exported as part of it's api, and we're not
+// part of the work library, so we can't use:
+//     extern TfEnvSetting<int> PXR_WORK_THREAD_LIMIT;
+extern std::variant<int, bool, std::string> const *
+Tf_GetEnvSettingByName(std::string const&);
+
+PXR_NAMESPACE_CLOSE_SCOPE
+
+namespace {
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+// This function always returns either 0 (meaning "no change") or >= 1
+//
+// Duplication of code from threadLimits.cpp - copied here to avoid having to
+// change API of work lib.  Will go away once we don't need to support tbb < 12
+// (ie, pre-OneTBB)
+static unsigned
+HdEmbree_NormalizeThreadCount(const int n)
+{
+    // Zero means "no change", and n >= 1 means exactly n threads, so simply
+    // pass those values through unchanged.
+    // For negative integers, subtract the absolute value from the total number
+    // of available cores (denoting all but n cores). If |n| >= number of cores,
+    // clamp to 1 to set single-threaded mode.
+    return n >= 0 ? n : std::max<int>(1, n + WorkGetPhysicalConcurrencyLimit());
+}
+
+
+// Returns the normalized thread limit value from the environment setting. Note
+// that 0 means "no change", i.e. the environment setting does not apply.
+//
+// Duplication of code from threadLimits.cpp - copied here to avoid having to
+// change API of work lib.  Will go away once we don't need to support tbb < 12
+// (ie, pre-OneTBB)
+static unsigned
+HdEmbree_GetConcurrencyLimitSetting()
+{
+    std::variant<int, bool, std::string> const *
+        variantValue = Tf_GetEnvSettingByName("PXR_WORK_THREAD_LIMIT");
+    int threadLimit = 0;
+    if (int const *value = std::get_if<int>(variantValue)) {
+        threadLimit = *value;
+    }
+    return HdEmbree_NormalizeThreadCount(threadLimit);
+}
+
+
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10340

pmolodo:
Updated to not modify the API of the work library

tcauchois:
Hey Paul, just to clarify: the root issue here is that we're spawning TBB threads from the hydra render thread, which does not have a tbb::task_scheduler_init because libWork only constructs one on the main thread at static initialization time?  Is that roughly correct?  I'm not sure having it be scoped is quite right; the render thread is long lived and we should be able to just initialize one the first time we get into the callback and keep it around.  I don't have a sense of how heavy they are to construct, though.

We will likely need to tweak this slightly to make it oneTBB-safe; I can provide suggestions on github, or just fix it up inline as I pull it down; let me know.

pmolodo:
> We will likely need to tweak this slightly to make it oneTBB-safe

Things already work with oneTBB - it handles the thread limits in a different / better  / more global way (though I no longer recall the exact details at this moment - can find it if you want, though).

This PR should be `#if` guarded to only change things for "old" TBB.

> Hey Paul, just to clarify: the root issue here is that we're spawning TBB threads from the hydra render thread, which does not have a tbb::task_scheduler_init because libWork only constructs one on the main thread at static initialization time? Is that roughly correct? I'm not sure having it be scoped is quite right; the render thread is long lived and we should be able to just initialize one the first time we get into the callback and keep it around. I don't have a sense of how heavy they are to construct, though.

Yeah, it's been a minute since I wrote this / investigated, but that sounds about right.  And absolutely, I think the "correct" thing to do would be to fix this in a more general way at a higher level - ie, libwork.  However, since this issue only affects "old" TBB (things already work nicely in oneTBB), and I was a bit leery of making changes to `libwork` (other than the documentation-only commit I added), I figured I'd just make my tweaks to `hdEmbree`.

If you want to fix it "right" at a higher level, though, go for it!

>  I can provide suggestions on github, or just fix it up inline as I pull it down; let me know.

Fixing inline as you go is fine with me!

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -22,6 +22,87 @@
...
+
+
+// Returns the normalized thread limit value from the environment setting. Note
+// that 0 means "no change", i.e. the environment setting does not apply.
+//
+// Duplication of code from threadLimits.cpp - copied here to avoid having to
+// change API of work lib.  Will go away once we don't need to support tbb < 12
+// (ie, pre-OneTBB)
+static unsigned
+HdEmbree_GetConcurrencyLimitSetting()
```
tcauchois, line 74: I think I have consensus internally to move this function back into "work".  If I make that change, do you want me to post it here for review or just check it in?

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -22,6 +22,87 @@
...
+
+// Make the calling context respect PXR_WORK_THREAD_LIMIT, if run from a thread
+// other than the main thread (ie, the renderThread)
+class _ScopedThreadScheduler {
+public:
+    _ScopedThreadScheduler() {
+        auto limit = HdEmbree_GetConcurrencyLimitSetting();
+        if (limit != 0) {
+            _tbbTaskSchedInit =
+                std::make_unique<tbb::task_scheduler_init>(limit);
```
tcauchois, line 94: I really wanted to find a way to jam this back into "work", but we're also simultaneously abstracting "work" to be able to use oneTBB as well as non-TBB threadpools, and there weren't any API contracts we felt good about adding.  So I think keeping the tbb::task_scheduler_init here is the lesser of two evils; the only negative side effect it has is creating a hard dependency on TBB, which is fine for a few years I think.

I don't think the task_scheduler_init can be scoped like you're doing here, but to give it the lifetime of the thread we'd need to either add it to HdRenderThread directly or subclass HdRenderThread, which seems like too much work, so I'm fine with leaving this as-is.

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -453,6 +534,9 @@ HdEmbreeRenderer::Render(HdRenderThread *renderThread)
 
         // Render by scheduling square tiles of the sample buffer in a parallel
         // for loop.
+#if TBB_INTERFACE_VERSION_MAJOR < 12
```
tcauchois, line 537: See comment above about scoping; anyway, at the least this should probably move to the top of the function.  I've done this internally but can post it back if you'd like.

pxr/imaging/plugin/hdEmbree/renderer.cpp
```diff
@@ -22,6 +22,87 @@
 #include <chrono>
 #include <thread>
 
+// -------------------------------------------------------------------------
+// Old TBB workaround - can remove once OneTBB is mandatory
```
tcauchois, line 26:
```suggestion
// Old TBB workaround - we plan to remove this once OpenUSD adopts
// oneTBB as a min spec. Note: for non-TBB-based work implementations,
// this code shouldn't cause any trouble.
```

tcauchois (commented):
A few things that came up in internal review of this.

Let me know (1) what you think of the changes; and (2) whether I should check the modified version in and then send you a commit link, or bring it back to github as suggestions for you to apply to the PR or what.

PR #3373, opened 2024-10-22
hio: Add CMake options to handled image file formats by plugins

theblackunknown:
### Description of Change(s)

This PR introduces CMake options to determine which image file formats are handled by each HioImage plugins.
Options defaults to the same behavior as before this PR for backward compatibility.

Those new options allow for an integration to control which plugins should handled which file formats.
For example it is possible to disable `Hio_StbImage` and let `Hio_Oiio` handle all image formats.

### DIFF
--- a/cmake/defaults/Options.cmake
+++ b/cmake/defaults/Options.cmake
@@ -240,3 +240,12 @@ if (${PXR_BUILD_PYTHON_DOCUMENTATION})
         set(PXR_BUILD_PYTHON_DOCUMENTATION "OFF" CACHE BOOL "" FORCE)
     endif()
 endif()
+
+set(PXR_IMAGING_HIO_STB_PRECEDENCE "30" CACHE STRING "Hio plugin precedence for Hio_StbImage.")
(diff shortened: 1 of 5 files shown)


meshula:
Hi @theblackunknown!

The built in mechanism for this is the pluginfo.json priority scheme. If you add exr and stb to your pluginfo for HioOIIO, and assign a higher priority there, you can get what you are after without the need to conditionally compile the library.

jesschimein:
Filed as internal issue #USD-10343

theblackunknown:
> Hi @theblackunknown!
> 
> The built in mechanism for this is the pluginfo.json priority scheme. If you add exr and stb to your pluginfo for HioOIIO, and assign a higher priority there, you can get what you are after without the need to conditionally compile the library.

Good observation !
This is partly why I have exposed both supported image file formats and precedence for each plugin, so I would say that this PR also support implementing what you suggest.

Another part which interest my use case is to entirely avoid embedding the vendored STB implementation: as a single source library the USD vendored one is out of sync with other part of our system also using STB.
And because `hio` embeds `Hio_StbImage` by default we have no way to remove it from a build.

PR #3398, opened 2024-11-04
usdShade: Add ShaderValidator

beersandrew:
### Description of Change(s)
- Added equivalent validation checks from ARShaderChecker in the new ValidationFramework, including validation implementation code, test code, tokens, and an entry in plugInfo.json

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65424945

### Checklist

### DIFF
--- a/pxr/usd/usdShade/validators.cpp
+++ b/pxr/usd/usdShade/validators.cpp
@@ -590,6 +590,157 @@ _SubsetsMaterialBindFamily(const UsdPrim& usdPrim)
     return errors;
 }
 
+static
+UsdValidationErrorVector
+_ShaderValidator(const UsdPrim& usdPrim)
+{
+    if (!usdPrim.IsA<UsdShadeShader>())
+    {
+        return {};
+    }
+
+    const UsdShadeShader shaderPrim(usdPrim);
+
+    const TfToken &implementationSource =
+        shaderPrim.GetImplementationSource();
+
+    UsdValidationErrorVector errors;
+    const UsdValidationErrorSites primErrorSite = {
+        UsdValidationErrorSite(usdPrim.GetStage(),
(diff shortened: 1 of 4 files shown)


pxr/usd/usdShade/validators.cpp
```diff
@@ -590,6 +590,157 @@ _SubsetsMaterialBindFamily(const UsdPrim& usdPrim)
...
+                UsdShadeValidationErrorNameTokens->multipleConnectionSources,
+                UsdValidationErrorType::Error,
+                primErrorSite,
+                TfStringPrintf(
+                "Shader input <%s> has %lu connection "
+                        "sources, but only one is allowed.",
+                    shaderInput.GetAttr().GetPath().GetText(),
+                    sources.size())
+            );
+            continue;
```
beersandrew, line 695: @tallytalwar one note here, I noticed this was not present in the other validator [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/complianceChecker.py#L871) however:
1. We do get a warning below when calling `GetConnectedSource` since with multiple sources we should call `GetConnectedSources`
2. It appears `GetConnectedSource` gets the first connected source, so the complianceChecker code would not be validating issues for other connected sources

Happy to change this and loop through the connected sources if you think it's better to add that logic in, it made sense to me to stop checking since no matter what the connections were, it seems invalid to have more than one. Let me know if that's incorrect.

jesschimein:
Filed as internal issue #USD-10396

beersandrew:
Closing this PR, this work is now done on https://github.com/PixarAnimationStudios/OpenUSD/pull/3447

PR #3403, opened 2024-11-06
usdUtils: Added MissingReferenceValidator

beersandrew:
### Description of Change(s)
- created implementation, test, as well as associated entries in plugInfo.json & validatorTokens.h

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65424971

### Checklist

### DIFF
--- a/pxr/usd/usdUtils/validators.cpp
+++ b/pxr/usd/usdUtils/validators.cpp
@@ -104,9 +104,44 @@ _PackageEncapsulationValidator(const UsdStagePtr& usdStage) {
     return errors;
 }
 
+static
+UsdValidationErrorVector
+_MissingReferenceValidator(const UsdStagePtr& usdStage) {
+    const SdfLayerRefPtr& rootLayer = usdStage->GetRootLayer();
+
+    SdfLayerRefPtrVector layers;
+    std::vector<std::basic_string<char>> assets, unresolvedPaths;
+    const SdfAssetPath& path = SdfAssetPath(rootLayer->GetIdentifier());
+
+    UsdUtilsComputeAllDependencies(path, &layers, &assets, &unresolvedPaths,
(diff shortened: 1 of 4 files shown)


pxr/usd/usdUtils/validators.cpp
```diff
@@ -104,9 +104,44 @@ _PackageEncapsulationValidator(const UsdStagePtr& usdStage) {
...
+            UsdValidationErrorType::Error,
+            UsdValidationErrorSites {
+                UsdValidationErrorSite(
+                    rootLayer, SdfPath(unresolvedPath))
+            },
+            TfStringPrintf(
+            ("Found unresolvable external dependency "
+                "'%s'."), unresolvedPath.c_str())
+        );
+    }
```
beersandrew, line 133: @tallytalwar One note here, it looks like `complianceChecker.py` is doing another check [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L181) which gets some diagnostic information, but it seems like this process of collecting diagnostic information should be started before opening the stage ([here](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L1086)). I tried doing this inside the validator with stage.Reload() but was unable to populate the diagnostics this way.

In my specific test, the errors were similar so I wanted your opinion on if it's possible to do something similar here, or if it's necessary to move the diagnostic part over in this case.

jesschimein:
Filed as internal issue #USD-10405

beersandrew:
Closing this PR, this work is now done on https://github.com/PixarAnimationStudios/OpenUSD/pull/3450

Issue #3401, opened 2024-11-05
Could NOT find Python3 (missing: Development Development.Embed) (found   version "3.10.15")


meghbhalerao:
### Description of Issue
I am facing the following error - 
```bash
Building with settings:
  USD source directory          /home/mb230/projects/pkgs/OpenUSD
  USD install directory         /home/mb230/projects/pkgs/build_usd
  3rd-party source directory    /home/mb230/projects/pkgs/build_usd/src
  3rd-party install directory   /home/mb230/projects/pkgs/build_usd
  Build directory               /home/mb230/projects/pkgs/build_usd/build
  CMake generator               Default
  CMake toolset                 Default
  Downloader                    built-in

  Building                      Shared libraries
    Variant                     Release
    Target                      
    Imaging                     On
      Ptex support:             Off
      OpenVDB support:          Off
      OpenImageIO support:      Off 
      OpenColorIO support:      Off 
      PRMan support:            Off
    UsdImaging                  On
      usdview:                  On
    MaterialX support           On
    Python support              On
      Python Debug:             Off
      Python docs:              Off
    Documentation               Off
    Tests                       Off
      Mayapy Tests:             Off
      AnimX Tests:              Off
    Examples                    On
    Tutorials                   On
    Tools                       On
    Alembic Plugin              Off
      HDF5 support:             Off
    Draco Plugin                Off

  Dependencies                  None
STATUS: Installing USD...
2024-11-05 16:38
[...]

PR #3404, opened 2024-11-06
Fix fallthrough on failed lookup of pythonLibPath (Linux)

dgovil:
### Description of Change(s)

The Python library name lookup here uses a heuristic that is potentially fallible.
However it currently falls through if the path doesn't exist at all.

Additionally none of the paths are verified before returning, so I added some asserts for that.

I don't think this will fix https://github.com/PixarAnimationStudios/OpenUSD/issues/3401 but it will likely help reduce some of the debugging burden in the future by failing earlier in a clearer way.

### Checklist

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -234,6 +234,10 @@ def GetPythonInfo(context):
     else:
         raise RuntimeError("Platform not supported")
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10408

Issue #3414, opened 2024-11-07
Crash saving USD file on Windows to removable volume mounted as NTFS folder


marktucker:
### Description of Issue
Tf_HasAttribute on Window has the potential to dive into an infinite recursion when faced with a directory on an NTFS volume that has been used as the mount point for a removable volume (USB drive, for example).

### Steps to Reproduce
1. On Windows, plug in a USB drive that is safe to reformat/delete all contents.
2. From the Start Menu Choose "Create and Format Hard Disk Partitions".
3. Find the USB Drive, RMB on it, and Choose "Delete Volume". Confirm the deletion.
4. RMB on the unallocated USB volume and choose "New Simple Volume".
5. On the "Assign Drive Letter or Path" page, set the USB volume to mount on "C:\UsbDrive" (C: has to be NTFS formatted to allow this).
6. Accept the rest of the dialogs to reformat the USB drive, and verify that C:\UsbDrive exists and can be written to.
7. Run python in a USD-capable environment and run the following code:
```
from pxr import Usd
s = Usd.Stage.CreateInMemory()
s.Export("C:/UsbDrive/output.usda")
```
8. Python will run for a long time, and eventually crash with a stack overflow.

### System Information (OS, Hardware)
Windows only.

### Package Versions
USD 24.03 and 24.08 have both been tested and fail the same way.

jesschimein:
Filed as internal issue #USD-10418

PR #3415, opened 2024-11-07
Fix infinite recursion writing to removable drives mounted to NTFS folders

marktucker:
Fix another issue with unresolvable reparse points on Windows. ArchReadLink
can return the passed in path if it refers to an "NT Object Manager" path, and we need to handle this as if it isn't a reparse point, because we can't follow the link.

This generalizes the fix by @seando-adsk in f6c2e82, by handling mounted volumes on Windows and avoiding infinitely recursive calls to TfReadLink by having TfIsLink return false in cases where this bad behavior could arise.

### Description of Change(s)
Changed Tf_HasAttribute to remove the reparse point attribute on a directory if TfReadLink returns the original path or an empty string (i.e. we couldn't follow the link). This means TfIsLink returns false in these cases, preventing recursive calls to TfReadLink because we always call TfIsLink before calling TfReadLink.

### Fixes Issue(s)
#3414

### Checklist

[ ] I have added unit tests that exercise this functionality
Unit testing this is basically impossible given the complex steps required to set up the USB drive appropriately.

[ X ] I have verified that all unit tests pass with the proposed changes

### DIFF
--- a/pxr/base/tf/fileUtils.cpp
+++ b/pxr/base/tf/fileUtils.cpp
@@ -80,13 +80,22 @@ Tf_HasAttribute(
 
     // Ignore reparse points on network volumes. They can't be resolved
     // properly, so simply remove the reparse point attribute and treat
-    // it like a regular file/directory.
+    // it like a regular file/directory. Also ignore reparse points where
+    // TfReadLink returns the passed in path. As described in ArchReadLink,
+    // this will happen in cases where the reparse points are NT Object
+    // Manager paths, which cannot be used as file paths. Or if TfReadLink
+    // returns an empty string, that means we could not resolve the link
+    // at all, so treat it as if it is not a link.
+    std::string linkPath;
     if ((attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0) {
         // Calling PathIsNetworkPath sometimes sets the "last error" to an
         // error code indicating an incomplete overlapped I/O function. We
         // want to ignore this error.
         DWORD olderr = GetLastError();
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10419

pxr/base/tf/fileUtils.cpp
```diff
@@ -106,8 +115,7 @@ Tf_HasAttribute(
     }
 
     // Read symlinks until we find the real file.
-    return Tf_HasAttribute(TfReadLink(path.c_str()), resolveSymlinks,
-                           attribute, expected);
+    return Tf_HasAttribute(linkPath, resolveSymlinks, attribute, expected);
```
anwang2009, line 124: It's a bit tricky to tell that linkPath is non-empty from reading the code; it'd be great if that could be made more obvious

anwang2009 (commented):
Thanks @marktucker, this PR looks great - just one comment

pxr/base/tf/fileUtils.cpp
marktucker: Is an additional comment around line 117 sufficient? Something like:

// At this point we know (attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0
// or we would have returned in the if block above. This means linkPath will
// be holding the result of calling TfReadLink(path.c_str()). The code is separated
// in this way to avoid calling TfReadLink twice.

anwang2009:
Yeah, that’s fair - thanks!

Issue #3285, opened 2024-09-10
Boost build config problem


teachmain:
### Description of Issue
FindBoost module was removed from cmake since version 3.30, see [CMP0167](https://cmake.org/cmake/help/latest/policy/CMP0167.html).  I think the flag `Boost_NO_BOOST_CMAKE` should be OFF by default ? or find another solution.

### System Information (OS, Hardware)
Archlinux 
### Package Versions
cmake 3.30.3 
openusd release 24.08

jesschimein:
Filed as internal issue #USD-10101

sunyab:
I believe this has been addressed by 8d149bf and 9b566e659fce6bb46ede4723e7dc4c9e0a173893 currently in the dev branch. Closing this out.

Issue #3384, opened 2024-10-29
New USD tests fail on Python 3.11 due to changed annotations


dgovil:
### Description of Issue

We are noticing an issue with running some of the embedded Boost python tests when building USD for Python 3.11 (pxr/external/boost/python/test). These tests utilize doctest which executes a documentation block and compares the script output directly. Unfortunately, the output is not consistent between Python versions, so some of the tests fail on Python 3.11 (presumably the expected output was generated with 3.10 where we are seeing the tests pass).
Here is an example test that is failing:
https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/external/boost/python/test/properties.py

```
Failed example:
    x1.value_r = 2
Expected:
    Traceback (most recent call last):
      File "properties.py", line 49, in ?
        x1.value_r = 2
    AttributeError: can't set attribute
Got:
    Traceback (most recent call last):
      File ".../doctest.py", line 1355, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest __main__[16]>", line 1, in <module>
        x1.value_r = 2
        ^^^^^^^^^^
    AttributeError: property of 'X' object has no setter

```
The three tests we are seeing fail in Python 3.11 are:

- python_properties
- python_pickle1
- python_pickle4

Python 3.11 had a lot of work done to make tracebacks more friendly to devs, so the tests don't apply to it the same way.

jesschimein:
Filed as internal issue #USD-10369

sunyab:
Thanks @dgovil! It looks like there are two issues here:

- `python_properties` is failing due to the traceback differences in 3.11. Thankfully doctest has facilities for ignoring these differences.
- I think `python_pickle1` and `python_pickle4` are failing because of the change to `__getstate__` in 3.11 (see [docs](https://docs.python.org/3/library/pickle.html#object.__getstate__)). That's causing `__reduce__` to include an extra `None` value in the returned tuple. Functionally, I don't think this matters since the [docs for __reduce__](https://docs.python.org/3/library/pickle.html#object.__reduce__) say that `None` is an acceptable value.

Issue #3055, opened 2024-04-24
`usdchecker` tests depend on `usdShaders` plugin, causing test failures when building with `--no-imaging`


nvmkuruc:
### Description of Issue
`usdchecker` now has tests that depend on `usdShaders`, introducing an (indirect) dependency on `imaging`.

This means that tests cannot successfully complete when building without imaging.

### Steps to Reproduce
1. `python3 ./build_scripts/build_usd.py --no-imaging --no-materialx --no-examples --tests <INSTALL DIRECTORY>`
2. Run `ctest -R testUsd --output-on-failure` in the `<INSTALL DIRECTORY>/build/usd` directory
3. The following tests should fail.
```
The following tests FAILED:
        571 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
        610 - testUsdChecker1 (Failed)
        611 - testUsdChecker2 (Failed)
        612 - testUsdChecker3 (Failed)
        613 - testUsdChecker4 (Failed)
        614 - testUsdChecker5 (Failed)
        622 - testUsdChecker13 (Failed)
```
The error messages should all be associated with the `ShaderPropertyTypeConformanceChecker`
```
Shader </World/material/nodeGraph/PreviewShader> has invalid shader node.  (fails 'ShaderPropertyTypeConformanceChecker')
```

### System Information (OS, Hardware)
WSL2

jesschimein:
Filed as internal issue #USD-9590

PR #3430, opened 2024-11-19
Add support for building for Windows ARM64 devices

anthony-linaro:
### Description of Change(s)

Add support for Windows ARM64 devices.

This was tested using the following command line:
```
python build_scripts/build_usd.py --no-python --onetbb --tests --no-imaging C:/WoA/OpenUSD/build_dir
```

Python is currently switched off due to lack of support for PySide on these devices. I am looking at that seperately.

I also had to disable imaging, as building boost appears to be broken when using newer CMake versions (similar issue to #3285)

TBB is set to OneTBB, as 2020u3 doesn't support these devices.

I also had to make a small modification in `testWrapper.py`, and change line 134 to `pathPattern = pathPattern + r'(\S*)'`, as otherwise the escape sequence was invalid.

### Checklist

~~[ ] I have added unit tests that exercise this functionality (Reference: 
[testing guidelines](https://openusd.org/release/api/_page__testing__guidelines.html))~~


jesschimein:
Filed as internal issue #USD-10450

anthony-linaro:
CLA sorted - approved by Tim Palmer

pmolodo:
Any reason why the testWrapper.py change you mentioned isn't included here? Otherwise, lgtm

anthony-linaro:
I figured it was more related to the version of python I have installed (`3.12.5`), rather than the one used for validation on other devices - as I didn't have those devices/versions to hand, and knew the script already worked as-was on them, I opted not to change it

build_scripts/build_usd.py
```diff
@@ -393,7 +393,10 @@ def RunCMake(context, force, extraArgs = None):
 
     # Note - don't want to add -A (architecture flag) if generator is, ie, Ninja
     if IsVisualStudio2019OrGreater() and "Visual Studio" in generator:
-        generator = generator + " -A x64"
+        if "ARM" in os.environ.get('PROCESSOR_IDENTIFIER'):
```
nvmkuruc: Could the `platform` module be used here?
https://docs.python.org/3/library/platform.html#module-platform

Mushroom: It could be, but problems arise when running an emulated x64 version of python (which is easy to do by mistake, as that's what the website gives you by default).

This way checks the name of the processor directly, which is not affected by emulation.

anthony-linaro: (repost from correct account)

It could be, but problems arise when running an emulated x64 version of python (which is easy to do by mistake, as that's what the website gives you by default).

This way checks the name of the processor directly, which is not affected by emulation.

asluk:
"Python is currently switched off due to lack of support for PySide on these devices. I am looking at that seperately."

Hi @anthony-linaro , per our call yesterday, you also mentioned that the python bindings are failing to compile -- can you post the error messages you are seeing there? 

"I also had to disable imaging, as building boost appears to be broken... "

@sunyab , is it expected that the command line

`python build_scripts/build_usd.py --no-python --onetbb --tests C:/WoA/OpenUSD/build_dir
`

would attempt to build boost in v24.11+ ?

Thanks all!

anthony-linaro:
Hi @asluk - I have tried to reproduce the compilation errors, but am unable, so will chalk it down to an env issue - with things set up properly, I have most of the tests passing.

However, when I run `ctest`, I get this:
```
The following tests FAILED:
          2 - python_properties (Failed)
         47 - python_pickle1 (Failed)
         50 - python_pickle4 (Failed)
        225 - testPlug (Failed)
        666 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
        708 - testUsdChecker1 (Failed)
        709 - testUsdChecker2 (Failed)
        710 - testUsdChecker3 (Failed)
        711 - testUsdChecker4 (Failed)
        712 - testUsdChecker5 (Failed)
        720 - testUsdChecker13 (Failed)
        778 - testUsdResolverExample (Failed)
```

I have attached the output file from the tests to this comment. Are these tests expected to fail? If not, do you have any idea why they might be? All env paths are set as instructed after build (`bin` and `lib` on `path`, `python` on `pythonpath`)

[LastTest.log](https://github.com/user-attachments/files/18346860/LastTest.log)

asluk:
@anthony-linaro  I don't think those test failures are expected -- can you try rebasing your branch to be on top of the "dev" branch? That is the most up to date commit. Thanks and happy new year!

sunyab:
> @sunyab , is it expected that the command line
> 
> `python build_scripts/build_usd.py --no-python --onetbb --tests C:/WoA/OpenUSD/build_dir `
> 
> would attempt to build boost in v24.11+ ?
> 

(sorry for the late reply, was out on vacation)

Yes, I would expect that to build boost in v24.11+. When imaging is enabled the `--tests` option requires the `idiff` tool from OpenImageIO for image diffs, and the version of OIIO the build script uses still requires boost.

sunyab:
> However, when I run `ctest`, I get this:
> 
> ```
> The following tests FAILED:
>           2 - python_properties (Failed)
>          47 - python_pickle1 (Failed)
>          50 - python_pickle4 (Failed)
>         225 - testPlug (Failed)
>         666 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
>         708 - testUsdChecker1 (Failed)
>         709 - testUsdChecker2 (Failed)
>         710 - testUsdChecker3 (Failed)
>         711 - testUsdChecker4 (Failed)
>         712 - testUsdChecker5 (Failed)
>         720 - testUsdChecker13 (Failed)
>         778 - testUsdResolverExample (Failed)
> ```

Hi @anthony-linaro, the `python_*` test failures are due to #3384, which is fixed in the dev branch. The other failures don't look familiar to me, so I'm not sure what's going on there.

anthony-linaro:
Hi both, thanks for the responses!

@sunyab You're right, those python issues are now fixed, thanks to the suggestion from @asluk to merge dev.

Sadly, the list of errors has now changed, and I now get this:
```
The following tests FAILED:
        225 - testPlug (Failed)
        656 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
        764 - testUsdChecker1 (Failed)
        765 - testUsdChecker2 (Failed)
        766 - testUsdChecker3 (Failed)
        767 - testUsdChecker4 (Failed)
        768 - testUsdChecker5 (Failed)
        776 - testUsdChecker13 (Failed)
        777 - testUsdChecker14 (Failed)
        778 - testUsdChecker15 (Failed)
        779 - testUsdChecker16 (Failed)
        780 - testUsdChecker17 (Failed)
        781 - testUsdChecker18 (Failed)
        788 - testUsdChecker26 (Failed)
        789 - testUsdChecker27 (Failed)
        799 - testUsdChecker37 (Failed)
        802 - testUsdResolverExample (Failed)
```

I have once again attached the logs: [LastTest.log](https://github.com/user-attachments/files/18350647/LastTest.log)

Any suggestions of where I could start to look? It looks like it's not seeing the "Tf" python module. The path variable is correct, and added as specified when building (ie, modified PATH and PYTHONPATH) - is there another one that got missed somewhere? Was there an extra step required between `build_usd.py` and `ctest` for python to be able to see the module?

build_scripts/build_usd.py
dgovil: It might be good to follow the convention in apple_utils.py and extract this into GetHostArch https://github.com/PixarAnimationStudios/OpenUSD/blob/release/build_scripts/apple_utils.py#L66

build_scripts/build_usd.py
```diff
@@ -393,7 +393,10 @@ def RunCMake(context, force, extraArgs = None):
...
+            generator = generator + " -A arm64"
```
dgovil: I'm not super familiar with how its configured, but you might also need to support ARM64EC on Windows too ? https://learn.microsoft.com/en-us/windows/arm/arm64ec

pxr/base/arch/defines.h
```diff
@@ -35,15 +35,17 @@
 #if defined(i386) || defined(__i386__) || defined(__x86_64__) || \
     defined(_M_IX86) || defined(_M_X64)
 #define ARCH_CPU_INTEL
-#elif defined(__arm__) || defined(__aarch64__) || defined(_M_ARM)
+#elif defined(__arm__) || defined(__aarch64__) || defined(_M_ARM) || \
+    defined(_M_ARM64)
```
mattyjams, line 39: Along the lines of @dgovil's comment in `build_usd.py`, I think we'd also want to include `defined(_M_ARM64EC)` here and below?

anthony-linaro:
I didn't add support for ARM64EC, as I didn't have a use-case for it - if
you guys have one internally I'm unaware of, I can try and add support.

The biggest issue with ARM64EC will likely be around the use of any
intrinsics - as it defines both _M_ARM64EC, and _M_X64, use of intrinsics
that could be considered "safely" gated behind _M_X64 (or similar) are
suddenly not so safe, so I'd need to go over the source with a fine tooth
comb. Not impossible, but requires substantially more work than a plain old
ARM64 enablement.

dgovil:
That's fair. At the very least, I'd suggest then adding a caveat to the README somewhere to document that EC isn't supported yet?

build_scripts/build_usd.py
anthony-linaro: Is the suggestion here to write a whole new `windows_utils.py` (and extract/refactor some things like `IsVisualStudioXXXXOrGreater()` which are windows only), or just add a function to `build_usd.py` called something along the lines of `GetWindowsHostArch()`?

anthony-linaro:
Good idea - I have updated the README.md file to reflect the work done on this PR

build_scripts/build_usd.py
dgovil: I'm not sure I'd advocate for another file necessarily (though it is the route I would take myself)

But just extracting the heuristics for detecting arm into a function somewhere in this file itself should be fine?

I just imagine that in the future, we'll need to modify the build parameters for certain dependencies etc based on the current arch, and so it'd be nice to have it as a function from the get go.

anthony-linaro: I have now moved it out to a seperate function - it should now reliably detect ARM or x64

Critique very much welcome - I've never been much of a python dev

asluk:
> Hi both, thanks for the responses!
> 
> @sunyab You're right, those python issues are now fixed, thanks to the suggestion from @asluk to merge dev.
> 
> Sadly, the list of errors has now changed, and I now get this:
> 
> ```
> The following tests FAILED:
>         225 - testPlug (Failed)
>         656 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
>         764 - testUsdChecker1 (Failed)
>         765 - testUsdChecker2 (Failed)
>         766 - testUsdChecker3 (Failed)
>         767 - testUsdChecker4 (Failed)
>         768 - testUsdChecker5 (Failed)
>         776 - testUsdChecker13 (Failed)
>         777 - testUsdChecker14 (Failed)
>         778 - testUsdChecker15 (Failed)
>         779 - testUsdChecker16 (Failed)
>         780 - testUsdChecker17 (Failed)
>         781 - testUsdChecker18 (Failed)
>         788 - testUsdChecker26 (Failed)
>         789 - testUsdChecker27 (Failed)
>         799 - testUsdChecker37 (Failed)
>         802 - testUsdResolverExample (Failed)
> ```
> 
> I have once again attached the logs: [LastTest.log](https://github.com/user-attachments/files/18350647/LastTest.log)
> 
> Any suggestions of where I could start to look? It looks like it's not seeing the "Tf" python module. The path variable is correct, and added as specified when building (ie, modified PATH and PYTHONPATH) - is there another one that got missed somewhere? Was there an extra step required between `build_usd.py` and `ctest` for python to be able to see the module?

I think the testUsdResolverExample failure is expected, and/or might need the example resolver to be bootstrapped in PATH and PYTHONPATH as well (@sunyab or @pmolodo  , do you remember the incantation for that?)

@tallytalwar , looks like the usdchecker rules are failing (also affects testUsdUtilsCreateNewUsdzPackageEditInPlace), but I'm not sure why-- does anything in the log stand out to you or have a clear root cause in your experience? (e.g., Shader </World/simpleMaterial/boardMat/PBRShader> has invalid shader node.  (fails 'ShaderPropertyTypeConformanceChecker'))

It's not clear to me what the testPlug failure is.

tallytalwar:
@asluk all the usdchecker tests (both old <13 and new framework >=13) are failing because its not able to find any of the usdShaders, example UsdPreviewSurface etc in sdr:

New compliance checker failing here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usdValidation/usdShadeValidators/validators.cpp#L341-L353
Old compliance checerk failing here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L712

Both these point to a build issue which is not picking up some plugins example, usdShaders here: https://github.com/PixarAnimationStudios/OpenUSD/tree/dev/pxr/usdImaging/plugin/usdShaders

Maybe the testPlug failure also points to that :/

asluk:
Thanks @tallytalwar ! Is there something additional in PATH or PYTHONPATH etc that @anthony-linaro  needs to do to detect the shaders?

@anthony-linaro , can you post the relevant bits from your PATH and PYTHONPATH when running tests?

tallytalwar:
Ahh I see.. I think the "--no-imaging" arg to the build script doesn't build pxr/usdImaging which is where these shader plugins live.

From the PR description:

> python build_scripts/build_usd.py --no-python --onetbb --tests --no-imaging C:/WoA/OpenUSD/build_dir

> I also had to disable imaging, as building boost appears to be broken when using newer CMake versions (similar issue to https://github.com/PixarAnimationStudios/OpenUSD/issues/3285)

asluk:
Thanks @tallytalwar .... @anthony-linaro , those failures are expected then ... @tallytalwar , is it worth a separate PR disabling some cases in the usdchecker tests when Imaging is not enabled in the build?

tallytalwar:
@asluk yes definitely. I was planning to do that, but a PR will also be appreciated. Will sync it in for internal consumption promptly.

anthony-linaro:
So, we would consider this ready to go then? Or is there a requirement to fix the imaging first?

asluk:
> So, we would consider this ready to go then? Or is there a requirement to fix the imaging first?

@anthony-linaro  can you try running just testPlug to see if you can get more output from it? something like ctest -R testPlug

Overall I think this is generally ready to go, but defer to @sunyab  as to next steps. Thanks all!

anthony-linaro:
@asluk Full output:
```
C:\WoA\OpenUSD\build_dir\build\OpenUSD>ctest -C Release -R testPlug --output-on-failure
Test project C:/WoA/OpenUSD/build_dir/build/OpenUSD
    Start 226: testPlug
1/1 Test #226: testPlug .........................***Failed    0.21 sec
Warning: in _DeclareAliases at line 526 of C:\WoA\OpenUSD\pxr\base\plug\plugin.cpp -- Expected string for alias name, but found int
.Error: return code 3221225477 doesn't match expected 0 (EXPECTED_RETURN_CODE).chdir: C:\Users\Anthony\AppData\Local\Temp\tmp4oyg3b4d
cmd: ['C:/Program', 'Files/Python311-arm64/python.exe', 'C:/WoA/OpenUSD/build_dir/tests/testPlug']


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.31 sec

The following tests FAILED:
        226 - testPlug (Failed)
Errors while running CTest
```

asluk:
> @asluk Full output:
> 
> ```
> C:\WoA\OpenUSD\build_dir\build\OpenUSD>ctest -C Release -R testPlug --output-on-failure
> Test project C:/WoA/OpenUSD/build_dir/build/OpenUSD
>     Start 226: testPlug
> 1/1 Test #226: testPlug .........................***Failed    0.21 sec
> Warning: in _DeclareAliases at line 526 of C:\WoA\OpenUSD\pxr\base\plug\plugin.cpp -- Expected string for alias name, but found int
> .Error: return code 3221225477 doesn't match expected 0 (EXPECTED_RETURN_CODE).chdir: C:\Users\Anthony\AppData\Local\Temp\tmp4oyg3b4d
> cmd: ['C:/Program', 'Files/Python311-arm64/python.exe', 'C:/WoA/OpenUSD/build_dir/tests/testPlug']
> 
> 
> 0% tests passed, 1 tests failed out of 1
> 
> Total Test time (real) =   0.31 sec
> 
> The following tests FAILED:
>         226 - testPlug (Failed)
> Errors while running CTest
> ```

Thanks @anthony-linaro  -- I can't think of something that would cause this, and I don't *think* that the warning is what is causing the failure. Also I'd expect many more tests to be broken if plugins were broken.

@sunyab , any ideas on how we can get more diagnostics on the failure? Maybe run with TF_DEBUG=PLUG* in the environment? Thanks!

asluk:
@anthony-linaro  can you set TF_DEBUG to be "PLUG*" in the environment ? Here is my LastTest.log file that gets output from ctest, when running the test successfully on Windows x86_64

[LastTest.log](https://github.com/user-attachments/files/18760640/LastTest.log)

anthony-linaro:
@asluk Attached:
[LastTestWithPluginDebug.log](https://github.com/user-attachments/files/18766149/LastTestWithPluginDebug.log)

Looks like it's failing on `TestPlugDsoUnloadable` - I noticed I'm using Python 3.11, and you 3.10 - could this be the issue, maybe 3.11 is more strict on the unloadable plugin?

asluk:
> @asluk Attached: [LastTestWithPluginDebug.log](https://github.com/user-attachments/files/18766149/LastTestWithPluginDebug.log)
> 
> Looks like it's failing on `TestPlugDsoUnloadable` - I noticed I'm using Python 3.11, and you 3.10 - could this be the issue, maybe 3.11 is more strict on the unloadable plugin?

Thanks @anthony-linaro -- I suspect it's something else-- I do see usdShadeValidators in your output so I wonder if it's what @tallytalwar suspects as well, that running the tests in a no-imaging build will fail due to the shaders not being available.

I'll try to repro on Windows x86_64 with imaging disabled.

nvmkuruc:
There's an open issue for checkers failing without imaging for reference

https://github.com/PixarAnimationStudios/OpenUSD/issues/3055

asluk:
confirmed that testPlug does pass for me with a --no-imaging --tests build on Windows x86_64 with python 3.11 -- I am at commit 67e851b1b0a67807dea8aff7ea2569e2c2e64e57 , which is the current Pixar "dev" branch -- @anthony-linaro , can you rebase on top of that?

(these tests fail for me)

The following tests FAILED:
        554 - testUsdZipFile (Failed)
        657 - testUsdUtilsCreateNewUsdzPackageUdim (Failed)
        658 - testUsdUtilsCreateNewUsdzPackageRemapClipsDir (Failed)
        659 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
        767 - testUsdChecker1 (Failed)
        768 - testUsdChecker2 (Failed)
        769 - testUsdChecker3 (Failed)
        770 - testUsdChecker4 (Failed)
        771 - testUsdChecker5 (Failed)
        779 - testUsdChecker13 (Failed)
        780 - testUsdChecker14 (Failed)
        781 - testUsdChecker15 (Failed)
        782 - testUsdChecker16 (Failed)
        783 - testUsdChecker17 (Failed)
        784 - testUsdChecker18 (Failed)
        791 - testUsdChecker26 (Failed)
        792 - testUsdChecker27 (Failed)
        802 - testUsdChecker37 (Failed)
        805 - testUsdResolverExample (Failed)

asluk:
All tests pass for me with --imaging with python 3.10 in dev, windows x86_64

with python 3.11 in dev, windows x86_64 these --imaging tests fail

The following tests FAILED:
        554 - testUsdZipFile (Failed)
        657 - testUsdUtilsCreateNewUsdzPackageUdim (Failed)
        867 - testUsdviewNavigationKeys (Failed)
        883 - testUsdResolverExample (Failed)
        
@anthony-linaro , if testPlug still fails for you after rebasing, can you try commenting out blocks of test_ErrorCases in testPlug.py until it passes, to keep zeroing in on specifics?

It looks like TestPlugDsoUnloadable is expected to fail to load, but not to crash, and I'm not sure if any ARM64 specifics with how TestPlugDsoUnloadable  gets compiled may be in play-- these compiler flags may need tweaking for ARM64 - https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/plug/CMakeLists.txt#L105

asluk:
small update-- I am now getting the same test results with --imaging builds on Windows x86_64 with python 3.10 and 3.11 in the dev branch -- testUsdZipFile , testUsdviewNavigationKeys , and testUsdResolverExample are the known failures for that. (My previous python 3.10 test run was implicitly excluding tests that are known to fail on Windows).

testPlug is the only real test that needs further investigation in this PR, and I think the approach I recommended for zeroing in on the error cases and compiler flags for the deliberately "bad" plugins being processed in that test should get the PR over the hump.

all the other failing tests in the PR are from the missing usdShaders in --no-imaging builds, that is tracked in https://github.com/PixarAnimationStudios/OpenUSD/issues/3055

anthony-linaro:
@asluk My error list after merging dev now looks like: 
```
The following tests FAILED:
        226 - testPlug (Failed)
        659 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
        767 - testUsdChecker1 (Failed)
        768 - testUsdChecker2 (Failed)
        769 - testUsdChecker3 (Failed)
        770 - testUsdChecker4 (Failed)
        771 - testUsdChecker5 (Failed)
        779 - testUsdChecker13 (Failed)
        780 - testUsdChecker14 (Failed)
        781 - testUsdChecker15 (Failed)
        782 - testUsdChecker16 (Failed)
        783 - testUsdChecker17 (Failed)
        784 - testUsdChecker18 (Failed)
        791 - testUsdChecker26 (Failed)
        792 - testUsdChecker27 (Failed)
        802 - testUsdChecker37 (Failed)
        805 - testUsdResolverExample (Failed)
```

So looks like `testPlug` wasn't fixed - I'll try and drill down to the specific issue in the suggested way

asluk:
> @asluk My error list after merging dev now looks like:
> 
> ```
> The following tests FAILED:
>         226 - testPlug (Failed)
>         659 - testUsdUtilsCreateNewUsdzPackageEditInPlace (Failed)
>         767 - testUsdChecker1 (Failed)
>         768 - testUsdChecker2 (Failed)
>         769 - testUsdChecker3 (Failed)
>         770 - testUsdChecker4 (Failed)
>         771 - testUsdChecker5 (Failed)
>         779 - testUsdChecker13 (Failed)
>         780 - testUsdChecker14 (Failed)
>         781 - testUsdChecker15 (Failed)
>         782 - testUsdChecker16 (Failed)
>         783 - testUsdChecker17 (Failed)
>         784 - testUsdChecker18 (Failed)
>         791 - testUsdChecker26 (Failed)
>         792 - testUsdChecker27 (Failed)
>         802 - testUsdChecker37 (Failed)
>         805 - testUsdResolverExample (Failed)
> ```
> 
> So looks like `testPlug` wasn't fixed - I'll try and drill down to the specific issue in the suggested way

thanks for the update @anthony-linaro  -- yep, I think once you start disabling error cases in testPlug.py, we'll get a more targeted sense of the actual severity and root causes of what it's reporting -- cc @sunyab 🙏🏼

anthony-linaro:
I've spent an amount of time debugging this, and have narrowed it down to an interesting difference in behaviour, which I'm not entirely sure on why it's happening - I will continue to try and debug, but thought I would update here.

The issue is actually coming from the `LoadLibrary` call in `library.cpp` when the DLL is loaded for `TestPlugDsoUnloadable` - on x64, the DLL fails to load as intended/documented, and execution continues with the handle being a NULLPTR.

On ARM64 however, there is an `EXCEPTION_ACCESS_VIOLATION` when the `LoadLibrary` call is made, that just doesn't seem to want to be caught. I've tried both regular `try` blocks, and also the MSVC extension `__try` blocks. Regardless of those, the exception seems to immediately kill off the process, causing the error we can see of `return code 3221225477 doesn't match expected 0`, which comes from `testWrapper.py`, line 321. Using `LoadLibraryEx` makes no difference.

I'm not entirely sure what's going on here - by all documentation, this function should just fail, and return `NULLPTR` - however, it doesn't seem to be doing that at all. Is it because the DLL is so intentionally broken we're running into undefined behaviour?

asluk:
Thanks for digging into it @anthony-linaro -- I suspect someone from Microsoft would have more info on why the exception is not being caught. If you comment out 

        # try to load an unloadable plugin
        badPlugin = Plug.Registry().GetPluginForType('TestPlugUnloadable')
        self.assertIsNotNone(badPlugin)
        with self.assertRaises(RuntimeError):
            badPlugin.Load()
            
         
does the rest of testPlug pass? If so, @sunyab , is there a way to conditionally skip that case behind some sort of Arch query in Python?

sunyab:
Sure, that test case could certainly be skipped over using the same logic to detect whether we're in a Windows ARM64 environment as in build_usd.py -- we wouldn't need anything from the Arch library itself (which isn't wrapped to Python anyway).

The main trickiness with TestPlugDsoUnloadable is its intentional use of an unresolved external symbol. I didn't see anything in the LoadLibrary docs that explicitly mention that loading such a library will result in the function returning NULL, so I guess technically its undefined behavior? At the same time, it feels like it'd be pretty weird if the behavior differed between ARM and x64 for this particular case.

One thing I'm kind of curious about is whether moving this code in `TestPlugDsoUnloadable.cpp`:

https://github.com/PixarAnimationStudios/OpenUSD/blob/62847554c9abce257729f9f6c7071165ccd680a4/pxr/base/plug/testenv/TestPlugDsoUnloadable.cpp#L22-L23

to within the TF_REGISTRY_FUNCTION block here:

https://github.com/PixarAnimationStudios/OpenUSD/blob/62847554c9abce257729f9f6c7071165ccd680a4/pxr/base/plug/testenv/TestPlugDsoUnloadable.cpp#L43

makes a difference. I have no real reason to think it should, I'm just throwing stuff against the wall.

anthony-linaro:
The intended behaviour of the function is mentioned here:
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya#return-value

The intention, as I understand it, is that you call the function, and if it
returns a null pointer, then you call "GetLastError()" for more
information. This is how it has worked when I have used it previously (for
example, if a DLL had a DEPENDENTLOADFLAG=0x2000, which I encountered last
week). Catching an access violation exception from inside a Win32 API would
be unusual, AIUI, as they are designed to be called from C, too.

Perhaps it is undefined behaviour, but the documentation doesn't make it
clear either way.

I will try the rearrangement inside the plugin tomorrow.

Oddly - this all seems to happen before testPlug - in my testing I even
went as far as emptying testPlug.py of anything but a "hello world" print,
yet I still hit the error. Is the python compiled, and so requires a
rebuild for changes to apply? There is also the possibility I did something
wrong there, of course

asluk:
@anthony-linaro , yes OpenUSD needs to be incrementally "rebuilt" in order for modified tests to get re-installed to the location from which they are run.

anthony-linaro:
@sunyab Moving the code around did not help in this instance, sadly

@asluk Commenting out that section of code (and rebuilding) allowed the rest of the test to pass

asluk:
Thanks @anthony-linaro -- go ahead and put that test block behind an if block that excludes it on windows ARM. @sunyab anything else needed for this PR ? Can we get it into v25.05 ? Thanks!

README.md
```diff
@@ -174,9 +174,13 @@ Or for visionOS:
 
 ##### Windows:
 
-Launch the "x64 Native Tools Command Prompt" for your version of Visual Studio
-and run the script in the opened shell. Make sure to use the 64-bit (x64) 
-command prompt and not the 32-bit (x86) command prompt.
+On Windows, it is possible to compile for either x64, or ARM64.
```
asluk, line 177: Maybe clarify that for Windows ARM64, only non-imaging builds are supported

anthony-linaro: Done!

anthony-linaro:
Both done!

README.md
```diff
@@ -174,9 +174,14 @@ Or for visionOS:
...
+**Note**: ARM64EC is not compatible at this time.
+**Note 2**: Imaging is not supported on ARM64 at present
```
sunyab: Just wanted to follow-up on this. From the discussions it sounded like you were running into #3285, but I believe that's been resolved in the dev branch. Would you mind checking on this again?

build_scripts/build_usd.py
```diff
@@ -394,7 +406,8 @@ def RunCMake(context, force, extraArgs = None):
 
     # Note - don't want to add -A (architecture flag) if generator is, ie, Ninja
     if IsVisualStudio2019OrGreater() and "Visual Studio" in generator:
-        generator = generator + " -A x64"
+        windowsHostArch = GetWindowsHostArch()
+        generator = generator + " -A " + windowsHostArch
```
sunyab:
```suggestion
        generator = generator + " -A " + GetWindowsHostArch()
```

build_scripts/build_usd.py
```diff
@@ -153,6 +153,18 @@ def IsVisualStudio2017OrGreater():
     VISUAL_STUDIO_2017_VERSION = (14, 1)
     return IsVisualStudioVersionOrGreater(VISUAL_STUDIO_2017_VERSION)
 
+# Helper to get the current host arch on Windows
+def GetWindowsHostArch():
+    identifier = os.environ.get('PROCESSOR_IDENTIFIER')
+    # ARM64 identifiers currently start with "ARMv8 ...."
+    # Note: This could be modified in the future to distinguish between ARMv8 and ARMv9
+    if "ARM" in identifier:
```
sunyab: Just to make this as specific as possible:

```suggestion
    if "ARMv" in identifier:
```

pxr/base/arch/timing.h
```diff
@@ -52,7 +52,13 @@ ArchGetTickTime()
     return __rdtsc();
 #elif defined (ARCH_CPU_ARM)
     uint64_t result;
+    #if defined(ARCH_COMPILER_MSVC)
+    // MSVC does not support inline assembly on ARM64 platforms
+    // 0x5F02 == ARM64_CNTVCT - manually calculated value avoids <windows.h>
+    result = _ReadStatusReg(0x5F02);
```
sunyab, line 58: Avoiding the windows.h include makes sense to me. Could we add something to `Arch_ValidateAssumptions` in `pxr/base/arch/assumptions.cpp` that verifies `ARM64_CNTVCT == 0x5F02` (maybe a static_assert?), along with a comment that points back to this code? That way we'll know if this value ever changes for some reason.

pxr/base/plug/testPlug.py
```diff
@@ -275,10 +275,12 @@ def test_ErrorCases(self):
         self.assertEqual(listener.numReceived, 0)
 
         # try to load an unloadable plugin
-        badPlugin = Plug.Registry().GetPluginForType('TestPlugUnloadable')
-        self.assertIsNotNone(badPlugin)
-        with self.assertRaises(RuntimeError):
-            badPlugin.Load()
+        # NOTE: This test is broken on Windows ARM64, so is switched off
+        if not "ARM" in os.environ.get('PROCESSOR_IDENTIFIER'):
```
sunyab:
```suggestion
        if "ARMv" not in os.environ.get('PROCESSOR_IDENTIFIER', ''):
```

Matches `ARMv` check in `build_usd.py` above. Also needed to specify a default value to `get` to avoid errors on other platforms.

pxr/base/plug/testPlug.py
```diff
@@ -275,10 +275,12 @@ def test_ErrorCases(self):
         self.assertEqual(listener.numReceived, 0)
 
         # try to load an unloadable plugin
-        badPlugin = Plug.Registry().GetPluginForType('TestPlugUnloadable')
-        self.assertIsNotNone(badPlugin)
-        with self.assertRaises(RuntimeError):
-            badPlugin.Load()
+        # NOTE: This test is broken on Windows ARM64, so is switched off
```
sunyab:
```suggestion
        # XXX: This test is broken on Windows ARM64, so is switched off
```

We typically use `XXX:` comments to denote known issues that should be fixed. Also, could you please expand the comment a bit to include more info about the failure?

build_scripts/build_usd.py
sunyab, line 1: Since the build script requires oneTBB for Windows ARM64, I think it would be good to add a check for this around line 2478 where we do other compatibility checks. Something like:

```
if Windows() and GetWindowsHostArch() == "ARM64" and not context.oneTBB:
    PrintError("Windows ARM64 builds require oneTBB. Enable via the --onetbb argument")
    sys.exit(1)
```

I was weighing whether we should automatically switch to oneTBB if we detect this case, but since oneTBB is a non-standard configuration and not part of the VFX Reference Platform I thought it'd be better to make users acknowledge that switch directly. Open to discussion if folks feel otherwise!

sunyab (changes requested):
Hi @anthony-linaro, I've run this change through internal reviews and tests. I ran into one issue in `testPlug.py`, which I've noted in the review. I've also added other notes to address -- nothing major IMHO.

Let me know if you've got any questions. Thanks!

build_scripts/build_usd.py
anthony-linaro: Doesn't VFX 2025 now use OneTBB by default?: https://vfxplatform.com/

sunyab: Sorry, I should have been more specific. The latest VFX 2025 does, but build_usd.py currently adheres to VFX 2022 (soon 2023).

anthony-linaro: I have updated this to your suggestion!

pxr/base/arch/timing.h
anthony-linaro: I have addressed this in my latest commit - please let me know if that's what you had in mind

README.md
anthony-linaro: Just checked - imaging works, but usdview does not (as PySide isn't enabled for Windows ARM64 yet, working on that) - I will add the full command line in a seperate comment

anthony-linaro:
After merging dev, I am now able to get it compiling most of the rest of imaging, other than usdview, so the command line I used for compiling was:
```
python build_scripts/build_usd.py --onetbb --tests --no-usdview C:/WoA/OpenUSD/build_dir
```

There is now only a single test failure, which is `testUsdResolverExample` - I suspect this it due to me using a newer version of python, as @asluk saw it fail with a newer version too.

asluk:
yes, testUsdResolverExample  fails on Windows x86_84 as well for me - -if @sunyab  is good with the commits that address his notes, I think this PR is good to go?

README.md
sunyab: Ok great, in that case can we remove the note about imaging not being supported here?

pxr/base/arch/timing.h
sunyab: Yep that looks great, thanks!

pxr/base/plug/testPlug.py
```diff
@@ -275,10 +275,12 @@ def test_ErrorCases(self):
         self.assertEqual(listener.numReceived, 0)
 
         # try to load an unloadable plugin
-        badPlugin = Plug.Registry().GetPluginForType('TestPlugUnloadable')
-        self.assertIsNotNone(badPlugin)
-        with self.assertRaises(RuntimeError):
-            badPlugin.Load()
+        # XXX: This test is broken on Windows ARM64, so is switched off
```
sunyab: Just expanded this comment a bit, feel free to modify if anything sounds off here.

```suggestion
        # XXX:
        # On Windows ARM64 loading TestPlugUnloadable (which intentionally uses
        # an undefined external symbol) unexpectedly crashes in the underlying
        # LoadLibrary call with an EXCEPTION_ACCESS_VIOLATION instead of
        # returning NULL as it does on x64. For now we just skip this test
        # on that platform. More discussion at:
        # https://github.com/PixarAnimationStudios/OpenUSD/pull/3430.
```

build_scripts/build_usd.py
```diff
@@ -2528,6 +2540,10 @@ def ForceBuildDependency(self, dep):
         PrintError("Cannot build usdview when Python support is disabled.")
         sys.exit(1)
 
+if Windows() and GetWindowsHostArch() == "ARM64" and not context.buildOneTBB:
```
sunyab: Minor, but would you mind moving this to around line 2943 to put it alongside other oneTBB-related checks, and adding this comment?

```suggestion
# Windows ARM64 requires oneTBB. Since oneTBB is a non-standard option, we
# error out and require the user to explicitly specify --onetbb instead of
# silently switching to oneTBB for them.
if Windows() and GetWindowsHostArch() == "ARM64" and not context.buildOneTBB:
    PrintError("Windows ARM64 builds require oneTBB. Enable via the --onetbb argument")
    sys.exit(1)

```

sunyab:
@asluk @anthony-linaro Yep, I had some last little doc questions/sugs but aside from that this PR is ready to go and I'll merge it as soon as those last bits are taken care of.

build_scripts/build_usd.py
anthony-linaro: Done!

README.md
anthony-linaro: I reworded it slightly to indicate usdview doesn't work, rather than the entirety of imaging

pxr/base/plug/testPlug.py
anthony-linaro: I have now added this and tweaked it slightly

anthony-linaro:
Changes made, hopefully all okay!

sunyab:
Great, I've merged this PR internally and it'll hit the `dev` branch with the next push. This PR will get closed out then. Thanks again!

PR #3438, opened 2024-11-25
Patch usddiff for USDZ containing crate layers

andrewkaufman:
### Description of Change(s)

The underlying layers need to be converted to usda before they are diffed.

Previously a crate layer within a USDZ would produce the following failure when run through usddiff:

```
Differences in a.usdz[basic.usdc] and b.usdz[basic.usdc]:
'utf-8' codec can't decode byte 0xfe in position 16: invalid start byte
```

Note that there are 2 test changes here:
- A new test using the same composed stage as the original `testDiffToolIterateUsdz`, but with the internal `basic.usda` converted to `basic.usdc`.
  - This was done by unzipping the original `a.usdz` and `b.usdz`, using `usdcat` to convert the layer, then `usdzip` to make the new archives.
- A change to the expected output in `testDiffToolIterateUsdz`
  - This is due to all layers within a USDZ running through usdcat now, which did not occur previously.
  - This is an additional bug fix, as `_runDiff` is called unconditionally on non-zipped layers, so should also be called unconditionally on internal layers. Previously, whitespace diffs in USDA layers would cause failures when embedded in USDZ, but would not cause failures if extracted first & diff'd outside the USDZ. Now the diff result is consistent whether the usda is tested on its own or within an archive.

### Checklist


jesschimein:
Filed as internal issue #USD-10468

PR #3368, opened 2024-10-17
usdUtils: Added FileExtensionValidator

beersandrew:
### Description of Change(s)
- added strings into plugInfo.json & validatorTokens
- added implementation and test for FileExtensionChecker
- added test usdz files with all supported extensions and an invalid extension for tests

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65424964

- [X] I have verified that all unit tests pass with the proposed changes

- [X] I have submitted a signed Contributor License Agreement

PR #3444, opened 2024-12-02
usdValidation: usdUtils: FileExtensionValidator

beersandrew:
### Description of Change(s)
 - added strings into plugInfo.json & validatorTokens
 - added implementation and test for FileExtensionChecker
 - added test usdz files with all supported extensions and an invalid extension for tests

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3?pane=issue&itemId=65424964

Note: this PR is a replacement for https://github.com/PixarAnimationStudios/OpenUSD/pull/3368

### Checklist


jesschimein:
Filed as internal issue #USD-10487

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -97,12 +99,60 @@ _PackageEncapsulationValidator(const UsdStagePtr &usdStage)
     return errors;
 }
 
+static UsdValidationErrorVector
+_FileExtensionValidator(const UsdStagePtr& usdStage) {
+    UsdValidationErrorVector errors;
+
+    const std::vector<TfToken> validExtensions = {TfToken("usda"),
+        TfToken("usdc"), TfToken("usd"), TfToken("usdz"), TfToken("png"),
+        TfToken("jpg"), TfToken("jpeg"), TfToken("exr")};
```
tallytalwar: Maybe a set of TfToken will be better here.

Also there is a TfTokenVector.

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -113,11 +114,48 @@ TestPackageEncapsulationValidator()
...
+    UsdValidationRegistry& registry = UsdValidationRegistry::GetInstance();
+
+    // Verify the validator exists
+    const UsdValidationValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->fileExtensionValidator);
+
+    TF_AXIOM(validator);
+
+    const UsdStageRefPtr& stage = UsdStage::Open("failFileExtension.usdz");
+    UsdValidationErrorVector errors = validator->Validate(stage);
```
tallytalwar, line 144: if you scope it, then you can make this const, and also the below one on line 147 a const.

tallytalwar (commented):
Looks good to me, a couple of small notes to be addressed though. Thanks @beersandrew

tallytalwar:
Apologies, I take my previous approval back. I think this validator is also in the league of validators which should be provided by reality kit. Its perfectly valid for usdz to contain formats (and shouldn't be limited to the formats supported by reality kit: .usda, .usdc, .usdz and .exr, .jpg, .jpeg, .png).

beersandrew:
@tallytalwar for these reality kit specific validators, should there be some equivalent of `--arkit` for the new validation framework, something like `usdPortableUsdzValidation` or should these be closed as they aren't part of coreUSD?

tallytalwar:
@beersandrew Yeah I feel these should exist as a plugin in scenekit or usd framework apple provides but not in the core openUSD.

Again cc`ing @dgovil as fyi, as without these in the apple environment, --arkit option with the new validation system won't fully do whats its supposed to validate.

dgovil:
Actually, we're not the ones who are limiting the list of formats used by usdz. It's the actual usdz spec by Pixar which we're adhering to. 

We'd prefer to be able to use more formats (see my Profiles proposal)

I think it's fine to get rid of the ARKit flag, but this one's a generic USDZ validator.

tallytalwar:
Ahh I see: https://openusd.org/release/spec_usdz.html#file-types (I never knew about this, was looking for evidence in zipFile code :/). Ok I agree this should be in Usdz validator's domain. Thanks for the clarification @dgovil.

@beersandrew could you include extensions mentioned in the link above?

tallytalwar (approved):
Thanks

beersandrew:
@tallytalwar I learned there are a few more JPEG extensions (`.jpe`, `.jif`, `.jfif`) listed [here](https://en.wikipedia.org/wiki/JPEG), but only `.jpg` and `.jpeg` work for me in `usdview`. I assume it's not a problem that these extensions don't work in `usdview`, however I think it makes sense to be explicit on the spec page about which extensions are supported, as currently it lists `any of the multiple common extensions for jpeg`.

What do you think?

tallytalwar:
Lets stick with these for now. Thanks

meshula:
reference note: jpe is an old alias for JPEG for DOS. JIF and JFIF are part of the historical record, as the wikipedia article puts it, rather than live formats. https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format

PR #3400, opened 2024-11-05
usdUtils: CompressionValidator

beersandrew:
### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65425103

### Checklist

[X] I have created this PR based on the dev branch

[X] I have followed the [coding conventions](https://openusd.org/release/api/_page__coding__guidelines.html)

[ ] I have added unit tests that exercise this functionality (Reference: 
[testing guidelines](https://openusd.org/release/api/_page__testing__guidelines.html))

[ ] I have verified that all unit tests pass with the proposed changes

[X] I have submitted a signed Contributor License Agreement (Reference: 
[Contributor License Agreement instructions](https://openusd.org/release/contributing_to_usd.html#contributor-license-agreement))

PR #3449, opened 2024-12-03
usdValidation: usdUtils: CompressionValidator

beersandrew:
### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65425103

Note: This PR is a replacement for https://github.com/PixarAnimationStudios/OpenUSD/pull/3400

### Checklist


pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -97,12 +98,55 @@ _PackageEncapsulationValidator(const UsdStagePtr &usdStage)
     return errors;
 }
 
+static UsdValidationErrorVector
+_CompressionValidator(const UsdStagePtr& usdStage) {
```
beersandrew: @tallytalwar I've come across something in this validator that's not clear to me.

My expectation from reading the description from the existing [CompressionChecker](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/complianceChecker.py#L149) is that this test should make sure that files within the USDZ file itself were not compressed. So I've created a sample file like this (in this [repo](https://github.com/beersandrew/usd-test-assets/tree/main/Validation/examples/CompressionChecker) as compressedUsdaInside.usdz). However this file fails [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/sdf/textFileFormat.cpp#L161), before it can get to the usdchecker code.

If I try a different approach of compressing the usdz itself, it also fails early [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/usdzResolver.cpp#L217).

Do you know how to reproduce the nicer errors from usdchecker itself? (Similar situation with ByteAlignmentChecker as well)

jesschimein:
Filed as internal issue #USD-10492

pxr/usdValidation/usdUtilsValidators/validators.cpp
tallytalwar: 1. Why its failing on parsing the "compressed" "usda" file is because I think usda extension makes it load the usda file plugin, which when parsing tries to see if the file header has the usda magic cookie and if not it fails. If you stick with this approach only maybe use a different extension for this file.
2. usdz is an archival format, so the code checks if it has any compression on it or not.

Maybe add a compressed asset to usdz file like a compressed jpeg/png?

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -97,12 +98,55 @@ _PackageEncapsulationValidator(const UsdStagePtr &usdStage)
...
+    const SdfLayerHandle &rootLayer = usdStage->GetRootLayer();
+    const UsdZipFile zipFile = UsdZipFile::Open(rootLayer->GetRealPath().c_str());
+
+    std::string packagePath = ArSplitPackageRelativePathOuter(rootLayer->GetIdentifier()).first;
+    if (!zipFile)
+    {
+        return {};
+    }
```
tallytalwar, line 154: You can move this check above doing the ArSplitPackageRelativePathOuter

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -97,12 +98,55 @@ _PackageEncapsulationValidator(const UsdStagePtr &usdStage)
...
+            const std::string &fileName = *it;
+            errors.emplace_back(
+                UsdUtilsValidationErrorNameTokens->compressionDetected,
+                UsdValidationErrorType::Error,
+                UsdValidationErrorSites {
+                    UsdValidationErrorSite(
+                            rootLayer, SdfPath(fileName))
+                },
+                TfStringPrintf(
+                ("File '%s' in package '%s' has "
```
tallytalwar: Extra "()" here? TfStringPrintf(("File .... instead of TfStringPrintf("File..

tallytalwar (commented):
Logic is fine, but I was thinking we can split this up into a UsdzPackageValidator and a UsdzRootPackageValidator with appropriate keywords.

Both validators are stage level validators:
- RootPackageValidator: will do compression error logic, byle alignment logic and extension validation logic just for the rootLayer, if the rootLayer is a package.
- UsdzPackageValidator: will do all the same validation checks for all packages found in the stage.

A lot of the actual validation code itself can be shared between the 2 validator methods.

This way we can do this: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usdValidation/bin/usdchecker/usdchecker.cpp#L572-L575

Let me know what you think about this and feel free to ping on aswf slack if you want to discuss this more?

Thanks

pxr/usdValidation/usdUtilsValidators/validators.cpp
beersandrew: I've tried compressing a jpg and adding it to the usdz, and I've tried including a zip file as an asset and then including it in the usdz as well. I'm still unable to trigger the CompressionChecker in the existing `complianceChecker.py` usdchecker version

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -143,6 +144,101 @@ _FileExtensionValidator(const UsdStagePtr& usdStage) {
     return errors;
 }
 
+static UsdValidationErrorVector
+_GetUsdzPackageErrors(const SdfLayerHandle &rootLayer)
+{
+    const UsdZipFile zipFile = UsdZipFile::Open(rootLayer->GetRealPath().c_str());
+    if (!zipFile)
```
beersandrew, line 151: @tallytalwar I noticed testing locally that when passing a .usda file in here, this if statement does not filter out the .usda layer, which I anticipated that it would. All of the errors are still coming back correctly but I wanted to ask if there was some other condition to put here to exit out early as I couldn't find something on the UsdZipFile class that was applicable

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -160,12 +162,125 @@ TestFileExtensionValidator()
...
+    {
+        const UsdStageRefPtr &stage =
+            UsdStage::Open("sceneWithInvalidPackage.usda");
+
+        const UsdValidationErrorVector errors = validator->Validate(stage);
+
+        // Verify internal compression and byte misalignment errors occur
+        TF_AXIOM(errors.size() == 3);
+
+        const std::string fullErrorPath = ArchGetCwd() +
```
beersandrew: @tallytalwar Two things about this line. 
1. It seems in the validator itself, if I pass in a root level myFile.usdz, the `ArSplitPackageRelativePathOuter` will get the package path that i want, as in, only `myFile.usdz` however passing in `parentFile.usda` that references `myFile.usdz` the error message has the full real path in the error message
2. If the above is okay as is, I wanted to join these paths here in a platform agnostic way and I thought there was some library function to join these paths so it would work everywhere as this as written seems it's the wrong direction for Windows, and it seemed wrong to write some helper function here to deal with the correct slashes. I did find other tests that used only this slash, but let me know if this should be changed.

pxr/usdValidation/usdUtilsValidators/validators.cpp
tallytalwar: Yeah looks like from the code it works even if the layer is a usda layer, as it goes through ArAsset. I will confirm internally if thats anticipated.

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
tallytalwar: I am not sure I fully understand, 1.

But for 2, there is: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/tf/stringUtils.h#L677-L689 which you can use to do a platform agnostic way of concatenating paths, etc.

beersandrew: To rephrase 1, this line: https://github.com/PixarAnimationStudios/OpenUSD/pull/3449/files#diff-d9adee7401f280af5ec4cc8aa666b4fb0d767ea269bedc2750dcc333503af6aaR156 behaves differently whether it's the root package or an internal package.
- When it's a root package the value will be : "rootPackage.usdz" (or whatever the name of the file is)
- When it's an internal package the value will be : "/fully/qualified/path/of/package/internalPackage.usdz" (where this is the path on disk or would be the path on disk of the internal package.

I think this is less of an issue now that you've reminded me of 2. (Thanks!) but I wanted to point it out as the error messages are different in these cases where sometimes the fully qualified path is there and other times it's not

tallytalwar: I see, yeah I think I can reason about that (will also confirm internally), but the reference path to internal package gets fully resolved relative to its referenced location, in order to locate the asset, and that causes the paths to be absolute.

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -143,6 +144,101 @@ _FileExtensionValidator(const UsdStagePtr& usdStage) {
...
+    {
+        return {};
+    }
+
+    std::string packagePath = ArSplitPackageRelativePathOuter(rootLayer->GetIdentifier()).first;
```
tallytalwar: const

pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -143,6 +144,101 @@ _FileExtensionValidator(const UsdStagePtr& usdStage) {
...
+static UsdValidationErrorVector
+_UsdzPackageValidator(const UsdStagePtr& usdStage) {
+
+    SdfLayerRefPtrVector layers;
+    std::vector<std::basic_string<char>> assets, unresolvedPaths;
+    const SdfLayerHandle &rootLayer = usdStage->GetRootLayer();
+    const SdfAssetPath &path = SdfAssetPath(rootLayer->GetIdentifier());
+
+    UsdUtilsComputeAllDependencies(path, &layers, &assets, &unresolvedPaths,
+                                   nullptr);
```
tallytalwar: For the purposes of this validation test, you don't need assets and unresolvedPaths, so you can simply pass these as nullptr. Implementation for this handles this appropriately: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdUtils/dependencies.cpp#L174-L182

tallytalwar (commented):
Just couple of more small notes to be addressed; will sync in when done. Thanks Andy

PR #3450, opened 2024-12-03
usdValidation: usdUtils: MissingReferenceValidator

beersandrew:
### Description of Change(s)
- created implementation, test, as well as associated entries in plugInfo.json & validatorTokens.h

### Fixes Issue(s)
- https://github.com/orgs/PixarAnimationStudios/projects/3/views/1?pane=issue&itemId=65424971

Note: This PR is a replacement for https://github.com/PixarAnimationStudios/OpenUSD/pull/3403

### Checklist

### DIFF
--- a/pxr/usdValidation/usdUtilsValidators/plugInfo.json
+++ b/pxr/usdValidation/usdUtilsValidators/plugInfo.json
@@ -4,10 +4,13 @@
             "Info": {
                 "Validators": {
                     "FileExtensionValidator": {
-                        "doc": "Only valid core layer extensions (.usd, .usda, .usdc, .usdz), valid core texture extensions (.exr, .jpg, .jpeg, .png) and embedded audio files (.M4A, .MP3, .WAV) are allowed in a package.",
-                        "keywords": [
-                            "UsdzValidators"
-                        ]
+                      "doc": "Only valid core layer extensions (.usd, .usda, .usdc, .usdz), valid core texture extensions (.exr, .jpg, .jpeg, .png) and embedded audio files (.M4A, .MP3, .WAV) are allowed in a package.",
+                      "keywords": [
+                        "UsdzValidators"
+                      ]
+                    },
+                    "MissingReferenceValidator": {
+                        "doc": "The composed USD stage should not contain any unresolvable asset dependencies (in every possible variation of the asset), when using the default asset resolver."
                     },
                     "PackageEncapsulationValidator": {
                         "doc": "If the root layer is a package, then its recommended for the composed stage to not contain references to files outside the package. The package should be self-contained, warn if not.", 
--- a/pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
+++ b/pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
@@ -7,8 +7,9 @@
 
 #include "pxr/base/arch/systemInfo.h"
 #include "pxr/base/tf/pathUtils.h"
-#include "pxr/usd/usdGeom/xform.h"
+#include "pxr/usd/usd/editContext.h"
 #include "pxr/usd/usd/variantSets.h"
+#include "pxr/usd/usdShade/shader.h"
 #include "pxr/usdValidation/usdUtilsValidators/validatorTokens.h"
 #include "pxr/usdValidation/usdValidation/error.h"
 #include "pxr/usdValidation/usdValidation/registry.h"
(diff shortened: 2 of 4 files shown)


pxr/usdValidation/usdUtilsValidators/validators.cpp
```diff
@@ -97,9 +97,44 @@ _PackageEncapsulationValidator(const UsdStagePtr &usdStage)
...
+            UsdValidationErrorType::Error,
+            UsdValidationErrorSites {
+                UsdValidationErrorSite(
+                    rootLayer, SdfPath(unresolvedPath))
+            },
+            TfStringPrintf(
+            ("Found unresolvable external dependency "
+                "'%s'."), unresolvedPath.c_str())
+        );
+    }
```
beersandrew, line 172: @tallytalwar One note here, it looks like complianceChecker.py is doing another check [here](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L181) which gets some diagnostic information, but it seems like this process of collecting diagnostic information should be started before opening the stage ([here](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdUtils/complianceChecker.py#L1086)). I tried doing this inside the validator with stage.Reload() but was unable to populate the diagnostics this way.

In my specific test, the errors were similar so I wanted your opinion on if it's possible to do something similar here, or if it's necessary to move the diagnostic part over in this case.

jesschimein:
Filed as internal issue #USD-10493

pxr/usdValidation/usdUtilsValidators/validators.cpp
tallytalwar: I think we might not need an explicit validator for this now! https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usdValidation/usdValidation/coreValidators.cpp#L16-L29 should provide coverage for this? Do you want to try running it through the CoreValidator and see if it gives you the appropriate composition error wrapped by a CoreValidation error?

pxr/usdValidation/usdUtilsValidators/plugInfo.json
```diff
@@ -3,6 +3,9 @@
         {
             "Info": {
                 "Validators": {
+                    "MissingReferenceValidator": {
+                        "doc": "The composed USD stage should not contain any unresolvable asset dependencies (in every possible variation of the asset), when using the default asset resolver."
```
tallytalwar, line 13: Lets remove the "every possible variation of the asset" thats part of the usdchecker to iterate through various variants. Validator should just be checking a stage/variant.

pxr/usdValidation/usdUtilsValidators/validatorTokens.h
```diff
@@ -16,18 +16,21 @@
...
+#define USD_UTILS_VALIDATOR_NAME_TOKENS                                       \
+    ((missingReferenceValidator,                                              \
+      "usdUtilsValidators:MissingReferenceValidator"))                        \
+    ((packageEncapsulationValidator,                                          \
       "usdUtilsValidators:PackageEncapsulationValidator"))
 
-#define USD_UTILS_VALIDATOR_KEYWORD_TOKENS                                     \
-    (UsdUtilsValidators)                                                       \
+#define USD_UTILS_VALIDATOR_KEYWORD_TOKENS                                    \
+    (UsdUtilsValidators)                                                      \
```
tallytalwar, line 28: Not sure if UsdUtilsValidators should be the name of it! (I also want to move this pxr/usdValidation/usdUtilsValidators to be called something else). :/

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -113,11 +115,54 @@ TestPackageEncapsulationValidator()
...
+    const UsdValidationValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->missingReferenceValidator);
+
+    TF_AXIOM(validator);
+
+    // Create stage with a reference that does not exist
+    const UsdStageRefPtr& stage = UsdStage::CreateInMemory();
+
+    const UsdGeomXform xform = UsdGeomXform::Define(stage, SdfPath("/Xform"));
+    const SdfReference badReference("doesNotExist.usd");
```
tallytalwar: Can you also try adding a non-layer asset which is unresolved / missing and see if the validator reports that?

pxr/usdValidation/usdUtilsValidators/validators.cpp
beersandrew: Yep, it does! awesome. I'll close this one out then. Thanks

beersandrew:
Closing since this work is already complete, see Varun's comment above

pxr/usdValidation/usdUtilsValidators/plugInfo.json
tallytalwar: Lets leave it as is, since UsdUtilsComputeAllDependencies operates at Sdf level and will give unresolved references across variants.

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
tallytalwar: Might also make sense for the test to include an unresolved layer in a variant which isn't selected in the stage (this is when the composition validator will not report this reference layer in a not-selected variant).

tallytalwar (commented):
A few notes, but otherwise lgtm.

pxr/usdValidation/usdUtilsValidators/validatorTokens.h
beersandrew: From my perspective, the naming started from "which folder in the usd code base does this validator depend on?". Now that the validation code is moved out of those folders, this feels like a broader problem of defining validation domains, which may be fine for the others already (usdGeom, usdShade... etc). 

Maybe this validator could go into `pxr/usd/usdValidation`? I'm not sure if it's important for that project to not have external dependencies on usdUtils or not.

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
beersandrew: I think i've added the non-layer asset test, but take a look and make sure it's what you wanted. I'll have the variant test added tomorrow

pxr/usdValidation/usdUtilsValidators/validatorTokens.h
tallytalwar: Slightly correction the suffix here is the name of the plugin (the directory name happens to be matching the plugin name :D).

I do plan to move the usdUtilsValidators plugin sometime later so lets leave it as-is for now.

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -113,11 +126,102 @@ TestPackageEncapsulationValidator()
...
+
+    // Verify the validator exists
+    const UsdValidationValidator *validator = registry.GetOrLoadValidatorByName(
+            UsdUtilsValidatorNameTokens->missingReferenceValidator);
+
+    TF_AXIOM(validator);
+
+    const UsdStageRefPtr& stage = UsdStage::CreateInMemory();
+
+    const UsdGeomXform xform = UsdGeomXform::Define(stage, SdfPath("/Xform"));
```
tallytalwar: Though this is fine, but you can also create a prim like:
```
prim = stage.DefinePrim("/Prim", "Xform")
```
which won't require you to include usdGeom/xform.h

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -113,11 +126,102 @@ TestPackageEncapsulationValidator()
...
+            "usdUtilsValidators:MissingReferenceValidator.UnresolvableDependency");
+        TF_AXIOM(errors.size() == 1);
+
+        ValidateError(errors[0], expectedErrorMessage,
+                expectedIdentifier);
+    }
+
+    xform.GetPrim().GetReferences().RemoveReference(badLayerReference);
+    const SdfReference badAssetReference("doesNotExist.jpg");
+    xform.GetPrim().GetReferences().AddReference(badAssetReference);
```
tallytalwar: Ahh don't think its legal to have a non-usd as a reference like this.

An asset would be something like this: https://github.com/PixarAnimationStudios/OpenUSD/blob/9b0c13b2efa6233c8a4a4af411833628c5435bde/extras/usd/tutorials/simpleShading/simpleShading.usda#L48

So what you can do is:
```python
>>> from pixar import Sdf, Usd, UsdShade
>>> stage = Usd.Stage.CreateInMemory()
>>> shader = UsdShade.Shader.Define(stage, "/Shader")
>>> assetInput = shader.CreateInput("notFoundAsset", Sdf.ValueTypeNames.Asset)
>>> assetInput.Set("NotFoundTexture.png")
True
>>> stage.ExportToString()
'#usda 1.0\n(\n    doc = """Generated from Composed Stage of root layer \n"""\n)\n\ndef Shader "Shader"\n{\n    asset inputs:notFoundAsset = @NotFoundTexture.png@\n}\n\n'
>>>
```

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
```diff
@@ -113,11 +126,102 @@ TestPackageEncapsulationValidator()
...
+    }
+
+    xform.GetPrim().GetReferences().RemoveReference(badAssetReference);
+    // Set an unresolved dependency on an unused variant
+
+    UsdVariantSets variantSets = xform.GetPrim().GetVariantSets();
+    UsdVariantSet colorsVariantSet = variantSets.AddVariantSet("myVariant");
+
+    UsdGeomXform selectedXform = UsdGeomXform::Define(stage, xform.GetPath().AppendChild(TfToken("selected")));
+    UsdGeomXform notSelectedXform = UsdGeomXform::Define(stage, xform.GetPath().AppendChild(TfToken("notSelected")));
```
tallytalwar: Don't think this is correct.

You can refer the authoring variant tutorials here: https://openusd.org/docs/Authoring-Variants.html

to see how you can add a variantSet with 2 variants on a prim, with one having a  good reference and other a bad reference. Have the one which has a good reference be the default set on the prim.

Does that make sense?

beersandrew: Yep, thanks for the help here. Should be good now

pxr/usdValidation/usdUtilsValidators/testenv/testUsdUtilsValidators.cpp
beersandrew: Thanks! This has been updated

Issue #1475, opened 2021-03-11
Missing visibility macros on core classes cause runtime TfType equality issues


jerry-huxtable-foundry:
### Description of Issue

We've identified a problem with USD classes which is causing us much distress on macOS with clang. When trying to set attributes on prims we get Coding Error messages like this one:

`Type mismatch for </TestMesh1.points>: expected 'VtArray<GfVec3f>', got 'VtArray<GfVec3f>'
`
After a lot of investigation we've determined the cause of this. The problem is USD is comparing the pointers of type_infos (in tf/safeTypeCompare.h), and we have two type_infos for this type, one in our library and one in USD. The underlying cause is that many of the classes in USD, such as GfVec3f aren't marked as having default visibility so don't get exported. This isn't a problem most of the time because nearly all of the code is inlined and a few methods are marked as being exported. However, this means that the type_info for GfVec3f doesn't get exported properly. It does get exported as a weak-external symbol, which is good, but the problems come when you then try to use GfVec3f from another dynamic library.

If you build a dylib that uses GfVec3f and build it with -fvisibility=default, everything is fine because the linker generates the type_info as weak-external. At runtime, the linker will merge all the weak-external type_infos into one symbol and everything works OK. If, however, you build your library -fvisibility=hidden (as we do to avoid namespace pollution and for performance reasons), the type_info gets marked as non-external and doesn't get merged at runtime. You now have two type_infos for the same type and USD says they're different. This problem also transfers itself to templates such as VtArray<GfVec3f>. 

The fix is to add the GF_API macro to the declaration of the GfVec3f class. Of course, GfVec3f is only one of many classes that suffer from this problem (TfToken is another which has caused us pain), and the appropriate export macro should be added to all of them.

### Steps to Reproduce
1. Create a source file that uses GfVec3f
2. Build a dynamic library, test.dylib with it using -fvisibility=hidden
3. Examine the resulting symbols using nm:

```
$ nm -mo test.dylib| c++filt | grep typeinfo | grep GfVec3f
test.dylib: 0000000000020800 (__DATA,__const) non-external (was signed char private external) typeinfo for pxrInternal_v0_21__pxrReserved__::GfVec3f

```
The output shows the problem: That type_info should be "weak external", not "non-external" and the library has its own private copy of the type_info, which doesn't equal the one in libgf.

```
Build it again with -fvisibility=default and you'll get this:
$ nm -mo test.dylib| c++filt | grep typeinfo | grep GfVec3f
[...]

Issue #665, opened 2018-10-16
Building with -fvisibility=hidden fails on linking


sopvop:
### Description of Issue

If you try to build USD with gcc or clang with `-fvisility=hidden` it will fail when linking.

### Steps to Reproduce
```console
$ cmake -DCMAKE_CXX_VISIBILITY_PRESET=hidden ..
$ make
....
undefined reference to `typeinfo for pxrInternal_v0_18__pxrReserved__::somethingSomething
```
### Why is this important.
Importance of builiding with hidden visibility is well described in GCC Wiki [here](https://gcc.gnu.org/wiki/Visibility).

tldr: It makes startup faster and binaries smaller.

Unlike MSVC, gcc does not export typeinfo and vtable of class by default. MSVC exports it if any member has attribute `__declspec(dllexport)` but gcc (and clang IIRC) require class to be declared with visibility attribute:
```cxx
class __attribute__ ((visibility ("default"))) Foo {};
//or in pxr sources
class SOME_API Foo {};
```
In USD sources FOO_API macro expand to `__attribute__ ((visibility ("default")))` on gcc or `__declspec(dllexport)` on windows, and those have different semantics when applied to types.
GCC will only export necessary rtti machinery (vtable, typeinfo and the like), but msvc will export all members as if all of them have `__declspec(dllexport)` attribute.

### Fix

I propose to add `FOO_CLASS_API` or similarly named macro to `api.h` headers which will expand to `__attribute__ ((visibility ("default")))` on gcc/clang and to nothing on windows. And add it to codegen templates and all public classes.

c64kernal:
Hi @sopvop, thanks so much for the feedback.  Building with hidden visibility is definitely something we'd like to do.  Please give us a little time to review this proposal and get back to you.  (For example, one thing that jumps out at me right away is that, while we try really hard not to, we sometimes use FOO_API to export whole classes on Windows (e.g., in the Houdini plugin code) -- so I'd be concerned about having to explain when to use FOO_CLASS_API and when to use FOO_API when exporting whole types).

jtran56:
Filed as internal issue #USD-4842.

c64kernal:
Hi again @sopvop,

Thanks so much for your patience with this, we had a chance to vet your proposal and overall we like it a lot.  We have a few suggestions to run by you as well.  Here's our current proposal (just small tweaks over your proposal, with a naming suggestion as well):

[...]

PR #3452, opened 2024-12-03
Autodesk: Enable -fvisibility=hidden and -fvisibility-inlines-hidden to hide symbols on Linux and macOS

erikaharrison-adsk:
### Description of Change(s)

This change intends to enable building with `-fvisibility=hidden` to hide the symbols. for these benefits:
1. It allows compiler and linker to do more optimizations, makes startup faster and binaries smaller
2. Minimization of the set of exported symbols in order to track ABI changes
3. Compliance with some specification or standard

Post on AOUSD: [Support building with -fvisibility=hidden](https://forum.aousd.org/t/support-building-with-fvisibility-hidden/1965)

### General technical comment
 
We have added ARCH_EXPORT_TYPE to classes, structs, enums, and some instantiated templates that require visible symbols, to instruct the compiler to generate visible symbols.

On windows, the macro ARCH_EXPORT_TYPE is expanded to empty in MSVC because the current symbol export works fine and does not need to be modified. The reason for not replacing the member function export with a class-level export in this PR is to reduce the workload and to avoid cumbersome and complex technical constraints, which would require a lot of non-patterned modifications. 

It's worth noting that visible and export are not equivalent terms, even though we may not be aware of the difference in our day-to-day work. It helps to understand some seemingly strange problems. For example, Clang can optimize and not to generate code and symbols for inline functions marked as visible, which can lead to symbol lookups failure when linking across modules.

Further cleanup can be done by following up with smaller PRs, which do not have to be as large as this one, so the effort and risk is more manageable.

In Clang, the visibility of an instantiated type of a template depends on both the template and template arguments. 
```C++
template<typename T> class __attribute__((visibility("default"))) Foo{};
class __attribute__((visibility("hidden"))) Bar{};
```
the visibility of `Foo<Bar>` would be hidden other than default, even Foo is marked as default. That is why this PR added "default" to some enums and lots of simple struct types.

- N/A

### Fixes Issue(s)

Related issues:
- https://github.com/PixarAnimationStudios/OpenUSD/issues/1475
- https://github.com/PixarAnimationStudios/OpenUSD/issues/665

### Checklist

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -37,6 +37,11 @@ from shutil import which
 # Helpers for printing output
 verbosity = 1
 
+# Hide symbols
+HIDDEN_SYMBOLS = ['-DCMAKE_CXX_VISIBILITY_PRESET=hidden',
+                  '-DCMAKE_C_VISIBILITY_PRESET=hidden',
+                  '-DCMAKE_VISIBILITY_INLINES_HIDDEN=ON']
+
 def Print(msg):
     if verbosity > 0:
         print(msg)
@@ -1801,6 +1806,8 @@ def InstallUSD(context, force, buildArgs):
         # system installed boost
         extraArgs.append('-DBoost_NO_BOOST_CMAKE=On')
         extraArgs.append('-DBoost_NO_SYSTEM_PATHS=True')
+
+        extraArgs += HIDDEN_SYMBOLS
         extraArgs += buildArgs
 
         RunCMake(context, force, extraArgs)
--- a/extras/imaging/examples/hdui/dataSourceTreeWidget.h
+++ b/extras/imaging/examples/hdui/dataSourceTreeWidget.h
@@ -17,7 +17,7 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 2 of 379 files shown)


jesschimein:
Filed as internal issue #USD-10497

AlexSchwank:
Have you tried this on arm64 macOS? This might break type comparison where `-fvisibility=hidden` sets a uniqueness bit. This is what the workaround [here](https://github.com/PixarAnimationStudios/OpenUSD/pull/1668) attempts to fix.

wingfiring:
> Have you tried this on arm64 macOS? This might break type comparison where `-fvisibility=hidden` sets a uniqueness bit. This is what the workaround [here](https://github.com/PixarAnimationStudios/OpenUSD/pull/1668) attempts to fix.

Yes, I tested on arm64 MacOS and passed all (~1300) cases we enabled. 
For the problem that [PR-1668](https://github.com/PixarAnimationStudios/OpenUSD/pull/1668) tries to solve, it is because the type information of the same type comes from different modules. The most related fix is:
image

Some similar issues discovered by test cases are also fixed. Most of issues are template related.
I don't doubt that there are still leaked issues, and if a test detects it or there is a reproduce step, I will continue to fix it.

PR #3467, opened 2024-12-17
Fix shader compile with metal in case decimal separator is comma

mnaehrig:
GLSL version string was converted to floating point using sscanf. But sscanf expects decimal separator to match with current locale, while GLSL version string always uses period as decimal separator. This caused Storm to fail e.g. on german systems.

### DIFF
--- a/pxr/imaging/hgiInterop/metal.mm
+++ b/pxr/imaging/hgiInterop/metal.mm
@@ -14,6 +14,7 @@
 #include "pxr/imaging/hgiMetal/hgi.h"
 
 #include "pxr/base/tf/diagnostic.h"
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10515

meshula:
Thanks for finding this problem!

I was thinking it would be way simpler and local independent to parse like this, there's no need to use floating point numbers at all, really.

```
int major, minor;
sscanf(versionString, "%d.%d", major, minor);
```

mnaehrig:
You're right. Independent parsing of major and minor is way simpler. Thanks for this advise! Just changed it.

PR #3469, opened 2024-12-18
Initial wasm commit including pxr/base

kohakukun:
### Description of Change(s)

- Add support for building pxr/base with Emscripten

Test can be run using ctest with the result:
78% tests passed, 23 tests failed out of 106

Some details about the failed tests:

  | Test Name                        | Error Message                                                      |
  |----------------------------------|--------------------------------------------------------------------|
  | testArchAbi                      | Dynamic library limitation                                         |
  | testArchError                      | Throws but it is not caught                                          |
  | testArchFileSystem               | Problem creating a file due to environment                         |
  | testArchStackTrace               | Problem creating a file due to environment                         |
  | testArchSymbols                  | Dynamic library limitation                                         |
  | testArchSystemInfo               | Problem accessing a file in filesystem due to environment          |
  | TfAtomicOfstreamWrapper          | Cannot create a log file                                           |
  | TfDl                             | Dynamic library limitation                                         |
  | TfEnvSetting                     | Error: return code 7 doesn't match expected 0 (EXPECTED_RETURN_CODE)|
  | TfException                      | Throws but it is not caught                                        |
  | TfFileUtils                      | Cannot create a log file                                           |
  | TfPathUtils                      | Assumes preloaded files                                            |
  | TfSafeOutputFile                 | Issues accessing filesystem                                        |
  | testJsIO_PlugInfo                | Fails because it cannot read the baseline                          |
  | testJsIO_Values                  | Fails because it cannot read the baseline                          |
  | testJsIO_Errors_BadCommaPlacement| Cannot open json file                                              |
  | testJsIO_Errors_MismatchedBraces | Cannot open json file                                              |
  | testJsConverter                  | Fails because it cannot read the baseline                          |
  | testJsWriter                     | Fails because it cannot read the baseline                          |
  | testJsWriter_Pretty              | Fails because it cannot read the baseline                          |
  | testWorkThreadLimitsRawTBBMax    |                                                                    |
  | testWorkThreadLimitsRawTBB2      |                                                                    |
  | testVtCpp      | Issue after including this commit https://github.com/PixarAnimationStudios/OpenUSD/commit/631862416dd0d539cd1aedd29d96c0686f498a53                                                                    |

The change in pxr/base/arch/attributes.h was required since static constructors dont support parameters with an error
`wasm-ld: error: constructor functions cannot take arguments: _Tf_RegistryAdd91(RegRegistry*)`

### Checklist


jesschimein:
Filed as internal issue #USD-10518

✅  CLA on file

pxr/base/arch/timing.cpp
```diff
@@ -270,9 +270,13 @@ Arch_MeasureExecutionTime(uint64_t maxTicks, bool *reachedConsensus,
     // Since measured times are +/- 1 quantum, we multiply by 2000 to get the
     // desired runtime, and from there figure number of iterations for a sample.
     const uint64_t minTicksPerSample = 2000 * ArchGetTickQuantum();
+    #if defined(ARCH_OS_WASM_VM)
+    const int sampleIters = 1; // FIXME
```
matthewcpp: Do we still need an actual implementation for this?

pxr/base/arch/errno.cpp
```diff
@@ -37,7 +51,7 @@ ArchStrerror(int errorCode)
     //   (the string may be truncated if buflen is too small and errnum is
     //   unknown). The string always includes a terminating null byte.
     //
-    return strerror_r(errorCode, msg_buf, 256);
+    return (char*) strerror_r(errorCode, msg_buf, 256);
```
matthewcpp: I wonder if we could make the whole function a bit tidier by being explicit with construction rather than c style casting?
something along the lines of:
```
std::string
ArchStrerror(int errorCode)
{
    char msg_buf[256];

#if defined(_GNU_SOURCE)
    char* result = strerror_r(errorCode, msg_buf, sizeof(msg_buf));
    return std::string(result);  // Explicit conversion from char* to std::string
#elif !defined(ARCH_COMPILER_MSVC)
    strerror_r(errorCode, msg_buf, sizeof(msg_buf));
    return std::string(msg_buf);
#else
    strerror_s(msg_buf, sizeof(msg_buf), errorCode);
    return std::string(msg_buf);
#endif
}
```

pxr/base/tf/atomicRenameUtil.cpp
```diff
@@ -77,8 +77,14 @@ Tf_AtomicRenameFileOver(std::string const &srcFileName,
     if (stat(dstFileName.c_str(), &st) != -1) {
         fileMode = st.st_mode & DEFFILEMODE;
     } else {
-        const mode_t mask = umask(0);
-        umask(mask);
+        #if defined(ARCH_OS_WASM_VM)
+            // For some reason, EMSCRIPTEN has the mask set to 0777, which
+            // results in files we cannot read
+            const mode_t mask = 0;
```
matthewcpp: this is the pattern to mask out, wonder if 0111 is more reasonable? (mask out executable bit)

pxr/base/tf/fileUtils.h
```diff
@@ -235,6 +235,11 @@ TfReadDir(std::string const &dirPath,
 TF_API
 bool TfTouchFile(std::string const &fileName, bool create=true);
 
+#ifdef ARCH_OS_WASM_VM
+    #define ACCESSPERMS 0777
+    #define DEFFILEMODE 0777
+#endif
```
matthewcpp, line 241: can we use 0666? This isn't for scripts or executables right?

pxr/base/tf/mallocTag.cpp
```diff
@@ -115,8 +115,11 @@ struct Tf_MallocBlockInfo {
     Tf_MallocPathNode *pathNode = nullptr;
 };
 
-#if !defined(ARCH_OS_WINDOWS)
-static_assert(sizeof(Tf_MallocBlockInfo) == 16, 
+#if defined EMSCRIPTEN
+static_assert(sizeof(Tf_MallocBlockInfo) == 8,
```
matthewcpp, line 122: Seems potentially incorrect if we are supporting 64 bit WASM builds?

pxr/base/tf/mallocTag.cpp
```diff
@@ -115,8 +115,11 @@ struct Tf_MallocBlockInfo {
     Tf_MallocPathNode *pathNode = nullptr;
 };
 
-#if !defined(ARCH_OS_WINDOWS)
-static_assert(sizeof(Tf_MallocBlockInfo) == 16, 
+#if defined EMSCRIPTEN
```
matthewcpp: Should this be `ARCH_OS_WASM_VM`?

extras/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
-add_subdirectory(performance)
+if (${PXR_BUILD_PERFORMANCE})
+    add_subdirectory(performance)
+endif()
```
matthewcpp, line 3: For now we just skip this directory if the target is wasm instead of adding this new option

cmake/defaults/Options.cmake
```diff
@@ -38,6 +39,7 @@ option(PXR_ENABLE_NAMESPACES "Enable C++ namespaces." ON)
 option(PXR_PREFER_SAFETY_OVER_SPEED
        "Enable certain checks designed to avoid crashes or out-of-bounds memory reads with malformed input files.  These checks may negatively impact performance."
         ON)
+option(PXR_BUILD_PERFORMANCE "Build Performance Component." ON)
```
matthewcpp: For now can we omit this option? It seems it is used to skip components that don't work with the wasm build.  can we just omit them outright?

cmake/macros/pxrStaticConfig.cmake
matthewcpp, line 1: this file is called static config but is all about loading plugins, can we add a comment explaining how it relates to our actual static build configurations?

build_scripts/build_usd.py
```diff
@@ -2583,12 +2683,17 @@ def ForceBuildDependency(self, dep):
...
+if context.targetWasm:
+    if not which("emcc"):
+        PrintError(" Wasm compiler emcc not found -- please install a compiler")
+        sys.exit(1)
+else:
+    if (not which("g++") and
+        not which("clang") and
+        not GetXcodeDeveloperDirectory() and
+        not GetVisualStudioCompilerAndVersion()):
+        PrintError(" C++ compiler not found -- please install a compiler")
```
matthewcpp: extra ' '  at beginning of message

build_scripts/build_usd.py
```diff
@@ -2583,12 +2683,17 @@ def ForceBuildDependency(self, dep):
...
 # Verify toolchain needed to build required dependencies
-if (not which("g++") and
-    not which("clang") and
-    not GetXcodeDeveloperDirectory() and
-    not GetVisualStudioCompilerAndVersion()):
-    PrintError("C++ compiler not found -- please install a compiler")
-    sys.exit(1)
+if context.targetWasm:
+    if not which("emcc"):
+        PrintError(" Wasm compiler emcc not found -- please install a compiler")
```
matthewcpp: extra ' ' at beginning of message

build_scripts/build_usd.py
```diff
@@ -2463,9 +2538,34 @@ def ForceBuildDependency(self, dep):
...
+    disabled = []
+    components = [
+        ('buildPython', 'Python'),
+        ('buildTools', 'tools'),
+        ('buildExamples', 'examples'),
+        ('buildTutorials', 'tutorials'),
+        ('buildUsdview', 'usdview'),
+        ('buildMaterialX', 'materialX'),
+        ('buildImaging', 'Imaging'),
+        ('buildUsdImaging', 'UsdImaging'),
```
matthewcpp, line 2510: General Question: will Imaging and UsdImaging be enabled later on in this process?

build_scripts/build_usd.py
```diff
@@ -1877,6 +1923,27 @@ def InstallUSD(context, force, buildArgs):
...
+            extraArgs.append('-DTBB_INCLUDE_DIRS="{}"'.format(os.path.join(context.usdInstDir, 'include')))
+            extraArgs.append('-DTBB_tbb_LIBRARY_DEBUG="{}"'.format(os.path.join(context.usdInstDir, 'lib/libtbb_debug.a')))
+            extraArgs.append('-DTBB_tbb_LIBRARY_RELEASE="{}"'.format(os.path.join(context.usdInstDir, 'lib/libtbb.a')))
+
+            extraArgs.append('-DOPENSUBDIV_INCLUDE_DIR="{}"'.format(os.path.join(context.usdInstDir, 'include')))
+            extraArgs.append('-DOPENSUBDIV_OSDCPU_LIBRARY="{}"'.format(os.path.join(context.usdInstDir, 'lib/libosdCPU.a')))
+
+            extraArgs.append('-DPXR_ENABLE_GL_SUPPORT=ON')
+            extraArgs.append('-DBUILD_SHARED_LIBS=OFF')
+            extraArgs.append('-DPXR_BUILD_PERFORMANCE=OFF')
```
matthewcpp: As per other BUILD_PERFORMANCE related comments, i think this can be removed?

build_scripts/build_usd.py
```diff
@@ -1044,7 +1082,9 @@ def InstallOneTBB(context, force, buildArgs):
     TBB_URL = "https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2020.3.1.zip"
 
 def InstallTBB(context, force, buildArgs):
-    if Windows():
+    if context.targetWasm:
+        raise RuntimeError("TBB is no longed used for WebAssembly builds. Use oneTBB instead.")
```
matthewcpp: can we go with ("OneTBB is required for WebAssembly builds.") ?

build_scripts/build_usd.py
```diff
@@ -1017,13 +1046,22 @@ def InstallBoost(context, force, buildArgs):
 ############################################################
 # Intel oneTBB
 
-ONETBB_URL = "https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.9.0.zip"
+ONETBB_URL = "https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.12.0.zip"
```
matthewcpp, line 1012: Do we need to update to this version specifically?  If so we might need to break this out to a separate change.

build_scripts/build_usd.py
matthewcpp, line 1: There are a number of instances in this file where the code is conflating target with build host.  We think it would be best to key everything wasm specific on its own.

I feel like all the way through this test should be if Wasm() or WasmTarget() because when we try to get this working on windows and linux, it's gonna be a bit of work to detangle.

The suggestion is to use WasmTarget() because this script in general thinks you are building and targeting the same system. And wasm is not the system we are building on

build_scripts/build_usd.py
```diff
@@ -847,7 +876,7 @@ def InstallBoost_Helper(context, force, buildArgs):
 
         macOSArch = ""
 
-        if MacOS():
+        if MacOS() and not context.targetWasm:
```
matthewcpp, line 868: Refer to comment regarding WasmTarget() above

build_scripts/build_usd.py
```diff
@@ -402,7 +424,7 @@ def RunCMake(context, force, extraArgs = None):
 
     # On MacOS, enable the use of @rpath for relocatable builds.
     osx_rpath = None
-    if MacOS():
+    if MacOS() and not context.targetWasm:
```
matthewcpp, line 428: Refer to comment regarding WasmTarget() above

build_scripts/build_usd.py
```diff
@@ -392,8 +413,9 @@ def RunCMake(context, force, extraArgs = None):
     if generator is not None:
         generator = '-G "{gen}"'.format(gen=generator)
 
+    # EMSCRIPTEN doesn't use VS On Windows
     # Note - don't want to add -A (architecture flag) if generator is, ie, Ninja
-    if IsVisualStudio2019OrGreater() and "Visual Studio" in generator:
+    if not context.targetWasm and IsVisualStudio2019OrGreater() and "Visual Studio" in generator:
```
matthewcpp, line 419: Refer to comment regarding WasmTarget() above

build_scripts/build_usd.py
```diff
@@ -1867,7 +1913,7 @@ def InstallUSD(context, force, buildArgs):
         else:
             extraArgs.append('-DPXR_BUILD_ANIMX_TESTS=OFF')
 
-        if Windows():
+        if Windows() and not context.targetWasm:
```
matthewcpp, line 1880: Refer to comment regarding WasmTarget() above

matthewcpp (commented):
I added comments from the initial internal review.  Thanks!

build_scripts/build_usd.py
kohakukun: Yes. Just wanted to leave this first pr digestible

build_scripts/build_usd.py
kohakukun: yeah, if not we need to make several patches in place, which is not ideal. I'll create an accompanying PR just updating onetbb version

build_scripts/build_usd.py
kohakukun: Not 100% sure what you are suggesting based on the comment. Should this change be `if MacOS() and not WasmTarget():`? Or you mean a different change?

build_scripts/build_usd.py
kohakukun: Created pr https://github.com/PixarAnimationStudios/OpenUSD/pull/3570

pxr/base/arch/timing.cpp
kohakukun: Good catch. I think this was legacy. I ran testHdSortedIdsPerf.cpp removing this and seems to return sensible results. Also `sampleIters` always evaluates to 1 in emscripten without the need of the macros. Will remove this change

build_scripts/build_usd.py
matthewcpp: Sent a follow up on slack, let me know if what i wrote makes sense!

pxr/base/arch/testenv/testArchUtil.cpp
```diff
@@ -141,8 +142,17 @@ ArchTestCrash(ArchTestCrashMode mode)
         _exit(0);
     }
     else if (childPid == -1) {
+#if defined(ARCH_OS_WASM_VM)
```
matthewcpp: It seems that fork is not implemented for pthreads in emscripten, however is this try/catch  actually needed here?

pxr/base/arch/attributes.h
```diff
@@ -250,15 +250,14 @@ struct Arch_ConstructorEntry {
     static void _name(__VA_ARGS__)
 
 #elif defined(ARCH_COMPILER_GCC) || defined(ARCH_COMPILER_CLANG)
-
 // The used attribute is required to prevent these apparently unused functions
 // from being removed by the linker.
 #   define ARCH_CONSTRUCTOR(_name, _priority, ...) \
         __attribute__((used, section(".pxrctor"), constructor((_priority) + 100))) \
-        static void _name(__VA_ARGS__)
+        static void _name()
```
matthewcpp, line 256: Are the args here not needed for wasm specifically?  Is it possible this will break TF_REGISTRY_FUNCTION?

pxr/base/arch/defines.h
```diff
@@ -45,6 +46,12 @@
 
 #if defined(__x86_64__) || defined(__aarch64__) || defined(_M_X64)
 #define ARCH_BITS_64
+#elif defined(__wasm64__)
+#define ARCH_BITS_64
+#define __emscripten__ 1
```
matthewcpp: are these defines of `__emscripten__` needed?  could we use `ARCH_WASM_VM` or `_EMSCRIPTEN` instead?

pxr/base/arch/math.h
```diff
@@ -29,7 +29,7 @@ PXR_NAMESPACE_OPEN_SCOPE
 /// \addtogroup group_arch_Math
 ///@{
 
-#if defined (ARCH_CPU_INTEL) || defined (ARCH_CPU_ARM) || defined (doxygen)
+#if defined (ARCH_CPU_INTEL) || defined (ARCH_CPU_ARM) || defined (doxygen) || defined(ARCH_OS_WASM_VM)
```
matthewcpp: nit: Can you please move the `ARCH_OS_WASM_VM` check before the `doxygen` check?

pxr/base/arch/symbols.cpp
```diff
@@ -28,7 +27,10 @@ ArchGetAddressInfo(
     std::string* objectPath, void** baseAddress,
     std::string* symbolName, void** symbolAddress)
 {
-#if defined(_GNU_SOURCE) || defined(ARCH_OS_DARWIN)
+#if defined(ARCH_OS_WASM_VM)
```
matthewcpp: I think we should be able to remove this directive.  In that case it should fall through to the return false at the bottom

pxr/base/tf/testenv/dl.cpp
matthewcpp, line 1: This file is meant to be arch independent.  Is it possible we could rework this a bit to not introduce `#if`'s on `ARCH_WASM_VM` here?

pxr/base/tf/envSetting.h
```diff
@@ -154,6 +156,10 @@ bool Tf_ChooseEnvSettingType(bool);
 int Tf_ChooseEnvSettingType(int);
 std::string Tf_ChooseEnvSettingType(char const *);
 
+TF_API
```
matthewcpp: Is there a specific need to expose this?

pxr/base/tf/pointerAndBits.h
```diff
@@ -12,10 +12,137 @@
...
+        : _ptr(p), _bits(bits)
+    {
+    }
+
+    constexpr uintptr_t GetMaxValue() const {
+        return std::numeric_limits<uint32_t>::max();
+    }
+
+    constexpr uintptr_t GetNumBitsValues() const {
+        return 32;
```
matthewcpp: This should be 2**32

pxr/base/tf/pointerAndBits.h
```diff
@@ -12,10 +12,137 @@
...
+    constexpr T *Get() const noexcept {
+        return _GetPtr();
+    }
+
+    /// Retrieve the raw underlying value.  This can be useful for doing literal
+    /// equality checks between two instances.  The only guarantees are that
+    /// this has the same bit pattern as the pointer value if the bits are 0,
+    /// and will compare equal to another instance when both have identical
+    /// pointer and bits values.
+    constexpr uintptr_t GetLiteral() const noexcept {
```
matthewcpp: Does this work correctly? Isn't sizeof(uintptr_t) == 32 on 32-bit wasm?

pxr/base/tf/stackTrace.h
```diff
@@ -14,6 +14,7 @@
 #include <ctime>
 #include <iosfwd>
 #include <string>
+#include <sys/types.h>
```
matthewcpp: Is this explicitly needed?

matthewcpp (commented):
I added some additional feedback that came in on our side.  Thanks!

pxr/base/tf/stackTrace.h
kohakukun: I think it is a leftover

pxr/base/tf/testenv/dl.cpp
kohakukun: Oh, your are right. This one should be in the failed test list without a proper solution yet

kohakukun: Good point. I'll rollback the cahnge

cmake/macros/pxrStaticConfig.cmake
kohakukun: Maybe the name is not very clear but, this is the explanation of this file and the initial commit:

> Description of Change(s)
When building with emscripten, a number of packages have to be wrapped in --whole-archive. Prior to this PR a consuming application would have to wrap packages it included in --whole-archive as well. After this change, that's handled in the generated cmake. See changes to usdviewweb/CMakeLists.txt for an example.

> A note about --whole-archive: When not using emscripten, it's possible to use $<LINK_LIBARAY:WHOLE_ARCHIVE,${lib}> to wrap that lib in the whole-archive flag. This isn't supported by CMake when using emscripten so instead we wrote our own version $<LINK_LIBRARY:LOAD_PLUGIN,${lib}> that does the same thing

Do you have any suggestions regarding the how to rename it?

pxr/base/tf/atomicRenameUtil.cpp
kohakukun: I think we had some issues with some situations but, I cannot reproduce with the current tests. I'll keep it in our local changes but rollback the change in the PR for now

pxr/base/arch/testenv/testArchUtil.cpp
kohakukun: You are right. This is the same issue and some other tests and the exception. I'll rollback the change and it should be handled differently and in a more general way

pxr/base/arch/attributes.h
kohakukun: Yeah, the compiler doesn't support variable arguments for a static constructor. We are currently building for all platforms and doesn't seem to have caused issues. No runtime errors that we have seen nor failing tests nor compilation issues also on linux

pxr/base/arch/testenv/testArchPRead.cpp
matthewcpp, line 1: The format string changes in this file cause build errors for us on linux.  Perhaps using the `PRId64` [format string macro](https://en.cppreference.com/w/c/types/integer#Format_macro_constants) would do the trick?  This worked for me when running this though our internal test gauntlet.

pxr/imaging/hd/testenv/testHdSortedIdsPerf.cpp
matthewcpp, line 1: Did this file creep in by accident?  In any event I think the use of the format string macro i mentioned in the comment for `testArchPRead.cpp` would be good to apply here.  The same error resulted when compiling this file.

pxr/base/arch/attributes.h
matthewcpp: While this change does not cause any failures in the open source build, it does unfortunately break other pixar code that relies on this macro.  Would you mind guarding the change you made behind  `ARCH_OS_WASM_VM`?

pxr/imaging/hd/testenv/testHdSortedIdsPerf.cpp
kohakukun: Not really. I used this file to test the timing api if I remember correctly. I know it is not enabled by default but I used for sanity check. On mac, it was also complaining about these lines, although the test didn't fail. Wanted to also make the change together with the other file

pxr/base/arch/testenv/testArchPRead.cpp
kohakukun: Works

pxr/base/tf/pointerAndBits.h
kohakukun: Good catch. I think this is kinda legacy, I managed to remove it and the test `TfPointerAndBits` now passes. I was reading the original code and seems to be safe to think that it should work for 32 and 64 bits

PR #3482, opened 2025-01-09
Autodesk: Set Target Colorspace on context

erikaharrison-adsk:
### Description of Change(s)

We found a Color Space diff in testUsdImagingGLMaterialXCustomNodes_customMaterial.
The issue is that target colorspace should be set on the generator context instead of the document.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/imaging/hdSt/materialXFilter.cpp
+++ b/pxr/imaging/hdSt/materialXFilter.cpp
@@ -224,10 +224,8 @@ HdSt_GenMaterialXShader(
     cms->loadLibrary(stdLibraries);
     mxContext.getShaderGenerator().setColorManagementSystem(cms);
 
-    // Set the colorspace
-    // XXX: This is the equivalent of the default source colorSpace, which does
-    // not yet have a schema and is therefore not yet accessable here 
-    mxDoc->setColorSpace("lin_rec709");
+    // Set the target colorspace
+    mxContext.getOptions().targetColorSpaceOverride = "lin_rec709";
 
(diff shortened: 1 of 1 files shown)


meshula:
@ashwinbhat This looks right, I think, but could explain what error you encountered? Was it something like the document color space was acescg, but the colors got interpreted as lin_rec709? How did you detect the problem? is it possible to catch the issue in a unit test to accompany the fix?

ashwinbhat:
> @ashwinbhat This looks right, I think, but could explain what error you encountered? Was it something like the document color space was acescg, but the colors got interpreted as lin_rec709? How did you detect the problem? is it possible to catch the issue in a unit test to accompany the fix?

@meshula I have been trying to integrate the material datalibrary referencing concept that was introduced in MaterialX. See [PR 2054](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2054) 
In our internal testing this has shown improvements in document load time.  Datalibrary referencing is not a required change for USD-Mtlx 1.39.x integration, the import based based workflow is still available in MaterialX.
We started seeing glimaging mtlx unit test failures with Colorspace issues. The root cause is that the datalibrary is referenced and not imported.  It highlighted how we were incorrectly setting up the target colorspace for shader generation. Setting the colorspace on the context establishes the target color space for shader generation, while setting the colorspace on the document potentially modifies the source color spaces without intending to.

jesschimein:
Filed as internal issue #USD-10561

meshula:
Got it, thank you for the explanation!

PR #3486, opened 2025-01-17
[sdf] Fix Sdf_Children<ChildPolicy>::Find to return correct value

nvidia-jomiller:
In most cases, Sdf_Children<ChildPolicy>::Find would return the incorrect value of 0 instead the size of _childNames. This would incorrectly index into the first element when a match was not found

### Checklist

[x] I have added unit tests that exercise this functionality (Reference: 

### DIFF
--- a/pxr/usd/sdf/children.cpp
+++ b/pxr/usd/sdf/children.cpp
@@ -86,19 +86,18 @@ size_t
 Sdf_Children<ChildPolicy>::Find(const KeyType &key) const
 {
     if (!TF_VERIFY(IsValid())) {
-        return 0;
+        return _childNames.size();
     }
 
     _UpdateChildNames();
 
     const FieldType expectedKey(_keyPolicy.Canonicalize(key));
-    size_t i = 0;
-    for (i=0; i < _childNames.size(); i++) {
+    for (size_t i=0; i < _childNames.size(); i++) {
         if (_childNames[i] == expectedKey) {
-            break;
+            return i;
         }
     }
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10583

sunyab:
Hey @nvidia-jomiller, 

I'm not quite seeing the bug you mentioned in your PR description. Prior to your change, in the case where the given `key` isn't found, the index `i` would have been incremented to `_childNames.size()` and then returned. The test case you added worked both with and without your change. AFAICT, the only functional change is that in the failed verify case we now return `_childNames.size()` instead of 0.

Am I missing something?

nvidia-jomiller:
Hey @sunyab , thanks for taking a look!

The change was mostly facilitated by another task but yeah, now that I read through it line by line I see functionally it's equivalent. What I was mostly going from is checks in [SdfChildrenView](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/sdf/childrenView.h#L555) and code docs in [Sdf_Children](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/sdf/children.h#L65). I didn't see any tests specifically testing it so I modified and added a test. I think this might just be a readability thing where I was expecting to see `_childNames.size()` returned. But I'm happy to close this out

sunyab:
Ok, thanks for confirming! Please leave this open for now, if nothing else I think we can merge in the test since more tests are always appreciated.

Issue #3492, opened 2025-01-23
Hydra scene index mode doesn't resolve UDIM paths


marktucker:
Try to render the attached scene with USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1.
If you start the render process from some directory other than the one containing the USDA file, the UDIM maps will not be found (if the cwd for the process is where the USDA file is, chances are the renderer will find the UDIMs using the unresolved relative paths).

I can see no evidence that when using scene index plugins that UDIM paths are run through UsdShadeUdimUtils::ResolveUdimPath as they are when using the legacy hydra delegate.

[udim.zip](https://github.com/user-attachments/files/18525005/udim.zip)

marktucker:
If this is not currently on your radar to be fixed for 25.05, I can try to come up with a PR.

jesschimein:
Filed as internal issue #USD-10592

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

tcauchois:
This is the first I've seen this, although it doesn't look too hard to fix.  I'll gladly accept any PRs though!

We could consider specializing UsdImagingDataSourceAttribute<SdfAssetPath> to check for the "<UDIM>" string and call this; or have UsdImagingDataSourceMaterial wrap asset parameters and call this...  While I don't particularly expect UDIMs outside of shader inputs, there's something nice about handling it uniformly in our baseline asset path handling.  Let me know what you think!

Regardless, we'll try to address this for 25.05.  It seems relatively high value for effort.

marktucker:
Just adding a comment to make sure you see that I did submit a PR for this...

tcauchois:
Ahh thanks, missed that!

PR #3494, opened 2025-01-24
Fix UDIM path resolution for scene index plugins

marktucker:
### Description of Change(s)

Implement a specialization of UsdImagingDataSourceAttribute<SdfAssetPath>
that properly resolves UDIM relative paths.

Note that I have not created a unit test, or run the unit tests yet. I'm posting this PR early because there is an obvious issue that I want to discuss, namely the correct "location" for this code change. This change, as is, will affect _all_ SdfAssetPath attributes read from USD and accessed by hydra. It is not clear to me if this is a good thing or a bad thing.

If it's a bad thing, I can take this UsdImagingDataSourceAttribute<SdfAssetPath> sepcialization and move it to dataSourceMaterial.cpp instead. Then in _UsdImagingDataSourceShadingNodeParameters::Get, I can conditionally create this data source for SdfAssetPath attributes instead of calling UsdImagingDataSourceAttributeNew to get the "standard" SdfAssetPath handler.

Alternatively, UsdImagingDataSourceAssetPathAttribute could be defined and implemented in dataSourceAttribute.h/cpp, but the standard factory would still return a vanilla UsdImagingDataSourceAttribute<SdfAssetPath>. But in dataSourceMaterial.cpp I could introduce the condition to use UsdImagingDataSourceAssetPathAttribute  for SdfAssetPath attributes. This has the nice feature that if any other scene index code wants to support UDIM translation, the specialized subclass can be shared.

I personally think the universality might be a good thing as if some renderer supports UDIMs on a procedural or other custom prim type, they won't have to jump through hoops or create a custom data source to handle that case. The downside is every time hydra accesses an SdfAssetPath attribute, we will scan that string looking for "<UDIM>". To me this seems like a small price, but happy to accept Pixar's judgement on this and alter this PR as needed.

### Fixes Issue(s)
#3492 

### Checklist

### DIFF
--- a/pxr/usdImaging/usdImaging/dataSourceAttribute.cpp
+++ b/pxr/usdImaging/usdImaging/dataSourceAttribute.cpp
@@ -8,9 +8,77 @@
 
 #include "pxr/imaging/hd/dataSourceLocator.h"
 
+#include "pxr/usd/usdShade/udimUtils.h"
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
+
+// We need to find the first layer that changes the value
+// of the parameter so that we anchor relative paths to that.
+static
+SdfLayerHandle
+_FindLayerHandle(const UsdAttribute& attr, const UsdTimeCode& time) {
+    for (const auto& spec: attr.GetPropertyStack(time)) {
+        if (spec->HasDefaultValue() ||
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10594

PR #3495, opened 2025-01-26
Build without Boost when using OpenVDB 12 or newer

brechtvl:
### Description of Change(s)

In OpenVDB 12 there is no more header dependency on Boost (and it can optionally be built entirely without Boost). Therefore OpenUSD itself does not need a direct dependency on Boost when using this version.

This adds OpenVDB version detection based on the header files, and moves CMake OpenVDB finding before Boost so that the version can be checked.

### Checklist


dgovil:
Thanks , Brecht. This is super useful. 

I'll also soon be putting up a PR to upgrade to VDB 12.0.1 (or 12.1? Not sure what it'll be) which will include the clang fixes as well. So this is fortuitous timing.

jesschimein:
Filed as internal issue #USD-10595

PR #3501, opened 2025-01-29
Add ZLIB to requiredDependencies if HDF5 is enabled

mhamid3d:
### Description of Change(s)

Alembic requires ZLIB, but only in the case that HDF5 support is enabled. Since we are selectively adding zlib as a dependency, we need to append zlib to `requiredDependencies` if HDF5 is enabled in the build context. This fixes alembic build failures on Windows.

### Fixes Issue(s)

This is an extension of #2988 which was selectively adding zlib as a dependency based on other active dependencies.

### Checklist

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2464,7 +2464,7 @@ if context.buildBoostPython:
 
 if context.buildAlembic:
     if context.enableHDF5:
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10613

PR #3505, opened 2025-01-30
APPLE: Fix iOS not clearing background buffer when animating/navigating.

dgovil:
### Description of Change(s)

There was an errant availability macro that was causing the path to be taken on only non-macOS platforms. This caused issues where the background would retain the previous background, causing a Solitaire win like effect.

This patch makes the color pass use the same pattern as the depth pass, and adequately clear the background.
I believe this resolves to the same LoadAction, but something about the availability macro was causing an issue here.

Either way, at least it cleans it up.

Tested on iOS 18.3

### Checklist


jesschimein:
Filed as internal issue #USD-10617

Issue #3508, opened 2025-01-31
primvars:velocities actively pruned in Hydra 2


robp-sidefx:
The change https://github.com/PixarAnimationStudios/OpenUSD/commit/c55f1384b37ffd74fac58b6a18ec33b2699b4687#diff-4058391680f6b738982c851862d178868840850626cbaa3ad17c597b6c21014f added rejection of certain primvars in `UsdImagingDataSourcePrimvars` such that they aren't passed along to the renderer, specifically `points`, `velocities`, `accelerations`.

It's unclear to me from the commit message and/or comments added what the issue with `primvars:velocities` was, but its active rejection is an issue for us, where we will apply this primvar to `Xform` prims as a means of expressing the overall velocity of an aggregate of objects.

We'd like to see this rejection trimmed back to just `points` but, again, admit there may be context we don't understand/appreciate.

jesschimein:
Filed as internal issue #USD-10623

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

spiffmon:
Hi @robp-sidefx , this was done to keep Hydra true to the UsdGeom OM, which explicitly calls out that `normals` and `widths` can be **overridden** by primvars in certain schemas, but not other intrinsic attributes.  We were concerned that you might wind up getting different results from (e.g.) `UsdGeomPointBased::ComputePointsAtTime()` and the points a renderer actually uses (calculated using a primvars:velocities).

And in what you describe, aren't you effectively saying you're going to get different computed affine transformations over time (in subframes) in the renderers that subscribe to this use of primvars than you would get with standard  UsdGeomXformable evaluation?

robp-sidefx:
> the UsdGeom OM, which explicitly calls out that normals and widths can be overridden by primvars in certain schemas, but not other intrinsic attributes.

Sorry, I've somehow missed that, and am now struggling to find it. What should I be reading? :)

> aren't you effectively saying you're going to get different computed affine transformations over time (in subframes) in the renderers that subscribe to this use of primvars than you would get with standard UsdGeomXformable evaluation?

Potentially yes, but I don't find that concerning. I'd expect that any values authored were authored for a reason, regardless how minimally or enormously they affect the visual result.

> We were concerned that you might wind up getting different results from (e.g.) UsdGeomPointBased::ComputePointsAtTime() and the points a renderer actually uses (calculated using a primvars:velocities).

[...]

PR #3509, opened 2025-01-31
Do not reject 'velocities' or 'accelerations' as primvars for the renderer

robp-sidefx:
### Description of Change(s)

Selectively undoing part of https://github.com/PixarAnimationStudios/OpenUSD/commit/c55f1384b37ffd74fac58b6a18ec33b2699b4687#diff-4058391680f6b738982c851862d178868840850626cbaa3ad17c597b6c21014f to restore `velocities` and `accelerations` as valid primvars for the renderer.

### Fixes Issue(s)

#3508 

### Checklist

### DIFF
--- a/pxr/usdImaging/usdImaging/dataSourcePrimvars.cpp
+++ b/pxr/usdImaging/usdImaging/dataSourcePrimvars.cpp
@@ -40,18 +40,11 @@ _GetInterpolation(const UsdAttribute &attr)
 
 // Reject primvars:points since we always want to get the value from
 // the points attribute.
-// Similar for velocities and accelerations.
 static
 bool
 _RejectPrimvar(const TfToken &name)
 {
     if (name == UsdGeomTokens->points) {
-        return true;
-    }
-    if (name == UsdGeomTokens->velocities) {
-        return true;
-    }
-    if (name == UsdGeomTokens->accelerations) {
         return true;
     }
 


jesschimein:
Filed as internal issue #USD-10624

unhyperbolic:
"velocities" instead of "primvars:velocities" should still work.

Did the UsdImaging Delegate actually take "primvars:velocities" for a point instancer? I am looking at UsdImagingPointInstancerAdapter::Get and I don't see that it does.

robp-sidefx:
> "velocities" instead of "primvars:velocities" should still work.

To clarify, the issue (as it shows up as a problem for us) in https://github.com/PixarAnimationStudios/OpenUSD/issues/3508 is that the primvar is being blocked when specified on an `Xform`. Specifying "velocities" as an attribute on an `Xform` also, as best I can tell, doesn't flow through Hydra.

Feedback from @spiffmon on the issue suggested reluctance to revert this, so we'll likely proceed with Houdini-specific primvar naming to achieve our desired workflow.

spiffmon:
Hi @robp-sidefx , just wanted to clarify rationale here while closing the issue.  Irrespective of Hydra 1 behavior, we don't want the UsdGeomPointBased to support primvar "overrides" for `velocities` or `accelerations` because, unlike `normals` and `widths`, they do not support the `interpolation` feature of primvars - they are very specifically expressing opinions about the `points` property and cannot interpolate across edges/faces, so it'd be setting ourselves up for "bad/meaningless data" that we'd need to validate against.  Since you mention inheriting from an xform, I'm guessing you're talking about `constant`  velocities... I think I'd argue that constant velocity/acceleration/angularVelocity would be something better expressed at the Xformable level (which would automatically address the inheritance aspect), although that would add considerable complexity to ctm computations.

PR #3520, opened 2025-02-05
[Tf] Update TF_FOR_ALL to C++11 iterator 

Cewein:
### Description of Change(s)

Update files under pxr/base/tf by replacing TF_FOR_ALL with C++11-style range-based for each loop.
Additionally, updated pointer usage to align with the new iterator approach, since we now handle references directly rather than raw pointers.

No modification to `iterator.h` was done nether to the part of `bits.h` and `compressedBits.h` that provide support for `TF_FOR_ALL` as it break the compilation and because the modification of the iterators regarding `bits` and `compressedBits` is out of scope of this PR.

File in sub folder have not been touch yet except `testenv/bits`.

### Ongoing Issue(s)

This PR addresses issue https://github.com/PixarAnimationStudios/OpenUSD/issues/80, though many more instances still need to be resolved. 

### Checklist

### DIFF
--- a/pxr/base/tf/errorMark.cpp
+++ b/pxr/base/tf/errorMark.cpp
@@ -109,11 +109,11 @@ TfReportActiveErrorMarks()
         localStacks = TfErrorMark_GetActiveMarkStacks();
     }
 
-    TF_FOR_ALL(i, localStacks) {
+    for(const auto& i: localStacks) {
         printf("== TfErrorMark @ %p created from ===========================\n",
-               i->first);
+               i.first);
         std::stringstream ss;
-        ArchPrintStackFrames(ss, i->second);
+        ArchPrintStackFrames(ss, i.second);
         printf("%s\n", ss.str().c_str());
     }
 }
--- a/pxr/base/tf/mallocTag.cpp
+++ b/pxr/base/tf/mallocTag.cpp
@@ -190,8 +190,8 @@ Tf_MallocTagStringMatchTable::SetMatchList(const std::string& matchList)
 {
     _matchStrings.clear();
     std::vector<std::string> items = TfStringTokenize(matchList, ",\t\n");
-    TF_FOR_ALL(i, items) {
-        _matchStrings.push_back(_MatchString(TfStringTrim(*i, " ")));
+    for(std::string i: items) {
+        _matchStrings.push_back(_MatchString(TfStringTrim(i, " ")));
     }
 }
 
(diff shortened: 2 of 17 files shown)


nvmkuruc:
Minor note-- It looks like your change description is referencing ` pxr/base/plug` instead of ` pxr/base/tf`.

Cewein:
> Minor note-- It looks like your change description is referencing ` pxr/base/plug` instead of ` pxr/base/tf`.

Thank you for noticing, updated the description, added also that the file under `pxr/base/tf/testenv` where almost not modified. I reserve that for another PR since it would have pilled up to too many file modified at the same time.

If you said this is alright would be happy to push the modification here.

jesschimein:
Filed as internal issue #USD-10653

pxr/base/tf/mallocTag.cpp
```diff
@@ -190,8 +190,8 @@ Tf_MallocTagStringMatchTable::SetMatchList(const std::string& matchList)
 {
     _matchStrings.clear();
     std::vector<std::string> items = TfStringTokenize(matchList, ",\t\n");
-    TF_FOR_ALL(i, items) {
-        _matchStrings.push_back(_MatchString(TfStringTrim(*i, " ")));
+    for(std::string i: items) {
```
meshula, line 193: could be a reference here, `std::string&`. could reserve `_matchStrings` to `matchList.size` even though the existing code didn't do it. `emplace_back` would allow construction in place. Another optimization that didn't originally exist, but since we're touching the loop :)

pxr/base/tf/errorMark.cpp
```diff
@@ -109,11 +109,11 @@ TfReportActiveErrorMarks()
         localStacks = TfErrorMark_GetActiveMarkStacks();
     }
 
-    TF_FOR_ALL(i, localStacks) {
+    for(const auto& i: localStacks) {
```
meshula, line 112: Throughout the diff, let's match the style of the surrounding code, with a space between for and the paren. 

In general the convention with auto is to only use it when the type is visible in the context. Like if `std::vector<float>` was declared on a previous line, `auto` would obvious from context. In the case of the _ActiveMarkStacksMap we have to go digging to discover that the iteration will be on a series of `vector<uintptr_t>` so it would help readability if the line read

`for (const vector::<uintptr_t>& i: localStacks) {`
Since the goal of this PR is to improve readability by expanding the TF_FOR_ALL macro, I think it's worthwhile to think about each replacement.

meshula:
Thanks for addressing the compile issue. Since we are using this PR to figure out the best way to update the TF_FOR_ALL macros, I've left a couple of suggestions in line to help iterate ideas.

pxr/base/tf/type.cpp
```diff
@@ -1096,9 +1096,9 @@ TfType::CastFromAncestor(TfType ancestor, void* addr) const
         return addr;
 
     ScopedLock regLock(GetRegistryMutex(), /*write=*/false);
-    TF_FOR_ALL(it, _info->baseTypes) { 
-        if (void* tmp = it->CastFromAncestor(ancestor, addr)) {
-            if (_CastFunction *castFunc = _info->GetCastFunc(it->GetTypeid()))
+    for(const auto& it: _info->baseTypes) { 
+        if (void* tmp = it.CastFromAncestor(ancestor, addr)) {
+            if (_CastFunction *castFunc = _info->GetCastFunc(it.GetTypeid()))
```
Cewein, line 1101: Since recent discussion, this is a case where TF_FOR_ALL should be keep as of now in my opinion.

`it->GetTypeid()` require direct access to the iterator and will loose it when switching to range based iterator

pxr/base/tf/errorMark.cpp
Cewein: this make sense, I was not sure on how to update with `auto` or the actual type since sometime the `auto` as sometime it might translate to lengthy types.

PR #3528, opened 2025-02-10
Add test to reproduce multi-threading issue of computing extent and relationship

roggiezhang-nv:
### Description of Change(s)
We found a very strange behavior that `ComputeExtent` along with `GetForwardedTargets` in a multi-threading way would influence the behavior of getting material binding with API `UsdRelationship::GetForwardedTargets`. See the changes of this PR about the test to reproduce this. In this test, it's trying to traverse the whole stage with multi-threads. All operations are supposed to be read-only to the stage, and the test should be successful. However, it fails unexpectedly sometimes. Adding `TfRegistryManager::GetInstance().SubscribeTo<UsdGeomBoundable>();` to the first line of `main()` will pass the test, which we still have no idea why.

You need to run the test with `--repeat until-fail:1000` to increase the possibility of failure.  

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/3529

### Checklist

### DIFF
--- a/pxr/usd/usdGeom/CMakeLists.txt
+++ b/pxr/usd/usdGeom/CMakeLists.txt
@@ -116,6 +116,20 @@ pxr_build_test(testUsdGeomHasAPI
         testenv/testUsdGeomHasAPI.cpp
 )
 
+pxr_build_test(testUsdGeomExtentThreading
+    LIBRARIES
+        usd
+        usdGeom
+        usdShade
+    CPPFILES
+        testenv/testUsdGeomExtentThreading.cpp
+)
+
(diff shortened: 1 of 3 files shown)


roggiezhang-nv:
@nvmkuruc for vis.

nvmkuruc:
It was a little harder to reproduce on Linux, but I was able to reproduce with `--repeat-until-fail 10000`.

jesschimein:
Filed as internal issue #USD-10665

nvmkuruc:
@matthewcpp Now that #3529 has been closed, should we close this?

PR #3535, opened 2025-02-19
APPLE: Fix Missing cstdef

dgovil:
### Description of Change(s)

An upstream change to Clang will require that size_t be explicitly brought in from a header, rather than implicitly used, as it allows delineation between different declarations of size_t.

The compiler versions are free to move around and/or remove unnecessary implicit includes internally to improve build times regularly and breaks folks accidentally relying on transitive includes quite often. libc++ 20.0.0 removed some transitive includes which causes breakages in code that was accidentally relying on those, violating the "include-what-you-use" principle

This change brings them in from cstddef and prefixes them with the std namespace

### Checklist

### DIFF
--- a/pxr/base/gf/half.h
+++ b/pxr/base/gf/half.h
@@ -18,6 +18,8 @@
 #include "pxr/base/gf/ilmbase_halfLimits.h"
 #include "pxr/base/gf/traits.h"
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10686

PR #3543, opened 2025-02-25
Autodesk: HgiPresent

erikaharrison-adsk:
### Description of Change(s)

- Add `HgiPresent`
    - For now only implement Vulkan presentation for window presentation
    - Metal presentation will come later
- Make `HgiInterop` an implementation of `HgiPresent`
- Update `HgiInterop` to be more consistent about its composition parameters
- Update `HdxPresentTask` to use `HgiPresent` instead of `HgiInterop`
- Update `HdxTaskController` and `UsdImagingGLEngine` to support `HgiPresent`
- Update all clients of `UsdImagingGLEngine` to default to `HgiInterop` presentation
- Fix various symbol export issues I came across

Tested with:
- Windows / WIN32 / NVIDIA
- Windows / WIN32 / Lavapipe
- Linux / X11 / Lavapipe
- macOS / Metal / Lavapipe

- N/A

### Fixes Issue(s)

- N/A

### Checklist


jesschimein:
Filed as internal issue #USD-10700

pxr/imaging/hdx/presentTask.cpp
```diff
@@ -103,11 +88,7 @@ HdxPresentTask::Execute(HdTaskContext* ctx)
         // framebuffer contents.
         // Eg. This allows us to render with HgiMetal and present the images
```
meshula: This comment could be improved as well, since this change is no longer focused specifically on GL interop. Something like ~

```
Accordingly, we may render with an Hgi backend that differs from the rendering system used by a particular application. for example, using Vulkan to render, and OpenGL to display in an application.
```

pxr/imaging/hdx/presentTask.h
```diff
@@ -82,6 +55,12 @@ class HdxPresentTask : public HdxTask
     HDX_API
     void Execute(HdTaskContext* ctx) override;
 
+    /// Returns true if the format is supported for presentation. This is useful
+    /// for upstream tasks to prepare the AOV data accordingly, and keeps the
+    /// interop step simple.
```
meshula: interop is jargonny, since we're eliminating the use of HgInterop here, I think we should say

```
/// ... and keeps
/// api-interoperable presentation simple.
```

pxr/imaging/hdx/taskController.cpp
```diff
@@ -1930,26 +1950,63 @@ void
...
 
-        GetRenderIndex()->GetChangeTracker().MarkTaskDirty(
-            _colorCorrectionTaskId, HdChangeTracker::DirtyParams);
+        if (!_presentTaskId.IsEmpty()) {
+            auto presentParams = _delegate.GetParameter<HdxPresentTaskParams>(
+                    _presentTaskId, HdTokens->params);
+            if (const auto windowDst = std::get_if<HgiWindowPresentParams>(
+                &presentParams.destination)) {
+
+                if (params.colorCorrectionMode.IsEmpty() ||
```
meshula: It would be preferable that we track the Hgi target render color space, and the HgiWindow color space at a higher level, rather than hard coding  the values here, but I realize that such a change should be out of scope since there is more work to be done to ensure that Hgi works in the right color space throughout the rest of the code.

pxr/imaging/hgi/types.cpp
```diff
@@ -246,14 +246,18 @@ HgiIsFloatFormat(
...
     case HgiFormatFloat32:
-    case HgiFormatFloat32UInt8: // Unclear
     case HgiFormatBC6FloatVec3:
     case HgiFormatBC6UFloatVec3:
     case HgiFormatPackedInt1010102:
         return true;
     case HgiFormatInt16:
     case HgiFormatUInt16:
     case HgiFormatInt32:
+    // While this looks like a float with packed integer format,
```
meshula, line 256: I wonder if this is an opportune moment to name this what it is, which is HgiFormatDepth32UInt8. The mapping from Hgi to graphics APIs would be far less ambiguous.

pxr/imaging/hgiInterop/hgiInterop.h
```diff
@@ -73,10 +67,16 @@ class HgiInterop final
     void TransferToApp(
         Hgi *srcHgi,
         HgiTextureHandle const &srcColor,
+        HgiBlendFactor colorSrcBlendFactor,
```
meshula: I'm very curious about why GL style fixed function parameters are passed here? In general wouldn't you just want a blit to a target? Fancy compositing really seems like something in the app domain and something an app could take care of in window/widget composition. ie, the viewport widget could just draw a textured quad, as opposed to trying to set up Hydra to do something a bit fancy? Minimally could we at least also have a not-fancy plain old blit? This set up seems like something I wouldn't want to have to deal with in my own apps, for example.

pxr/imaging/hgiInterop/metal.h
```diff
@@ -42,9 +42,16 @@ class HgiInteropMetal final
     HGIINTEROP_API
     void CompositeToInterop(
         HgiTextureHandle const &color,
+        HgiBlendFactor colorSrcBlendFactor,
```
meshula: ditto on retaining a simple variant

pxr/imaging/hgiInterop/metal.h
```diff
@@ -77,8 +84,16 @@ class HgiInteropMetal final
         void* pointer;
     };
 
-    void _BlitToOpenGL(VtValue const &framebuffer, GfVec4i const& compRegion,
-                       int shaderIndex);
+    void _BlitToOpenGL(uint32_t framebuffer,
+        GfRect2i const& compRegion,
```
meshula: ditto on retaining a simple variant

pxr/imaging/hgiInterop/metal.mm
```diff
@@ -679,41 +681,48 @@
 }
 
 void
-HgiInteropMetal::_BlitToOpenGL(VtValue const &framebuffer,
-                               GfVec4i const &compRegion,
-                               int shaderIndex)
+HgiInteropMetal::_BlitToOpenGL(uint32_t framebuffer,
+    GfRect2i const &compRegion,
```
meshula: ditto on retaining a simple variant

pxr/imaging/hgiPresent/metal.mm
```diff
@@ -0,0 +1,318 @@
...
+kernel void flipAndCopyTexture(
+    texture2d<float, access::read> inTexture,
+    texture2d<float, access::write> outTexture,
+    uint2 inCoord [[thread_position_in_grid]])
+{
+    if(inCoord.x >= outTexture.get_width() ||
+        inCoord.y >= outTexture.get_height()) {
+        return;
+    }
+
```
meshula, line 113: was the intent to insert a matrix here to convert the color space to the presentation color space? GfColorSpace will give you the proper matrix, and it would be very inexpensive to insert it here. GfColorSpace will also give you parameters for the EOTF in the case of SRGB conversion from linear.

pxr/imaging/hgiPresent/vulkan.cpp
```diff
@@ -0,0 +1,756 @@
...
+        return GfColorSpaceNames->LinearRec709;
+    case VK_COLOR_SPACE_SRGB_NONLINEAR_KHR:
+    case VK_COLOR_SPACE_EXTENDED_SRGB_NONLINEAR_EXT:
+        return GfColorSpaceNames->SRGBRec709;
+    case VK_COLOR_SPACE_DISPLAY_P3_NONLINEAR_EXT:
+        return GfColorSpaceNames->SRGBP3D65;
+    case VK_COLOR_SPACE_DISPLAY_P3_LINEAR_EXT:
+        return GfColorSpaceNames->LinearDisplayP3;
+    case VK_COLOR_SPACE_BT709_NONLINEAR_EXT:
+        // Neither a gamma of 1.8 nor 2.2 are good approximations
```
meshula, line 46: Since this refers to BT709 nonlinear, I cannot see a reason to deviate from SRGBRec709 here. The only difference I can think of between SRGB and BT709 transfer functions is in the number of digits of precision each specification records. I think we should just use SRGBRec709, and there is no need for a comment.

pxr/imaging/hgiPresent/vulkan.cpp
```diff
@@ -0,0 +1,756 @@
...
+    std::vector<VkSurfaceFormatKHR> formats{formatCount};
+    HGIVULKAN_VERIFY_VK_RESULT(
+        vkGetPhysicalDeviceSurfaceFormatsKHR(physicalDevice, _surface, &
+            formatCount, formats.data()));
+
+    using CandidateFormat = std::pair<VkSurfaceFormatKHR, _ColorVolume::Match>;
+    std::vector<CandidateFormat> candidateFormats;
+    for (const auto formatAndColorSpace : formats) {
+        const auto [format, colorSpace] = formatAndColorSpace;
+        if (_VkColorSpaceEnumToName(colorSpace) != _params.surfaceColorSpace) {
```
meshula, line 488: similarly to the metal case, although the hardware can help us with some conversions, it can't help with all. So I think there should also be a use of the GfColorSpace matrices and coefficients in a shader to help with those cases. VK doesn't know about ACES for example.

pxr/imaging/hdx/taskController.cpp
DDoS: It's not so much as hardcoded as these are the only cases Storm supports right now. Rendering is in linear extended sRGB, and optional colour correction converts it to non-linear (extended?) sRGB or some arbitrary user defined colour space.
I don't think there's anything wrong with using extended linear sRGB for rendering, since it simplifies the rendering code around one standard space, and doesn't loose colour information. But I think that the output colour space could be a property of the render index, like it is for the pixel format. I'm not sure if the window needs a separate colour space, it should really make a best effort at preserving the colour information according to the system constraints, which might mean doing extra transformations. To make things easier, the render index needs to have not just some colour space name tokens, but the full transformation data to make it simple to implement the necessary conversions.
Anyway I think this is out of scope of this PR. It does the best it can with the current situation.

pxr/imaging/hgi/types.cpp
DDoS: I could, but it's 19 usages in 10 new files not in the PR. Also I would recommend naming it `HgiFormatDepthFloat32Stencil8` since some APIs like Metal and Vulkan distinguish between colour and depth floats (`VK_FORMAT_D32_SFLOAT` vs `VK_FORMAT_R32_SFLOAT`, `MTLPixelFormatDepth32Float` vs `MTLPixelFormatR32Float`), and stencil is a bit mask not an integer.

pxr/imaging/hgiInterop/hgiInterop.h
DDoS: These parameters were there before, just hardcoded (except for the depth function), so I exposed them and unified the interop backend capabilities. It makes sense that interop would let you configure composition since it might be drawing in a framebuffer that already has data in it, or will be drawn into again (so you might want depth values for example).
And while we could change this to assume the framebuffer is "cleared", just do a plain blit, and ask that the application composite it itself after that; it would add an extra framebuffer copy and break backwards compabitlity for those who rely on the existing composition behaviour.

DDoS: Also worth noting that these operators are still part of modern APIs, for example [`VkPipelineColorBlendAttachmentState`](https://docs.vulkan.org/spec/latest/chapters/framebuffer.html#VkPipelineColorBlendAttachmentState) and [`VkPipelineDepthStencilStateCreateInfo`](https://docs.vulkan.org/spec/latest/chapters/fragops.html#VkPipelineDepthStencilStateCreateInfo), since that part of the fragment pipeline, while more customizable, is still pretty much "fixed function".

pxr/imaging/hgiPresent/metal.mm
DDoS: No, this shader is needed for Metal because it doesn't support blitting to a different format or coordinate systems like Vulkan does. So we need to read the AOV to floats, flip vertically, then write back to the surface format.
But if we do decide that the presentation layer should do a best effort conversion to a supported window colour space, then yes it would go here.

pxr/imaging/hgiPresent/vulkan.cpp
DDoS: How would we know the output of the colour correction task is, for example, ACES? As far as I can tell, only linear extended sRGB, non linear (extended?) sRGB, and some user defined OCIO spaces are supported. The user defined colour space could be ACES, but the strings used by OCIO aren't standardized; they're only IDs to the actual transformation in a config file. At least that's what I've been told.

pxr/imaging/hgiPresent/vulkan.cpp
DDoS: Wikipedia and the Vulkan documentation say the OETF are different.
BT.709:
BT 709
sRGB:
sRGB
> [sRGB](https://en.wikipedia.org/wiki/SRGB) was created after the early development of Rec.709. The creators of sRGB chose to use the same primaries and white point as Rec.709, but changed the tone response curve (sometimes referred to as [gamma](https://en.wikipedia.org/wiki/Gamma_correction)) to better suit the intended use in offices and brighter conditions than television viewing in a dark living room.

pxr/imaging/hgiInterop/hgiInterop.cpp
```diff
@@ -47,19 +47,19 @@ HgiInterop::~HgiInterop() = default;
 void HgiInterop::TransferToApp(
     Hgi *srcHgi,
     HgiTextureHandle const &srcColor,
+    HgiBlendFactor colorSrcBlendFactor,
```
clach: Can we put all of these configuration parameters into a struct instead for better future-proofing?

clach:
We talked about this during the meeting as a potential followup, but it'd be great to add a test executing the window present codepath.

pxr/usdImaging/usdImagingGL/engine.h
```diff
@@ -501,19 +501,22 @@ class UsdImagingGLEngine
...
-    /// itself and skip the engine's presentation.
+    /// Disable the presentation task. An application may choose to manage the
+    /// AOVs that are rendered into itself and skip the task controller's
+    /// presentation.
+    USDIMAGINGGL_API
+    void DisablePresentation();
+
+    /// Enable the presentation task, and configure it to present to a window.
     USDIMAGINGGL_API
-    void SetEnablePresentation(bool enabled);
```
clach, line 514: We don't want to delete this API immediately, and instead should mark it deprecated for at least 1 or 2 releases before deletion. If possible, we could implement the existing functions using the new API.

pxr/imaging/hdx/taskController.h
```diff
@@ -133,14 +135,6 @@ class HdxTaskController final
     HDX_API
     HdAovDescriptor GetRenderOutputSettings(TfToken const& name) const;
 
-    /// The destination API (e.g., OpenGL, see hgiInterop for details) and
-    /// framebuffer that the AOVs are presented into. The framebuffer
-    /// is a VtValue that encoding a framebuffer in a destination API
-    /// specific way.
-    /// E.g., a uint32_t (aka GLuint) for framebuffer object for OpenGL.
-    HDX_API
-    void SetPresentationOutput(TfToken const &api, VtValue const &framebuffer);
```
clach, line 142: Similar to comments in UsdImagingGLEngine, we don't want to delete this API immediately, and instead should mark it deprecated.

pxr/usdImaging/usdImagingGL/engine.h
clach, line 514: Should also consider bumping the UsdImagingGL api version number.

pxr/imaging/hgiPresent/present.h
```diff
@@ -0,0 +1,214 @@
...
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+class Hgi;
+class HgiTexture;
+
+/// Configure the presentation to simply do nothing.
+/// This is the default presentation since it doesn't
+/// require any external resources.
+struct HgiNoOpPresentParams
```
clach: I believe our coding style is everything in a library should be prefaced by the library name.  So this struct name would be "HgiPresentNoOpPresentParams," and the same pattern for everything else in HgiPresent.

pxr/imaging/hgiPresent/metal.mm
```diff
@@ -0,0 +1,318 @@
...
+// HgiMetal: it just adds overhead. Unless we go the full mile abstracting
+// over surfaces, swapchains and other presentation concerns, using the Hgi
+// layer doesn't have any real benefits. Implementing it would be a lot of
+// additional work to replace not that many lines of API specific code.
+
+namespace {
+PXR_NAMESPACE_USING_DIRECTIVE
+
+std::optional<MTLPixelFormat>
+_FindPreferredSurfaceFormatMatch(HgiFormat format, bool needs_sRGB_Transform)
```
clach, line 35: Could this function go in HgiMetal/conversions.mm?

pxr/imaging/hgiPresent/vulkan.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include "pxr/imaging/hgiPresent/presentImpl.h"
+
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+class HgiVulkan;
+
+class HgiPresentVulkan final: public HgiPresentImpl
```
clach: Can we make this name (and the Metal version) more indicative of the fact that this is the Vulkan _window_ present impl?

pxr/imaging/hgiPresent/interopGL.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "pxr/imaging/hgiInterop/hgiInterop.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+/// \class HgiPresentInteropGL
+///
+/// Forward the textures to \class HgiInterop.
+///
+class HgiPresentInteropGL final : public HgiPresentImpl
```
clach: This naming is confusing. This existing `HgiInteropVulkan` means Vulkan-to-GL interop, but `HgiPresentInteropGL` means interop to GL.

pxr/imaging/hdx/presentTask.h
```diff
@@ -22,32 +21,12 @@ PXR_NAMESPACE_OPEN_SCOPE
...
-
-    // Subrectangular region of the framebuffer over which to composite aov
-    // contents. Coordinates are (left, BOTTOM, width, height).
-    GfVec4i dstRegion;
-
-    // When not enabled, present task does not execute, but still calls
-    // Hgi::EndFrame.
-    bool enabled;
+    /// Parameters for the presentation destination. Forwarded to hgiPresent.
+    HgiPresentDestinationParams destination;
```
clach: Personal preference: call this `destinationParams` so it's clear it's more params.

pxr/imaging/hdx/taskController.cpp
```diff
@@ -1930,26 +1950,63 @@ void
 HdxTaskController::SetColorCorrectionParams(
     HdxColorCorrectionTaskParams const& params)
 {
-    if (_colorCorrectionTaskId.IsEmpty()) {
-        return;
-    }
+    if (!_colorCorrectionTaskId.IsEmpty()) {
```
clach: Any reason we got rid of the early-out? Would save us a level of indenting.

pxr/imaging/hgiInterop/vulkan.h
```diff
@@ -36,9 +37,16 @@ class HgiInteropVulkan final
     HGIINTEROP_API
     void CompositeToInterop(
         HgiTextureHandle const &color,
+        HgiBlendFactor colorSrcBlendFactor,
```
clach: Same comment about putting configuration params into struct for here and similar functions.

pxr/imaging/hgiPresent/debugCodes.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "pxr/base/tf/debug.h"
+#include "pxr/base/tf/registryManager.h"
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+TF_REGISTRY_FUNCTION(TfDebug)
+{
+     TF_DEBUG_ENVIRONMENT_SYMBOL(HGIPRESENT_DUMP_CANDIDATE_SURFACE_FORMATS,
+        "Dump candidate VkSurfaceFormatKHR in order of match");
```
clach: This needs to be updated to be non-Vulkan-specific.

pxr/imaging/hgiPresent/interopHandle.h
```diff
@@ -0,0 +1,66 @@
...
+
+    bool operator!=(const HgiNullInteropHandle& other) const
+    {
+        return !(*this == other);
+    }
+};
+
+#if defined(PXR_GL_SUPPORT_ENABLED)
+/// Interop to OpenGL
+struct HgiGLInteropHandle
```
clach: Needs different name, this looks like an `HgiGL` struct. Should match whatever we come to after resolving the `HgiPresentInteropGL` name.

pxr/imaging/hgiInterop/hgiInterop.cpp
DDoS: I wanted to do this, but wasn't sure where to put the struct. If I put it in HgiPresent, then there's a circular dependency. If I put it in HgiInterop then I need to include an HgiInterop header in HgiPresent. Not as bad as a circular dependency, but I think it leaks HgiInterop into HgiPresent too much.

pxr/imaging/hgiPresent/present.h
DDoS: I originally did this, but the names got very long, so I shortened to just Hgi. Could we shorten it to "HgiPr" or something like that?

pxr/imaging/hgiPresent/metal.mm
DDoS: I'm not sure, this feels very specific to presentation. I'm not sure it would ever get any other caller than HgiPresent Metal.

pxr/imaging/hgiPresent/interopGL.h
DDoS: Would something like `HgiPresentGLInterop` make more sense?

DDoS:
> We talked about this during the meeting as a potential followup, but it'd be great to add a test executing the window present codepath.

I could try to implement a basic test for X11 specifically, as a proof-of-concept, and add other platform tests in another PR.

pxr/usdImaging/usdImagingGL/engine.h
DDoS: How do I do that?

pxr/imaging/hdx/taskController.cpp
meshula: I expect there'll be the flexibility to pick a rendering space as well ultimately, since it's true that different groups have different needs. Thanks for the thoughts, definitely OOS.

pxr/imaging/hgi/types.cpp
meshula: Since you called it out in the comment, it seemed worth asking. I do agree that not spilling out into more files is preferable.

pxr/imaging/hgiInterop/hgiInterop.h
meshula: Gotcha. @clach commented on having them in a struct, and if the struct defaults to the existing case, then I'm happy with that.

pxr/imaging/hgiPresent/vulkan.cpp
meshula: This comment followed on the previous one, in the case that we are intending to do color space conversions. As discussed there, currently the target is just versus what the system currently does in linear rec709, so the question of supporting conversions here is out of scope :)

pxr/imaging/hgiInterop/hgiInterop.h
DDoS: I did add a struct, but I set the defaults as to fully overwrite instead of compositing. I think the default should be the "null" state, and "null composition" should be no composition. Least surprising behaviour.

pxr/usdImaging/usdImagingGL/engine.h
clach: See `usdImagingGL/version.h`; you'd just bump the number and add a comment.

pxr/imaging/hgiPresent/interopGL.h
clach: I think so.

pxr/imaging/hgiInterop/hgiInterop.h
meshula: Least surprising is the best :D

pxr/imaging/hgiPresent/vulkan.cpp
meshula: background notes which I'm recording here simply because I looked it up to double check my understanding and for anyone else who happens on this conversation and is wondering about it: the OETF in "Broadcast Television" recommendation 709 refers to a calibrated conversion of levels of light to numbers of HDTV video cameras available when the spec was written in 1991, and that the conversion of numbers to light was BT1886, which mapped onto SDR CRTs in tv's of the time.  They aren't symmetrical for that reason. 

Looking at this https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#TRANSFER_CONVERSION I can see why chose your approximation. VK_COLOR_SPACE_BT709_NONLINEAR_EXT does seem to be meant to encode camera values, perhaps out of some sense of completeness, but it isn't something we should ever encounter inside Hgi. 

In some pedantic sense I agree that the switch statement in HgiPresent should handle all the enumerations, but from a systems perspective, I think we should never actually get here, and there in principle shouldn't be any way to contrive that a need for BT709 ever wends its way through the system to hit the switch!

DDoS: Any changes needed?

meshula: No. If this enum is ever really invoked for real in the future, we should just implement the official function, but I can't think of any reason to do so without a motivating scenario.

dgovil:
Hi @erikaharrison-adsk ,
This is a really interesting PR.
Would you be able to provide a description of what HgiPresent is and aims to achieve? 

I believe I understand it from the submitted code, but having a high level overview would help to loop in some other reviewers as well.

DDoS:
> Hi @erikaharrison-adsk , This is a really interesting PR. Would you be able to provide a description of what HgiPresent is and aims to achieve?
> 
> I believe I understand it from the submitted code, but having a high level overview would help to loop in some other reviewers as well.

Hello @dgovil, the goal of this PR is to add native window presentation to `HdxPresentTask`, and a new Hgi level API to abstract over presentation. In it's most basic form, a `UsdImagingGLEngine` user can call ` engine->EnableWindowPresentation(window)`, where `window` is a handle to a window created and managed by the user, and expect that ` engine->Render()` will present the final color AOV to the window without any additional intervention.

dgovil:
Thanks!

PR #3545, opened 2025-02-25
Fix ArchGetFileName on windows for file paths with special characters

mnaehrig:
### Description of Change(s)

ArchGetFileName called std::filesystem::canonical() with an utf8-encoded std::string, but on windows std::filesystem::path expects std::string with system encoding. To fix this issue the std::filesystem::path is now created from the utf16-encoded string, which is returned by GetFinalPathNameByHandleW().

### Checklist

### DIFF
--- a/pxr/base/arch/testenv/testFileSystem.cpp
+++ b/pxr/base/arch/testenv/testFileSystem.cpp
@@ -96,7 +96,12 @@ int main()
     // Open a file, check that the file path from FILE* handle is matched.
     ARCH_AXIOM((firstFile = ArchOpenFile(firstName.c_str(), "rb")) != NULL);
     std::string filePath = ArchGetFileName(firstFile);
+#if defined(ARCH_OS_WINDOWS)
+    ARCH_AXIOM(std::filesystem::equivalent(ArchWindowsUtf8ToUtf16(filePath),
+                   ArchWindowsUtf8ToUtf16(firstName)));
+#else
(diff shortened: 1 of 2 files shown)


nvmkuruc:
Do you have a test case that shows the failure state?

Simplifying things the function to use the existing `Arch` converter seems like a good change.

mnaehrig:
You just need a usd file with a non-ASCII character in it's name, e.g. "würfel.usd". If you try to open it in usdview on windows, std::filesystem::canonical raises an exception.

jesschimein:
Filed as internal issue #USD-10702

pxr/base/arch/testenv/testFileSystem.cpp
```diff
@@ -96,7 +96,12 @@ int main()
     // Open a file, check that the file path from FILE* handle is matched.
     ARCH_AXIOM((firstFile = ArchOpenFile(firstName.c_str(), "rb")) != NULL);
     std::string filePath = ArchGetFileName(firstFile);
+#if defined(ARCH_OS_WINDOWS)
```
nvmkuruc, line 99: I'm curious if there a solution that could use `u8path` to avoid forking the test behavior around the platform?

mnaehrig: Forking could be avoided using `u8path`. I didn't use it because it's deprecated in C++20 and may be removed with future versions (C++26 or later). But if this is no problem I would change it.

PR #3546, opened 2025-02-25
Fix file paths with non-ASCII characters on windows

mnaehrig:
### Description of Change(s)
It was not possible to run an OpenUSD application from a directory with non-ASCII characters in it's name on windows. This is because LoadLibrary and std::ifstream were used with UTF-8-encoded file paths, while only system encoding and UTF-16 are supported. File paths are converted to UTF-16 now.

### Checklist


jesschimein:
Filed as internal issue #USD-10703

pxr/imaging/hio/glslfx.cpp
```diff
@@ -232,7 +232,11 @@ static unique_ptr<istream>
 _CreateStreamForFile(string const& filePath)
 {
     if (TfIsFile(filePath)) {
+#if defined(ARCH_OS_WINDOWS)
+        return make_unique<ifstream>(ArchWindowsUtf8ToUtf16(filePath));
+#else
```
spiffmon: We prefer to keep architecture-related ifdefs internal to arch.  Could you add something like `std::string ArchGetOSEncodedPath(const std::string &filePath)` in `arch/fileSystem.cpp` and then use that in both affected codesites?

mnaehrig: As it's not recommended to use std::string with ANSI encoding any more, I would suggest to add two versions of `ArchGetOSEncodedPath` to `arch/fileSystem.h` with different return values. There is already an architecture-related ifdef-section at the bottom, which could be extended like this:


```cpp
#if defined(ARCH_OS_WINDOWS)

...

/// Converts a regular UTF-8 std::string to a UTF-16 windows string,
/// as UTF-16 is the native encoding for filenames on windows. Windows filesystem
/// APIs do not support UTF-8, but would expect ANSI encoded std::string instead.
/// As ANSI supports only a limited set of characters, it's strongly recommended
/// to use UTF-16 instead, as we do here.
inline std::wstring ArchGetOSEncodedPath(const std::string& filePath)
{
    return ArchWindowsUtf8ToUtf16(filePath);
}

#else

/// Returns given file path with encoding for POSIX systems.
/// As UTF-8 is the de facto standard on POSIX systems we don't have to change the encoding.
inline const std::string& ArchGetOSEncodedPath(const std::string& filePath)
{
    return filePath;
}

#endif
```

What do you think about this?

spitzak: I would not use the same function name but return different types.

Windows can be made to use UTF-8 for the filename encoding. Recommend this be used instead and no 16-bit characters used anywhere in USD.

mnaehrig: I agree that the proposed solution isn’t particularly elegant. But it’s the best I’ve found so far to support Windows applications with arbitrary ANSI code pages. In that case I don't see a way around UTF-16 to cover all possible scenarios.

If it is acceptable that developers have to ensure that their applications run with UTF-8 code page, the proposed changes would be obsolete. But UTF-8 is not the default on Windows. So there is a constant risk, that developers are not aware of this.

For applications with a binary executable, UTF-8 could just be enabled via a manifest file. Since this would also work for our use case, I would be fine with this. However, this would not work for Python applications like _usdview_. For Python, it would be necessary to enable UTF-8 for the entire system. There is a Windows setting in beta which allows this, but since it may have unexpected side effects on legacy applications, I don't see this as a viable solution for most users.

nvmkuruc: Have you considered using `std::filesystem::path`?  It seems like you could use `u8path` to avoid introducing new `Arch` functionality.

```
return make_unique<ifstream>(std::filesystem::u8path(filePath));
```

mnaehrig: `u8path` would solve this issue. I just didn't use it as it is deprecated since C++20.

nvmkuruc: Once `u8path` is deprecated, I believe you could use the `std::u8string`.

```
return make_unique<ifstream>(std::filesystem::path(std::u8string(std::cbegin(filePath), std::cend(filePath))));
```

mnaehrig: Ok, I'm convinced of u8path now. Also there seems to be no schedule for removal of u8path yet. So if there is no veto I would change it this way.

pxr/base/arch/library.cpp
```diff
@@ -27,7 +28,7 @@ void* ArchLibraryOpen(const std::string &filename, int flag)
 {
 #if defined(ARCH_OS_WINDOWS)
     arch_lastLibraryError = 0;
-    if (void* result = LoadLibrary(filename.c_str())) {
+    if (void* result = LoadLibraryW(ArchWindowsUtf8ToUtf16(filename).c_str())) {
```
spiffmon: Any reason not to use `u8path` here, too?  Thinking ahead to one day deprecating the Windows-only 8/16 conversion functions.

mnaehrig: Didn't use it there as code was architecture-related anyway. But you are right, u8path is better choice if conversion functions should be deprecated some day. u8path is used now.

PR #3549, opened 2025-02-26
Autodesk: Fix vtarray unecessary or hidden copies

erikaharrison-adsk:
### Description of Change(s)

The changes contained in this PR were done after checking VtArrays unintended copies with the environment variable `VT_LOG_STACK_ON_ARRAY_DETACH_COPY` on several scene including Caldera from Activision and the Pixar usdskel examples.
It also introduces a new function `GetJointInverseWorldBindTransforms` to benefit from the caching done in the skel definition.

This should reduce the number of allocations with scenes leveraging usdskel and instances.

- N/A

### Fixes Issue(s)

- N/A

### Checklist


pxr/imaging/hd/sceneIndexAdapterSceneDelegate.cpp
```diff
@@ -2577,7 +2577,7 @@ HdSceneIndexAdapterSceneDelegate::GetInstancerId(SdfPath const &id)
         }
 
         if (instancerIds.size() > 0) {
-            instancerId = instancerIds[0];
+            instancerId = instancerIds.AsConst()[0];
```
nvmkuruc: You could consider `cfront` as well.

https://openusd.org/dev/api/group___s_t_l___a_p_i.html#gaeea74d6d743836c57fef3d0d4e6c982b

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
```diff
@@ -2665,7 +2668,7 @@ UsdSkelImagingSkeletonAdapter::_SkelData::ComputePoints(
 
         if(TF_VERIFY(_boneMeshPoints.size() == _boneMeshJointIndices.size())) {
 
-            VtVec3fArray skinnedPoints(_boneMeshPoints);
```
nvmkuruc, line 2625: I'm surprised this is triggering a detach.  The copy constructor should be using a `const VtArray&`.

https://github.com/PixarAnimationStudios/OpenUSD/blob/62847554c9abce257729f9f6c7071165ccd680a4/pxr/base/vt/array.h#L281

pxr/usd/usdGeom/xformable.cpp
```diff
@@ -931,15 +931,15 @@ UsdGeomXformable::GetLocalTransformation(
     if (opOrderVec.size() == 0) {
         return true;
     }
-    
-    for (VtTokenArray::reverse_iterator it = opOrderVec.rbegin() ; 
-         it != opOrderVec.rend(); ++it) {
+    const VtTokenArray opOrderVecConst = opOrderVec.AsConst();
+    for (VtTokenArray::const_reverse_iterator it = opOrderVecConst.rbegin() ;
```
nvmkuruc: If you're open to the suggestion, `crbegin` and `crend` would be less invasive changes.

pxr/usd/usdGeom/pointInstancer.cpp
```diff
@@ -1155,7 +1155,7 @@ UsdGeomPointInstancer::_ComputeExtentAtTimePreamble(
     }
 
     // verify that all the protoIndices are in bounds.
-    TF_FOR_ALL(iter, *protoIndices) {
```
nvmkuruc, line 1158: Might be a good opportunity to transition to a ranged base for loop.

```
for (const int protoIndex : protoIndices->AsConst()){ ... }
```

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
cpichard: I think it's triggered later when `skinnedPoints` is used, not when it's constructed.

pxr/usd/usdGeom/pointInstancer.cpp
cpichard: is TF_FOR_ALL going away ?

nvmkuruc: I don't know if it's going away, but there are several open issues / pull requests that are focused on replacing it.

https://github.com/PixarAnimationStudios/OpenUSD/issues/80

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
nvmkuruc: So you're effectively moving the detach that would've happened here?

```
GfVec3f* points = skinnedPoints.data();
```

jesschimein:
Filed as internal issue #USD-10707

pxr/usd/usdGeom/pointInstancer.cpp
tcauchois: TF_FOR_ALL predates range-based for, and IIRC is slower to compile as well as slower at runtime.  Not such a big deal that we're hunting them down, but we try to switch them over when we can.  So +1 to the suggestion here.

pxr/usd/usdGeom/xformable.cpp
```diff
@@ -746,9 +746,9 @@ UsdGeomXformable::TransformMightBeTimeVarying() const
     if (opOrderVec.size() == 0) {
         return false;
     }
-
-    for (VtTokenArray::reverse_iterator it = opOrderVec.rbegin() ; 
-         it != opOrderVec.rend(); ++it) {
+    const VtTokenArray &opOrderVecConst = opOrderVec.AsConst();
+    for (VtTokenArray::const_reverse_iterator it = opOrderVecConst.rbegin() ;
```
gitamohr: I think you can just use `crbegin()` and `crend()` here.

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
```diff
@@ -908,8 +909,9 @@ UsdSkelImagingSkeletonAdapter::InvokeComputation(
         return;
     }
 
-    VtVec3fArray skinnedPoints = 
-        restPoints.UncheckedGet<VtVec3fArray>();
+    // Explicit copy of the skinnedPoints to avoid VT_LOG_STACK_ON_ARRAY_DETACH_COPY logs
+    VtVec3fArray skinnedPoints(restPoints.UncheckedGet<VtVec3fArray>().begin(),
```
gitamohr: Hmm.  I wonder if we should have an explicit way to state this in `VtArray`.  Like `MakeUnique` or something.  I would hate for this pattern to proliferate, since if the data _is_ uniquely owned already this is a pessimization.  I'm not suggesting you do this as part of this change -- just mentioning it for the sake of discussion.

pxr/usd/usdGeom/xformable.cpp
gitamohr: Ah, I see Matt commented the same on a subsequent instance, sorry for the repetition.

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
gitamohr: What do you think about removing the explicit copies, but leaving comments like `// XXX: MakeUnique()`?  I'll file an internal task for us to add that and replace the comments with calls to it.

nvmkuruc: Would `MakeUnique` still log as a detach?

spiffmon: No it would not.  The name of the env setting is slightly off (but VT_LOG_STACK_ON_ARRAY_POSSIBLY_UNINTENDED_DETACH_COPY is really long) but is documented as being to help find unintended detaches, and the point of MakeUnique() is that it is an intentional copy (only if needed.

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
cpichard: Exactly.

pxr/usdImaging/usdSkelImaging/skeletonAdapter.cpp
cpichard: Hi @gitamohr,  good idea ! I had the same reasoning as you, if there is only one copy then this change would create another, which is not ideal. So I looked at the potential code paths and couldn't find one where there would be only one copy. Adding a MakeUnique function is definitely the way to go as it's more explicit and avoid potential duplication.

gitamohr: Cool, thanks @cprichard!

pxr/usdImaging/usdSkelImaging/extComputations.cpp
```diff
@@ -141,8 +141,9 @@ UsdSkelImagingInvokeExtComputation(
         return;
     }
 
-    VtVec3fArray skinnedPoints =
-        restPointsValue.UncheckedGet<VtVec3fArray>();
+    // Explicit copy of the skinnedPoints to avoid VT_LOG_STACK_ON_ARRAY_DETACH_COPY logs
+    VtVec3fArray skinnedPoints(restPointsValue.UncheckedGet<VtVec3fArray>().begin(), // TODO: MakeUnique()
```
gitamohr: Thanks for adding the comment -- sorry if I wasn't clear in the earlier comments, but I'd like for us _not_ to add explicit copies, since if the data is already unique, this is a pessimization.

pxr/usd/usdSkel/skeletonQuery.cpp
```diff
@@ -356,7 +356,10 @@ UsdSkelSkeletonQuery::_ComputeSkinningTransforms(VtArray<Matrix4>* xforms,
 
         if (xforms->size() == inverseBindXforms.size()) {
             // xforms = inverseBindXforms * xforms
-            _MultTransforms<Matrix4>(inverseBindXforms, *xforms, *xforms);
+            // Explicit copy to avoid VT_LOG_STACK_ON_ARRAY_DETACH_COPY log
+            VtArray<Matrix4> xformsTmp(xforms->cbegin(), xforms->cend());
+            _MultTransforms<Matrix4>(inverseBindXforms, xformsTmp.AsConst(), xformsTmp);
```
gitamohr: Could you please add the `TODO: MakeUnique()` comment here too, and remove the explicit copy?

spiffmon: @cpichard , we don't want to introduce any - even temporary - regression here.  That's why @gitamohr asked you to **also** remove the new code you added, leaving only the TODO comment - looks like you did so below, but not here?

pxr/usdImaging/usdSkelImaging/extComputations.cpp
cpichard: Sorry @gitamohr , I didn't see this very comment. I'll do that.

pxr/usd/usdSkel/skeletonQuery.cpp
cpichard: Sure, I'll remove the code I added. I didn't see @gitamohr previous comment last week. The code below is just copying the temporary result into xforms, it's not what @gitamohr is asking.

pxr/usd/usdGeom/pointInstancer.cpp
Cewein: I am currently working on removing TF_FOR_ALL. There is a conversation on slack about it, ping me if you have better alternative that what I did. Open to any feedback.

cpichard:
Hi, 
I think I have addressed all the comments, I am not sure, but is there anything left for me to do to get this PR through ? 
Thanks,
Cyril

PR #3551, opened 2025-02-27
APPLE: Remove multiple instances of ZLIB

dgovil:
### Description of Change(s)

Due to a prior change in OpenUSD, ZLIB was moved to a conditional dependency, and then removed from the list if run on Linux or macOS

Unfortunately, the removal only removes the first found instance of ZLib, so doesn't remove it if multiple dependencies ask for it.

A future clang version also has issues with the ZLib version used here so this would be needed to pass builds soon. However, since clang is used primarily on Linux and Mac, it is preferred to just use the OS provided version of Zlib.

### Checklist

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2514,7 +2514,7 @@ if context.buildAnimXTests:
 # itself does not require it. The --no-zlib flag can be passed to the build
 # script to allow the dependency to find zlib in the build environment.
 if (Linux() or not context.buildZlib) and ZLIB in requiredDependencies:
-    requiredDependencies.remove(ZLIB)
+    requiredDependencies = [r for r in requiredDependencies if r != ZLIB]
 
 # Error out if user is building monolithic library on windows with draco plugin
 # enabled. This currently results in missing symbols.


build_scripts/build_usd.py
```diff
@@ -2514,7 +2514,7 @@ def ForceBuildDependency(self, dep):
 # itself does not require it. The --no-zlib flag can be passed to the build
 # script to allow the dependency to find zlib in the build environment.
 if (Linux() or not context.buildZlib) and ZLIB in requiredDependencies:
-    requiredDependencies.remove(ZLIB)
```
nvmkuruc, line 2517: Do you think `requiredDependencies` could be a `set` to avoid this issue?

sunyab: I believe we rely on `requiredDependencies` being ordered to ensure we build dependencies before their dependents.

dgovil: Unfortunately, not. It needs to preserve ordering and afaik that's still not guaranteed for a set in Python.
Edit: replied at the same time as Sunya

nvmkuruc: Makes sense.

jesschimein:
Filed as internal issue #USD-10713

Issue #1826, opened 2022-04-05
Performance issues with large amounts of materials


sirpalee:
### Description of Issue
When a stage has a large amount of materials (50k+), Hydra spends a significant amount of time in `Hd_PrimTypeIndex<HdSprim>::SyncPrims`. In the example file (with 250k cubes and materials), the function takes 30.39 ms, while the whole render call 31.737 ms. See attached image from the profiler:

![SyncPrims_perf](https://user-images.githubusercontent.com/582376/161678926-a449e264-9a33-45ec-8b91-9e0550922dd6.png)

### Steps to Reproduce
1. Open usd from [cubes_with_preview_250k.zip](https://github.com/PixarAnimationStudios/USD/files/8414940/cubes_with_preview_250k.zip) in usdview.
2. Make meshes invisible.
3. Move the camera around and observe render time.

### System Information (OS, Hardware)
- Windows 10
- AMD Threadripper 3970X & NVIDIA RTX 3090

### Package Versions
- Latest build of dev branch.

### Build Flags
- Default flags using `build_usd.py` and `--build-variant` set to `relwithdebuginfo`.

jilliene:
Filed as internal issue #USD-7313

tcauchois:
Just to provide some context, there are a few known issues with material Sync():

1. Currently, it's single-threaded; it's straightforward to thread, but some render delegates have thread hazards so it's tricky to roll out...
2. Unlike geometry, we loop through all materials checking if they're invalid; we skip the work of updating them if they're up-to-date, but at scale (as you can see) this still causes issues.

Both of these will be addressed in planned work.  It looks like most of the time in your trace is taken up by GetSprimDirtyBits, which is a lookup of prim path -> u32 invalidation flags, and there's not much to address there other than #1 and #2 above.

All of this said, 50k GLSL or OSL compiles seems like a nightmare as well.  Can I ask what's driving the shader variation?  If the materials are duplicates, it's worth deduplicating them in the scene, regardless of how we optimize hydra.  If you want to pass different parameters in per geometry, I'd recommend doing that with primvars on the geometry.  We've got some current limitations around texture bindings per-geometry, and around taking full advantage of USD instances of materials; if one of these is the issue, please let us know!

spiffmon:
[...]

PR #3567, opened 2025-03-13
Autodesk: [hdSt] Early parallel MaterialX codegen

erikaharrison-adsk:
### Description of Change(s)

This is an optimization in Storm of the MaterialX codegen overhead encountered when adding new materials to the stage, most importantly when opening a stage with numerous materials. The feature is controlled by env var `HDST_ENABLE_PARALLEL_MTLX_CODEGEN`, set to true by default.

When the optimization is on, creating a material Sprim starts "syncing" the prim early: converts the USD material network to the Hd counterpart, finds the material tag and the topological material hash and then launches a parallel TBB task with this input data, which performs MaterialX codegen. The parallel tasks are identified by the material hash, and managed by a cache, which prevents duplicate codegen tasks being launched.

The data computed before launching the parallel task includes things like parameters which don't affect the shader graph topology, so it's unique per Sprim, not per parallel codegen task. For this reason, it's cached per Sprim until its sync is fully completed, in a heap-allocated "filter task". In the original design, this data lived on the stack in the Sync() call tree.

Finally, the results of the codegen tasks are used in the regular Sync() calls to finish the sync. For synchronization, we wait for all of the parallel codegen tasks to complete before syncing all of the material sprims.

When the optimization is disabled (the feature switch is off or the material is not a MaterialX material), the sync process follows the the original code path, except that a filter task is allocated on the stack to hold the intermediate data, in order to share code between the two code paths.

- N/A

### Fixes Issue(s)

- N/A

### Checklist


pxr/imaging/hd/material.h
```diff
@@ -26,9 +26,10 @@ class HdMaterial : public HdSprim
     enum DirtyBits : HdDirtyBits {
         Clean                 = 0,
         // XXX: Got to skip varying and force sync bits for now
+        DirtyInit             = 1 << 1, // Sprim created, but never synced
```
ppenenko, line 29: We need to differentiate between materials which are dirty because they've just been created but never synced, and those that were dirtied by editing. When we use the optimization, we cache the state as of the material's creation, so if it's edited after the creation and before `Sync()`, that state will become stale. This was caught by animated material unit tests. I'll comment below how we handle the stale state.

pxr/imaging/hd/primTypeIndex.cpp
```diff
@@ -96,7 +96,8 @@ Hd_PrimTypeIndex<PrimType>::InsertPrim(const TfToken    &typeId,
     }
 
 
-    PrimType *prim = _RenderDelegateCreatePrim(renderDelegate, typeId, primId);
+    PrimType *prim = _RenderDelegateCreatePrim(
+        sceneDelegate, renderDelegate, typeId, primId);
```
ppenenko, line 100: Need to pass the scene delegate here - more on the motivation below.

pxr/imaging/hd/primTypeIndex.cpp
```diff
@@ -383,10 +384,15 @@ Hd_PrimTypeIndex<PrimType>::SyncPrims(HdChangeTracker  &tracker,
 
     _dirtyPrimDelegates.clear();
     HdSceneDelegate *prevDelegate = nullptr;
+
     for (size_t typeIdx = 0; typeIdx < numTypes; ++typeIdx) {
 
+        TfToken const& primType = _primTypeNames[typeIdx];
+
+        renderDelegate->WaitForEarlyInitTasks(primType);
```
ppenenko, line 392: New render delegate API: optionally wait for parallel tasks associated with the particular prim type.

pxr/imaging/hd/primTypeIndex.cpp
```diff
@@ -493,11 +499,14 @@ Hd_PrimTypeIndex<HdSprim>::_TrackerMarkPrimClean(
...
 HdSprim *
-Hd_PrimTypeIndex<HdSprim>::_RenderDelegateCreatePrim(HdRenderDelegate *renderDelegate,
+Hd_PrimTypeIndex<HdSprim>::_RenderDelegateCreatePrim(HdSceneDelegate *sceneDelegate, 
+                                                     HdRenderDelegate *renderDelegate,
                                                      const TfToken &typeId,
                                                      const SdfPath &primId)
 {
-    return renderDelegate->CreateSprim(typeId, primId);
+    HdSprim *const sprim = renderDelegate->CreateSprim(typeId, primId);
+    renderDelegate->StartEarlySprimInit(sceneDelegate, sprim);
```
ppenenko, line 508: The hook for starting the early sync.

pxr/imaging/hdSt/material.cpp
```diff
@@ -36,6 +40,11 @@ TF_DEFINE_PRIVATE_TOKENS(
     (opacity)
 );
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+TF_DEFINE_ENV_SETTING(HDST_ENABLE_PARALLEL_MTLX_CODEGEN, true,
```
ppenenko, line 44: The feature switch.

pxr/imaging/hdSt/material.cpp
```diff
@@ -167,6 +176,72 @@ HdStMaterial::_ProcessTextureDescriptors(
                                        doublesSupported);
 }
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+void
+HdStMaterial::StartEarlyInit(
+    HdSceneDelegate *sceneDelegate,
+    HdSt_MaterialXSyncDispatcher *syncDispatcher)
```
ppenenko, line 183: The `syncDispatcher` is the global cache of parallel tasks, deduplicating them by hash and implementing waiting for them.

pxr/imaging/hdSt/material.cpp
```diff
@@ -167,6 +176,72 @@ HdStMaterial::_ProcessTextureDescriptors(
...
+{
+    // The task should be created by this method just once
+    TF_VERIFY(!_filterTask);
+    _filterTask.reset();
```
ppenenko, line 187: The state that's cached until the sync completes.

pxr/imaging/hdSt/material.cpp
```diff
@@ -167,6 +176,72 @@ HdStMaterial::_ProcessTextureDescriptors(
...
+
+    if (!TfGetEnvSetting(HDST_ENABLE_PARALLEL_MTLX_CODEGEN)) {
```
ppenenko, line 189: Checking the feature switch.

pxr/imaging/hdSt/material.cpp
```diff
@@ -167,6 +176,72 @@ HdStMaterial::_ProcessTextureDescriptors(
...
+
+    HdMaterialNetworkMap const& hdNetworkMap =
+        vtMat.UncheckedGet<HdMaterialNetworkMap>();
+
+    if (hdNetworkMap.terminals.empty() || hdNetworkMap.map.empty()) {
+        return;
+    }
+
+    bool isVolume = false;
+    auto filterTask = std::make_unique<HdSt_MaterialFilterTask>();
```
ppenenko, line 208: Creating the filter task locally. If some checks fail below, then we don't move it to the Sprim's member and destroy the local.

pxr/imaging/hdSt/material.cpp
```diff
@@ -180,15 +255,20 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
...
         return;
     }
 
-    bool needsRprimMaterialStateUpdate = false;
-    bool markBatchesDirty = false;
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+    if (maskedBits != DirtyInit) {
+        // This means that the material has been modified since creation,
+        // so the state of the filter task is stale
+        _filterTask.reset();
```
ppenenko, line 269: This is where we handle cases when the material has changed between creation and the first sync. We'll create a new filter task later, but we can still use the shader produced by parallel codegen, if the changes in parameters haven't changed the shader hash.

pxr/imaging/hdSt/material.cpp
```diff
@@ -200,21 +280,45 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
...
-            params = _networkProcessor.GetMaterialParams();
-                textureDescriptors = _networkProcessor.GetTextureDescriptors();
+            // Initialize the material network
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+            if (_filterTask) {
+                // We have a filter task created by early parallel
+                // initialization
+                constexpr bool isVolume = false; // We only use filter tasks
+                                                 // for non-volume materials
+                _hdStMaterialNetwork.ProcessFilterTask(
```
ppenenko, line 295: Renamed `_networkProcessor`  to `_hdStMaterialNetwork` . "Processor" sounds confusing here IMO because the object `_hdStMaterialNetwork` is simply initialized with those `Process` methods from `hdNetworkMap`, it doesn't process some external objects.

pxr/imaging/hdSt/material.cpp
```diff
@@ -225,6 +329,8 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
         materialMetadata = _fallbackGlslfx->GetMetadata();
     }
 
+    bool markBatchesDirty = false;
```
ppenenko, line 332: Just narrowed its scope.

pxr/imaging/hdSt/material.cpp
```diff
@@ -259,6 +365,8 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
 
     bool hasDisplacement = !(displacementSource.empty());
 
+    bool needsRprimMaterialStateUpdate = false;
```
ppenenko, line 368: Just narrowed its scope.

pxr/imaging/hdSt/material.cpp
```diff
@@ -383,7 +491,7 @@ HdStMaterial::_GetHasLimitSurfaceEvaluation(VtDictionary const & metadata) const
 HdDirtyBits
 HdStMaterial::GetInitialDirtyBitsMask() const
 {
-    return AllDirty;
+    return DirtyInit;
```
ppenenko, line 494: Change in the initial dirty flags.

pxr/imaging/hdSt/material.h
```diff
@@ -123,7 +131,11 @@ class HdStMaterial final: public HdMaterial
     TfToken _materialTag;
     size_t _textureHash;
 
-    HdStMaterialNetwork _networkProcessor;
+    HdStMaterialNetwork _hdStMaterialNetwork;
+
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+    std::unique_ptr<HdSt_MaterialFilterTask> _filterTask;
```
ppenenko, line 137: The in-flight sync state cached on the prim.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -159,8 +159,8 @@ _GetGlslfxForTerminal(
     }
 }
 
-static HdMaterialNode2 const*
-_GetTerminalNode(
+HdMaterialNode2 const*
+HdSt_GetTerminalNode(
```
ppenenko, line 163: Now need to call it from outside this source file.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -1097,6 +1097,43 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
...
+        HdSt_GetTerminalNode(
+            filterTask.hdNetwork,
+            terminalName,
+            &filterTask.terminalNodePath);
+
+    if (!filterTask.terminalNode) {
+        return;
+    }
+
+    ProcessFilterTask(
```
ppenenko, line 1127: `ProcessMaterialNetwork()` is implemented via `ProcessFilterTask()`

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -399,10 +396,8 @@ _AddDefaultMtlxTextureValues(
     }
 }
 
-static void
-_AddFallbackDomeLightTextureNode(
-    HdMaterialNetwork2* hdNetwork,
-    SdfPath const& hdTerminalNodePath)
+void
+HdSt_MaterialFilterTask::AddFallbackDomeLightTextureNode()
```
ppenenko, line 400: This and a few other functions are now methods of `HdSt_MaterialFilterTask`.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -412,14 +407,14 @@ _AddFallbackDomeLightTextureNode(
             HdStPackageFallbackDomeLightTexture(), 
             HdStPackageFallbackDomeLightTexture()));
     const SdfPath domeTexturePath = 
-        hdTerminalNodePath.ReplaceName(_tokens->domeLightFallback);
-    hdNetwork->nodes.insert({domeTexturePath, hdDomeTextureNode});
+        terminalNodePath.ReplaceName(_tokens->domeLightFallback);
```
ppenenko, line 410: This and a few other related pieces of data are now encapsulated as members of `HdSt_MaterialFilterTask`.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1055,37 +1052,30 @@ _ReplaceFilenameInput(
     hdNetwork->nodes[hdTerminalNodePath].
         inputConnections[terminalTextureConn.upstreamOutputName] =
             {terminalTextureConn};
-
 }
 
-// Gather the Material Params from the glslfx ShaderPtr
```
ppenenko, line 1061: Not technically a GLSLFX shader yet - it's just generated by MaterialX and then later wrapped in GLSLFX.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1204,68 +1194,89 @@ _AddMaterialXParams(
...
+                    &hdNetwork, terminalNodePath, variable->getVariable());
             }
         }
     }
 }
 
-static mx::ShaderPtr
-_GenerateMaterialXShader(
-    HdMaterialNetwork2 const& hdNetwork,
+HdSt_MaterialXGeneratorTask::HdSt_MaterialXGeneratorTask(
```
ppenenko, line 1221: This is a different kind of "task" - the necessary data passed to the lambda executing in parallel in TBB and wrapping the MaterialX codegen.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1290,27 +1301,26 @@ _IsTopologicalShader(TfToken const& nodeId)
...
 }
 
-// Build the topoNetwork, equivalent to the given hdNetwork but anonymized and 
-// stripped of non-topological parameters to better re-use the generated shader.  
-size_t _BuildEquivalentMaterialNetwork(
-    HdMaterialNetwork2 const& hdNetwork,
-    HdMaterialNetwork2* topoNetwork,
-    HdAnnonNodePathMap* annonNodePathMap)
+size_t
+HdSt_MaterialFilterTask::BuildAnonymizedMaterialNetwork(
```
ppenenko, line 1305: Unified the terminology around "anonymized". The input parameters are replaced by `HdSt_MaterialFilterTask`'s members.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1426,96 +1438,109 @@ size_t _BuildEquivalentMaterialNetwork(
...
-    // Use the Resource Registry to cache the generated MaterialX 
-    // glslfx Shader
-    Tf_HashState shaderHash;
-    TfHashAppend(shaderHash, topoHash);
-    TfHashAppend(shaderHash, materialTag);
-    HdInstance<mx::ShaderPtr> glslfxInstance =
-        resourceRegistry->RegisterMaterialXShader(shaderHash.GetCode());
-
-    if (glslfxInstance.IsFirstInstance()) {
+    HdSt_MaterialXGeneratorTask generatorTask(
```
ppenenko, line 1467: Creating the generator task on the stack in the case when parallel codegen is off.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1426,96 +1438,109 @@ size_t _BuildEquivalentMaterialNetwork(
...
+    HD_TRACE_FUNCTION();
+
+	if (!syncDispatcher
+        || !filterTask
+        || !filterTask->IsMaterialX(&SdrRegistry::GetInstance())) {
+        return;
     }
 
-    return glslfxShader;
+    auto generatorTask = std::make_unique<HdSt_MaterialXGeneratorTask>(
```
ppenenko, line 1536: Creating the generator task on the heap for parallel codegen.

pxr/imaging/hdSt/materialXSyncDispatcher.cpp
```diff
@@ -0,0 +1,108 @@
...
+    HdStResourceRegistrySharedPtr& resourceRegistry)
+{
+    HD_TRACE_FUNCTION();
+
+    // Just in case, pass the registry to the parallel task by a weak pointer,
+    // to avoid keeping it alive if something goes wrong with synchronization.
+    std::weak_ptr<HdStResourceRegistry> resourceRegistryWeakPtr
+        = resourceRegistry;
+
+    _dispatcher.Run(
```
ppenenko, line 65: Launching the generator task.

pxr/imaging/hdSt/materialXSyncDispatcher.cpp
```diff
@@ -0,0 +1,108 @@
...
+    );
+}
+
+void
+HdSt_MaterialXSyncDispatcher::Wait()
+{
+	HD_TRACE_FUNCTION();
+
+    _dispatcher.Wait();
+	_generatorTaskSet.clear();
```
ppenenko, line 105: Waiting for all generator tasks.

pxr/imaging/hdSt/materialXSyncDispatcher.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+void
+HdSt_MaterialXSyncDispatcher::AddGeneratorTask(
+    std::unique_ptr<HdSt_MaterialXGeneratorTask> generatorTask,
+    HdStResourceRegistrySharedPtr& resourceRegistry)
+{
+    HD_TRACE_FUNCTION();
+
+
+    if ( resourceRegistry->ContainsMaterialXShader(
```
ppenenko, line 33: First check that the shader hasn't been registered before the early parallel init. This is a new method which, unlike   `RegisterMaterialXShader`, doesn't add a new instance to the registry if an existing one isn't found.

Reusing `RegisterMaterialXShader` in combination with `IsFirstInstance` would be problematic because the intended usage pattern, in the case when there's no existing value, seems to be:
1. Insert a `nullptr` value and return a registry instance pointing to that value and holding a lock to the whole MaterialX registry.
2. The client code is supposed to populate the instance with a non-`nullptr` value while holding the lock. The instance's `IsFirstInstance` method returns `true` if the value is `nullptr`.

So the above pattern can't be used for parallel tasks, when the uniqueness check has to happen before launching the task, the non-`nullptr` value is known only when the task completes and the lock has to be released ASAP.

pxr/imaging/hdSt/resourceRegistry.cpp
```diff
@@ -665,6 +665,15 @@ HdStResourceRegistry::RegisterMaterialXShader(
 {
     return _materialXShaderRegistry.GetInstance(id);
 }
+
+bool
+HdStResourceRegistry::ContainsMaterialXShader(
+        HdInstance<MaterialX::ShaderPtr>::ID id)
+{
+    bool found = false;
```
ppenenko, line 673: The new method just wraps existing functionality.

pxr/imaging/hdSt/materialXSyncDispatcher.cpp
```diff
@@ -0,0 +1,108 @@
...
+        generatorTask->GetShaderHash() )) {
+        // We already have a shader for this topology.
+        return;
+    }
+
+    // Use a separate concurrent container for tracking unique generator tasks.
+    // The generated shader will be registered in the resource registry when
+    // the task completes.
+    auto insertResult =
```
ppenenko, line 42: Another uniqueness check, this time between the different parallel tasks, using a dedicated data structure.

ppenenko:
Addresses issue #1826

jesschimein:
Filed as internal issue #USD-10789

pxr/imaging/hdSt/material.cpp
```diff
@@ -36,6 +40,11 @@ TF_DEFINE_PRIVATE_TOKENS(
     (opacity)
 );
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+TF_DEFINE_ENV_SETTING(HDST_ENABLE_PARALLEL_MTLX_CODEGEN, false,
+    "Enable early parallelized MaterialX codegen");
```
ppenenko, line 45: The feature switch, disabled by default for now, to err on the side of caution. One problem we've noticed with multithreaded MaterialX codegen is that formatting of `float` literals varies. This will soon be fixed in MaterialX: https://github.com/AcademySoftwareFoundation/MaterialX/pull/2454

ppenenko:
Some belated performance figures. The optimization reduces the time to first pixel for https://github.com/usd-wg/assets/tree/main/full_assets/OpenChessSet from 840ms to 550ms in our measurements.
In more complex scenes with hundreds of materials, the improvement is measured in multiple seconds.

jesschimein:
Hi @ppenenko  , thank you for the contribution!  Per the discussion at the ASWF USD/MaterialX sub-working group, could you look at using data sources and scene indices instead of scene and render delegates for this change?  🙏
Also, would it be possible to break the change down into smaller atomic PRs, ideally touching fewer files per PR?  It would really help to simplify a review.  Thank you!

ppenenko:
Hi @jesschimein, thanks for getting back to me!

I certainly understand your concern about this optimization relying on Scene Delegate APIs, because all of us are committed to the long-term Hydra 2.0 direction. However, this change is solving a very concrete problem with Hydra Storm, which has been known for a while (see e.g. #1826). Storm is Scene Delegate-based, and I think it's fair to say that the community is going to rely on Storm for a while.

On a related note, I've taken a look at how `USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX` used together with `HD_ENABLE_SCENE_INDEX_EMULATION` might affect this optimization, and it seems completely transparent with regard to these settings. I.e. the conversion from the material data source to the material network happens upstream, and the expensive MaterialX codegen that's being parallelized happens later. I honestly don't think it would be feasible to move the codegen process earlier in the pipeline, given Storm's assumptions about the lifetimes of prims and the timing of the `Sync` phase.

Lastly, in terms of breaking this up into smaller PRs - I wish it was possible, but as far as I can tell, all of these changes are necessary for the end goal of the optimization, with some minor stylistic improvements and typo fixes around the same lines of code. If you're concerned about the amount of effort it would take to merge after a few months of `dev` diverging from this branch, I would certainly volunteer to merge the latest from `dev` before the you guys can review it.

jesschimein:
Thanks for the careful consideration!  I'll take this back to the team and see what they say.  cc: @klucknav

pxr/imaging/hd/material.h
tcauchois: Kind of subtle... Using bit 1 here isn't a bug, but might aggravate existing bugs, since bit 1 for geometry is special.  Hence the line about skipping the varying bit.  I might use 1 << 4 for this.

pxr/imaging/hd/renderDelegate.h
```diff
@@ -343,6 +343,13 @@ class HdRenderDelegate
     virtual HdSprim *CreateSprim(TfToken const& typeId,
                                  SdfPath const& sprimId) = 0;
 
+    //
+    /// Concrete render delegates can override this method to implement early
+    /// initialization of Sprims that have just been created
+    ///
+    virtual void StartEarlySprimInit(HdSceneDelegate *sceneDelegate,
```
tcauchois, line 350: I like the idea of this in general, but I'm worried about pulling data this early; while USD is pretty random access by the time you get a UsdStage, scene graphs with computed values can be much more particular and this has come up as a challenge in several different app integrations.

I'm also strongly biased against adding new API to HdRenderDelegate if it's possible to avoid, just since it's subclassed by a lot of people with a lot of use cases.

Two alternatives to the new API here and in PrimTypeIndex:

1. Scene index implementation.

Write a scene index filter parameterized on backend (OSL/GLSL/etc).  On PrimsAdded, check if it's a materialX material, and if so kick off codegen.  Override the material so that if you try to pull it, it waits for materialX codegen to finish for that material and produces a new material with just a GLSL node (and primvar/texture nodes as needed).

(You'd want to run this downstream of any scene indices which e.g. prune unused materials).

2. Implement this entirely in Storm

In HdStMaterial::Sync, kick off codegen.  You don't need the results until you're building the draw batch, so in the draw batch when you go to fetch the surface shader code at that point wait for codegen to finish.

... Note, if codegen is async I think this doesn't matter much because hopefully material sync gets faster, but also note that we can turn on parallel prim sync for materials via HdStRenderDelegate::IsParallelSyncEnabled, as long as HdStMaterial::Sync doesn't have any threading hazards.  The main threading hazard in the past was render delegate calls to the change tracker, which could be replaced by new dependency entries in the dependencySceneIndex.

(Happy to explore these ideas further with you on github/aswf slack if they're confusing).

pxr/imaging/hdSt/material.cpp
tcauchois: Not sure why DirtyInit is a useful bit here?  It seems like you're solving a more general problem, of you've already launched parallel codegen but it's out of date, at which point the two solutions are:
1. The nodes changed, so scrap everything and restart.
2. The change was just a parameter edit, so we'd like to keep the codegen output but update the parameter buffer.
... in which case, the interesting thing is not whether this prim has been updated since edit, but whether the edit was meaningful for codegen or not.  That's a thing we're hoping to add, of course, outside of the context of this review.

Is there another reason why DirtyInit is useful?

pxr/imaging/hdSt/material.cpp
tcauchois: AFAIK nobody will be setting DirtyInit except you, but there's plenty of code that might get super confused if materials don't have DirtyResource set on them on load.  I'd leave this as AllDirty?

pxr/imaging/hdSt/materialXSyncDispatcher.cpp
tcauchois: cc @davidgyu

pxr/imaging/hd/renderDelegate.h
ppenenko: Thanks for the review @tcauchois! Let me digest your feedback and get back to you.

pxr/imaging/hd/material.h
ppenenko: Sure, I'll change the bit, I just wanted to use the first available one.

pxr/imaging/hdSt/material.cpp
ppenenko: `DirtyInit`'s purpose is exactly how you describe. We need to be able to tell apart the following three situations:
1. The prim is clean - nothing to do, we should return early from `Sync`
2. The prim was created, never modified since creation, and never synced (only `DirtyInit` is set). In this case we we can use the result of early codegen (if there is one).
3. The prim has been modified since creation (some other dirty bits are set). In this case, we can't be sure that the result of the early codegen task matches the state of the prim.

In the last case, the "filter task" is destroyed and the `Sync` method falls back to the old synchronous code path for codegen and other updates. In practice however, it's likely that it's just a parameter change which doesn't affect codegen. This will not incur significant overhead, because:
1. Codegen is, by far, the most expensive part of the sync.
2. Codegen results are cached in the `HdStResourceRegistry`.
3. Codegen results are identified by a hash which only depends on material properties which affect codegen (see #2330).
4. So, codegen will be skipped.

It would definitely be nice to have a mechanism for determining changes that affect codegen as those changes happen, i.e. reactively. But that seems like a huge change from the current mechanism that involves processing the whole material network at once in `_BuildEquivalentMaterialNetwork`/`BuildAnonymizedMaterialNetwork`, called once per `Sync`. So perhaps more of a long-term goal?

pxr/imaging/hdSt/material.cpp
ppenenko: `AllDirty` here would defeat the purpose of `DirtyInit` as described in https://github.com/PixarAnimationStudios/OpenUSD/pull/3567#discussion_r2274263482
If we're talking about the OpenUSD codebase, I can find about a dozen of explicit uses of `HdMaterial::DirtyResource` and all of them seem to be in setter (`|=`) situations. `HdMaterial::DirtyResource` is never tested individually, and the only test against the material dirty bits seems to be in our `Sync` function, where they're tested against `AllDirty` and `DirtyInit`. Can you point me at some counter-examples?

pxr/imaging/hd/renderDelegate.h
ppenenko: Could you please clarify your concern about pulling the MaterialX data too early? I'm assuming it's not a thread safety issue? Is it about the data being modified (e.g. due to animated materials) after the parallel tasks are launched, so that codegen results potentially get out of date? If so, then I believe the `DirtyInit` bit together with the MaterialX cache solve that problem without a noticeable performance overhead, as described earlier.

I agree that adding new functions to the `HdRenderDelegate` API is not an elegant solution but I hoped it could be a reasonable tradeoff. At the very least, the two new functions have no-op implementations in the base class, so any other render delegates don't have to override them and can remain blissfully unaware of them. Anyways, I'm all for finding a better API.

Before we discuss the alternatives, let me write down some observations that led me to the current implementation.

1. For multithreaded performance, it's best to launch the parallel tasks as early as possible, and to wait for them as late as possible. This way, they can overlap with arbitrary work on other threads, and the duration of waiting should be minimized.
2. MaterialX+Storm codegen is only one part of syncing a material sprim, e.g. [AddMaterialXParams()](https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#diff-c5e0ffdbaf94bdfaffbc73413cb05018d1a3115640d2c53d51ddff9c886be346R1059) and [the rest of the Sync function](https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#r1995723681) are executed after codegen.
3. The smaller and the better encapsulated the task, the lower the threading risks. So, parallelizing just the codegen should be safer than parallelizing the whole `Sync`. Ideally, a codegen task should be self-contained just like a compilation task with a modern compiler like shaderc or DXC. The general TBB advice is also to parallelize at the lowest granularity level possible.
4. At the same time, codegen takes the vast majority of the duration of a `Sync`, which makes it the top priority for parallelization.
5. There's a many-to-one relationship between materials and unique shaders, which is maintained by the topology hashing mechanism and the MaterialX cache. Thus, it's best to create one parallel task per unique shader, not per material sprim. This eliminates duplication of work without the need for synchronization between the tasks, which is also along the lines of good TBB practices.

So with the above points in mind, starting codegen tasks in the `Sync` phase looks suboptimal to me because:
* If the sync is still single-threaded, then waiting for the last sprim's task may take a while.
* If sprim syncs are parallelized, then:
   * It would be harder to guarantee thread-safety of these bigger tasks.
   * If the results are waited on while building draw batches, then we'd need to encapsulate and persist all the necessary state till then. I suspect it would be harder to do than with the filter task in the current implementation.
   * Again, since there are multiple sprims per unique shader, we would need to solve the duplication of work and complicate the synchronization.

On the other hand, the idea of starting the early sync from a scene index sounds very interesting. We could inject an `HdSt`- specific scene index which would start the tasks, without adding new strongly-typed APIs. My two concerns would be:
1. How can we associate the per-prim data (similar to the filter task) with the sprim? I guess we can always use Sdf paths for that purpose.
2. I do believe that we should still wait for the tasks as late in the pipeline as possible, i.e. in the `Sync` phase, as in the current implementation, for the sake of the minimization of waiting.

Would appreciate your thoughts on the above.

pxr/imaging/hdSt/material.cpp
tcauchois: We have some internal to pixar and in general the material dirty bits are an area of active development by multiple people at multiple companies.  I'd also be somewhat surprised if messing with this didn't cause problems with the Hydra 2 transition (in particular, around https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/imaging/hd/dirtyBitsTranslator.cpp#L247) and what various scene indices are expecting.

Taking a step back, you're overloading the meaning of dirtybits a bit, both as a descriptive tool to say which signals have been invalidated and also as a (set-once) version counter, and I don't think dirty bits are expressive enough for that semantic load.  Even if it works in this PR, it feels brittle to me.

If the thing you're trying to capture is whether we've gotten any material invalidations since you launched the codegen, a more robust tool we use elsewhere (though less targeted) is to use the sprim or material version counter on HdChangeTracker (if there isn't a material version counter, it's easy to add one following the pattern).  This gets invalidated by any material change, but assuming you mostly care about startup, and startup is unlikely to have spurious edits, you still get the speedup you want with correct handling of spurious edits.

pxr/imaging/hd/renderDelegate.h
tcauchois: Sequencing concern: this has gotten better lately but especially with computed scene graphs, you might know the scene topology earlier than you know the scene values, which means when you're populating the scene (sending PrimsAdded), you might not be able to respond to all data queries (GetPrims/etc).  It's come up a few times as a pain point, and with the Hydra 2 rollout we're generally fixing the bridge so you can do early queries (needed for stuff like procedurals), but putting early data fetches in a scene index that we can enable/disable, rather than the heart of the render index, feels a little safer to me.

Parallel sprim sync: we've eliminated most/all of the hazards here and I think are planning to turn this on internally soon if nobody gets to it before us.  If MaterialX codegen is truly geometry independent (we don't codegen different versions for different geomprop availability/etc, which I think is true?) then getting started before Sync() might buy us a bit of time.  In the long run, pushing the join() as late as possible will similarly be helpful, since I don't know that we need the actual codegen (as opposed to just metadata) until we get to DrawBatch::CompileShader, though we might need to do some refactoring for that (e.g. passing material param codegen as a separate mixin).

APIs - one of the reasons we're phasing out HdSceneDelegate is that with the number of folks developing for different plugin points in Hydra, API changes can be pretty disruptive, which raises the bar for when it's worth it to add new entrypoints.  That's part of why I'm being annoying about this, even though the concept is great :).

Associating per-prim-data with the sprim: I'm imagining something like the following:
- Early codegen scene index computes the hash of the material for deduplication purposes.  Using a mechanism similar to the HdInstance<> API, we grab a container for that hash key, and if we're the first to it we launch a thread to fill it in; otherwise we just grab a copy of the container.  When we need to pull material data from an sprim, we figure out its material hash (either recompute it, or store it per-sprim), grab the container, and join() on the codegen task before pulling the codegen out.

Anyway, hope this makes sense!

pxr/imaging/hdSt/material.cpp
ppenenko: Thanks for the explanation. I made some assumptions based on OpenUSD, but fair enough, it's easy to see how bitfields can be a leaky abstraction in dependent code. It's good to know that `HdChangeTracker` is an alternative.

To put things into perspective, `DirtyInit` is there to prevent a relatively minor (dozens of microseconds) performance overhead of redoing `BuildAnonymizedMaterialNetwork`, which is necessary to produce the input for codegen and to compute the material hash. Without it, `BuildAnonymizedMaterialNetwork` would be called once before launching a parallel task, and then again, with the same inputs, when syncing the prim. The purpose of `HdSt_MaterialFilterTask` is to persist the result of the first call to `BuildAnonymizedMaterialNetwork` till material prim sync.

But since that overhead is relatively minor, the parallelization would still be a net performance gain in the absence of `DirtyInit`. So maybe the initial integration of this optimization into OpenUSD could go in without `DirtyInit` and `HdSt_MaterialFilterTask`, with a mechanism based on `HdChangeTracker` added later? That would help us a lot with reducing the divergence between our branch and OpenUSD. We could leave the runtime feature switch off at first to err on the side of caution. Of course, the question of the early task launch API will need to be resolved before this first integration (more on that later).

pxr/imaging/hdSt/materialXFilter.cpp
ppenenko: I should retract this statement - it is in fact a GLSLFX shader because the codegen task includes `HdSt`-specific codegen.

pxr/imaging/hd/renderDelegate.h
ppenenko: OK, so it sounds like refactoring this with a scene index that would launch codegen tasks would alleviate two concerns at once:
1. Sequencing.
2. Avoiding new Render Delegate API additions.

This looks like the way forward, and I'll look into this refactoring soon.

Regarding the parallel material sprim sync - it's great to hear that it's close to becoming a reality, and we seem to agree that the parallel codegen tasks should be a separate mechanism, primarily due to the many-to-one relationship between the two kinds of items.

In terms of synchronization between the two kinds of tasks: unfortunately, since both would be TBB tasks, an individual sprim task can't block waiting on an individual codegen task, to avoid deadlocks in TBB. In my current implementation, I wait for all codegen tasks at once with a `WorkDispatcher` (TBB task group), before syncing any material sprims. This is coarse but is still a net performance win. In many cases, all of the codegen tasks would be done by that point anyway.

In an ideal world, the scheduling of the codegen tasks and the sprim tasks should be handled with a `tbb::flow::graph` or a `WorkTaskGraph`, but I'm not sure how feasible it would be to extend the existing parallel sync mechanism to support something like that.

erikaharrison-adsk:
Approach being replaced by the one at https://github.com/PixarAnimationStudios/OpenUSD/pull/3847 . Closing.

PR #3576, opened 2025-03-21
Autodesk: CMake: export feature compile definition

erikaharrison-adsk:
### Description of Change(s)

- The CMake package should export the feature compile definitions so clients can check what's supported by the binaries
- Fix a bug where the targets file is appended on every build, which causes it grown indefinitely

- N/A

### Fixes Issue(s)

- N/A

### Checklist


cmake/defaults/Packages.cmake
```diff
@@ -268,19 +273,24 @@ endif()
...
 endif()
 
 if (PXR_BUILD_ANIMX_TESTS)
     find_package(AnimX REQUIRED)
 endif()
 
+# Add the definitions globally. These end up in the target
+# COMPILE_DEFINITIONS property, but not in the INTERFACE_COMPILE_DEFINITIONS,
+# which means that they are not exported targets. We have to add them manually.
+add_compile_definitions(${PXR_FEATURE_DEFINITIONS})
```
theblackunknown, line 292: How hard would it be to support `INTERFACE_COMPILE_DEFINITIONS` on OpenUSD targets ?
This seems like a more robust solution as we would get those transitively, and we can introspect those by querying CMake target properties.

pxr/usd/plugin/usdAbc/CMakeLists.txt
```diff
@@ -10,7 +10,6 @@ set(optionalLibs "")
 set(optionalIncludeDirs "")
 
 if (PXR_ENABLE_HDF5_SUPPORT)
-   add_definitions(-DPXR_HDF5_SUPPORT_ENABLED)
```
theblackunknown, line 13: Is this simply removed because it is not used at all ?

cmake/defaults/Packages.cmake
DDoS: I agree that this would be a nicer solution, and I tried at first to add a single top-level `INTERFACE` library to contain the compile definitions. And while I could get it to build, I couldn't figure out the export. It never showed up in the targets file. Pixar does a lot of wrapping around the `add_library` code, the logic is complex, and I had a hard time debugging the issue, so I abandoned that idea for something much more simple...

pxr/usd/plugin/usdAbc/CMakeLists.txt
DDoS: No, I moved it to `Packages.cmake` so it's consistent with the rest of the package support definitions: https://github.com/PixarAnimationStudios/OpenUSD/pull/3576/files#diff-4af9a512717a73d6c1ef832e7e6a48b6ddab7568cb393063d48a0b8656b393c8L261

jesschimein:
Filed as internal issue #USD-10818

PR #3580, opened 2025-03-26
Add Conda environment check in GetPythonInfo for Linux

flipphillips:
### Description of Change(s)

There are several anaconda-related build issues, reported (see #3577, #2371). At leas on Linux builds, some of the problems are due to the conda python reporting that it is built with a static libpython. In the macOS section of the `build_usd` the test is more 'liberal' than simply testing for `LDLIBRARY`. 

This change adds a CondaQ predicate to the build script, and, if true, uses the macOS logic in the build section. Otherwise it remains the 'old way' using `LDLIBRARY`.

### Fixes Issue(s)

Possibly #3577, #2371 on Linux

### Checklist

- [-] I have added unit tests that exercise this functionality (Reference: 
  [testing guidelines](https://openusd.org/release/api/_page__testing__guidelines.html)) NB - this is in a build script and there are no tests published. 

- [-] I have verified that all unit tests pass with the proposed changes NB - I have tested this in multiple environments

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -78,6 +78,9 @@ if MacOS():
 def MacOSTargetEmbedded(context):
     return MacOS() and apple_utils.TargetEmbeddedOS(context)
 
+def CondaEnvQ():
+    return "CONDA_PREFIX" in os.environ
+
 def GetLocale():
     if Windows():
         # Windows handles encoding a little differently then Linux / Mac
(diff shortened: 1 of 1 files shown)


nvmkuruc:
Is it reasonable to just make the `Linux` always do these checks and not just when `conda` is being used?  That would avoid specializing `build_usd` for a particular package manager and reduce the differences between Linux and macOS behavior.

jesschimein:
Filed as internal issue #USD-10829

flipphillips:
> Is it reasonable to just make the `Linux` always do these checks and not just when `conda` is being used? That would avoid specializing `build_usd` for a particular package manager and reduce the differences between Linux and macOS behavior.

Absolutely but I didn’t want to be presumptuous:) in case something depended on the non-Conda-behavior.

PR #3587, opened 2025-04-03
[usdLux] normalized lights divide by surfaceArea expressed in meters^2

pmolodo:
### Description of Change(s)

This allows normalized lights to compose correctly into stages with a different metersPerUnit setting, at a backwards-compatibility cost of a change in luminance of normalized lights before this, in layers with metersPerUnit != 1

(See below for a full discussion.)

https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/77

### Checklist

# Intro + Background

Neither the current UsdLux implementation, or the current version of the UsdLux clarifications and changes in [this OpenUSD proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/77) [(md)](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/pr/usdlux-clarification/proposals/usdlux-clarification/README.md) [(pdf)](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/pr/usdlux-clarification/proposals/usdlux-clarification/README.pdf), explicitly specify how the behavior of `LightAPI.inputs:normalize` relates to linear stage units (`metersPerUnit`), and what happens when the linear stage units change (ie, by being composed into another stage with different `metersPerUnit`).  We wish to further update the UsdLux clarification to address this gap.

## Terms

**lsu = linear stage unit(s)**
- When used with a numerical value - ie, "the cube is 3 lsu" - it means the raw binary numerical value stored in the scene / computer, before any unit-aware compensation is done.
- When used with a unit name - ie, "the lsu is centimeters", it means the setting stored in the scene, via `metersPerUnit` - ie, `metersPerUnit=.01` implies "the lsu is cm"

**nit**
- while not a formal SI unit, it is a common alias for the SI unit of luminance, candela per square meter (`cd/m^2`)

## Common usage of the normalize attribute

In theory, you might desire to have normalized lights and non-normalized lights behave differently when composed into a scene with different units... but I struggled to come up with a credible common scenario for that.

Most commonly, normalize is used to allow lighting artists to adjust the size of a light without affecting it's overall illumination output on the surrounding scene.  ie, you might want to make shadows softer without dramatically altering the overall light level of the scene.  It's used as a way to alter the way you sculpt the light while designing it - not as a way to mark it as a fundamentally different type of light, that should behave in a fundamentally different way from other lights, when placed into a new setting, or when automated, non-artistic operations are done on it.

As transforms associated with changes in scene scale are "non-artistic" (and ideally, automated) changes, it follows that artists will generally want normalized and non-normalized lights to behave similarly in this situation.  That is, if they have two lights that appear identical in two_lights.usd, except one is normalized and one isn't, they will want the two lights to *stay* identical to each other when placed into a new scene.

# Assumptions and Observations

## Assumption 1: Light dimensions must be in lsu

If this were not the case, then a layer with both lights + geometry would not retain the same relative size when composed into a scene with a different `metersPerUnit` - which is unlikely to be the wanted behavior.

## Assumption 2: For non-normalized lights, `inputs:intensity` represents luminance in `nits`

This document is intended as a further exploration to the proposed UsdLux clarifications and changes from this [OpenUSD proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/77) [(md)](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/pr/usdlux-clarification/proposals/usdlux-clarification/README.md) [(pdf)](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/pr/usdlux-clarification/proposals/usdlux-clarification/README.pdf), and that proposal assumes (non-normalized) `inputs:intensity` is in a fixed units, `nits`.

## Observation 1: Direct light sampling for non-normalized, non-ies lights are generally not affected by `metersPerUnit`

For non-normalized lights, we've assumed `inputs:intensity` [is always a measure of base luminance in `nits`](#assumption-2-for-non-normalized-lights-inputsintensity-represents-luminance-in-nits).  Because measurement of luminance is unaffected by distance, it is thus unaffected by a scale change.  Most other UsdLux light properties are modifiers to the luminance, whose formulas do not depend on any values which are in scene units. (One possible exception are lights with attached ies files, whose measurements are defined to be luminous intensity values expressed in candela. However, we defer discussion of unit interaction with ies files for another disucssion.)  Therefore, direct light sampling for non-normalized, non-ies lights are unaffected by scale changes.

By "direct light sampling", I mean the sampling of a light for a ray known to intersect the light.  The [light dimensions are in lsu](#assumption-1-light-dimensions-must-be-in-lsu), so a `metersPerUnit` change can/will affect their size and position in the scene, and therefore can affect whether a given ray will intersect it, or the exact sampling pattern used. But, once it's known a ray intersects a non-normalized, non-ies light, the `metersPerUnit` setting is unimportant.

Therefore, other than ensuring the size scale of a non-normalized light is as desired, they will generally not present any special problems when undergoing unit changes.

-------------------------------------------------------------------------------

# Normalized Area Light Option 1: Normalized lights divide by scene-relative surface area

## Summary of behavior
- The `sizeFactor` for normalized area lights (where `luminance = inputs:intensity / sizeFactor`)
  is in scene-relative units of `lsu^2`
- ie, we divide by the "raw" binary surface area, not taking scene units / `metersPerUnit`
  into account
- Believe this is what most renderers currently do

## When rendering the composed stage, renderers need to:
- **no extra action required**
- because all relevant quantities are expressed in lsu, no conversions are necessary, and in theory, the renderer shouldn't even have to read the `metersPerUnit` setting

## When composing lighting scenes with different `metersPerUnit`, artists / end users need to:
- artist action always required: **apply compensating scale xform**
    - ie, the current "standard behavior" taken to preserve real-world units for any reference with different `metersPerUnit`
- additonal artist action required for normalized area lights: **per light overrides must be manually calculated and applied**
    - `inputs:intensity` is in lsu-relative units, and so remains unchanged
    - ...but because of the compensating scale xform, the dimensions of the light in lsu changes
    - because the dimensions of the light in lsu change, the calculated normalization scale factor changes, and thefore the luminance (in `nits`) of the light changes
    - put another way - the compensating scale xform will be treated just like any other scale xform,
      and the light will try to automatically adjust it's luminance without knowledge of the scene
      scale causing the brightness of the light to be incorrect relative to the intended size of the
      asset
    - HOWEVER - this change will not reflect how a "real world" light would behave, and therefore is unlikely to be what an end user "desires" - regardless of [whether the light is marked as "normalized" or not](#common-usage-of-the-normalize-attribute)
        - if the comosed asset also has non-normalized lights, their relative behavior will change
        - if the composed asset also has "normal" geo, the light interaction will change
    - therefore, the end user will need to manually apply new intensity override values on all normalized lights, which are precalculated to offset the scale change
        - also, any subsequent changes to intensity in the original source would be overriden/lost
    - Can we put the burden on USD or renderer, to handle this automatically?
        - Any solutions I could envision require significant additions or changes to USD

## Backwards compatibility behavior

- This is generally the current behavior adopted by most renderers, so therea are no backward compatibility issues.

-------------------------------------------------------------------------------

# Normalized Area Light Option 2: Normalized lights divide by fixed-unit surface area (meters^2)

## Summary of behavior
- The `sizeFactor` for normalized area lights (where `luminance = inputs:intensity / sizeFactor`)
  is in fixed units of `meters^2`

## When rendering the composed stage, renderers need to:
- for non-normalized area lights:
    - **no extra action required**
- for normalized lights:
    - for "standard" area lights with non-zero surface area:
        - **normalization sizeFactor multiplied by `metersPerUnit`^2**
        - ie, the "raw" sizeFactor would be an area in units of lsu^2
        - by multiplying by `metersPerUnit`^2, we would get an area in units of m^2
    - for "line-like" lights with with zero-surface area, but a "length":
        - **normalization sizeFactor multiplied by `metersPerUnit`**
        - ie, the "raw" sizeFactor would be a length in units of lsu
        - by multiplying by `metersPerUnit`, we would get a length in units of m
        - lights that meet this condition:
            - CylinderLight with length != 0 and treatAsLine == true
    - for other lights:
        - **normalization sizeFactor unaffected by `metersPerUnit`**

## When composing lighting scenes with different `metersPerUnit`, artists / end users need to:
- artist action always required: **apply compensating scale xform**
    - ie, the current "standard behavior" taken to preserve real-world units for any reference with different `metersPerUnit`
- additonal artist action required for normalized area lights: **None**
    - The compensatory scale xform already ensured that the size of the lights, in real-world units, remained unchanged
    - The renderer will automatically apply the additional factor required to scale the normalization sizeFactor by the `metersPerUnit`, at render time (see above)
    - So no extra action needed on part of artist

## Backwards compatibility behavior

- Area lights with normalize on, and a scale other than `metersPerUnit=1`, will be significantly brighter or darker in "old" vs "new" UsdLux
- `newLuminance = oldLuminance / metersPerUnit^2`
- ie, for scenes with default cm-scale (`metersPerUnit=.01`), normalized lights in "new" usdLux will have 10,000x higher luminance than with "old" UsdLux

-------------------------------------------------------------------------------

# Summary and Analysis

## Summary of differences

Here's a summary of differences, with my subjective rating of which option has the superior behavior in each area:

| Normalized Area Units| 1: Scene-Relative (lsu^2) | 2: Fixed (m^2) |
|---------------------------|----------------|-----------|
| Additional renderer action | None | Multiply area by metersPerUnit^2 when normalizing |
| Artist actions needed when composing lights into a stage with different `metersPerUnit` | <li> Add compensating scale xform (as normal) <li>  Manually override intensity on all normalized lights | <li> Add compensating scale xform (as normal) |
| Backwards Compatibility Issues | Full (current behavior) | Normalized lights significantly change luminance compared to old behavior (ie, scenes with default cm-scale become 10,000x brighter)  |

### Big win for [Option 2 (Normalized Fixed Units)][Option2]: no manual overrides on lights to preserve correct real-world units
- The need for [Option 1][Option1] to treat normalized units as different, and to have to manually calculate and apply overrides PER-NORMALIZED-LIGHT, is a significant extra burden - thus, [Option 2][Option2] - is a big winner here.

### Big win for [Option 1 (Normalized Scene-relative units)][Option1]: better backwards compatibility of normalized light intensity for "most" scenes
- Unless a stage has a LSU of meters (ie, `metersPerUnit=1`), the brightness of normalized lights under [Option 2][Option2] will [change by orders of magnitude](#backwards-compatibility-behavior-1)
- As the default LSU is centimeters (`metersPerUnit=1`), it is assumed that "many" / "most" scenes will have their normalized lights change significantly under [Option 2][Option2]

### Minor win for [Option 1 (Normalized Scene-relative units)][Option1]: renderers need no additional unit-aware changes

- In the absence of other factors, obviously the fewer changes we require the renderers to make, the better
- So the fact that [Option 1][Option1] requires no renderer-side changes is a win for it
- ...however, it feels a relatively minor win, as it's a relatively straightforward / easy change, and we're already requiring a lot of other renderer-side changes

## Conclusion

 - We have a choice between backwards compatibility ([Option 1][Option1]) vs better / more correct behavior going forward ([Option 2][Option2])

- The UsdLux clarification project is about providing better behavior going forward, at the cost of some backward compatibility, so overall we feel [Option 2][Option2] is the better choice long-term.

[Option1]: #normalized-area-light-option-1-normalized-inputsintensity-is-in-scene-relative-power-units-current-behavior
[Option2]: #normalized-area-light-option-2-normalized-inputsintensity-is-in-fixed-power-units

### DIFF
--- a/pxr/usd/usdLux/generatedSchema.usda
+++ b/pxr/usd/usdLux/generatedSchema.usda
@@ -167,7 +167,7 @@ class "LightAPI" (
         displayName = "Normalize Power"
         doc = """Normalizes the emission such that the power of the light
         remains constant while altering the size of the light, by dividing the
-        luminance by the world-space surface area of the light.
+        luminance by the world-space surface area of the light, in meters².
 
         This makes it easier to independently adjust the brightness and size
         of the light, by causing the total illumination provided by a light to
@@ -196,18 +196,34 @@ class "LightAPI" (
         ### Area Lights:
 
         For an area light, the `sizeFactor` is the surface area (in world
-        space) of the shape of the light, including any scaling applied to the
-        light by its transform stack. This includes the boundable light types
+        space, expressed in meters²) of the shape of the light, including any
+        scaling applied to the light by its transform stack. The area should
+        then be multiplied by the scene's `metersPerUnit^2` to convert the
+        surface area into meters². Area lights include the boundable light types
         which have a calculable surface area:
 
         - MeshLightAPI
         - DiskLight
         - RectLight
         - SphereLight
-        - CylinderLight
+        - CylinderLight with treatAsLine == false
 
         <center><b>
-                        sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light)
+                sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light) * metersPerUnit²
+        </b></center>
+
+        ### Line Lights:
+
+        Lights that simulate 1-dimensional emissive \"lines\" have a `sizeFactor`
+        equal to the length (in world space, expressed in meters) of the shape
+        of the light, including any scaling applied to the light by its
+        transform stack. The length should then be multiplied by the scene's
+        `metersPerUnit` to convert the length into meters. Line lights include:
+
+        - CylinderLight with length != 0 and treatAsLine == True
+
+        <center><b>
+                sizeFactor<sub>line</sub> = worldSpaceLength(light) * metersPerUnit
         </b></center>
 
         ### DistantLight:
--- a/pxr/usd/usdLux/lightAPI.h
+++ b/pxr/usd/usdLux/lightAPI.h
@@ -424,7 +424,7 @@ public:
     // --------------------------------------------------------------------- //
     /// Normalizes the emission such that the power of the light
     /// remains constant while altering the size of the light, by dividing the
-    /// luminance by the world-space surface area of the light.
+    /// luminance by the world-space surface area of the light, in meters².
     /// 
     /// This makes it easier to independently adjust the brightness and size
     /// of the light, by causing the total illumination provided by a light to
@@ -453,18 +453,34 @@ public:
     /// ### Area Lights:
     /// 
     /// For an area light, the `sizeFactor` is the surface area (in world
-    /// space) of the shape of the light, including any scaling applied to the
-    /// light by its transform stack. This includes the boundable light types
+    /// space, expressed in meters²) of the shape of the light, including any
+    /// scaling applied to the light by its transform stack. The area should
+    /// then be multiplied by the scene's `metersPerUnit^2` to convert the
+    /// surface area into meters². Area lights include the boundable light types
     /// which have a calculable surface area:
     /// 
     /// - MeshLightAPI
     /// - DiskLight
     /// - RectLight
     /// - SphereLight
-    /// - CylinderLight
+    /// - CylinderLight with treatAsLine == false
     /// 
     /// <center><b>
-    /// sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light)
+    /// sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light) * metersPerUnit²
+    /// </b></center>
+    /// 
+    /// ### Line Lights:
+    /// 
+    /// Lights that simulate 1-dimensional emissive "lines" have a `sizeFactor`
+    /// equal to the length (in world space, expressed in meters) of the shape
+    /// of the light, including any scaling applied to the light by its
+    /// transform stack. The area should then be multiplied by the scene's
+    /// `metersPerUnit` to convert the length into meters. Line lights include:
+    /// 
+    /// - CylinderLight with length != 0 and treatAsLine == True
+    /// 
+    /// <center><b>
+    /// sizeFactor<sub>line</sub> = worldSpaceLength(light) * metersPerUnit
     /// </b></center>
     /// 
     /// ### DistantLight:
--- a/pxr/usd/usdLux/schema.usda
+++ b/pxr/usd/usdLux/schema.usda
@@ -252,7 +252,7 @@ class "LightAPI" (
         displayName = "Normalize Power"
         doc = """Normalizes the emission such that the power of the light
         remains constant while altering the size of the light, by dividing the
-        luminance by the world-space surface area of the light.
+        luminance by the world-space surface area of the light, in meters².
 
         This makes it easier to independently adjust the brightness and size
         of the light, by causing the total illumination provided by a light to
@@ -281,18 +281,34 @@ class "LightAPI" (
         ### Area Lights:
 
         For an area light, the `sizeFactor` is the surface area (in world
-        space) of the shape of the light, including any scaling applied to the
-        light by its transform stack. This includes the boundable light types
+        space, expressed in meters²) of the shape of the light, including any
+        scaling applied to the light by its transform stack. The area should
+        then be multiplied by the scene's `metersPerUnit^2` to convert the
+        surface area into meters². Area lights include the boundable light types
         which have a calculable surface area:
 
         - MeshLightAPI
         - DiskLight
         - RectLight
         - SphereLight
-        - CylinderLight
+        - CylinderLight with treatAsLine == false
 
         <center><b>
-                        sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light)
+                sizeFactor<sub>area</sub> = worldSpaceSurfaceArea(light) * metersPerUnit²
+        </b></center>
+
+        ### Line Lights:
+
+        Lights that simulate 1-dimensional emissive "lines" have a `sizeFactor`
+        equal to the length (in world space, expressed in meters) of the shape
+        of the light, including any scaling applied to the light by its
+        transform stack. The length should then be multiplied by the scene's
+        `metersPerUnit` to convert the length into meters. Line lights include:
+
+        - CylinderLight with length != 0 and treatAsLine == True
+
+        <center><b>
+                sizeFactor<sub>line</sub> = worldSpaceLength(light) * metersPerUnit
         </b></center>
 
         ### DistantLight:


jesschimein:
Filed as internal issue #USD-10863

spitzak:
I'm pretty certain everybody wants option 1. The light does not change when the meters per unit changes, even if you are working with "real world light units". Yes the total amount of light emitted will increase by the square of the scale, but the distance of that light from any objects will also increase by the scale and thus drop off by the square of the scale. 
For every light an "area" is computed, and if "normalize" is on then when a ray hits the light, the color/intensity/etc settings are divided by this "area" to get the emitted value. What is missing in USD is the settings where "area" is 1.0. It would also be useful to pin down exactly how the "area" is computed (for instance does the cone angle change it?). Also any indication of how accurate the "area" of a mesh light needs to be, I suspect assuming it is a bounding-box-filling ellipsoid is good enough.

anderslanglands:
Hi @spitzak, could you explain a bit more what you'r suggesting here? You seem to be saying that option 1 preserves brightness when a normalized light is scaled, which is not the case.

roelandschoukens:
@spitzak I am not sure about that, many people are used to specifying their light source power in _Watt_. This much more closely resembles option 2, where the same number always specifies the same total amount of light.

With option 1, if I understand it correctly, the thing that remains invariant with `metersPerUnit` is the brightness of everything. Consider what happens if you change that value from 1/100 to 1, (aka cm → m) without changing any other numbers:

 1. every light will be 100 times as large, and it will have 10,000 as much surface area.
 2. option 1 keeps the luminance of the light emission constant: this is trivially true for non-normalized lights, and for normalized lights, _sizeFactor_ did not change
 3. Due to 1 and 2, every light will emit 10,000 times as much power
 4. Illumination follows inverse square law: everything is 100 times further away. The illumination cast by the light is spread over a larger area, exactly compensating for the increased total power. The brightness of the illumination stays the same.

What is the specification for line lights? You cannot specify luminance for these. Presumably you can specify the total emission to be equivalent to that of a 1×1 area light for a point light, and that of a 1×N area light for a line-like light? (where 1 represents one unit in Option 1, and 1 metre in option 2)

Finally, I would expect that regardless of the choice being made here, DCC tools or renderers would automatically do the right thing when loading an USD file: scale distances properly, and translate those light settings to whatever representation they internally use. So ideally artists should not notice anything or need to do anything special right? Do USD files have a version field so those tools know which behaviour to use for a certain file?

EDIT: apparently some programs will not correct for different units when composing multiple USD files. I don’t think there is any sensible way to standardise this behaviour unless correction for units is also mandatory when composing USDs.

It would also be good to specify what happens if an artist defined transform is applied to the light geometry. Eg. does applying a 2x scale to a _normalized_ light quadruple its power? For non-uniform transforms the situation is a bit hairy, eg. with a [1, 1, 2] scaling transform, a quad light may stay the same, or it may double in area, depending on orientation. For a sphere the area will change according to some hairy mathematical formula.

roelandschoukens:
And about putting the specification in words, maybe this can be done in terms of units. Assuming option 2 is chosen (please correct me if I’m wrong here):

With normalized _enabled_:
 * For all types of light, the intensity specifies luminous intensity (cd).

With normalized _disabled_:
 * For area lights the intensity value specifies luminance (nit, aka cd/m²)
 * For line lights the intensity specifies luminous intensity per metre of length (cd/m)
 * For point lights the intensity still specifies luminous intensity (cd)

PR #3589, opened 2025-04-04
Autodesk: Vulkan validation errors

erikaharrison-adsk:
### Description of Change(s)

* Fix Vulkan validation errors related to volume rendering.
* A cull compute shader sometime has a `topologyVisibility` binding. The `pipelineDrawBatch` doesn't bind this buffer for culling. It's not actually used, so not binding the buffer doesn't break anything, but Vulkan mandates that any buffer in a descriptor set layout (created by reflection) must be bound, regardless of its use. This caused a validation error. I moved the binding call so it now executes for both cull and draw.
* `SubmitLayoutChange` should return the old layout so restoring it isn't as error prone
* Sometimes old layout can be depth instead of color.
* Don't enable alpha-to-coverage if there's no color attachment

- N/A

### Fixes Issue(s)

- N/A

### Checklist


jesschimein:
Filed as internal issue #USD-10870

PR #3603, opened 2025-04-16
Autodesk: build_usd.py now installs Vulkan dependencies

erikaharrison-adsk:
### Description of Change(s)

- Using `build_usd.py`, build and install:
  - Vulkan-Headers
  - Vulkan-Utility-Libraries
  - Vulkan-Loader
  - VulkanMemoryAllocator
  - shaderc
- Only the required build and runtime files are installed. Tools and debug binaries (like validation layers) are expected to be installed externally (by the Vulkan SDK).
- All versions match the ones from Vulkan SDK 1.3.296.0 (last 1.3.x release)
- Removed third party files from hgiVulkan, they are now handled by `build_usd.py`
  - We had to do some small code changes in hgiVulkan in order to better use the third party dependencies
- Installing the Vulkan ICD loader will prevent load-time link issues if no Vulkan enabled graphics driver is installed
  - This is an uncommon situation, but we've had it happen, especially in VMs
  - This is an allowed use case for distributing the ICD
- Using the Vulkan SDK is still supported
  - When using the new argument `--use-vulkan-sdk`
  - When not using `build_usd.py`
- Use `BUILD_LOCAL_INTERFACE` to keep the hgiVulkan private dependencies out of the `pxr` target export.
  - This makes it simpler for client projects to import the `pxr` targets. You don't need to link against as many targets.
  - It's still possible to manually do `find_package` (using the USD install path) and link against the installed Vulkan dependencies.
- Add the Vulkan dependencies to `pxrConfig.cmake`.
  - Make it even simpler for client projects. It also takes care of the Vulkan SDK and `build_usd.py` differences.
  - Only the public dependencies (not wrapped in `BUILD_LOCAL_INTERFACE`) need to be found.

- N/A

### Fixes Issue(s)

- N/A

### Checklist


jesschimein:
Filed as internal issue #USD-10914

DarkDefender:
I've tested this out and it works really well besides two issues:

1. If I build VulkanMemoryAllocator myself using the default build options for it (not using the `build_usd.py` script), building openUSD will fail. I would change this so that we always use `#include <vk_mem_alloc.h>`. Then when using the VULKAN_SDK, we add `$ENV{VULKAN_SDK}/include/vma` to the default include directories. This way people can use the default/standard VulkanMemoryAllocator without any issues.

2. When building with GCC 14 on my machine, `pxr/imaging/hgiVulkan/sampler.cpp` fails to compile because of a missing include. Simply adding `#include <algorithm>` in the file will solve the build issue.

DDoS:
> I've tested this out and it works really well besides two issues:
> 
>     1. If I build VulkanMemoryAllocator myself using the default build options for it (not using the `build_usd.py` script), building openUSD will fail. I would change this so that we always use `#include <vk_mem_alloc.h>`. Then when using the VULKAN_SDK, we add `$ENV{VULKAN_SDK}/include/vma` to the default include directories. This way people can use the default/standard VulkanMemoryAllocator without any issues.
> 
>     2. When building with GCC 14 on my machine, `pxr/imaging/hgiVulkan/sampler.cpp` fails to compile because of a missing include. Simply adding `#include <algorithm>` in the file will solve the build issue.

Thanks for the feedback, [I made some changes](https://github.com/PixarAnimationStudios/OpenUSD/compare/3808f5897ff73d39c9aa4f17237d4c22c4d5cac5..901ee01502e35d089ebb6da3dad264c176a4f1fb), let me know if that fixes your issues.

DarkDefender:
I can confirm that it fixes my issues! Thanks! :)

DarkDefender:
@DDoS the new commit breaks compilation for me.
The build now errors out with spirv_reflect not being able to find the spirv headers anymore:
```
pxr/imaging/hgiVulkan/shaderCompiler.cpp
In file included from pxr/imaging/hgiVulkan/shaderCompiler.cpp:15:
/usr/include/spirv_reflect.h:37:10: fatal error: ./include/spirv/unified1/spirv.h: No such file or directory
   37 | #include "./include/spirv/unified1/spirv.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
If I skip the new commit and just apply the `build_usd.py now installs Vulkan dependencies` commit, everything compiles fine again.

Seems like your changes somehow broke the logic for passing `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` to spirv_reflect

DDoS:
@DarkDefender can you please share your `spirv-reflect_*` cache variable values?

DarkDefender:
@DDoS if I grep the cmake cache file I get these two:
```
spirv-reflect_INCLUDE_DIR:PATH=/usr/include
spirv-reflect_LIBRARY:FILEPATH=/usr/lib/libspirv-reflect-static.a
```

DDoS:
@DarkDefender can you add some debug `message()` to this condition `if (SpirvReflect_FOUND AND NOT EXISTS "${spirv-reflect_INCLUDE_DIR}/include/spirv/unified1/spirv.h")` in `FindSpirvReflect.cmake` to see what the result is?

cmake/defaults/Packages.cmake
```diff
@@ -187,38 +187,60 @@ if (PXR_BUILD_IMAGING)
...
+            find_package(VulkanMemoryAllocator CONFIG REQUIRED)
+            list(APPEND VULKAN_LIBS GPUOpen::VulkanMemoryAllocator)
+
+            find_package(ShaderC MODULE REQUIRED)
+            list(APPEND VULKAN_LIBS $<BUILD_LOCAL_INTERFACE:shaderc_combined>)
         endif()
+
+        # In either case we have custom logic for SPIRV-Reflect
+        find_package(SpirvReflect MODULE REQUIRED)
+        list(APPEND VULKAN_LIBS $<BUILD_LOCAL_INTERFACE:spirv-reflect>)
```
DarkDefender: If I change this back to `list(APPEND VULKAN_LIBS spirv-reflect)` it seems to work.
I'm guessing that the `SpirvReflect_IS_SOURCE_DEP` check is needed here.

DarkDefender:
> @DarkDefender can you add some debug message() to this condition if (SpirvReflect_FOUND AND NOT EXISTS "${spirv-reflect_INCLUDE_DIR}/include/spirv/unified1/spirv.h") in FindSpirvReflect.cmake to see what the result is?

The debug message() gets triggered for me. `target_compile_definitions(spirv-reflect INTERFACE SPIRV_REFLECT_USE_SYSTEM_SPIRV_H)` should be executed.

cmake/defaults/Packages.cmake
DDoS: That's strange, I didn't expect this to make a difference. I'll have to try harder and reproduce this issue locally to debug. Thanks for helping out.

DDoS:
@DarkDefender Which CMake version are you using? Which build system?

I really can't reproduce this issue. When I build with `build_usd.py`, `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` is enabled, and `spirv.h` is found.

DarkDefender:
@DDoS I'm using cmake 3.31.7 and ninja.

Are you sure that the `build_usd.py` script doesn't copy in the `spirv/unified1/spirv.h` header?
The `CopyDirectory(context, 'include/spirv', 'include/spirv')` line makes me a bit suspicious that this might be the case.
If this is the case you wouldn't really notice that `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` doesn't get picked up as it will find the `spirv.h` header that is bundled with spirv-reflect.

DDoS:
@DarkDefender The relevant `build_usd.py` include paths are
- include
    - spirv_reflect.h
    - spirv
        - unified1
            - spirv.h

When `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` is undefined, the path relative to `spirv_reflect.h` needs to start with `include` (`./include/spirv/unified1/spirv.h`), which here it doesn't. Furthermore I can check the compiler command line in build.ninja and see that `-DSPIRV_REFLECT_USE_SYSTEM_SPIRV_H` is given when building `shaderCompiler.o`.

Since it appears you're using a system package manager install of spirv-reflect, can you confirm the header does support this define? It was added three years ago: https://github.com/KhronosGroup/SPIRV-Reflect/commit/f2241d55af5404a2d17c9cb8a73d131fdc3ce202 (sdk-1.3.216.0).

DarkDefender:
@DDoS I am using the latest git version of https://github.com/KhronosGroup/SPIRV-Reflect
It does have `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` as it wouldn't have worked before if that wasn't the case.

It seems like on my end `list(APPEND VULKAN_LIBS $<BUILD_LOCAL_INTERFACE:spirv-reflect>)` will ignore that we set `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` with `target_compile_definitions(spirv-reflect INTERFACE SPIRV_REFLECT_USE_SYSTEM_SPIRV_H)` while `list(APPEND VULKAN_LIBS spirv-reflect)` doesn't.

I'm not sure why though.

DDoS:
> @DDoS I am using the latest git version of https://github.com/KhronosGroup/SPIRV-Reflect It does have `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` as it wouldn't have worked before if that wasn't the case.
> 
> It seems like on my end `list(APPEND VULKAN_LIBS $<BUILD_LOCAL_INTERFACE:spirv-reflect>)` will ignore that we set `SPIRV_REFLECT_USE_SYSTEM_SPIRV_H` with `target_compile_definitions(spirv-reflect INTERFACE SPIRV_REFLECT_USE_SYSTEM_SPIRV_H)` while `list(APPEND VULKAN_LIBS spirv-reflect)` doesn't.
> 
> I'm not sure why though.

This almost feels like a CMake bug. I'm using 3.31.6, we're only one patch version off. Could you downgrade and test again?

DarkDefender:
I tried with the cmake 3.31.6 version, no change.
I also tried to see if I can reproduce the issue with the `build_usd.py` and I can.
This is the command I use to reproduce the issue:
```
python build_scripts/build_usd.py --build-monolithic --no-python --vulkan --no-openvdb --no-embree --no-openimageio --no-opencolorio --no-alembic --no-hdf5 --no-draco --no-materialx --no-mayapy-tests --no-animx-tests --no-tutorials --no-docs --no-python-docs --no-examples --no-tests /tmp/usd_build/
```

@DDoS Can you see if it also triggers on your machine with the same flags to `build_usd.py`?
If it makes any difference, the build_usd.py uses `make` as the build system here.

build_scripts/build_usd.py
```diff
@@ -1643,6 +1643,102 @@ def InstallAnimX(context, force, buildArgs):
...
+        extraArgs = buildArgs + [
+            '-DSPIRV_REFLECT_EXECUTABLE=OFF',
+            '-DSPIRV_REFLECT_STATIC_LIB=ON',
+            '-DSPIRV_REFLECT_BUILD_TESTS=OFF',
+            '-DSPIRV_REFLECT_ENABLE_ASSERTS=OFF',
+            '-DSPIRV_REFLECT_ENABLE_ASAN=OFF',
+        ]
+        RunCMake(context, force, extraArgs)
+        CopyFiles(context, 'spirv_reflect.h', 'include')
+        CopyDirectory(context, 'include/spirv', 'include/spirv')
```
DarkDefender: I think we should remove this as the spirv header file is not used and is not supposed to be used.

DarkDefender:
@DDoS Seems like the `--build-monolithic` is what show cases the issue.
If I build a non monolithic library, it works.

So I'm guessing that there are some flags that doesn't get passed on properly in the monolithic build or something.

DDoS:
@DarkDefender Thanks for digging deeper into this! I never tried a monolithic build before, I'll give it a try, and fingers crossed that will reproduce the issue. I think I already have an idea of what the problem may be.

Update: I can confirm this reproduces the issue!

build_scripts/build_usd.py
DDoS: This header is the one included by `spirv_reflect.h`. It's needed.

DarkDefender: Ah, you are right. Never mind. I somehow got the idea that the build script also installed the spirv headers. But it doesn't. Sorry for the noise.

DDoS:
@DarkDefender [this](https://github.com/PixarAnimationStudios/OpenUSD/pull/3603/files#diff-2bce2a745085745a286ecd010f9f85c4cabb032cff8e6de3e9f199124dd57f28R826-R831) should fix the problem. Let me know if it works now, thanks.

DarkDefender:
Seems to build fine for me now again! :)

I just have one last bit of feedback. (See my review comment)

cmake/modules/FindSpirvReflect.cmake
```diff
@@ -0,0 +1,52 @@
...
+        target_include_directories(spirv-reflect PUBLIC "${spirv-reflect_INCLUDE_DIR}")
+        set_target_properties(spirv-reflect PROPERTIES
+            ARCHIVE_OUTPUT_DIRECTORY "third_party")
+    endif()
+endif()
+
+if (SpirvReflect_FOUND AND NOT EXISTS "${spirv-reflect_INCLUDE_DIR}/include/spirv/unified1/spirv.h")
+    # Most probably not from the Vulkan SDK, so use a "system path."
+    # If installed from build_usd.py this will be found in the USD install path.
+    if (SpirvReflect_IS_SOURCE_DEP)
```
DarkDefender: I don't think this is defined anymore so this will always be false?

DDoS: Fixed! Good catch, I think this somehow got lost in the cherry-pick.

pxr/imaging/hgiVulkan/sampler.cpp
```diff
@@ -12,7 +12,7 @@
 #include "pxr/imaging/hgiVulkan/sampler.h"
 #include "pxr/imaging/hgiVulkan/diagnostic.h"
 
-#include <float.h>
```
DarkDefender, line 15: Just one last thing I noticed.
On older gcc versions (GCC 11), this include or `#include <cfloat>` is needed because it can't find the definition of `FLT_MAX` otherwise. Don't know why this isn't needed in newer gcc versions, but I don't think we do any harm by explicitly including `<cfloat>`

DDoS: re-added

DarkDefender:
All seems to be working fine on my side now! Thank you so much for all the back and forth! :)

davidgyu:
Have you looked into doing this without adding additional custom cmake Find modules?
CMake already has built-in support for locating Vulkan SDK components and it would preferable to have the core build system depend on this standard mechanism.

DDoS:
> Have you looked into doing this without adding additional custom cmake Find modules? CMake already has built-in support for locating Vulkan SDK components and it would preferable to have the core build system depend on this standard mechanism.

The `FindVulkan.cmake` logic that comes with CMake doesn't handle SPRIV-Reflect, that one I don't see how we could do without.

For shaderc there's a pkg-config file that can be used, but that would require adding pkg-config as a dependency, so I didn't consider it. If that's not actually a problem then I could maybe remove it.

I understand that adding "find modules" is to avoid, so I tried to keep them simple and validate them on as many platforms as possible (macOS, Windows, Linux, Android).

I considered laying out the files as in the Vulkan SDK so `FindVulkan.cmake` could be used, but the file layout is quite different per platform, so it would just shift the complexity to the `build_usd.py` and probably end up more brittle.

DarkDefender:
> I considered laying out the files as in the Vulkan SDK so `FindVulkan.cmake` could be used, but the file layout is quite different per platform, so it would just shift the complexity to the `build_usd.py` and probably end up more brittle.

Most Linux distributions do not ship the Vulkan SDK in their package managers either. Only the individual components are available.

It also seems to me like most other Vulkan libraries/projects do not depend on the vulkan SDK directly.

DDoS:
This PR would help a lot in maintaining the vcpkg port of USD. A lot of the patching required by the port is made obsolete by this PR.

meshula:
@DDoS also notes that Blender has cherrypicked this PR. https://github.com/blender/blender/commit/eb96c2dfb30289a8449ac6802de61ddb7ece4879

PR #3606, opened 2025-04-21
Autodesk: Set Default Path of TBB and OpenSubdiv in pxrConfig.cmake.in

erikaharrison-adsk:
### Description of Change(s)

Similar to MaterialX exported targets, adding these default paths makes pxrConfig.cmake easy to use, especially when TBB and OpenSubdiv are passed to USD cmake step. And user doesn't need to specify TBB_DIR and OpenSubdiv_DIR when using pxr exported targets.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/pxrConfig.cmake.in
+++ b/pxr/pxrConfig.cmake.in
@@ -53,6 +53,13 @@ if (NOT DEFINED PXR_FIND_TBB_IN_CONFIG)
     set(PXR_FIND_TBB_IN_CONFIG "@PXR_FIND_TBB_IN_CONFIG@")
 endif()
 if (PXR_FIND_TBB_IN_CONFIG)
+    # Set TBB_DIR to where the TBBConfig.cmake file was located.
+    # This can be overridden by specifying a different TBB_DIR when running cmake.
+    if (NOT DEFINED TBB_DIR)
+        if (NOT [[@TBB_DIR@]] STREQUAL "")
+            set(TBB_DIR [[@TBB_DIR@]])
+        endif()
+    endif()
     find_dependency(TBB @TBB_VERSION@ CONFIG)
 endif()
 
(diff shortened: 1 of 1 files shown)


pmolodo:
I tested this with `build_usd.py`, and with the default values (ie, `--no-onetbb`), I got a `pxrConfig.cmake` with:

```cmake
if (NOT DEFINED PXR_FIND_TBB_IN_CONFIG)
    set(PXR_FIND_TBB_IN_CONFIG "OFF")
endif()
if (PXR_FIND_TBB_IN_CONFIG)
    # Set TBB_DIR to where the TBBConfig.cmake file was located.
    # This can be overridden by specifying a different TBB_DIR when running cmake.
    if (NOT DEFINED TBB_DIR)
        if (NOT [[TBB_DIR-NOTFOUND]] STREQUAL "")
            set(TBB_DIR [[TBB_DIR-NOTFOUND]])
        endif()
    endif()
    find_dependency(TBB 2020.3 CONFIG)
endif()
```

I'm unsure of Cmake's behavior if `${TBB_DIR} STREQUAL "TBB_DIR-NOTFOUND"` - if this causes an immedate fail of `find_package(TBB)`, this could create problems for downstream deps that want to use OneTBB (even though USD didn't use it).
(If `find_package(TBB)` still checks all the "standard places," even if `${TBB_DIR} STREQUAL "TBB_DIR-NOTFOUND"`, though, then this should be harmless.)

If I built with `build_usd.py --onetbb`, I got:

```cmake
if (NOT DEFINED PXR_FIND_TBB_IN_CONFIG)
    set(PXR_FIND_TBB_IN_CONFIG "ON")
endif()
if (PXR_FIND_TBB_IN_CONFIG)
    # Set TBB_DIR to where the TBBConfig.cmake file was located.
    # This can be overridden by specifying a different TBB_DIR when running cmake.
    if (NOT DEFINED TBB_DIR)
        if (NOT [[/src/USD/usd_build/USD/_build/Release/usd0.25.05/py3.10/install/lib/cmake/TBB]] STREQUAL "")
            set(TBB_DIR [[/src/USD/usd_build/USD/_build/Release/usd0.25.05/py3.10/install/lib/cmake/TBB]])
        endif()
    endif()
    find_dependency(TBB 2021.9.0 CONFIG)
endif()
```
...which means we've ended up with a non-portable `pxrConfig.cmake` file.  Is there a way we can check to see if TBB_DIR is underneath the install directory, and if so, set it to, ie, `${_IMPORT_PREFIX}/lib/cmake/TBB"?

I'm actually a bit surprised that, if building with `TBBConfig.cmake`, cmake's automatic `XXXConfig.cmake` generation doesn't handle this more gracefully.  Is there a way to "tell" cmake that a dependency will be packaged along with the package being built?  Or a more standard / cmake-ish way to get portable `XXXConfig.cmake` files?

pmolodo:
To be fair this PR - this same issue is also happening with other parts of `pxrConfig.cmake`, like `MaterialX_DIR`... and this PR was presumably just following the already established pattern.  Unfortunately, this established pattern seems somewhat problematic...

jesschimein:
Filed as internal issue #USD-10931

Issue #3311, opened 2024-09-23
Improve `Tf.Status` python binding


roggiezhang-nv:
### Description of Issue

`Tf.Status` accepts a verbose argument in python https://github.com/PixarAnimationStudios/OpenUSD/blob/59992d2178afcebd89273759f2bddfe730e59aa8/pxr/base/tf/__init__.py#L181

This has no equivalent in C++, so the binding attempts to set a "null context" by passing empty strings for module, functio, and filename.

Those arguments are passed to `Tf_PythonCallContext`

https://github.com/PixarAnimationStudios/OpenUSD/blob/59992d2178afcebd89273759f2bddfe730e59aa8/pxr/base/tf/wrapStatus.cpp#L32

which joins the module & function name without checking if they are empty

https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/base/tf/pyCallContext.cpp#L42

resulting in a `TfCallContext` with a function "." instead of the intended "" (or ideally None/nullptr).

This is highly confusing on the c++ side (e.g. in a TfDiagnosticMgr Delegate implementation).

Hopefully this is a non-contentious fix to check for empty string before joining.

But I also wonder if it wouldn't be better to use TfCallContext::hide() instead so c++ callsites could check `diagnostic.GetContext().IsHidden()` rather than `diagnostic.GetSourceFunction().empty()`

### Steps to Reproduce
1.

roggiezhang-nv:
@nvmkuruc for vis.

jesschimein:
Filed as internal issue #USD-10179

nvmkuruc:
@jesschimein I believe this issue can be closed.  It was fixed with #3609.

jesschimein:
Thank you!  Closing :)

PR #3609, opened 2025-04-22
Provide quiet status specialization for python

nvmkuruc:
### Description of Change(s)
`Tf_PythonCallContext` assumes that a valid `moduleName` and `functionName` are available and joins them with `.`.  However, when `Tf.Status` is called with `verbose=False`, `moduleName` and `functionName` are both empty.

This yields a non-empty `fullName` of `.` when consumed by downstream diagnostic delegates that could be confusing.

This PR introduces a "quiet" version of `_Status` which bypasses `Tf_PyCallContext` and provides an explicitly empty `TfCallContext`.

### Fixes Issue(s)
#3311 

### Checklist

### DIFF
--- a/pxr/base/tf/__init__.py
+++ b/pxr/base/tf/__init__.py
@@ -190,7 +190,7 @@ def Status(msg, verbose=True):
         codeInfo = GetCodeLocation(framesUp=1)
         _Status(msg, codeInfo[0], codeInfo[1], codeInfo[2], codeInfo[3])
     else:
-        _Status(msg, "", "", "", 0)
+        _Status(msg)
 
 def RaiseCodingError(msg):
     """Raise a coding error to the Tf Diagnostic system."""
--- a/pxr/base/tf/wrapStatus.cpp
+++ b/pxr/base/tf/wrapStatus.cpp
@@ -54,6 +54,15 @@ _Status(string const &msg, string const& moduleName, string const& functionName,
         Post(msg);
 }
 
+static void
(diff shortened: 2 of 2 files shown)


jesschimein:
Filed as internal issue #USD-10934

pxr/base/tf/__init__.py
```diff
@@ -190,7 +190,7 @@ def Status(msg, verbose=True):
         codeInfo = GetCodeLocation(framesUp=1)
         _Status(msg, codeInfo[0], codeInfo[1], codeInfo[2], codeInfo[3])
     else:
-        _Status(msg, "", "", "", 0)
+        _Status(msg)
```
anwang2009, line 193: `msg` alone without default string arguments causes `TfCallContext` _function and _file to be initialized to `nullptr`, introducing a segfault here:

https://github.com/PixarAnimationStudios/OpenUSD/blob/ceab7e6541c02c9376a03de84bb1174e95e7264e/pxr/base/tf/diagnosticMgr.cpp#L639

We'll need more modifications to the diagnostic manager to get this change through, I think

nvmkuruc: I noticed that default initialized `TfCallContext` appears in the OpenUSD code base, but few things seem to check for `GetFile` or `GetFunction` for `nullptr`.  Would the simplest and most robust solution be to just default `_function`, `_file`, and `_prettyFunction` to an empty string?

PR #3616, opened 2025-05-03
Fix for usdview not creating a context prior to opening stage

sdobbs-blz:
### Description of Change(s)
This fixes the issue of usdview not creating it's context prior to reading the initial layer.
Prior to this fix the stage rootlayer would not respect the given context.

This issue is discussed here - https://forum.aousd.org/t/best-practices-opening-usdstage-versus-usdview/2367

I will submit another PR (shortly) for the usdcat/usdedit fix

### Checklist

### DIFF
--- a/pxr/usdImaging/usdviewq/appController.py
+++ b/pxr/usdImaging/usdviewq/appController.py
@@ -1223,13 +1223,15 @@ class AppController(QtCore.QObject):
             popMask = (None if populationMaskPaths is None else
                        Usd.StagePopulationMask())
 
-            # Open as a layer first to make sure its a valid file format
-            try:
-                layer = Sdf.Layer.FindOrOpen(usdFilePath)
-            except Tf.ErrorException as e:
-                sys.stderr.write(_GetFormattedError(
-                    [err.commentary.strip() for err in e.args]))
-                sys.exit(1)
+            ctx = self._resolverContextFn(usdFilePath)
+            with Ar.ResolverContextBinder(ctx):
+                # Open as a layer first to make sure its a valid file format
+                try:
+                    layer = Sdf.Layer.FindOrOpen(usdFilePath)
+                except Tf.ErrorException as e:
+                    sys.stderr.write(_GetFormattedError(
+                        [err.commentary.strip() for err in e.args]))
+                    sys.exit(1)
 
             if sessionFilePath:
                 try:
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-10966

sdobbs-blz:
Here's the associated usdcat fix - https://github.com/PixarAnimationStudios/OpenUSD/pull/3624

jesschimein:
Hi @sdobbs-blz !  Would you mind resending your CLA?  I can't find it on file as either individual or part of a corporate CLA 🙇‍♀️

sdobbs-blz:
I just requested that our (Blizzard) corporate CLA be resent, it was sent to "usd-cla@pixar.com" please let me know if you have any issues retrieving it, thanks!

jesschimein:
Thank you!  We've received it :)

pxr/usdImaging/usdviewq/appController.py
```diff
@@ -1223,13 +1223,15 @@ def _GetFormattedError(reasons=None):
...
-                sys.exit(1)
+            ctx = self._resolverContextFn(usdFilePath)
+            with Ar.ResolverContextBinder(ctx):
+                # Open as a layer first to make sure its a valid file format
+                try:
+                    layer = Sdf.Layer.FindOrOpen(usdFilePath)
+                except Tf.ErrorException as e:
+                    sys.stderr.write(_GetFormattedError(
+                        [err.commentary.strip() for err in e.args]))
+                    sys.exit(1)
```
sunyab, line 1230: We should bind the context with the call to `Resolve` at line 1213 as well. I'd suggest moving that block from 1213-1215 down here after the `try/except` and inside the `Ar.ResolverContextBinder` context.

```
    if not layer:
        if not Ar.GetResolver().Resolve(usdFilePath):
           sys.stderr.write(_GetFormattedError(["File not found"]))
           sys.exit(1)
```

That way, we just try to open the layer and if it succeeds then we've avoided an unnecessary call to `Resolve`. If opening the layer fails but we don't get an error exception, then we can check if it failed because the file wasn't found and report the same error message as before.

sunyab (changes requested):
Hi @sdobbs-blz , had one note for you about this change. Let me know what you think, thanks!

pxr/usdImaging/usdviewq/appController.py
sdobbs-blz: sounds good, I'll update.

sdobbs-blz: Updated, I assume we do not need to worry about layer possibly being null but the resolver somehow being able to resolve `usdFilePath`?

sunyab: Yep, I think that's fine -- we previously didn't cover that case either, and `Usd.Stage.Open` will itself issue an error if it's somehow passed an invalid layer.

sunyab (approved):
I've merged this internally and it'll be part of the next push to dev. Thanks again!

sdobbs-blz:
@sunyab  just an FYI, there's another PR with the same fix for usdcat - https://github.com/PixarAnimationStudios/OpenUSD/pull/3624

PR #2605, opened 2023-08-17
Fix compilation issues when building USD using the C++20 standard

mattyjams:
Hello! :)

This may be a ways out on the road map still since the VFX reference platform hasn't made it there yet, but I took an initial crack at addressing some C++20 compilation issues and wanted to offer this small handful of changes that worked for me and still maintains compatibility with the currently chosen standard (C++14).

With the changes here, I was able to build USD with the C++ standard set forward to C++20 using the following patch:

```
diff --git a/cmake/defaults/CXXDefaults.cmake b/cmake/defaults/CXXDefaults.cmake
index e970dfbe2..48c132ff7 100644
--- a/cmake/defaults/CXXDefaults.cmake
+++ b/cmake/defaults/CXXDefaults.cmake
@@ -25,8 +25,8 @@ include(CXXHelpers)
 include(Version)
 include(Options)

-# Require C++14
-set(CMAKE_CXX_STANDARD 14)
+# Require C++20
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
```

I tested on the following configurations:
- Windows 10, Visual Studio 2022 v17.6.5/msvc 19.36.32537, Python 3.9.13
- macOS 13.5, Apple clang 14.0.3, Python 3.9.6

I unfortunately don't have a Linux environment I can test this with at the moment. Interestingly, I did **not** need to incorporate #2242 (which aims to address #2183), so that particular issue may be specific to using gcc.

Here are some details about the changes:

---

The first three commits replace the use of `std::shared_ptr`'s `unique()` with `use_count() == 1` instead. `unique()` was deprecated in C++17 and removed in C++20, and should be equivalent to `use_count() == 1` which is available back to C++11. I broke them out into individual commits updating `hd`, `hdSt`, and `pcp`, but I'm also happy to create separate pull requests for each if that's preferable.

I did want to flag the note on [cppreference.com](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique) though:

> This function was deprecated in C++17 and removed in C++20 because use_count is only an approximation in multithreaded environment (see Notes in use_count).

I'm not sure whether that's of concern for the instances of `unique()` currently in place, but switching to `use_count()` at least shouldn't be any less safe.

---

[...]

PR #3619, opened 2025-05-05
Update stage.cpp

metaremi:
### Description of Change(s)

fix compiler error: 
invalid operands to binary expression ('const std::optional<SdfLayerHandle>' (aka 'const optional<TfWeakPtr<SdfLayer>>') and 'const std::optional<SdfLayerHandle>')

### Checklist

  no behavior change

### DIFF
--- a/pxr/usd/usd/stage.cpp
+++ b/pxr/usd/usd/stage.cpp
@@ -1196,9 +1196,10 @@ public:
         // the session layer or it matches, and we either don't care about the
         // path resolverContext or it matches.
         return _rootLayer == req->_rootLayer &&
-            (!_sessionLayer || (_sessionLayer == req->_sessionLayer)) &&
-            (!_pathResolverContext || (_pathResolverContext ==
-                                       req->_pathResolverContext));
+            (!_sessionLayer || 
+             (req->_sessionLayer && *_sessionLayer == *req->_sessionLayer)) &&
+            (!_pathResolverContext || 
+             (req->_pathResolverContext && *_pathResolverContext == *req->_pathResolverContext));
     }
     virtual UsdStageRefPtr Manufacture() {
         return UsdStage::_InstantiateStage(


JuneMeta:
Note - a signed CLA has been submitted.

FlorianZ:
Hi @metaremi. Can you provide details on how you're building OpenUSD? I am a little surprised to see this fail given that std::optional's operator== should just work here?!

metaremi:
its using internal tools that hide this level of detail :-)
but basically its clang 17 on linux. I'll ask around to see where I can get more details.

this code generate an error, but there are plenty of warnings following that show up with the -Wambiguous-reversed-operator option as well

metaremi:
I put the error/warning messages here: https://pastebin.com/37gUKTY5

jesschimein:
Filed as internal issue #USD-10974

metaremi:
I collected the compiler arguments here: https://pastebin.com/jENC1LVA

FlorianZ:
I was chatting with the team about this. We are bound by the VFX reference platform (vfxplatform.com) and are not building or validating against C++20, yet. However, we did get some contributions to fix C++20 build issues (#2605).

We also think the `std::optional` build error here might be a red herring, and that the real issue is with `operator==` overload resolution on `TfWeakPtr`. Out of curiosity, have you instead tried turning `TfWeakPtr::operator==` into a non-member friend function, to replace the member method it currently is? That might fix the problem at the root. Again, it's just a guess as we don't have easy access to a C++20 build.

metaremi:
I had tried this and this creates different errors.
I do agree the issue is deeper and would require more work to clean up for c++20, however this pull request is enough a bandaid to fix c++20 build for the library (this fixes and the changes merged in #2605), so it would be good to consider to merge this meanwhile.

nvmkuruc:
@FlorianZ Do you think the `std::optional` is necessary for `_sessionLayer`?  If I'm reading this correctly, the evaluation of the bool operator on `_sessionLayer` only checks that the optional has a value, not that the handle stored in the optional is valid.  Should `_sessionLayer` just be an `SdfLayerHandle` that defaults to an invalid handle?

FlorianZ:
@nvmkuruc the comment in that method is still accurate: 
> we either don't care about the session layer or it matches

So the optional is to disambiguate which case we're trying to match:

1. We don't care to match the session layer
2. We care to match, and want it to match to "invalid handle" on both sides

spiffmon:
> I had tried this and this creates different errors. I do agree the issue is deeper and would require more work to clean up for c++20, however this pull request is enough a bandaid to fix c++20 build for the library (this fixes and the changes merged in #2605), so it would be good to consider to merge this meanwhile.

@metaremi , would you possibly be able to post the different errors you get when turning `TfWeakPtr::operator==` into a non-member friend function?

PR #3637, opened 2025-05-16
Apple: Handle a possible nullptr when checking layer dependencies

dgovil:
### Description of Change(s)

When trying to package this asset, we experience a nullptr dereference while traversing the dependencies.
It's a simple check in the DFS to avoid the crash. Note that the localization will still faile, but it won't crash anymore.

[palmtree.usdc.zip](https://github.com/user-attachments/files/20258336/palmtree.usdc.zip)

I don't have a great asset to share that can be added to the unit tests, so I haven't written one unfortunately. If folks had a suggestion for something that would similarly trigger it, that would be great.

Thanks to Damien Courtois for discovering this and suggesting the workaround here.

### Checklist


jesschimein:
Filed as internal issue #USD-11015

PR #3640, opened 2025-05-20
Update info.cpp

metaremi:
### Description of Change(s)
Add support for long file paths on Windows

### Fixes Issue(s)
windows build fails if the path is too long

### Checklist


jesschimein:
Filed as internal issue #USD-11022

spiffmon:
@metaremi , plug claims that only `*` and `**` are given special treatment in its pattern, so we think `?` and `+` anywhere in the pattern might be problematic for all platforms. In which case, we suggest instead amending `_TranslateWildcardToRegex()` to add those chars to the cases that receive escaping, which it looks like should resolve the windows long-path issue as well?

spiffmon:
And actually the full list that should receive escaping is: `.[]+(){}?^$|`.  Let us know if you would rather we take care of this... we just likely wouldn't be testing specifically with Windows long-paths .

PR #3661, opened 2025-06-03
Autodesk: [hdSt] Persistent run-time-populated MaterialX codegen cache

erikaharrison-adsk:
### Description of Change(s)

This change extends the existing MaterialX shader registry (an in-memory cache) with a persistent on-disk cache as a performance optimization. If we've generated a MaterialX shader with a unique ID on a previous run of the application and encounter the same ID on a subsequent run, we avoid redoing the expensive codegen process and instead restore the necessary data from cached files.

The cache directory can be set with the `HDST_MTLX_CODEGEN_CACHE_DIR_PATH` environment variable or the `hdstMtlxCodegenCacheDirPath` render setting. It's left up to the application to manage the cache, such as:

- determine where to place the cache directory;
- cleaning the directory if the cached files become invalidated, e.g. due to upgrading the versions of USD and/or MaterialX;
- cleaning up cached files over a certain age or once a storage limit has been reached.

For reference, the optimization reduces the duration of `HdRenderIndex::SyncAll` for https://github.com/usd-wg/assets/tree/main/full_assets/OpenChessSet from about 550ms to about 250ms in our measurements.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/imaging/hd/instanceRegistry.h
+++ b/pxr/imaging/hd/instanceRegistry.h
@@ -20,32 +20,37 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+// Forward declarations
+template <typename VALUE>
+class HdInstanceRegistry;
+
+using HdInstanceKey = uint64_t;
+using HdInstanceRegistryMutex = std::mutex;
+using HdInstanceRegistryLock = std::unique_lock<HdInstanceRegistryMutex>;
 
 /// \class HdInstance
 ///
-/// This class is used as an interface to a shared instance in
-/// HdInstanceRegistry.
+/// This class is used as an interface to a shared instance stored in
+/// a REGISTRY object.
 ///
-/// KeyType is a hashable index type and VALUE is shared_ptr. In most use
-/// cases, the client computes a hash key which represents large bulky data
-/// (like topology, primvars) and registers it into HdInstanceRegistry. If the
-/// key has already been registered, the registry returns HdInstance and the
+/// HdInstanceKeyType is a hashable index type and VALUE is shared_ptr. In most
+/// use cases, the client computes a hash key which represents large bulky data
+/// (like topology, primvars) and registers it into REGISTRY. If the
+/// key has already been registered, the registry returns an HdInstance and the
 /// client can use GetValue() without setting/computing actual bulky data. If
 /// it doesn't exist, IsFirstInstance() returns true for the first instance
 /// and the client needs to populate an appropriate data VALUE into the
 /// instance by SetValue().
 ///
-/// In order to support concurrent access to HdInstanceRegistry, this
-/// class holds a lock to a mutex in HdInstanceRegistry. This lock will
+/// In order to support concurrent access to REGISTRY, this
+/// class holds a lock to a mutex in REGISTRY. This lock will
 /// be held until the instance of this interface class is destroyed.
 ///
-template <typename VALUE>
+template <typename VALUE, class REGISTRY = HdInstanceRegistry<VALUE>>
 class HdInstance {
 public:
-    typedef uint64_t KeyType;
     typedef VALUE ValueType;
-
-    typedef KeyType ID;
+    typedef HdInstanceKey ID;
 
     struct ValueHolder {
         ValueHolder(ValueType const & value = ValueType())
@@ -59,46 +64,44 @@ public:
         ValueType value;
         int recycleCounter;
     };
-    typedef tbb::concurrent_unordered_map<KeyType, ValueHolder> Dictionary;
-
-    typedef std::mutex RegistryMutex;
-    typedef std::unique_lock<RegistryMutex> RegistryLock;
 
     HdInstance() = delete;
 
     /// Construct an instance holding a registry lock, representing a value
     /// held in a registry container.
-    explicit HdInstance(KeyType const &key,
-                        ValueType const &value,
-                        RegistryLock &&registryLock,
-                        Dictionary *container)
+    explicit HdInstance(HdInstanceKey           key,
+                        ValueType const         &value,
+                        HdInstanceRegistryLock  &&registryLock,
+                        REGISTRY                *registry)
         : _key(key)
         , _value(value)
         , _registryLock(std::move(registryLock))
-        , _container(container)
+        , _registry(registry)
         , _isFirstInstance(!bool(_value))
     { }
 
     /// Construct an instance with no lock or registry container. This
     /// is used to present a consistent interface to clients in cases
     /// where shared resource registration is disabled.
-    explicit HdInstance(KeyType const &key)
+    explicit HdInstance(HdInstanceKey key)
         : _key(key)
         , _value(ValueType())
         , _registryLock()
-        , _container(nullptr)
+        , _registry(nullptr)
         , _isFirstInstance(!bool(_value))
     { }
 
     /// Returns the key
-    KeyType const &GetKey() const { return _key; }
+    HdInstanceKey GetKey() const { return _key; }
 
     /// Returns the value
     ValueType const &GetValue() const { return _value; }
 
-    /// Update the value in dictionary indexed by the key.
+    /// Update the value in the registry
     void SetValue(ValueType const &value) {
-        if (_container) (*_container)[_key] = ValueHolder(value);
+        if (_registry) {
+            _registry->_SetValue(_key, value);
+        }
         _value = value;
     }
 
@@ -108,45 +111,55 @@ public:
     }
 
 private:
-    KeyType       _key;
-    ValueType     _value;
-    RegistryLock  _registryLock;
-    Dictionary   *_container;
-    bool          _isFirstInstance;
+    HdInstanceKey           _key;
+    ValueType               _value;
+    HdInstanceRegistryLock  _registryLock;
+    REGISTRY                *_registry = nullptr;
+    bool                    _isFirstInstance;
 };
 
-/// \class HdInstanceRegistry
+/// \class HdInstanceRegistryBase
 ///
-/// HdInstanceRegistry is a dictionary container of HdInstance.
-/// This class is almost just a dictionary from key to value.
-/// For cleaning unused entries, it provides GarbageCollect() API.
-/// It sweeps all entries in the dictionary and erase unreferenced entries.
+/// HdInstanceRegistryBase is a dictionary container of HdInstance.
+/// This class is mostly just a dictionary from key to value, ensuring
+/// thread-safe access to the values.
+/// The DERIVED template parameter can be used in a Curiously Recurring
+/// Template Pattern to extend this class with a persistent cache backing the
+/// in-memory dictionary.
+/// For cleaning up unused entries, the GarbageCollect() API is provided.
+/// It sweeps all entries in the dictionary and erases unreferenced entries.
 /// When HdInstance::ValueType is shared_ptr, it is regarded as unreferenced
 /// if the shared_ptr is unique (use_count==1). Note that Key is not
 /// involved to determine the lifetime of entries.
 ///
-template <typename VALUE>
-class HdInstanceRegistry {
+template <typename VALUE, class DERIVED>
+class HdInstanceRegistryBase {
 public:
-    typedef HdInstance<VALUE> InstanceType;
-
-    HdInstanceRegistry() = default;
+    friend class HdInstance<VALUE, DERIVED>;
+    using InstanceType = HdInstance<VALUE, DERIVED>;
     
-    /// Copy constructor.  Need as HdInstanceRegistry is placed in a map
-    /// and mutex is not copy constructable, so can't use default
-    HdInstanceRegistry(const HdInstanceRegistry &other)
+    using Dictionary = tbb::concurrent_unordered_map<
+        HdInstanceKey,
+        typename InstanceType::ValueHolder
+    >;
+
+    HdInstanceRegistryBase() = default;
+
+    /// Copy constructor.  Need as HdInstanceRegistryBase is placed in a map
+    /// and mutex is not copy-constructible, so can't use default
+    HdInstanceRegistryBase(const HdInstanceRegistryBase &other)
         : _dictionary(other._dictionary)
-        , _registryMutex()  // mutex is not copied
+        , _mutex()  // mutex is not copied
     { }
 
-    /// Returns a shared instance for given key.
+    /// Returns a shared instance for a given key.
     InstanceType GetInstance(
-        typename InstanceType::KeyType const &key);
+        HdInstanceKey key);
 
     /// Returns a shared instance for a given key
     /// only if the key exists in the dictionary.
     InstanceType FindInstance(
-        typename InstanceType::KeyType const &key, bool *found);
+        HdInstanceKey key, bool *found);
 
     /// Removes unreferenced entries and returns the count
     /// of remaining entries. When recycleCount is greater than zero,
@@ -166,7 +179,7 @@ public:
 
     /// Returns a const iterator being/end of dictionary. Mainly used for
     /// resource auditing.
-    typedef typename InstanceType::Dictionary::const_iterator const_iterator;
+    typedef typename Dictionary::const_iterator const_iterator;
     const_iterator begin() const { return _dictionary.begin(); }
     const_iterator end() const { return _dictionary.end(); }
 
@@ -180,28 +193,93 @@ private:
         return value.use_count() == 1;
     }
 
-    typename InstanceType::Dictionary _dictionary;
-    typename InstanceType::RegistryMutex _registryMutex;
+    /// Store the value in the in-memory dictionary and write it to the 
+    /// persistent cache if implemented by the derived class.
+    void _SetValue(
+        HdInstanceKey key,
+        VALUE const& value);
 
-    HdInstanceRegistry &operator =(HdInstanceRegistry &) = delete;
+    /// Look a value up by the key in the in-memory dictionary. If the key is
+    /// not found there and the derived class implements a persistent cache,
+    /// try to load the value from there.
+    typename Dictionary::iterator _LookUp(
+        HdInstanceKey key);
+
+    Dictionary              _dictionary;
+    HdInstanceRegistryMutex _mutex;
+
+    HdInstanceRegistryBase &operator =(HdInstanceRegistryBase &) = delete;
+};
+
+/// \class HdInstanceRegistry
+///
+/// The default implementation of HdInstanceRegistryBase without persistent
+/// cache support.
+///
+template <typename VALUE>
+class HdInstanceRegistry
+    : public HdInstanceRegistryBase<VALUE, HdInstanceRegistry<VALUE>>
+{
+public:
+    friend class HdInstanceRegistryBase<VALUE, HdInstanceRegistry<VALUE>>;
+
+    void SaveToDisk(
+        HdInstanceKey key,
+        VALUE const& value)
+    {
+    }
+
+    VALUE LoadFromDisk(HdInstanceKey key)
+    {
+        return nullptr;
+    }
 };
 
 // ---------------------------------------------------------------------------
-// instance registry impl
+// Implementations
 
-template <typename VALUE>
-HdInstance<VALUE>
-HdInstanceRegistry<VALUE>::GetInstance(
-        typename HdInstance<VALUE>::KeyType const &key)
+template <typename VALUE, class DERIVED>
+typename HdInstanceRegistryBase<VALUE, DERIVED>::Dictionary::iterator
+HdInstanceRegistryBase<VALUE, DERIVED>::_LookUp(
+    HdInstanceKey key)
+{
+    typename Dictionary::iterator it = _dictionary.find(key);
+    if (it != _dictionary.end()) {
+        return it;
+    }
+
+    VALUE value = static_cast<DERIVED*>(this)->LoadFromDisk(key);
+    if (value) {
+        it = _dictionary.insert(
+            std::make_pair(key,
+                typename InstanceType::ValueHolder(value))).first;
+    }
+    return it;
+}
+
+template <typename VALUE, class DERIVED>
+void
+HdInstanceRegistryBase<VALUE, DERIVED>::_SetValue(
+    HdInstanceKey key,
+    VALUE const& value)
+{
+    _dictionary[key] = typename InstanceType::ValueHolder(value);
+    static_cast<DERIVED*>(this)->SaveToDisk(key, value);
+}
+
+template <typename VALUE, class DERIVED>
+HdInstance<VALUE, DERIVED>
+HdInstanceRegistryBase<VALUE, DERIVED>::GetInstance(
+    HdInstanceKey key)
 {
     HD_TRACE_FUNCTION();
     HF_MALLOC_TAG_FUNCTION();
 
     // Grab Registry lock
     // (and don't release it in this function, return it instead)
-    typename InstanceType::RegistryLock lock(_registryMutex);
+    HdInstanceRegistryLock lock(_mutex);
 
-    typename InstanceType::Dictionary::iterator it = _dictionary.find(key);
+    typename Dictionary::iterator it = _LookUp(key);
     if (it == _dictionary.end()) {
         // not found. create new one
         it = _dictionary.insert(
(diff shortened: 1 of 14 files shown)


pxr/imaging/hd/instanceRegistry.h
```diff
@@ -20,32 +20,37 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+// Forward declarations
+template <typename VALUE>
+class HdInstanceRegistry;
+
+using HdInstanceKey = uint64_t;
+using HdInstanceRegistryMutex = std::mutex;
+using HdInstanceRegistryLock = std::unique_lock<HdInstanceRegistryMutex>;
```
ppenenko, line 29: Previously, these were typedefs in the templated class definition of `HdInstance` but they never depended on template parameters. They've also been referenced in the `HdInstanceRegistry` implementation below, where they had to be namespaced with `HdInstance::`. I'm extending the `HdInstanceRegistry` implementation in order to optionally support persistence, and these verbose typedefs were hurting the readability in that code, so I'm making them standalone.

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -59,46 +64,44 @@ class HdInstance {
...
     /// Construct an instance holding a registry lock, representing a value
     /// held in a registry container.
-    explicit HdInstance(KeyType const &key,
-                        ValueType const &value,
-                        RegistryLock &&registryLock,
-                        Dictionary *container)
+    explicit HdInstance(HdInstanceKey           key,
+                        ValueType const         &value,
+                        HdInstanceRegistryLock  &&registryLock,
+                        REGISTRY                *registry)
```
ppenenko, line 75: Previously, it was enough for the instance to hold a pointer to the hash map container owned by the registry, because it would just directly put values in it. But now the registry can implement a custom persistence mechanism, so we need the instance to hold a pointer to the registry and call its methods instead.

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -59,46 +64,44 @@ class HdInstance {
...
-        , _container(container)
+        , _registry(registry)
         , _isFirstInstance(!bool(_value))
     { }
 
     /// Construct an instance with no lock or registry container. This
     /// is used to present a consistent interface to clients in cases
     /// where shared resource registration is disabled.
-    explicit HdInstance(KeyType const &key)
+    explicit HdInstance(HdInstanceKey key)
```
ppenenko, line 86: We know that the key is always just an integer, so it's better to pass it around by value.

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -108,45 +111,55 @@ class HdInstance {
...
-/// HdInstanceRegistry is a dictionary container of HdInstance.
-/// This class is almost just a dictionary from key to value.
-/// For cleaning unused entries, it provides GarbageCollect() API.
-/// It sweeps all entries in the dictionary and erase unreferenced entries.
+/// HdInstanceRegistryBase is a dictionary container of HdInstance.
+/// This class is mostly just a dictionary from key to value, ensuring
+/// thread-safe access to the values.
+/// The DERIVED template parameter can be used in a Curiously Recurring
+/// Template Pattern to extend this class with a persistent cache backing the
+/// in-memory dictionary.
```
ppenenko, line 128: BTW, this opens up the possibility to use the same pattern for other types of resources.

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -108,45 +111,55 @@ class HdInstance {
...
-    HdInstanceRegistry(const HdInstanceRegistry &other)
+    using Dictionary = tbb::concurrent_unordered_map<
+        HdInstanceKey,
+        typename InstanceType::ValueHolder
+    >;
+
+    HdInstanceRegistryBase() = default;
+
+    /// Copy constructor.  Need as HdInstanceRegistryBase is placed in a map
+    /// and mutex is not copy-constructible, so can't use default
```
ppenenko, line 149: This comment doesn't seem to be correct: these objects are never stored by value in the USD codebase, so this constructor shouldn't be necessary and this class should comply with the [rule of zero](https://en.cppreference.com/w/cpp/language/rule_of_three) (right now, it doesn't: the copy operator is deleted below but it's copy-constructible).

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -180,73 +193,142 @@ class HdInstanceRegistry {
...
+typename HdInstanceRegistryBase<VALUE, DERIVED>::Dictionary::iterator
+HdInstanceRegistryBase<VALUE, DERIVED>::_LookUp(
+    HdInstanceKey key)
+{
+    typename Dictionary::iterator it = _dictionary.find(key);
+    if (it != _dictionary.end()) {
+        return it;
+    }
+
+    VALUE value = static_cast<DERIVED*>(this)->LoadFromDisk(key);
```
ppenenko, line 251: Loading from disk, if implemented by the derived class.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ if (${PXR_ENABLE_MATERIALX_SUPPORT})
     list(APPEND optionalPrivateClasses
          materialXFilter
          materialXShaderGen
+         materialXShaderRegistry
```
ppenenko, line 29: This is to make the header visible to the tests.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -626,6 +617,28 @@ pxr_register_test(testHdStQualifiers
...
+    SRC testenv/testHdStMaterialXShaderGen
+    DEST testHdStMaterialXShaderGen
+)
+pxr_register_test(testHdStMaterialXShaderGen_testCacheSerialization
+    COMMAND "${CMAKE_INSTALL_PREFIX}/tests/testHdStMaterialXShaderGen --testCacheSerialization"
+    EXPECTED_RETURN_CODE 0
+    STDOUT_REDIRECT shadergen_testCacheSerialization.out
+    DIFF_COMPARE shadergen_testCacheSerialization.out
+    TESTENV testHdStMaterialXShaderGen
+)
```
ppenenko, line 639: A new test that tests if metadata is serialized correctly by the new MaterialX codegen cache. It reuses an existing test executable `testHdStMaterialXShaderGen` with a new command-line argument.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -1121,7 +1121,7 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
 
 #ifdef PXR_MATERIALX_SUPPORT_ENABLED
         if (!isVolume) {
-            _materialXGfx = HdSt_ApplyMaterialXFilter(&surfaceNetwork, materialId,
+            _materialXCodegenResult = HdSt_ApplyMaterialXFilter(&surfaceNetwork, materialId,
```
ppenenko, line 1124: This used to return a shared pointer to a `MaterialX::Shader` object. Unfortunately, a `MaterialX::Shader` is not serializable, and source code is only one part of its state. So, to make caching feasible, I extract all the data that the downstream pipeline needs from `MaterialX::Shader`, encapsulate it in an object called "codegen result" and cache that to disk.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1109,126 +1108,56 @@ _AddMaterialXParams(
         }
     }
 
-    const mx::ShaderStage& pxlStage = glslfxShader->getStage(mx::Stage::PIXEL);
-    const auto& paramsBlock = pxlStage.getUniformBlock(mx::HW::PUBLIC_UNIFORMS);
-    for (size_t i = 0; i < paramsBlock.size(); ++i) {
+    HdSt_MaterialParamVector const& fallbackParams =
+        mxCodegenResult.GetFallbackParams();
```
ppenenko, line 1112: Fallback parameters are one necessary part of the codegen result.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1109,126 +1108,56 @@ _AddMaterialXParams(
...
 
-        // MaterialX parameter Information
-        const auto* variable = paramsBlock[i];
-        const auto varType = HdStMaterialXHelpers::GetMxTypeDesc(variable);
+    for (HdSt_MaterialParam const& fallbackParam : fallbackParams) {
```
ppenenko, line 1114: The old code was populating the fallback parameters from the MaterialX shader.

In the new code, they are retrieved from MaterialX in the codegen result's constructor if codegen takes place at run time, or read from disk if the codegen happened on a previous run and was cached, and stored in the codegen result object.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1109,126 +1108,56 @@ _AddMaterialXParams(
...
         // If it was not found in the mapping use the value from the MaterialX
         // variable to get the value. 
-        // Note that if the network was upgraded in Hdmtlx, and the node's 
+        // Note that if the network was upgraded in HdMtlx, and the node's 
         // parameter names changed they will not be found through the above 
         // mapping and instead need to be found from the variables in the 
-        // MaterialX glslfxShader. 
-        else {
-            std::string separator;
-            const auto varValue = variable->getValue();
```
ppenenko, line 1143: All of this has moved to the codegen result's constructor.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1109,126 +1108,56 @@ _AddMaterialXParams(
...
-                }
-            }
-        }
+        // MaterialX shader.
+    }
 
-        if (!param.fallbackValue.IsEmpty()) {
-            materialParams->push_back(std::move(param));
-        }
+    for (TfToken textureParamName : mxCodegenResult.GetTextureParams()) {
```
ppenenko, line 1132: Similarly, texture parameters are cached in the codegen result.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1264,8 +1193,10 @@ _GenerateMaterialXShader(
     mxHdInfo.bindlessTexturesEnabled = bindlessTexturesEnabled;
 
     // Generate the glslfx source code from the mtlxDoc
-    return HdSt_GenMaterialXShader(
+    mx::ShaderPtr mxShader = HdSt_GenMaterialXShader(
         mtlxDoc, stdLibraries, searchPaths, mxHdInfo, apiName);
+
+    return std::make_shared<HdSt_MaterialXCodegenResult>(*mxShader);
```
ppenenko, line 1199: Creating a codegen result from the MaterialX Shader, and destroying the latter because we've already extracted everything we will ever need from it.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1296,8 +1227,37 @@ _IsTopologicalShader(TfToken const& nodeId)
...
+    // A VtValue may store a TfToken, in which case we need to convert it to a
+    // string in order to avoid the same pitfall as above
+    void _AppendPersistentHash(Tf_HashState& hashState, VtValue const& vtValue)
+    {
+        if (vtValue.IsHolding<TfToken>()) {
+            TfHashAppend(hashState, vtValue.Get<TfToken>().GetString());
+        } else {
+            TfHashAppend(hashState, vtValue.GetHash());
+        }
+    }
```
ppenenko, line 1258: Without these workarounds, the cache always generated new, random hashes, and could never find an existing file in the cache directory.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1453,6 +1429,13 @@ HdSt_ApplyMaterialXFilter(
 
     TRACE_FUNCTION_SCOPE("ApplyMaterialXFilter: Found Mtlx Node.")
 
+    HdSt_MaterialXShaderRegistry* const materialXShaderRegistry =
+        resourceRegistry->GetMaterialXShaderRegistry();
```
ppenenko, line 1433: This is a specialized instance registry that supports a persistent cache. To minimize include dependencies, I expose it to clients of `HdStResourceRegistry` by pointer.

pxr/imaging/hdSt/materialXShaderRegistry.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+#include <fstream>
+
+namespace mx = MaterialX;
+namespace fs = std::filesystem;
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+TF_DEFINE_ENV_SETTING(HDST_MTLX_CODEGEN_CACHE_DIR_PATH, "",
+    "Path to the directory of the persistent MaterialX codegen cache");
```
ppenenko, line 28: Environment variable which can be used to set the path to the directory where the cached files are written. It serves as the default value for an `HdSt` render setting.

pxr/imaging/hdSt/materialXShaderRegistry.cpp
```diff
@@ -0,0 +1,542 @@
...
+    static const _DeserializerMap _deserializerMap{
+        {
+            "bool",
+            {
+                false,
+                [](std::istringstream& issValue) -> VtValue
+                {
+                    bool val = false;
+                    issValue >> val;
+                    return VtValue(val);
```
ppenenko, line 135: These implementations mostly moved here from `materialXFilter.cpp`

pxr/imaging/hdSt/materialXShaderRegistry.cpp
```diff
@@ -0,0 +1,542 @@
...
+    for (HdSt_MaterialParam const& param : _fallbackParams) {
+
+        std::ostringstream osValue;
+        osValue.imbue(std::locale::classic());
+        osValue << param.fallbackValue;
+        
+        JsObject jsParam{
+            { _tokens->name,    JsValue(param.name) },
+            { _tokens->type,    JsValue(param.fallbackValue.GetTypeName()) },
+            { _tokens->value,   JsValue(osValue.str()) } };
```
ppenenko, line 494: Serializing parameters to the JSON file.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -205,7 +219,13 @@ HdStRenderDelegate::HdStRenderDelegate(HdRenderSettingsMap const& settingsMap)
         HdRenderSettingDescriptor{
             "Dome light camera visibility",
             HdRenderSettingsTokens->domeLightCameraVisibility,
-            VtValue(true) }
+            VtValue(true) },
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+        HdRenderSettingDescriptor{
+            "Path to the directory of the persistent MaterialX codegen cache",
+            HdStRenderSettingsTokens->hdstMtlxCodegenCacheDirPath,
+            VtValue(HdSt_MaterialXShaderRegistry::GetCacheDirPathEnvSetting()) }
```
ppenenko, line 227: The cache directory exposed as a Storm render setting. The env var supplies the default value.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -258,8 +278,19 @@ HdStRenderDelegate::SetDrivers(HdDriverVector const& drivers)
...
+        HdStRenderSettingsTokens->hdstMtlxCodegenCacheDirPath,
+        HdSt_MaterialXShaderRegistry::GetCacheDirPathEnvSetting());
+#endif
+
     _resourceRegistry =
-        _HgiToResourceRegistryMap::GetInstance().GetOrCreateRegistry(_hgi);
+        _HgiToResourceRegistryMap::GetInstance().GetOrCreateRegistry(
+            _hgi
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+            , cacheDirPath.c_str()
```
ppenenko, line 291: Passing the render setting to the registry.

pxr/imaging/hdSt/resourceRegistry.cpp
```diff
@@ -122,6 +128,10 @@ HdStResourceRegistry::HdStResourceRegistry(Hgi * const hgi)
     // default aggregation strategy for single buffers (for nested instancer)
     , _singleAggregationStrategy(
         std::make_unique<HdStVBOSimpleMemoryManager>(this))
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+    , _materialXShaderRegistry(
+        std::make_unique<HdSt_MaterialXShaderRegistry>(mtlxCacheDirPath))
```
ppenenko, line 133: The MaterialX shader registry is now owned by unique pointer to simplify include dependencies.

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen.cpp
```diff
@@ -141,16 +147,123 @@ void TestShaderGen(
...
+
+    auto addFallbackParam = [&fallbackParams](
+        const char* name, VtValue value) -> void
+    {
+        fallbackParams.emplace_back(
+            HdSt_MaterialParam::ParamTypeFallback,
+            TfToken(name), value);
+    };
+
+    addFallbackParam("BoolParamTrue", VtValue(true));
```
ppenenko, line 163: Testing the serialization/deserialization of all supported data types.

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen.cpp
```diff
@@ -141,16 +147,123 @@ void TestShaderGen(
...
+            std::numeric_limits<int>::min(),
+            std::numeric_limits<int>::max())));
+
+    HdSt_MaterialXCodegenResult codegenResult0(
+        "",
+        std::move(fallbackParams),
+        std::move(textureParams));
+
+    std::stringstream ss;
+    codegenResult0.SaveMetadata(ss);
```
ppenenko, line 208: Serializing the codegen result we've just populated to a string.

pxr/imaging/hdSt/testenv/testHdStMaterialXShaderGen.cpp
```diff
@@ -141,16 +147,123 @@ void TestShaderGen(
...
+    ss.seekg(0);
+
+    JsValue jsMetadata = JsParseStream(ss);
+
+    HdSt_MaterialXCodegenResult codegenResult1(
```
ppenenko, line 213: Deserializing into a new codegen result object.

pxr/imaging/hdSt/tokens.h
```diff
@@ -82,12 +82,19 @@ PXR_NAMESPACE_OPEN_SCOPE
 #define HDST_RENDER_BUFFER_TOKENS                       \
     ((stormMsaaSampleCount, "storm:msaaSampleCount"))
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+#define HDST_MTLX_CODEGEN_CACHE_DIR_PATH_TOKEN (hdstMtlxCodegenCacheDirPath)
```
ppenenko, line 86: A token for the render setting's name.

jesschimein:
Filed as internal issue #USD-11070

pxr/imaging/hd/instanceRegistry.h
```diff
@@ -180,73 +193,142 @@ class HdInstanceRegistry {
...
+/// cache support.
+///
+template <typename VALUE>
+class HdInstanceRegistry
+    : public HdInstanceRegistryBase<VALUE, HdInstanceRegistry<VALUE>>
+{
+public:
+    friend class HdInstanceRegistryBase<VALUE, HdInstanceRegistry<VALUE>>;
+
+    void SaveToDisk(
```
tcauchois, line 226: We're anticipating the need for a more sophisticated management layer for the cache, and think that should live outside of hdSt.  We have a big requirements list for it (including stuff like location on disk, size limit, eviction policy, security concerns, fingerprinting, hooks for testing); we don't expect you to hit all of that, but we also don't think the instance registry is a good place to park that complexity, since it's just a bit of memoization code.

If the control flow works out (which it looks like it does) I think the idea of the instance registry calling out to the persistent cache seems neat, though!  But I wonder if there's a way to get that extensibility without all of the recurrent templates.

pxr/imaging/hdSt/materialXFilter.cpp
tcauchois: Awesome! :)

pxr/imaging/hdSt/materialXFilter.cpp
tcauchois: I want to dig around internally and see if we have a function that does this already.  If not, I may try to promote something like this to pxr/base.  I think there are definitely some other cases where the TfHash isn't what you want for a fingerprint, though I'm not sure whether you'd hit them in a material network traversal.

pxr/imaging/hdSt/materialXShaderRegistry.cpp
tcauchois: So as my guiding light for this kind of feature, I'm using the NVidia GL shader disk cache, which I think gets configured by environment variable or the magic nvidia control panel.

From Autodesk's perspective, what kind of control interface would you like for the cache?  Env var?  Something in C++ so you can put it in a settings panel?  Plugin-based so people can configure weird site-specific multilevel caches or something?  Curious to hear what the long term plans for something like this would be.

tcauchois (commented):
Good stuff!  I think the MaterialXCodeGenResult refactor is landable now, but left some comments about the disk-backed instance registry.

pxr/imaging/hdSt/materialXShaderRegistry.cpp
ppenenko: Ultimately, I would defer these decisions up to the application/product developers. But I think this cache can't be system-level like the Nvidia one, because the codegen results depend on the specific implementation, including the MaterialX and USD versions. So, the cache directory should be local to a specific version of the application, and should probably be managed by its install process out of the box.

Great point about studios potentially looking to manage this directory in their pipelines - I think it would definitely be doable with something like Rez configuring a directory based on the versions of Maya, USD and MaterialX. But I'm speculating.

ppenenko: In terms of integration into the application, I believe this env setting is enough for the app to pass the path to Storm, just like it's currently enough to expose this path to usdview via the data-driven settings mechanism.

At the same time, for functionality like clearing the cache, maintaining a storage quota, LRU, invalidation on app updates, the file system itself could serve as the API surface. These actions could happen at different times than storing/restoring, e.g. on application startup, shutdown on installation/deployment. In a studio setting, these may even be a responsibility of a pipeline script.

pxr/imaging/hd/instanceRegistry.h
ppenenko: I understand that there's a list of potential features to be added to this mechanism. As I argued earlier, a few of them don't require integration with the store/restore APIs and could instead separately manipulate the file system. Others, admittedly, might require such an integration, e.g. a version could be passed to the cache together with the hash. So I agree that the disk cache API should be configurable.

At the same time, I believe that this PR is proof that the instance registry API is a natural design choice from the perspective of client code. The instance registry backed by an optional disk cache acts as a write-though cache, and the client code isn't exposed to the API of the disk cache. I really don't see an alternative here.

I agree that the current templated design with CRTP is heavy. Virtual interfaces would be an alternative, but I went with CRTP initially because the existing instance registry was already templated, and for good reason - it's parameterized by the value type. Mixing virtual interfaces with templates would be complicated as well - we'd either need to make the abstract bases templated as well or to derive value classes from an `ISerializeableValue` base. That being said, I think refactoring this with composition instead of inheritance would be nice in the future.

Regarding implementing the disk cache outside of `HdSt` - I agree completely. In fact, as I mentioned in our meetings, I had to move the instance registry itself to `Tf` for my follow-up SPIR-V cache work, because it needed to be accessible from `HgiVulkan`. It was perfectly feasible, because this code doesn't really have dependencies.

So to sum it up, while I agree that the current implementation can and should be extended, I hope it can be merged in the near future to:
1. Remove a significant merging overhead on our end.
2. Benefit some of our products near-term.
3. Open the way for a follow-up SPIR-V cache PR, which we both agreed was a major optimization for Vulkan.

In other words, I hope we can iterate on this feature.

pxr/imaging/hdSt/materialXFilter.cpp
ppenenko: Hey @tcauchois, I just realized that the pointer hashing is not only problematic for the persistent cache in this PR.

While testing an integration of #3847, I noticed that some of material hashes differed between the early parallel codegen and `HdStMaterial::Sync` and tracked it down to some of the `TfToken`s having identical string values but different addresses in those two phases.

It turns our that while static `TfToken`s are "immortal" (persist for the lifetime of the process), the dynamically-creates ones are refcounted and can be destroyed. So in #3847, the same string can be interned once, hashed by pointer, deallocated, then interned again at a different address and hashed again by pointer, producing a different hash.

So I would argue that it's a global footgun which should ideally be fixed on the `TfToken` side rather than in material hashing specifically.

PR #3710, opened 2025-07-07
APPLE: Make codesigning improvements

dgovil:
### Description of Change(s)

This PR overhauls the current code signing setup with the following changes:

1. Developers can now pass in code signing identifiers directly to the build_usd.py script.
2. Codesigning identifiers are now passed to CMake for Xcode targets if they want to sign anything.
3. The code signing identifier is calculated less often
4. Codesigning now supports signing frameworks
5. Codesigning supports checking previous signing status and only re-signing as needed. Note: Since USD's cmake always installs all targets again, they're always re-signed. Future work might help optimize this, but at the least we're not re-signing every dependency.
6. Codesigning is validated earlier to prevent running through a whole build just to find code signing fails

This makes code signing much more reliable and faster overall. 

This work is needed to enable properly building OpenUSD as a framework

### Checklist

### DIFF
--- a/build_scripts/apple_utils.py
+++ b/build_scripts/apple_utils.py
@@ -181,11 +181,8 @@ def GetCodeSigningIdentifiers() -> Dict[str, str]:
     XcodeVersion = GetXcodeVersion()[0]
     codeSignIDs = _GetCodeSignStringFromTerminal()
 
-    if not codeSignIDs:
-        return {"-": None}
-
     identifiers = {}
-    for codeSignID in codeSignIDs.splitlines():
+    for codeSignID in (codeSignIDs or "").splitlines():
         if "CSSMERR_TP_CERT_REVOKED" in codeSignID:
             continue
         if ")" not in codeSignID:
@@ -200,9 +197,7 @@ def GetCodeSigningIdentifiers() -> Dict[str, str]:
 
             identifiers[identifier] = identifier_hash
 
-    if not identifiers:
-        raise RuntimeError("Could not find a valid codesigning ID. Try re-logging into your Xcode developer account.")
-
+    identifiers["-"] = None
     return identifiers
 
 
@@ -225,6 +220,8 @@ def GetDevelopmentTeamID(identifier=None):
 
     if not identifier:
         identifier = GetCodeSignID()
+    if identifier == "-":
+        return None
 
     identifier_hash = GetCodeSigningIdentifiers().get(identifier)
     if not identifier_hash:
--- a/build_scripts/apple_utils.py
+++ b/build_scripts/apple_utils.py
@@ -365,6 +365,9 @@ def ConfigureCMakeExtraArgs(context, args:List[str]) -> List[str]:
     if TargetEmbeddedOS(context):
         system_name = context.buildTarget
 
+    if context.macOSCodesign:
+        args.append(f"-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY={context.macOSCodesign}")
+
     if system_name:
         args.append(f"-DCMAKE_SYSTEM_NAME={system_name}")
         args.append(f"-DCMAKE_OSX_SYSROOT={GetSDKRoot(context)}")
--- a/build_scripts/apple_utils.py
+++ b/build_scripts/apple_utils.py
@@ -4,7 +4,6 @@
 # Licensed under the terms set forth in the LICENSE.txt file available at
 # https://openusd.org/license.
 #
-
 # Utilities for managing Apple OS build concerns.
 #
 # NOTE: This file and its contents may change significantly as we continue
@@ -19,6 +18,7 @@ import re
 import platform
 import shlex
 import subprocess
+import glob
 from typing import Optional, List, Dict
 
 TARGET_NATIVE = "native"
@@ -278,7 +278,7 @@ def Codesign(install_path, identifier=None, force=False, verbose_output=False) -
     if not MacOS():
         return False
 
-    codeSignID = identifier or GetCodeSignID()
+    identifier = identifier or GetCodeSignID()
 
     if verbose_output:
         global devout
@@ -286,22 +286,35 @@ def Codesign(install_path, identifier=None, force=False, verbose_output=False) -
         print(f"Code-signing files in {install_path} with {identifier}", file=devout)
 
     try:
-        team_identifier = GetDevelopmentTeamID(codeSignID)
+        team_identifier = GetDevelopmentTeamID(identifier)
     except:
+        if verbose_output:
+            print("Could not get team_identifier")
         team_identifier = None
 
-    for root, dirs, files in os.walk(install_path, topdown=True):
-        for f in files:
+    codesignPaths = [
+        os.path.join(install_path, 'lib'),
+        os.path.join(install_path, 'plugin'),
+        os.path.join(install_path, 'share/usd'),
+        os.path.join(install_path, "frameworks")
+    ]
+        
+    for basePath in codesignPaths:
+        if not os.path.exists(basePath):
+            continue
 
-            _, ext = os.path.splitext(f)
-            if ext in (".dylib", ".so"):
-                path = os.path.join(root, f)
-                result = CodesignPath(path, identifier, team_identifier=team_identifier, force=force, is_framework=False)
-                if verbose_output:
-                    if result:
-                        print(f"Code-signed binary: {path}")
-                    else:
-                        print(f"Did not code-sign binary: {path}")
+        for root, dirs, files in os.walk(basePath, topdown=True):
+            for f in files:
+
+                _, ext = os.path.splitext(f)
+                if ext in (".dylib", ".so"):
+                    path = os.path.join(root, f)
+                    result = CodesignPath(path, identifier, team_identifier=team_identifier, force=force, is_framework=False)
+                    if verbose_output:
+                        if result:
+                            print(f"Code-signed binary: {path}")
+                        else:
+                            print(f"Did not code-sign binary: {path}")
 
         # Bit annoying to have to do this twice, but seems the fastest way to skip traversing frameworks
         frameworks = [d for d in dirs if d.endswith(".framework")]
--- a/build_scripts/apple_utils.py
+++ b/build_scripts/apple_utils.py
@@ -215,7 +215,7 @@ def GetCodeSignID() -> str:
 
 
 def GetDevelopmentTeamID(identifier=None):
-    if os.environ.get("DEVELOPMENT_TEAM"):
+    if "DEVELOPMENT_TEAM" in os.environ:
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-11198

build_scripts/apple_utils.py
```diff
@@ -163,36 +162,159 @@ def ExtractFilesRecursive(path, cond):
...
+    if not force:
+        codesigning_info = GetCommandOutput(["codesign", "-vd", path])
+        if not codesigning_info:
+            resign = True
+        else:
+            for line in codesigning_info.splitlines():
+                if line.startswith("TeamIdentifier="):
+                    current_team_identifier = line.split("=")[-1]
+                    if not team_identifier and "not set" in team_identifier:
+                        break
```
sunyab: It seems like this block of code doesn't handle the case where `team_identifier` is `None`?

There's a condition at line 256 that looks like it was meant to handle this, but it seems like it will never be met -- If the first clause is true, then the string must be empty so "not set" would never be found. Was that meant to be an `or` instead of an `and`?

build_scripts/apple_utils.py
```diff
@@ -163,36 +162,159 @@ def ExtractFilesRecursive(path, cond):
...
+                    elif team_identifier == current_team_identifier:
+                        break
+            else:
+                resign = True
```
sunyab, line 263: Can this all be simplified to something like:
```
if (not codesigning_info or
    not team_identifier or "not set" in team_identifier or
    f"TeamIdentifier={team_identifier}" not in codesigning_info):
    resign = True
```
?

build_scripts/apple_utils.py
```diff
@@ -240,6 +362,9 @@ def ConfigureCMakeExtraArgs(context, args:List[str]) -> List[str]:
     if TargetEmbeddedOS(context):
         system_name = context.buildTarget
 
+    if context.macOSCodesign:
+        args.append(f"-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY={context.macOSCodesign}")
+
```
sunyab: Is this necessary? IIUC, all of the code signing is happening outside of cmake.

build_scripts/build_usd.py
```diff
@@ -2002,6 +2002,7 @@ def InstallUSD(context, force, buildArgs):
                        default=codesignDefault, action="store_true",
                        help=("Enable code signing for macOS builds "
                              "(defaults to enabled on Apple Silicon)"))
+    group.add_argument("--codesign-id", dest="macos_codesign_id", type=str)
```
sunyab: Need documentation for this argument

build_scripts/apple_utils.py
```diff
@@ -163,36 +162,159 @@ def ExtractFilesRecursive(path, cond):
...
+            return env_signing_id
+        raise RuntimeError(
+            f"Could not find environment specified identifier {env_signing_id} in registered code signing identifiers")
+
+    return list(GetCodeSigningIdentifiers().keys())[0]
+
+
+def GetDevelopmentTeamID(identifier=None):
+    if os.environ.get("DEVELOPMENT_TEAM"):
+        return os.environ.get("DEVELOPMENT_TEAM")
```
sunyab:
```suggestion
    if "DEVELOPMENT_TEAM" in os.environ:
        return os.environ.get("DEVELOPMENT_TEAM")
```

sunyab: Equivalent but more Pythonic?

build_scripts/apple_utils.py
```diff
@@ -163,36 +162,159 @@ def ExtractFilesRecursive(path, cond):
...
+
+    if not resign:
+        return False
+
+    # Frameworks need to be signed with different parameters than loose binaries
+    if path.endswith(".framework"):
+        subprocess.check_output(
+            ["codesign", "--force", "--sign", identifier, "--generate-entitlement-der", "--verbose", path])
+    else:
+        subprocess.check_call(["codesign", "--force", "--sign", identifier, path], stdout=devout, stderr=devout)
```
sunyab: Any reason the framework codepath uses `check_output` while the other path uses `check_call`?

build_scripts/apple_utils.py
```diff
@@ -163,36 +162,159 @@ def ExtractFilesRecursive(path, cond):
...
+                    elif team_identifier == current_team_identifier:
+                        break
+            else:
+                resign = True
+
+    if not resign:
+        return False
+
+    # Frameworks need to be signed with different parameters than loose binaries
+    if path.endswith(".framework"):
```
sunyab: I'd suggest passing in an "isFramework" argument instead of re-parsing the path so that the logic of determining whether a path is a framework is centralized in the caller.

sunyab (changes requested):
Hi @dgovil , left some notes from a first pass review. Thanks!

dgovil:
Thanks, @sunyab !
I'll take a pass next week at resolving these

build_scripts/apple_utils.py
dgovil: Good catch. Indeed was meant to be an or. Fixed

build_scripts/apple_utils.py
dgovil: It's not currently. I was looking into moving the code signing into CMake itself , but didn't include it in the PR.
I'll remove this for now, and introduce it later if I revisit moving it into CMake.

The advantage there is that you only sign the dylibs that are rebuilt , so cuts down on build iterations.

build_scripts/apple_utils.py
dgovil: Ah I actually meant to use current_team_identifier in this line, so can't collapse it down unfortunately. I'll make the change though. Good callout

build_scripts/apple_utils.py
dgovil: Switched over. IDK if USD still supports Python 3.7  but would also be a good place for a walrus operator whenever that happens (introduced in 3.8)

build_scripts/apple_utils.py
dgovil: Nah, just force of habit for me to use check_output while working on stuff

build_scripts/apple_utils.py
dgovil: Yeah good call. Fixed

dgovil:
Thanks, @sunyab 
I pushed an update that also rebases on top of dev. 

Appreciate the notes. Hope you have a good thanksgiving break.

sunyab:
Hey @dgovil, I ran into an error when testing this change earlier. I was testing the process we follow internally when doing QA for USD releases:

- Create and activate a Python venv
- Install the necessary Python modules into the venv (PySide6 and PyOpenGL)
- Build USD into the venv via build_usd.py

This was the result:

```
sunya@usd-mac-08 ~ --> /usr/local/bin/python3 -m venv codesign_test    
sunya@usd-mac-08 ~ --> source codesign_test/bin/activate
(codesign_test) sunya@usd-mac-08 ~ --> python -m pip install PyOpenGL PySide6==6.3.1
(codesign_test) sunya@usd-mac-08 ~ --> python OpenUSD/build_scripts/build_usd.py --build-target universal --tests codesign_test 

<snip>

STATUS: Installing USD...
/Users/sunya/codesign_test/lib/python3.9/site-packages/PySide6/Qt/lib/QtWebEngineCore.framework: code object is not signed at all
In subcomponent: /Users/sunya/codesign_test/lib/python3.9/site-packages/PySide6/Qt/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app
  File "/Users/sunya/dev-cmake/build_scripts/build_usd.py", line 2982, in <module>
    apple_utils.Codesign(context.usdInstDir,
  File "/Users/sunya/dev-cmake/build_scripts/apple_utils.py", line 312, in Codesign
    result = CodesignPath(path, identifier, team_identifier=team_identifier, force=force, is_framework=True)
  File "/Users/sunya/dev-cmake/build_scripts/apple_utils.py", line 270, in CodesignPath
    subprocess.check_call(
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['codesign', '--force', '--sign', '-', '--generate-entitlement-der', '--verbose', '/Users/sunya/codesign_test/lib/python3.9/site-packages/PySide6/Qt/lib/QtWebEngineCore.framework']' returned non-zero exit status 1.
```

Since we're now code-signing frameworks, it looks like we're now picking up some Qt frameworks that are hidden deep in the PySide6 Python module.

I know we were previously recursing over the entire install directory and signing dylibs, but since we're now looking at frameworks too I wonder if we ought to limit the search that the `Codesign` method does. What do you think?

dgovil:
Oh good catch. Yeah we should probably only sign the OpenUSD.framework. 

I can add a check , but if you need to unblock your teams, just let me know if you add it first.

dgovil:
Okay, I added what I think should do the trick.

build_scripts/apple_utils.py
```diff
@@ -308,6 +308,9 @@ def Codesign(install_path, identifier=None, force=False, verbose_output=False) -
         dirs[:] = [d for d in dirs if not d.endswith(".framework")]
 
         for framework in frameworks:
+            framework_name = os.path.splitext(framework)[0]
+            if framework_name.lower() not in ["openusd", "opensubdiv", "materialx"]:
+                continue
```
sunyab: Instead of hardcoding a list of names, what do you think about just limiting the search to all `.dylib`, `.so`, and `.framework` files/directories in specific paths directory instead of walking the entire directory structure?

So something like:
```
codesignPaths = [
    os.path.join(install_dir, 'lib'),
    os.path.join(install_dir, 'plugin'),
    os.path.join(install_dir, 'share/usd')
]

for dir in codesignPaths:
    shared_libs = glob.glob(f"{os.path.join(dir, '*.dylib')")
    shared_libs += glob.glob(f"{os.path.join(dir, '*.so')")
    for lib in shared_libs:
        # Codesign..

    frameworks = glob.glob(f"{os.path.join(dir, '*.framework')")
    for framework in frameworks:
        # Codesign...
```

I like that this is more targeted -- the current implementation winds up signing files in unnecessary places like the `build` directory.

dgovil: Oh yeah that makes sense. I guess I hadn't considered that since I always install into a clean directory, but I'm assuming others may install into a communal directory?

dgovil: Oh I see. It's because the usdInstPath is not the final install directory but the full container directory.

dgovil: @sunyab Okay, I pushed an update that integrates your solution to filter. I still use my walk within the filtered set because there are some dylibs that are within nested folders that need to be signed.
But it should hopefully resolve what you're seeing

PR #3714, opened 2025-07-08
APPLE: Streamline Framework build headers

dgovil:
### Description of Change(s)

This PR allows for Framework builds (requires https://github.com/PixarAnimationStudios/OpenUSD/pull/3713 to be merged first) to use modified headers such that the frameworks can easily be used in Xcode or via `-framework OpenUSD` without any extra configuration.

It does so by modifying headers in the framework by changing them to `<OpenUSD/pxr/pxr.h>` instead of `<pxr/pxr.h>`.
These headers then work without having to configure anything, but if a developer were to also add the header search path, then they could use either form interchangeably too.

Of course, any code that uses the prefixed header format could not be used portably with a codebase that doesn't prefix its headers, so the option is opt-in.

### Checklist


jesschimein:
Filed as internal issue #USD-11202

PR #3725, opened 2025-07-14
Apple: Fix EXR write compression artifacts

dgovil:
### Description of Change(s)

We were finding that the EXR write out would often have weird artifacts like the red lines below if we wrote RGB instead of RGBA.

image

As such, we disable compression if writing fewer than 4 channels.
I couldn't think of a good way to test this unfortunately, so I don't have a test but I'm happy to add one if someone could think of something.

### Checklist

### DIFF
--- a/pxr/imaging/hio/OpenEXR/openexr-c.c
+++ b/pxr/imaging/hio/OpenEXR/openexr-c.c
@@ -355,7 +355,8 @@ exr_result_t nanoexr_write_exr(
     exr_set_longname_support(exr, 1);
 
     /// XXX In the future Hio may be able to specify compression levels
-    result = exr_set_zip_compression_level(exr, 0, 4);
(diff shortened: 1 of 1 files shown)


meshula:
Thanks for the report! Is this intermittent, or consistent for certain image sizes? Our test programs don't show this artifact, so maybe you've uncovered something new here.

dgovil:
Copying from our devs

> It reproduced in multiple resolutions.
The sample was rendered by 512x512 float32 exr

jesschimein:
Filed as internal issue #USD-11227

meshula:
@dgovil hio/testenv/testHioImage.cpp writes out textures whose color values are functions of x and y, then reads the texture back in and confirms the value. Maybe one of the existing tests can be tweaked to induce the failure?

PR #3727, opened 2025-07-15
Autodesk: [hdSt] Fix `CRLF` diffs in codegen unit tests on Windows

erikaharrison-adsk:
### Description of Change(s)

This PR reworks the handling of line-ending differences between platforms in codegen tests with a Git mechanism. This fixes test failures in Windows configurations where the `diff` tool is missing. As a way to verify the change, the `testHdStCodeGen` tests are enabled for Vulkan on Windows.

The original motivation for this work was to enable `testHdStMaterialXShaderGen` tests across multiple platforms, but those will have to wait for other fixes and a MaterialX update, which will finally remove all differences between platforms in the code generated by MaterialX.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/cmake/macros/testWrapper.py
+++ b/cmake/macros/testWrapper.py
@@ -162,7 +162,7 @@ def _diff(fileName, baselineDir, verbose, failuresDir=None):
     isWindows = platform.system() == 'Windows'
 
     diffTool = shutil.which('diff')
-    diffToolBaseArgs = ['--strip-trailing-cr']
+    diffToolBaseArgs = []
     if not diffTool and isWindows:
         diffTool = 'fc.exe'
         diffToolBaseArgs = ['/t']
--- a/pxr/imaging/hdSt/CMakeLists.txt
+++ b/pxr/imaging/hdSt/CMakeLists.txt
@@ -238,7 +238,7 @@ pxr_build_test(testHdStBasicDrawing
 )
 endif()
 
-if (X11_FOUND)
+if (X11_FOUND OR PXR_ENABLE_VULKAN_SUPPORT)
 pxr_build_test(testHdStBarAllocationLimit
     LIBRARIES
         hdSt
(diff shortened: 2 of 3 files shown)


jesschimein:
Filed as internal issue #USD-11229

cmake/macros/testWrapper.py
```diff
@@ -162,7 +162,7 @@ def _diff(fileName, baselineDir, verbose, failuresDir=None):
     isWindows = platform.system() == 'Windows'
 
     diffTool = shutil.which('diff')
-    diffToolBaseArgs = ['--strip-trailing-cr']
+    diffToolBaseArgs = []
     if not diffTool and isWindows:
         diffTool = 'fc.exe'
```
ppenenko, line 167: As a recap, shader codegen tests compare code generated by MaterialX and Storm against baseline files retrieved from source control. One issue that can cause false positives when running these tests across different platforms is that, following the general convention for text files, MaterialX and Storm generate `CRLF` for line endings on Windows and `LF` on other platforms.

With the `--strip-trailing-cr` command-line option, the `diff` tool ignores the line-ending differences, which solves the problem. Unfortunately, as we discussed earlier, we can't rely on the availability of `diff` on Windows machines. So we perform the check below and, if `diff` is not found, and fall back to `fc.exe`.

Unfortunately though, `fc.exe` doesn't support functionality equivalent to `diff --strip-trailing-cr`. That means that the behavior of the tests varies in this crucial aspect depending on the configuration of the individual machine. Typically, `diff` would be found on Windows workstations thanks to Git for Windows, but it may be missing on Windows CI machines. When tests fail because of these configuration differences, diagnosing the root cause is hard for the developer.

So, I propose to remove the option of ignoring line endings for both diff utilities, which will make the behavior of the tests consistent across all Windows machines, and to add an alternative Git-based mechanism below.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -238,7 +238,7 @@ pxr_build_test(testHdStBasicDrawing
 )
 endif()
 
-if (X11_FOUND)
+if (X11_FOUND OR PXR_ENABLE_VULKAN_SUPPORT)
```
ppenenko, line 241: In order to verify that the new diffing method works, we'd like to enable some codegen tests on Windows in the same PR. The OpenGL-specific variants of the tests fail on Windows for other reasons, so we propose to enable their Vulkan-specific variants first.

It should be enough to just check `PXR_ENABLE_VULKAN_SUPPORT` (without checking for Windows) here because `PXR_ENABLE_VULKAN_SUPPORT` would be `false` on Apple and these tests were enabled on other platforms before.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -626,6 +626,111 @@ pxr_register_test(testHdStQualifiers
         TF_DEBUG=HD_SAFE_MODE
 )
 
+function(register_codegen_tests suffix hgi_env)
```
ppenenko, line 629: Moving these tests up, to before the early `return()` statements for Apple and Windows.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -626,6 +626,111 @@ pxr_register_test(testHdStQualifiers
...
+    pxr_install_test_dir(
+        SRC testenv/testHdStCodeGen${suffix}
+        DEST testHdStCodeGen${suffix}
+    )
```
ppenenko, line 633: Moving the installation of tests here to make sure that it's controlled by the same conditions as the registration of the tests.

pxr/imaging/hdSt/testenv/.gitattributes
```diff
@@ -0,0 +1 @@
+*.out		text
```
ppenenko, line 1: The complete documentation on this mechanism is at https://git-scm.com/docs/gitattributes

This particular line means that all `*.out` (baseline) files under this specific directory will be treated as "text" by Git on checkout, i.e. they will be stored with `LF` line endings in Git but checked out with `CRLF` line endings on Windows.

This is similar to the local `git config core.autocrlf` setting that is typically configured on Windows machines to `true` by the Git installation. According to the docs, `git config core.autocrlf`, if configured, will take precedence over the setting in `.gitattributes`.

We have verified that the tests pass on Windows workstations with `git config core.autocrlf` configured to `true` and on our CI machines where `git config core.autocrlf` is left unconfigured, which is the expected behavior according to the docs.

PR #3745, opened 2025-07-21
Autodesk: [hd] Add some fast paths for triangles meshes in HdMeshUtil

erikaharrison-adsk:
### Description of Change(s)

- Don't create a redundant copy of face-varying primvars when already triangulated.
- Implement a fast path for triangle indices when we already have a triangle mesh.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/imaging/hd/meshUtil.h
+++ b/pxr/imaging/hd/meshUtil.h
@@ -22,6 +22,8 @@
 #include "pxr/base/vt/array.h"
 #include "pxr/base/vt/value.h"
 
+#include "pxr/base/tf/pxrCLI11/CLI11.h"
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 /// \class HdQuadInfo
@@ -56,6 +58,18 @@ struct HdQuadInfo {
     int maxNumVert;
     std::vector<int> numVerts;  // num vertices of non-quads
     std::vector<int> verts;     // vertex indices of non-quads
+};
+
+/// Return status for a computation
+enum class CLI11_NODISCARD HdMeshComputationResult
+{
+    /// Computation failed
+    Error,
+    /// Computation succeeded and a result was produced
+    Success,
+    /// Computation succeeded but no result was produced,
+    /// because it is the same as the input.
+    Unchanged
 };
 
 /// \class HdMeshUtil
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-11256

pxr/imaging/hd/meshUtil.h
```diff
@@ -104,7 +104,8 @@ class HdMeshUtil
     /// a buffer of size numElements and type corresponding to dataType
     /// (e.g. HdTypeFloatVec3); the result is a VtArray<T> of the
     /// correct type written to the variable "triangulated".
-    /// This function returns false if it can't resolve dataType.
```
clach, line 107: A return value of false is currently used to indicate some sort of error (see `ComputeTriangulatedFaceVaryingPrimvar`'s use in `HdEmbree`, where a return of false triggers a coding error). Does it make more sense for an already triangulated primvar to result in a return value of true (successful triangulation), and in such a case we set `triangulated` to the original data?

DDoS: I wanted to avoid the buffer copy, because triangulation can't construct a `VtArray` directly from the pointer it gets. In `HdSt_TriangulateFaceVaryingComputation::Resolve()` a `false` result will lead to `_SetResult(_source)`, which is exactly what I want. But I agree it would be good to differentiate from the error case. I can have this return a tri-state enum.

clach:
@DDoS Two things: 1) This change fails `testUsdImagingGLGeomSubsets` for me internally (looks like something wrong with the triangles).  And 2), an open source version of the build on linux fails with this error:
```
[...]/pxr/imaging/hd/meshUtil.cpp: In member function ‘void pxrInternal_v0_25_8__pxrReserved__::HdMeshUtil::ComputeTriangleIndices(pxrInternal_v0_25_8__pxrReserved__::VtVec3iArray*, pxrInternal_v0_25_8__pxrReserved__::VtIntArray*, pxrInternal_v0_25_8__pxrReserved__::VtIntArray*) const’:
[...]/pxr/imaging/hd/meshUtil.cpp:117:73: error: ‘void* memcpy(void*, const void*, size_t)’ copying an object of type ‘pxrInternal_v0_25_8__pxrReserved__::VtArray<pxrInternal_v0_25_8__pxrReserved__::GfVec3i>::ElementType’ {aka ‘class pxrInternal_v0_25_8__pxrReserved__::GfVec3i’} with ‘private’ member ‘pxrInternal_v0_25_8__pxrReserved__::GfVec3i::_data’ from an array of ‘const int’; use assignment or copy-initialization instead [-Werror=class-memaccess]
```

Issue #3754, opened 2025-07-25
Creating a truly portable OpenUSD install on Windows- hardcoded absolute paths in pxrConfig.cmake


panthuncia:
I'm building a portable OpenUSD binary so that I don't need to re-build it every time I rebuild my project. I noticed that there are some hard-coded paths in the output pxrConfig.cmake which stop it from being used if its original install location is gone.

````    
_add_property(INTERFACE_INCLUDE_DIRECTORIES "C:/install/include")
_add_property(INTERFACE_COMPILE_DEFINITIONS "$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:TBB_USE_DEBUG=1>")
_add_property(INTERFACE_LINK_DIRECTORIES "$<$<CONFIG:Debug>:C:/install/lib>;$<$<CONFIG:Release>:C:/install/lib>")
_add_property(IMPORTED_IMPLIB "C:/install/lib/tbb_debug.lib")
_add_property(IMPORTED_LOCATION "C:/install/bin/tbb_debug.dll")
````
I tried replacing these lines with the following:

```
    _add_property(INTERFACE_INCLUDE_DIRECTORIES "${PXR_CMAKE_DIR}/include")
    _add_property(INTERFACE_COMPILE_DEFINITIONS "$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:TBB_USE_DEBUG=1>")
    _add_property(INTERFACE_LINK_DIRECTORIES "$<$<CONFIG:Debug>:${PXR_CMAKE_DIR}/lib>;$<$<CONFIG:Release>:${PXR_CMAKE_DIR}/lib>")
    _add_property(IMPORTED_IMPLIB "${PXR_CMAKE_DIR}/lib/tbb_debug.lib")
    _add_property(IMPORTED_LOCATION "${PXR_CMAKE_DIR}/bin/tbb_debug.dll")
```

and this fixed the issue, but it seems odd that edits like this would be required.

 I'm not terribly good with CMake- is there a way that this can be resolved at build-time rather than by manually editing these files? If something doesn't already exist to fix this, and someone can point me to what would need to change, I could submit a PR to fix this.

nvmkuruc:
Related issue:  #1025

jesschimein:
Filed as internal issue #USD-11272

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

[...]

Issue #3610, opened 2025-04-23
build_usd.py favours system TBB over local one


robpieke:
In 37de070d481f7da8a5647c8c5d318e21340d4c18 the search for TBB was updated to use `find_package(TBB CONFIG)` as a first-pass. This tries to find `TBBConfig.cmake`, which isn't generated during the build&install of TBB via `build_usd.py`, and so it's easy to fallback to picking up a system version of TBB.

My suggestion is that `build_usd.py` should be updated to generate `TBBConfig.cmake`. For example, amending `InstallTBB_Linux` with the following:

```
diff --git a/build_scripts/build_usd.py b/build_scripts/build_usd.py
--- a/build_scripts/build_usd.py	(revision 8de23df83a98dfbe5e35577575fcb9333fa130cd)
+++ b/build_scripts/build_usd.py	(date 1745400338674)
@@ -1163,6 +1163,19 @@
         CopyDirectory(context, "include/serial", "include/serial")
         CopyDirectory(context, "include/tbb", "include/tbb")
 
+        # Generate TBBConfig.cmake to satisfy the call to
+        # `find_package(TBB CONFIG)` in `cmake/defaults/Packages.cmake`
+        makeTBBConfigCmd = ' '.join([
+            'cmake',
+            '-DINSTALL_DIR={}'.format(context.instDir+'/lib/cmake/TBB'),
+            '-DSYSTEM_NAME={}'.format('Linux'),
+            '-DTBB_VERSION_FILE={}'.format('include/tbb/tbb_stddef.h'),
+            '-DINC_REL_PATH={}'.format('../../../include'),
+            '-DLIB_REL_PATH={}'.format('../../../lib'),
+            '-DBIN_REL_PATH={}'.format('../../../bin'),
+            '-P cmake/tbb_config_installer.cmake'])
+        Run(makeTBBConfigCmd)
+
 TBB = Dependency("TBB", InstallTBB, "include/tbb/tbb.h")
 
 ############################################################
```

(obviously the other OSes would need handling as well)

jesschimein:
Filed as internal issue #USD-10942

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

michalpelka:
Thanks @robpieke , it helped me build OpenUSD next to existing TBB installation in my system!

paulo-maciel:
[...]

Issue #3534, opened 2025-02-17
Use config mode for TBB finder


darix:
Hi, 

I am currently working on enabling USD support in the openSUSE Blender package.

With the help of a friend I was able to figure out why the finder couldn't find our tbb libraries although we had the cmake files installed in the tbb-devel package.

This resulted in the patch below:

[system-tbb.patch.txt](https://github.com/user-attachments/files/18829423/system-tbb.patch.txt)

Would that be something you would incorporate also upstream?

jesschimein:
Filed as internal issue #USD-10679

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

darix:
and as a follow up question ... making the install process respect cmake install paths a bit more :) 

currently a few paths seem to use prefix/subdir and libdir is also ignored.

sunyab:
Closing out old issue. OpenUSD does use TBB's package config file if present, otherwise it will fall back to the old FindTBB cmake module. See #3207 and #3783 for more details.

PR #3783, opened 2025-08-20
Leverages TBB's own TBBConfig.cmake to find TBB

erslavin:
### Description of Change(s)

This change aims to treat TBB like MaterialX and OpenSubdiv by ensuring that TBB provides the proper `cmake` configuration that can be used by both the OpenUSD build and downstream builds built on top of `cmake` using OpenUSD builds via `pxrConfig.cmake`.  It effectively:

* Removes logic necessary to manually define the TBB targets for `cmake` by using TBB's own provided `cmake` configuration files
* Removes more complex `find_package` logic from `build_usd.py`
* Adds the ability to treat legacy TBB the same as oneTBB by patching the `cmake` config files provided by TBB to point to the vendored artifacts in `bin` and `lib`
* Ensures that `TBB_DEBUG=1` is set as an interface compile definition for debug builds in legacy TBB cases (`oneTBB` already defines this using a generator expression)
* Ensures that `oneTBB` configuration is generated after the `make` steps and that it is patched to point to the right vendored artifact locations
* Removes the additional includes added to the OpenExec libraries - these are covered via the `INTERFACE_INCLUDE_DIRECTORIES` for the `TBB::tbb` target and should not be additionally added here to prevent absolute paths leaking into `pxrTargets.cmake`

No additional unit test was added as the OpenUSD build consumes this configuration when building itself.

### Fixes Issue(s)

This should fix the following issues:

* https://github.com/PixarAnimationStudios/OpenUSD/issues/3610
* https://github.com/PixarAnimationStudios/OpenUSD/issues/3534

and partially address https://github.com/PixarAnimationStudios/OpenUSD/issues/3754 by making the TBB portion relocatable for downstream builds using `pxrConfig.cmake` and `pxrTargets.cmake`.

### Checklist


jesschimein:
Filed as internal issue #USD-11365

build_scripts/build_usd.py
```diff
@@ -1024,6 +1024,41 @@ def InstallTBB_Windows(context, force, buildArgs):
...
+                      (
+                        "${_tbb_root}/lib/${_tbb_arch_subdir}/${_tbb_compiler_subdir}/${_tbb_component}_debug.lib",
+                        "${_tbb_root}/lib/${_tbb_component}_debug.lib"
+                      ),
+                      (
+                        "if (EXISTS \"${_tbb_debug_lib}\")",
+                        "if (EXISTS \"${_tbb_debug_lib}\")\n" +
+                        "               set_property(TARGET TBB::${_tbb_component} APPEND PROPERTY INTERFACE_COMPILE_DEFINITIONS \"TBB_USE_DEBUG=1\")"
+                    )
+                  ])
```
nvidia-jomiller, line 1059: Do the copied CMake files (`cmake/TBBConfig.cmake` and `cmake/TBBConfigVersion.cmake`) also need to be removed on Windows similar to MacOS and Linux platforms?

build_scripts/build_usd.py
```diff
@@ -1024,6 +1024,41 @@ def InstallTBB_Windows(context, force, buildArgs):
         CopyFiles(context, "lib\\intel64\\vc14\\*.*", "lib")
         CopyDirectory(context, "include\\serial", "include\\serial")
         CopyDirectory(context, "include\\tbb", "include\\tbb")
+        CopyFiles(context, "cmake\\TBBConfig.cmake", "lib\\cmake\\TBB")
+        CopyFiles(context, "cmake\\TBBConfigVersion.cmake", "lib\\cmake\\TBB")
```
nvidia-jomiller, line 1028: nitpick but It might be nice to use `os.path.join` instead of `\\`

build_scripts/build_usd.py
```diff
@@ -1123,6 +1158,47 @@ def _RunBuild(arch):
...
+        # as well as adds TBB_DEBUG to the INTERFACE_INCLUDE_DEFINITIONS
+        cmakeInstallTBBCmd = 'cmake -DTBB_ROOT={root} -DTBB_OS=Darwin -P cmake/tbb_config_generator.cmake'.format(
+            root=context.instDir)
+        Run(cmakeInstallTBBCmd)
+        CopyFiles(context, os.path.join(context.instDir, "cmake", "TBBConfig.cmake"),
+                  os.path.join(context.instDir, "lib", "cmake", "TBB"))
+        CopyFiles(context, os.path.join(context.instDir, "cmake", "TBBConfigVersion.cmake"),
+                  os.path.join(context.instDir, "lib", "cmake", "TBB"))
+        
+        PatchFile(os.path.join(context.instDir, "lib/cmake/TBB/TBBConfig.cmake"),
```
nvidia-jomiller: nitpick but the path components should be a list of strings instead of a single string since you're already using `os.path.join`

build_scripts/build_usd.py
spitzak: Or use forward slashes!

erslavin: In this particular case, it was just mirroring what was already there.  Sure, I could change it for this, but I think if we did that we'd have to change the original ones as well that are outside of this PR.

build_scripts/build_usd.py
erslavin: No, this isn't required on Windows - Windows uses a pre-built binary zip that straight copies things out to the target directories.  Linux / Mac build TBB from source, and when TBB builds it will put the `cmake` artifacts in the `cmake` folder, which are then copied to `lib/cmake/TBB` to be consistent with `MaterialX` and `OpenSubdiv`

build_scripts/build_usd.py
erslavin: This one makes sense to me, it's a different case than above because there wasn't already code there to mirror, will change.

build_scripts/build_usd.py
```diff
@@ -1024,6 +1024,41 @@ def InstallTBB_Windows(context, force, buildArgs):
...
+                        "# TBBConfigVersion.cmake defines TBB_VERSION",
+                        "# TBBConfigVersion.cmake defines TBB_VERSION\n" +
+                        "#\n" +
+                        "# THIS FILE HAS BEEN MODIFIED FROM ITS ORIGINAL VERSION"      
+                      ),
+                      (
+                        "get_filename_component(_tbb_root \"${_tbb_root}\" PATH)",
+                        "get_filename_component(_tbb_root \"${_tbb_root}\" PATH)\n" +
+                        "get_filename_component(_tbb_root \"${_tbb_root}\" PATH)\n" +
+                        "get_filename_component(_tbb_root \"${_tbb_root}\" PATH)"
```
buddly27: nitpick: `PATH` is generally discouraged in favor of `DIRECTORY`, since it better expresses the intent. 

In this case, could this be simplified with the following?
```cmake
get_filename_component(_tbb_root "${_tbb_root}/../.." ABSOLUTE)
```
(same thing for other platforms)

erslavin:
@robpieke I think this one takes a similar path you were suggesting and makes it more comprehensive so everything works with legacy TBB, oneTBB, and removes the FindTBB module because the build can always use the config coming from TBB.  Would you mind taking a look at this and letting me know if it captures your original intent?  Thanks!

build_scripts/build_usd.py
erslavin: I agree on your statement about `PATH` - this was just replacing what was already in the supplied `cmake` file from TBB, so I would prefer that was fixed there.  However, your suggestion to use `get_filename_component` with `ABSOLUTE` seems appropriate and would also address your original concern.

robpieke:
> @robpieke I think this one takes a similar path you were suggesting ...  Would you mind taking a look at this and letting me know if it captures your original intent? Thanks!

Yes, in particular `cmakeInstallTBBCmd` looks to capture the intent of my POC perfectly.

tallytalwar:
Original issue reported in #3610 has been fixed by ecea8d8. Also a part of this PR has been incorporated in 9e27474.

As per the discussions on various other forums, once legacy TBB dependency is removed from OpenUSD, we can incorporate more changes from this PR. Closing this for now. Thanks

Issue #801, opened 2019-03-15
Disabling a display purpose in Embree does not stop it from being displayed.


abucior:
### Description of Issue
When using Embree, once a display purpose has been enabled, disabling the purpose does not stop it from being rendered in the viewport.

### Steps to Reproduce
1. Load the following into usdview:
```#usda 1.0
def Xform "test"
{
    def Sphere "sphere"
    {
        double3 xformOp:translate = (0, 0, 0)
        uniform token[] xformOpOrder = ["xformOp:translate"]
        uniform token purpose = "render"
    }
    def Cube "box"
    {
        double3 xformOp:translate = (3, 0, 0)
        uniform token[] xformOpOrder = ["xformOp:translate"]
        uniform token purpose = "proxy"
    }
}
2. Enable Embree rendering
3. Toggle on and off render and proxy display purposes. Note that once a purpose has been enabled, it does not stop rendering, even when disabled. The prim bounding box does seem to update, but it's GL rendered.

Switching back to GL rendering, the behaviour is correct.

### System Information (OS, Hardware)
Linux, Centos 7

### Package Versions
USD 19.01

abucior:
Apologies for double-creating this ticket. I've closed the previous one. Github was having issues and I couldn't edit the old ticket, so I created a new one, but now it all seems to be working fine. Sorry about that.

jilliene:
Filed as internal issue #USD-5164

c64kernal:
Thanks for the report @abucior -- our sample embree implementation currently lacks proper collection or renderTag suport.  We'll use this to track the work of adding it. Thanks again!

spitzak:
It looks like the same bug is in hdPrman

spitzak:
I got our delegate to work but it is rather kludgy. It looks like Sync() is only called on rprims that should now be visible when the set changes, so I had that turn on any that are turned off. And RenderPass also tracks the RenderTags being passed to it, and if any tags are removed it goes through all the rprims and turns off (using a delegate-specific api) any that have that removed purpose. There is no need for RenderPass to turn anything on, Hydra is at least partially doing the right thing.

I recommend that Hydra destroy the rprims with the wrong purpose, as this would return the index to the same state it would be if created initially with that purpose set.

spitzak:
[...]

PR #3786, opened 2025-08-25
Don't clear DirtyRenderTag bit and not skip sync of non-visible rprim when display purpose is changed

roggiezhang-nv:
### Description of Change(s)

When the display purpose of a prim is modified to a value that is not meant to be rendered, Hydra skips some syncs because it counts as invisible. This PR fixes it by not clearing DirtyRenderTag bit and not skipping sync of non-visible rprim when display purpose is changed.

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/801

### Checklist

### DIFF
--- a/pxr/imaging/hd/dirtyList.cpp
+++ b/pxr/imaging/hd/dirtyList.cpp
@@ -40,8 +40,11 @@ _DirtyRprimIdsFilterPredicate(
     
     if (mask == HdChangeTracker::Clean || bits & mask) {
         // Update the render tag if needed.
+        TfToken oldRenderTag = renderIndex->GetRenderTag(rprimID);
+        // Why we don't clear the dirty bit of DirtyRenderTag?
+        // Because rprim sync needs to detect the change of render tag.
         const TfToken& primRenderTag =
-            renderIndex->UpdateRenderTag(rprimID, bits);
+            renderIndex->_UpdateRenderTagInternal(rprimID, bits, false);
 
         // XXX An empty render tag set means everything passes the filter
         //     We should use an explicit token to indicate all render tags.
@@ -50,6 +53,14 @@ _DirtyRprimIdsFilterPredicate(
         //     Primary user is tests, but some single task render delegates
         //     that don't support render tags yet also use it.
         if (filterParam->renderTags.empty()) {
+            return true;
+        }
+        
+        // If the render tag has changed, we need to give rprim a chance to 
+        // sync the render tag so rprim can be updated.
+        // This is a hack to get around the fact that hydra skips the sync of
+        // rprims when the render tags are not meant to be rendered below.
+        if (oldRenderTag != primRenderTag) {
             return true;
         }
 
--- a/pxr/imaging/hd/renderIndex.cpp
+++ b/pxr/imaging/hd/renderIndex.cpp
@@ -1152,6 +1152,14 @@ TfToken
 HdRenderIndex::UpdateRenderTag(SdfPath const& id,
                                HdDirtyBits bits)
 {
+    return _UpdateRenderTagInternal(id, bits, true);
+}
+
+TfToken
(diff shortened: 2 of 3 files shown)


roggiezhang-nv:
@nvmkuruc for vis.

jesschimein:
Filed as internal issue #USD-11381

rajabala:
Could you provide some context on the underlying issue you're trying to solve?
The Hydra 1 Sync step intentionally skips Sync for Rprims with a purpose/render tag that isn't in any of the tasks submitted for execution. If the task's render tags were modified to include a skipped one, we recreate the dirty list, which would include the previously-skipped-prim.
If the prim's render tag were changed to one that _is_ in the tasks' render tags, then too, we rebuild the dirty list and will thus sync the prim.

nvmkuruc:
Thanks for taking a look @rajabala.  I think @roggiezhang-nv is trying to solve the issue reported here-- https://github.com/PixarAnimationStudios/OpenUSD/issues/801

roggiezhang-nv:
@rajabala 

A simple example that can be reproduced in usdview, you can create a cube or a mesh, and switch its display purpose from `default` or `render` to `proxy`. You'll see the mesh is not disappeared.

nvmkuruc:
@roggiezhang-nv Do you know if this reproduces now that Hydra 2 is enabled by default?

roggiezhang-nv:
> @roggiezhang-nv Do you know if this reproduces now that Hydra 2 is enabled by default?

They should be the same, but I can verify it again.

roggiezhang-nv:
@nvmkuruc It looks to me 25.11 has fixed this issue. Not sure which PRs made this. I tried both HD_ENABLE_SCENE_INDEX_EMULATION enabled/disabled.

rajabala:
Hmm, I just tested it locally and I see the issue with both PRMan and Embree. Not with Storm.
I'll look into it some more. I'm not convinced this is an issue in hd..

rajabala:
In Hydra 1, the Sync step as you've noticed, is called only for prims with a render tag that's in any of the submitted task's render tag lists (`HdTask::GetRenderTags`).

Because of this, `HdRprim` defines a virtual `UpdateRenderTag` method that render delegates can override to _know_ when the render tag has changed. This method is called _before_ Sync. 

From the looks of it, only `hdSt` overrides it to update an internal version counter that's consulted by the render passes.
I think `hdEmbree` and `hdPrman` would need to do a similar thing where they update their internal state and use this in `Hd*RenderPass::Execute` to generate the correct image.

This is unfortunate complexity in the Hydra 1 execution model that should become simple and straight-forward in Hydra 2, where the `HdRenderer` can observe the terminal scene index, and thus receive and process the notices as needed.

roggiezhang-nv:
> Hmm, I just tested it locally and I see the issue with both PRMan and Embree. Not with Storm. I'll look into it some more. I'm not convinced this is an issue in hd..

If I was correct, I can see this issue even with Storm with 25.02. I need to check it again.

roggiezhang-nv:
@rajabala You are right. I verified with hydra 1 of 25.02. I can repro the issue with Storm also. 

> This is unfortunate complexity in the Hydra 1 execution model that should become simple and straight-forward in Hydra 2, where the HdRenderer can observe the terminal scene index, and thus receive and process the notices as needed.

Are the fixes inside Hydra 2 already? Or not?

roggiezhang-nv:
@rajabala I have a novice question. Can I ask why we only sync those rprims that are in the render tag list, but not sync those rprims that have render tags changed even they are not in the render tags list anymore?

roggiezhang-nv:
@rajabala Thanks for your suggestion about how Hydra 1 works to sync display purpose change. We have fixed our issue with that. Closing this PR.

Issue #3746, opened 2025-07-21
USDA files encoded with BOM order marker for UTF-8 cannot be opened


erslavin:
When a `usda` file is UTF-8 encoded with the BOM signature, the code in the `UsdaFileFormat`'s `_CanReadImpl` that is invoked prior to spinning up the parser will fail to detect it as a valid USDA asset.  Re-encoding the file without the BOM signature solves the issue.

jesschimein:
Filed as internal issue #USD-11257

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

spiffmon:
GIven the recommendations **not** to include the BOM, and that it may cause confusion to many text editors/users... would we want the usda parser to silently accept it, or generate a warning?  Presumably such a file that is then force-Saved() or exported  would not contain the BOM?

Thanks for filing, @erslavin , and were you planning on tackling this one?

erslavin:
I think there might be two issues if OpenUSD wants to handle the BOM marker:

* The PEGTL parser code probably needs to be updated to have an optional rule in the LayerHeader to account for the presence of the BOM
* The pre-parser code also would need to be updated to account for its presence

This issue was specifically focused on addressing the second bullet above (as the parser isn't even invoked until it clears the `_CanReadImpl` method), but a brief examination of the PEGTL parser indicates to me that the first bullet would likely also have to be addressed.

I guess the question then would be do we want OpenUSD to support the BOM marker?  If so, does it generate a warning if it sees one, but still succeeds?

I checked the documentation, and didn't see any mention of the BOM marker in either of these places: 
* https://openusd.org/release/usdfaq.html
* https://openusd.org/release/api/_usd__page__u_t_f_8.html

If we don't want OpenUSD to handle the BOM marker, maybe an update to the documentation is all that is necessary?  The message that comes out of the pre-parser code is generic - it just says "is not a valid usda layer"

spiffmon:
Thanks, @erslavin ; we think non-support is the better place to be, and would be happy to take a PR to enhance the error message from the pre-parser code.  If that also included an update to the docs, that'd be great, but we could also handle that part.

meshula:
[...]

PR #3794, opened 2025-08-29
NVIDIA: Check BOM of USDA files and report errors if found

roggiezhang-nv:
### Description of Change(s)

When a usda file is UTF-8 encoded with the BOM signature, the code in the UsdaFileFormat's _CanReadImpl that is invoked prior to spinning up the parser will fail to detect it as a valid USDA asset. Re-encoding the file without the BOM signature solves the issue.

Spiff suggested to not support assets with BOM signature, and enhance the error message from the pre-parser code. This PR works for that. 

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/3746

### Checklist


roggiezhang-nv:
@nvmkuruc for vis.

spitzak:
I'm rather suprised that UTF-16 and UTF-32 are supported. Is this necessary, or can maybe they be deprecated and dropped as well?

nvmkuruc:
UTF-16 and UTF-32 aren't supported.  If those BOM's are detected, the error message recommends conversion to UTF-8.  Any suggestions on how to clarify that?

jesschimein:
Filed as internal issue #USD-11395

nvmkuruc:
@roggiezhang-nv, I think we may need to revise the proposed implementation.  It's possible that `CheckBOM` will now generate runtime errors during asset read that were previously captured in `_CanReadImpl`.  It might be less invasive to make BOM check a warning that gets emitted by `_CanReadImpl` when the file prefix is the BOM marker + the correct magic cookie.

pxr/usd/sdf/usdaFileFormat.cpp
```diff
@@ -177,25 +177,76 @@ SdfUsdaFileFormat::~SdfUsdaFileFormat()
...
+    }
+
+    return true;
+}
+
 bool
 _CanReadImpl(const std::shared_ptr<ArAsset>& asset,
-             const std::string& cookie)
+             const std::string& cookie,
+             bool bomCheckWarning = true)
```
spiffmon: We're calling this a "Warning", and yet, AFAICT, there is no warning or error issued anywhere.  Particularly for callers that are going to raise a RUNTIME_ERROR if this returns `false` (Like `_ReadFromAsset`) it would be extremely helpful to say why.  We could add an additional out-string `whyNot` parameter here that may return a non-empty info string... or issue a TfError that callers below would consume?

roggiezhang-nv: Improved according to your suggestions.

spiffmon: Thank you, @roggiezhang-nv - LGTM!

PR #3828, opened 2025-09-27
Physics updates for nested bodies

andrewkaufman:
### Description of Change(s)

These changes relate to https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/82 and should be merged after (or along with) https://github.com/PixarAnimationStudios/OpenUSD/pull/3827. Between these 2 PRs, and the internal change https://github.com/PixarAnimationStudios/OpenUSD/commit/ff68c3646784b666afe051fd519ce78a8b775eca I believe we have completed the proposal.

Here I have:
- Enabled articulations on kinematic bodies (e.g. mocap driven articulations)
  - One slight difference from the proposal; I did not implement the intricate logic about partially simulated articulations. I was advised that this concept isn't in-scope for the upcoming spec, so thought it better to leave out of the implementation as the simpler approach implemented here is all that is strictly required.
- Added a MassAPI validator
  - This wasn't called for explicitly in the proposal, but as the logic around mass is even more complex now, it seemed better to have one.
- Updated the schema and overview docs to allow for nested bodies
- Added a new nested articulation example (see screenshot below)
- Fixed a broken filepath in the existing BoxOnBox example (see screenshot below)

Note that in the interest of time I have put up the PR before coordinating with @AlesBorovicka. It would be good to wait for his review as well as the internal Pixar review before merging.

_Nested articulation example loaded in a reduced coordinate simulator_
image

_Current example is broken_
image

https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/82
https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/physics_nested_bodies

### Checklist

### DIFF
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -652,7 +652,7 @@ articulation.
 \subsection usdPhysics_examples Examples
 
 \subsubsection usdPhysics_box_on_box Box on Box
-\include examples/usdPhysicsBoxOnBox2.usda
+\include examples/usdPhysicsBoxOnBox.usda
 
 \subsubsection usdPhysics_box_on_quad Box on Quad
 \include examples/usdPhysicsBoxOnQuad.usda
@@ -671,5 +671,8 @@ articulation.
 
 \subsubsection usdPhysics_distance_joint_example Distance Joint
 \include examples/usdPhysicsDistanceJoint.usda
+
+\subsubsection usdPhysics_nested_articulation_example Nested Articulation
+\include examples/usdPhysicsNestedArticulation.usda
 
 */
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -213,15 +213,26 @@ multi-scene simulation scenario.
 \subsubsection usdPhysics_interaction_with_usd Interaction with the USD hierarchy
 
 If a prim in a USD scene graph hierarchy is marked with UsdPhysicsRigidBodyAPI, 
-the behavior is such that with UsdPhysicsRigidBodyAPI, all chidlren of the 
-marked prim are assumed to be part of this rigid body, and move rigidly along 
-with the body. This is consistent with the common behavior one expects during 
-hand-animation of a sub-tree. If aggregate properties of the entire rigid body 
-must be computed, such as total mass or the entirety of its collision volume, 
-then the contents of the entire subtree are considered. 
+all prims in the subtree of the marked prim are assumed to be part of this
+rigid body, and move rigidly along with the body. This is consistent with the
+common behavior one expects during hand-animation of a subtree.
+
+The exception to this is when a prim in a rigid body’s subtree has its own
+nested UsdPhysicsRigidBodyAPI. This prim creates a separate rigid body and forms
+the root of its own subtree. For all UsdPhysics purposes, this prim and all
+prims below it are not considered parts of bodies higher up in the hierarchy.
+The phrase “subtree belonging to a rigid body” is defined as all child prims
+of a rigid body that are not parts of other rigid bodies. Multiple rigid bodies
+in a subtree move completely independently, even if their transforms are stored
+relative to one another.  
+
+If aggregate properties of the entire rigid body must be computed, such as
+total mass or the entirety of its collision volume, then the contents of the
+entire subtree are considered, taking care to exclude the contents of any
+nested body subtrees.
 
 Note that it is of course permitted to change/animate the transforms in such 
-a sub-tree, in which case any derived quantities in the physics engine such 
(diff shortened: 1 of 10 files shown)


docs/wp_rigid_body_physics.rst
```diff
@@ -680,7 +680,9 @@ Here is our concrete schema proposal in full:
        doc = """Applies physics body attributes to any UsdGeomXformable prim and
        marks that prim to be driven by a simulation. If a simulation is running
        it will update this prim's pose. All prims in the hierarchy below this 
-       prim should move accordingly."""
```
andrewkaufman, line 683: I'm not entirely sure what this file is for. I updated the schema docs here just because of search/replace really. I noticed it also has some out-of-date information that presumably came from overview.dox at one time as well, but I haven't updated that. Is there some automated process to synchronize them?

tallytalwar: If I remember correctly white papers (wp) are the original proposals and we tend to not update these. cc. @spiffmon

pxr/usd/usdPhysics/overview.dox
```diff
@@ -627,21 +633,26 @@ rotors are attached to), or a direct or indirect parent prim. For fixed
...
-joint collections). If loops are found, they may be broken at an arbitrary 
-location. Alternatively, a joint in the loop may use its 
-excludeFromArticulation attribute flag to denote that it wishes to remain a 
-maximal joint, and at this point the loop is then broken.
+joints). If loops are found, they may be broken at an arbitrary location.
+Alternatively, a joint in the loop may use its excludeFromArticulation attribute
+flag to denote that it wishes to remain a maximal joint, and at this point the
+loop is then broken.
+
+It is possible to specify a kinematic articulation (e.g. a chain of bodies & joints
```
amoravanszky, line 647: "It is possible to specify a kinematic articulation....by setting the UsdPhysicsTokensType::physicsKinematicEnabled flag on the root body within the articulation." --> There is nothing in the spec that says that physicsKinematicEnabled has a different meaning for reduced coordinate solvers than for maximal solvers or that its behavior changes when combined with the usage of ArticulationRootAPI.  And ideally we would avoid this.  I would hope that if a body in an "articulation" is kinematic it behaves exactly as if it would would we not have any ArticulationRootAPI in the stage.  This in turn means that only the body with physicsKinematicEnabled  = true is kinematic, the other bodies are dynamic.  If you want the entire "articulation" to be kinematic, you should simply set physicsKinematicEnabled   = true on all component bodies.  I think that is what you meant with "partially simulated articulations" in the PR description.  If you put in this sentence above, you will shut the door to such partially simulated articulations also in the future because you are essentially saying the kinematic flag on all non-root bodies can be ignored.  I think that is problematic, and I think this paragraph starting on line 647 should simply be deleted.

andrewkaufman: The reason to include such a paragraph was that its explicitly asked for in the proposal:

> Add a description of fully or partially kinematic articulations within the Articulation subsection & update the Kinematic Bodies subsection to refer to it.

I buy your argument that saying nothing leaves us open to doing anything in the future, but at present (on `dev` before this PR) kinematic articulations were actively prevented (by the parser). So I'd prefer to make it clear in the docs that they should be expected to work now.

I think your main objection is that what I had written is actually inaccurate, setting kinematic on the root body isn't enough, it needs to be set on all bodies.

Would this diff also accomplish the clarification you're after? 

> It is possible to specify a kinematic articulation (e.g. a chain of bodies & joints
 which is driven by motion capture or other time sampled animation) by setting the
UsdPhysicsTokensType::physicsKinematicEnabled flag on ~~the root body~~ **all bodies** within the articulation.

docs/wp_rigid_body_physics.rst
andrewkaufman: @tallytalwar should I drop the changes to this file then?

pxr/usd/usdPhysics/overview.dox
amoravanszky: Yes it is fine if it says "on all bodies".  Thanks for the update.

jesschimein:
Filed as internal issue #USD-11481

pxr/usd/usdPhysics/overview.dox
andrewkaufman: I have pushed this change (rebased into the docs commit)

docs/wp_rigid_body_physics.rst
andrewkaufman: I dropped the changes from this file (rebased into the docs commit)

tallytalwar: Thanks @andrewkaufman

tallytalwar:
Thanks @andrewkaufman I have checked in the PR internally and will be available in the next dev push.

Issue #3771, opened 2025-08-06
USDZ asset resolution is case sensitive


dgovil:
We noticed today that OpenUSD will fail to read a `.USDZ` file but will read a `.usdz` file, even if the underlying file system is case insensitive.

Repro steps:
1. Download any USDZ from https://developer.apple.com/augmented-reality/quick-look/
2. Use usdcat to read it
3. Rename it to `foo.USDZ`
4. Use usdcat to read it again

You'll get this error `Failed to open "foo.USDZ" - Failed to open asset '/Users/dhruvgovil/Downloads/foo.USDZ[entry.usdc]'`

It looks like this is a bug in the asset resolver itself, because the file format plugins seem to handle case insensitivity fine.

jesschimein:
Filed as internal issue #USD-11330

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

sunyab:
Closing this out, this was addressed by #3843.

PR #3843, opened 2025-10-07
Make asset resolver case insensitive

danielenricocahall:
### Description of Change(s)

Ensure asset resolver can be run against files where the extension is capitalized e.g; `seahorse_anim_mtl_variant_2.USDZ` vs `seahorse_anim_mtl_variant_2.usdz` to address https://github.com/PixarAnimationStudios/OpenUSD/issues/3771.

Validated by running through steps detailed in issue:

- Download any USDZ from https://developer.apple.com/augmented-reality/quick-look/
- Use `usdcat` to read it
- Rename it to `foo.USDZ`
- Use `usdcat` to read it again

And indeed, `usdcat` works with the update.

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/3771
### Checklist

- [ ] I have submitted a signed Contributor License Agreement (Reference: 

### DIFF
--- a/pxr/usd/ar/resolver.cpp
+++ b/pxr/usd/ar/resolver.cpp
@@ -1444,7 +1444,7 @@ private:
 
         bool HandlesFormat(const std::string& extension) const
         {
-            return _packageFormat == extension;
+            return _packageFormat == TfStringToLowerAscii(extension);
         }
 
     private:
--- a/pxr/usd/ar/resolver.cpp
+++ b/pxr/usd/ar/resolver.cpp
@@ -1254,9 +1254,8 @@ private:
                 if (extension.empty()) {
                     continue;
                 }
-
                 _packageResolvers.push_back(std::make_shared<_PackageResolver>(
-                    extension, plugin, packageResolverType));
+                    TfStringToLowerAscii(extension), plugin, packageResolverType));
 
                 TF_DEBUG(AR_RESOLVER_INIT).Msg(
                     "ArGetResolver(): Using package resolver %s for %s "


jesschimein:
Filed as internal issue #USD-11521

pxr/usd/ar/resolver.cpp
```diff
@@ -1444,7 +1444,7 @@ class _DispatchingResolver final
 
         bool HandlesFormat(const std::string& extension) const
         {
-            return _packageFormat == extension;
+            return _packageFormat == TfStringToLowerAscii(extension);
```
sunyab, line 1446: For completeness we should also lower-case the extensions specified in the `plugInfo.json`, so that in the off chance that someone specifies a package resolver extension like `.UsDz` the case-insensitive comparison here still works.

The code that reads the extensions in is in `_InitializePackageResolvers` around line 1241. We should just make sure all of those extensions are also lower-cased.

sunyab (changes requested):
Thanks for filing this! The overall behavior seems good, I just left one note about an extra tweak we should make for completeness.

pxr/usd/ar/resolver.cpp
danielenricocahall: Hi @sunyab thank you for the feedback! I just committed an update to lower case each extension when it's added to the collection of package resolvers - let me know if this is what you were thinking!

sunyab: That looks great!

sunyab:
I've merged this internally, this'll go into the `dev` branch with the next push. Thanks!

PR #3844, opened 2025-10-07
Add texture support for geometric shader

kohakukun:
### Description of Change(s)

Add texture support for geometric shader. Currently texture is only supported in materialNetworkShader, lightingShader and renderPassShader. But we found that we may need texture support in the vertex or fragment shader which is not for the material or lighting. For example, we may use the texture to pass a large buffer to the shader. So the geometric shader also requires textures.


jesschimein:
Filed as internal issue #USD-11522

PR #3847, opened 2025-10-09
Autodesk: [hdSt] Early parallel MaterialX codegen, launched by a scene index

erikaharrison-adsk:
### Description of Change(s)

This is a refactoring of https://github.com/PixarAnimationStudios/OpenUSD/pull/3567 incorporating the code review feedback, namely:

1. Removed the `HdDirtyBits`-based mechanism to check if the filter task cached when the codegen was launched can be reused during material sync. Instead, we always derive the necessary data again, which is a minor overhead relative to the codegen process itself. This can be further optimized in the future using an alternative mechanism, e.g. based on the change tracker as suggested in the original code review.
2. Removed the new render delegate APIs for starting the early codegen and waiting for its results. Instead, the Storm render delegate injects a Storm-specific scene index which launches the parallel tasks when it receives material primitives. The waiting is exposed via a method of the scene index class.

- N/A

### Fixes Issue(s)

- N/A

### Checklist


jesschimein:
Filed as internal issue #USD-11537

pxr/imaging/hd/sceneIndexAdapterSceneDelegate.cpp
```diff
@@ -1308,29 +1308,37 @@ _ToMaterialNetworkMap(
...
-        GetRenderIndex().GetRenderDelegate()->GetMaterialRenderContexts();
-    HdMaterialNetworkSchema netSchema = matSchema.GetMaterialNetwork(renderContexts);
+    HdMaterialNetworkSchema netSchema =
+        matSchema.GetMaterialNetwork(renderContexts);
     if (!netSchema.IsDefined()) {
         return VtValue();
     }
 
     return VtValue(_ToMaterialNetworkMap(netSchema, renderContexts));
 }
```
ppenenko, line 1344: Extracted this static function for code reuse between the old synchronous code path and the new scene index which gets scene index prims without any conversions.

pxr/imaging/hdSt/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ if (${PXR_ENABLE_MATERIALX_SUPPORT})
     list(APPEND optionalPrivateClasses
          materialXFilter
          materialXShaderGen
+         materialXSyncSceneIndex
```
ppenenko, line 29: The new scene index implementation, which replaces `materialXSyncDispatcher` in the original PR.

pxr/imaging/hdSt/material.cpp
```diff
@@ -201,18 +219,19 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
...
-            materialTag = _networkProcessor.GetMaterialTag();
-            params = _networkProcessor.GetMaterialParams();
-                textureDescriptors = _networkProcessor.GetTextureDescriptors();
+            fragmentSource = _hdStMaterialNetwork.GetFragmentCode();
+            volumeSource = _hdStMaterialNetwork.GetVolumeCode();
+            displacementSource = _hdStMaterialNetwork.GetDisplacementCode();
+            materialMetadata = _hdStMaterialNetwork.GetMetadata();
+            materialTag = _hdStMaterialNetwork.GetMaterialTag();
+            params = _hdStMaterialNetwork.GetMaterialParams();
+                textureDescriptors = _hdStMaterialNetwork.GetTextureDescriptors();
```
ppenenko: I propose to rename `_networkProcessor`  to `_hdStMaterialNetwork`. IMHO "Processor" sounds confusing here because the object `_hdStMaterialNetwork` is simply initialized with those `Process` methods from `hdNetworkMap`, it doesn't process some external objects.

pxr/imaging/hdSt/material.cpp
```diff
@@ -191,6 +196,19 @@ HdStMaterial::Sync(HdSceneDelegate *sceneDelegate,
...
+        HdStRenderDelegate* stormDelegate = static_cast<HdStRenderDelegate*>(
+            sceneDelegate->GetRenderIndex().GetRenderDelegate());
+
+        if (HdSt_MaterialXSyncSceneIndex* sceneIndex =
+            stormDelegate->GetMaterialXSyncSceneIndex()) {
+
+            sceneIndex->Wait();
+        }
+    }
+#endif
```
ppenenko, line 228: Wait for the codegen tasks to complete. Previously, we were waiting by calling a new render delegate API before syncing the materials: https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#r1995705092

Now, we wait for the generator task group in each material's sync. In practice, only the first material sprim would ever wait for any measurable duration, and the overhead of waiting for all subsequent sprims is negligible (around 1 microsecond).

This is also where we could get the cached per-material state from the scene index, when we have a new mechanism to check the validity of that state - similar to https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#r1995719425

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -159,8 +159,8 @@ _GetGlslfxForTerminal(
     }
 }
 
-static HdMaterialNode2 const*
-_GetTerminalNode(
+HdMaterialNode2 const*
+HdSt_GetTerminalNode(
```
ppenenko, line 163: Now need to call it from outside this source file.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -1100,57 +1100,85 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
...
+        HdSt_GetTerminalNode(
+            filterTask->hdNetwork,
+            terminalName,
+            &filterTask->terminalNodePath);
+
+    if (!filterTask->terminalNode) {
+        return;
+    }
+
+    ProcessFilterTask(materialId, filterTask, isVolume, resourceRegistry);
```
ppenenko, line 1126: `ProcessMaterialNetwork` is now implemented via `ProcessFilterTask` for code reuse.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -1100,57 +1100,85 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
...
-    _displacementSource.clear();
-    _materialMetadata.clear();
-    _materialParams.clear();
-    _textureDescriptors.clear();
-    _materialTag = HdStMaterialTagTokens->defaultMaterialTag;
+    bool isVolume = false;
+
+    auto filterTask = std::make_shared<HdSt_MaterialFilterTask>();
+    filterTask->hdNetwork =
+        HdConvertToHdMaterialNetwork2(hdNetworkMap, &isVolume);
```
ppenenko, line 1107: The state necessary for the codegen process is now encapsulated in the `HdSt_MaterialFilterTask` class, because it needs to be created earlier than the material sprim and to persist until the parallel codegen task is finished.

`HdStMaterialNetwork::ProcessMaterialNetwork` is called in the old, synchronous code path, but we create a `HdSt_MaterialFilterTask` here in order to share code between the two code paths.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -387,10 +384,8 @@ _AddDefaultMtlxTextureValues(
     }
 }
 
-static void
-_AddFallbackDomeLightTextureNode(
-    HdMaterialNetwork2* hdNetwork,
-    SdfPath const& terminalNodePath)
+void
+HdSt_MaterialFilterTask::AddFallbackDomeLightTextureNode()
```
ppenenko, line 388: This and a few other functions are now methods of `HdSt_MaterialFilterTask`.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -401,13 +396,13 @@ _AddFallbackDomeLightTextureNode(
             HdStPackageFallbackDomeLightTexture()));
     const SdfPath domeTexturePath = 
         terminalNodePath.ReplaceName(_tokens->domeLightFallback);
-    hdNetwork->nodes.insert({domeTexturePath, hdDomeTextureNode});
+    hdNetwork.nodes.insert({domeTexturePath, hdDomeTextureNode});
```
ppenenko, line 399: This and a few other related pieces of data are now encapsulated as members of `HdSt_MaterialFilterTask`.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1329,28 +1343,26 @@ _IsTopologicalShader(TfToken const& nodeId)
...
-// Build the topoNetwork, equivalent to the given hdNetwork but anonymized and 
-// stripped of non-topological parameters to better re-use the generated shader.  
-static size_t
-_BuildEquivalentMaterialNetwork(
-    HdMaterialNetwork2 const& hdNetwork,
-    HdMaterialNetwork2* topoNetwork,
-    HdAnnonNodePathMap* annonNodePathMap)
+size_t
+HdSt_MaterialFilterTask::BuildAnonymizedMaterialNetwork(
+    HdMaterialNetwork2* anonNetwork)
```
ppenenko, line 1367: Unified the terminology around "anonymized". The input parameters are replaced by `HdSt_MaterialFilterTask`'s members.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1235,70 +1226,93 @@ _AddMaterialXParams(
...
+                    &hdNetwork, terminalNodePath, variable->getVariable());
             }
         }
     }
 }
 
-static mx::ShaderPtr
-_GenerateMaterialXShader(
-    HdMaterialNetwork2 const& hdNetwork,
+HdSt_MaterialXGeneratorTask::HdSt_MaterialXGeneratorTask(
```
ppenenko, line 1257: This is a different kind of "task" - the necessary data passed to the lambda executing in parallel in TBB and wrapping the MaterialX codegen.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1478,108 +1492,139 @@ _BuildEquivalentMaterialNetwork(
...
-    TfHashAppend(shaderHash, topoHash);
-    TfHashAppend(shaderHash, materialTag);
-    HdInstance<mx::ShaderPtr> glslfxInstance =
-        resourceRegistry->RegisterMaterialXShader(shaderHash.GetCode());
-
-    if (glslfxInstance.IsFirstInstance()) {
+    HdSt_MaterialXGeneratorTask generatorTask(
+        filterTask,
+        materialPath,
+        *resourceRegistry->GetHgi());
```
ppenenko, line 1584: Creating the generator task on the stack in the case when parallel codegen is off.

pxr/imaging/hdSt/materialXFilter.cpp
```diff
@@ -1478,108 +1492,139 @@ _BuildEquivalentMaterialNetwork(
...
+	if (   !filterTask->terminalNode
+        || !filterTask->IsMaterialX(&SdrRegistry::GetInstance())) {
+        return nullptr;
     }
 
-    return glslfxShader;
+    return std::make_unique<HdSt_MaterialXGeneratorTask>(
+        filterTask,
+        materialPath,
+        hgi);
```
ppenenko, line 1678: Creating the generator task on the heap for parallel codegen.

ppenenko: One important change from the original implementation is that the generator task now gets a shared pointer to the filter task and holds on to it. The generation process reads some material network data which is part of the filter task, so it's important to ensure that it's not destroyed before codegen is complete. In the original implementation, the sprim held ownership of the filter task for the duration of codegen.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,180 @@
+//
```
ppenenko, line 1: Replaces [pxr/imaging/hdSt/materialXSyncDispatcher.cpp](https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#diff-968f17a13de5562f49ae88052266f093f11aa87c9822aae1764b6ac1e9e36fc7) in the original implementation

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,180 @@
...
+    std::unique_ptr<HdSt_MaterialXGeneratorTask> generatorTask,
+    HdStResourceRegistrySharedPtr& resourceRegistry)
+{
+    HD_TRACE_FUNCTION();
+
+    if (resourceRegistry->ContainsMaterialXShader(
+        generatorTask->GetShaderHash())) {
+        // We already have a shader for this topology.
+        return;
+    }
```
ppenenko, line 149: First check that the shader hasn't been registered before the early parallel init. This is a new method which, unlike   `RegisterMaterialXShader`, doesn't add a new instance to the registry if an existing one isn't found.

Reusing `RegisterMaterialXShader` in combination with `IsFirstInstance` would be problematic because the intended usage pattern, in the case when there's no existing value, seems to be:
1. Insert a `nullptr` value and return a registry instance pointing to that value and holding a lock to the whole MaterialX registry.
2. The client code is supposed to populate the instance with a non-`nullptr` value while holding the lock. The instance's `IsFirstInstance` method returns `true` if the value is `nullptr`.

So the above pattern can't be used for parallel tasks, when the uniqueness check has to happen before launching the task, the non-`nullptr` value is known only when the task completes and the lock has to be released ASAP.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    // Use a separate concurrent container for tracking unique generator tasks.
+    // The generated shader will be registered in the resource registry when
+    // the task completes.
+    auto insertResult =
+        _generatorTaskSet.insert(generatorTask->GetShaderHash());
```
ppenenko, line 155: Another uniqueness check, this time between the different parallel tasks, using a dedicated data structure.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -54,6 +58,11 @@ TF_DEFINE_ENV_SETTING(HD_ENABLE_GPU_TINY_PRIM_CULLING, false,
 TF_DEFINE_ENV_SETTING(HDST_MAX_LIGHTS, 16,
                       "Maximum number of lights to render with");
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+TF_DEFINE_ENV_SETTING(HDST_ENABLE_PARALLEL_MTLX_CODEGEN, false,
+                      "Enable early parallelized MaterialX codegen");
+#endif
```
ppenenko, line 64: Moved this setting definition here for better encapsulation.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -610,4 +619,24 @@ HdStRenderDelegate::_ApplyTextureSettings()
...
+HdStRenderDelegate::SetTerminalSceneIndex(
+    const HdSceneIndexBaseRefPtr &terminalSceneIndex)
+{
+    if (!TfGetEnvSetting(HDST_ENABLE_PARALLEL_MTLX_CODEGEN)) {
+        return;
+    }
+
+    _materialXSyncSceneIndex = HdSt_MaterialXSyncSceneIndexRefPtr(
+        new HdSt_MaterialXSyncSceneIndex(terminalSceneIndex, *this));
+}
```
ppenenko, line 648: Injecting our scene index as the very last in the chain, and only if the optimization is enabled.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -610,4 +619,24 @@ HdStRenderDelegate::_ApplyTextureSettings()
...
+
+HdSt_MaterialXSyncSceneIndex*
+HdStRenderDelegate::GetMaterialXSyncSceneIndex()
+{
+    return get_pointer(_materialXSyncSceneIndex);
+}
+#endif
```
ppenenko, line 655: This is currently only used in `HdStMaterial::Sync` to wait for the parallel codegen, but could be used in the future to reuse some per-material state.

pxr/imaging/hdSt/resourceRegistry.cpp
```diff
@@ -666,6 +666,15 @@ HdStResourceRegistry::RegisterMaterialXShader(
...
 }
+
+bool
+HdStResourceRegistry::ContainsMaterialXShader(
+        HdInstance<MaterialX::ShaderPtr>::ID id)
+{
+    bool found = false;
+    _materialXShaderRegistry.FindInstance(id, &found);
+    return found;
+}
```
ppenenko, line 677: The new method just wraps existing functionality.

ppenenko:
FYI @tcauchois

pxr/imaging/hdSt/material.cpp
```diff
@@ -172,48 +177,91 @@ HdStMaterial::_ProcessTextureDescriptors(
...
+
+        // This scene index manages early parallel MaterialX codegen. It exists
+        // when that optimization is enabled.
+        if (HdSt_MaterialXSyncSceneIndex* sceneIndex =
+            stormDelegate->GetMaterialXSyncSceneIndex()) {
+
+            // Wait for all early parallel codegen tasks to complete and
+            // retrieve the state cached for this sprim when codegen was
+            // started
+            filterTask = sceneIndex->WaitAndExtractFilterTask(GetId());
```
ppenenko, line 214: Wait for the codegen tasks to complete. Previously, we were waiting by calling a new render delegate API before syncing the materials: https://github.com/PixarAnimationStudios/OpenUSD/pull/3567/files#r1995705092

Now, we wait for the generator task group in each material's sync. In practice, only the first material sprim would ever wait for any measurable duration, and the overhead of waiting for all subsequent sprims is negligible (around 1 microsecond).

This is also where we get the cached per-material state from the scene index. The filter task is removed from the scene index its ownership is transferred to the local variable.

Reusing the filter task allows us to avoid getting the material resource again below.

pxr/imaging/hdSt/material.cpp
```diff
@@ -172,48 +177,91 @@ HdStMaterial::_ProcessTextureDescriptors(
...
+        if (vtMat.IsHolding<HdMaterialNetworkMap>()) {
+
+            HdMaterialNetworkMap const& hdNetworkMap =
+                vtMat.UncheckedGet<HdMaterialNetworkMap>();
+
+            if (!hdNetworkMap.terminals.empty() && !hdNetworkMap.map.empty()) {
+                _hdStMaterialNetwork.ProcessMaterialNetwork(GetId(),
+                    hdNetworkMap, resourceRegistry.get());
+
+                processedMaterialNetwork = true;
```
ppenenko, line 243: Once the material network has processed the filter task, we don't need the task any more, so we destroy it by letting it go out of scope.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,241 @@
...
+        // If the dirtied prim is a material for which we've cached a filter
+        // task, remove the task since it's now stale. Any ongoing generator 
+        // tasks will keep their own shared pointers to the respective filter 
+        // tasks, so this is safe. If the element is not in the map, then the 
+        // lookup won't involve locking.
+        _FilterTaskMap::accessor accessor;
+        if (_filterTaskMap.find(accessor, entry.primPath)) {
+            _filterTaskMap.erase(accessor);
+        }
+    }
```
ppenenko, line 101: Deleting the filter task on dirtying.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,241 @@
...
+    auto generatorTask = HdSt_CreateMaterialXGeneratorTask(
+        id,
+        vtMat,
+        *resourceRegistry->GetHgi());
+
+    if (generatorTask) {
+        {
+            _FilterTaskMap::accessor accessor;
+            _filterTaskMap.insert(accessor, id);
+            accessor->second = generatorTask->GetFilterTask();
```
ppenenko, line 132: Caching the filter task in the scene index.

ppenenko:
> ### Description of Change(s)
> This is a refactoring of #3567 incorporating the code review feedback, namely:
> 
> 1. Removed the `HdDirtyBits`-based mechanism to check if the filter task cached when the codegen was launched can be reused during material sync. Instead, we always derive the necessary data again, which is a minor overhead relative to the codegen process itself. This can be further optimized in the future using an alternative mechanism, e.g. based on the change tracker as suggested in the original code review.

FYI @tcauchois @klucknav I've pushed another change to this PR reintroducing the reuse of filter tasks in `HdStMaterial::Sync`, but with invalidation implemented via a simpler mechanism based on scene index dirty notifications instead of dirty flags.

This addresses the other major feedback item from https://github.com/PixarAnimationStudios/OpenUSD/pull/3567.

To recap, filter tasks can become stale if the material changes after parallel codegen started but before the sprim is synced. Animated materials is the only known case of this, and USD has tests for animated materials.

This should be the last change in this branch.

pxr/imaging/hdSt/material.cpp
tcauchois: Is it possible instead to have the material prim datasource call wait on the scene index, so that Storm doesn't need to deal with the new scene index at all?  I'll try to sketch out a proposal in the scene index code.

pxr/imaging/hdSt/materialNetwork.cpp
```diff
@@ -1100,57 +1100,85 @@ HdStMaterialNetwork::ProcessMaterialNetwork(
...
-    bool isVolume = false;
     HdMaterialNetwork2 surfaceNetwork =
         HdConvertToHdMaterialNetwork2(hdNetworkMap, &isVolume);
     const TfToken &terminalName = (isVolume) ? HdMaterialTerminalTokens->volume 
                                             : HdMaterialTerminalTokens->surface;
 
-    SdfPath surfTerminalPath;
-    if (HdMaterialNode2 const* surfTerminal = 
-            _GetTerminalNode(surfaceNetwork, terminalName, &surfTerminalPath)) {
+    filterTask->terminalNode =
```
tcauchois, line 1116: Here & material.*, instead of bringing the "filterTask" API into a bunch of files in Storm, you could modularize things better by either:
(1) having the scene index produce a "glsl" material network that can be read by the non-materialx code; or
(2) having the scene index produce a digested material network as code snippets for each stage & params, and pass them through as datasources.  If material.cpp finds that, it uses it, and otherwise it falls back to network processing.

This splits the code up a little better, which is nice since the MaterialX code is an external dependency with an evolving API and hidden behind a build flag, so having a big API interface between the two makes me a bit worried.

pxr/imaging/hdSt/materialNetwork.h
```diff
@@ -20,9 +20,57 @@
...
 
+class SdrRegistry;
 class HdStResourceRegistry;
 using HioGlslfxSharedPtr = std::shared_ptr<class HioGlslfx>;
 using HdSt_MaterialParamVector = std::vector<class HdSt_MaterialParam>;
+struct HdSt_MaterialFilterTask;
+using HdSt_MaterialFilterTaskSharedPtr =
+    std::shared_ptr<HdSt_MaterialFilterTask>;
+
+extern HdMaterialNode2 const*
```
tcauchois, line 31: Don't think this needs an extern, but if you want it to be publicly visible (which maybe we don't care?) you could throw an HDST_API on it...

pxr/imaging/hdSt/materialNetwork.h
```diff
@@ -20,9 +20,57 @@
...
+    SdfPath * terminalNodePath);
+
+/// Encapsulates the input data for MaterialX codegen as well as metadata
+/// necessary for completing `HdStMaterial::Sync` based on the result of the
+/// codegen.
+/// This object can either live on the stack, if MaterialX codegen happens
+/// synchronously, or on the heap, owned by the respective Sprim, if the
+/// codegen happens in parallel tasks.
+///
+struct ARCH_EXPORT_TYPE HdSt_MaterialFilterTask final
```
tcauchois, line 44: This feels like an implementation detail of the MaterialX code, and I don't think this belongs in hdSt/materialNetwork.h.  Also curious why we need ARCH_EXPORT_TYPE here.

pxr/imaging/hdSt/materialXFilter.h
```diff
@@ -35,18 +44,68 @@ struct HdSt_MxShaderGenInfo {
...
+class HdSt_MaterialXGeneratorTask final
+{
+public:
+    /// Modifies the input filter task - stores mappings between original and
+    /// anonymized Sdf paths.
+    ///
+    HdSt_MaterialXGeneratorTask(
+        HdSt_MaterialFilterTaskSharedPtr filterTask,
+        SdfPath const& materialPath,
+        Hgi const& hgi);
```
tcauchois, line 58: I know you're improving on the status quo here (which is to pass in resource registry and pull hgi out of that), but it would be cleaner to just pass in the relevant info, namely bindless support and shading language.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
tcauchois: Makes sense.

pxr/imaging/hdSt/materialXSyncSceneIndex.cpp
```diff
@@ -0,0 +1,241 @@
...
+}
+
+HdSt_MaterialXSyncSceneIndex::~HdSt_MaterialXSyncSceneIndex()
+{
+}
+
+HdSceneIndexPrim
+HdSt_MaterialXSyncSceneIndex::GetPrim(const SdfPath &primPath) const
+{
+    // Just forward to input scene index - we don't modify prim data
```
tcauchois, line 46: I was envisioning that you call WaitAndExtractFilterTask here, either when you call GetPrim() on a materialx prim, or potentially when you call primDataSource->Get("material").  That way, you don't need all of the complicated logic in renderDelegate.h to pass the scene index pointer through.

pxr/imaging/hdSt/renderDelegate.cpp
```diff
@@ -626,4 +634,24 @@ HdStRenderDelegate::_ApplyTextureSettings()
         HdStTextureType::Field, 1048576 * memInMb);
 }
 
+#ifdef PXR_MATERIALX_SUPPORT_ENABLED
+void
+HdStRenderDelegate::SetTerminalSceneIndex(
```
tcauchois, line 639: As mentioned in the scene index, this bit breaks a bunch of encapsulation.  We need GetMaterialXSyncSceneIndex so we can call WaitAndExtractFilterTask in material.cpp, but if we do that in the scene index GetPrim() or a datasource instead we don't need this call anymore.  Meanwhile, rather than grafting the new scene index onto the terminal scene index like this, the more flexible/preferred way is to register a scene index plugin that inserts it somewhere relative to other scene index filters; it's a small bit of boilerplate you can grab from e.g. the dependencySceneIndexPlugin.* or something.

Then the only issue is your scene index's use of the render delegate, but it shouldn't really be getting a pointer to the render delegate anyway.  It does need a pointer to a registry of MaterialX codegen results, but we could either find a lower key way to pass in the resource registry, or just have a local registry on the scene index instead.

tcauchois (commented):
Hey Pavlo, this is great progress!  Moving the task launching to Hydra 2 PrimsAdded looks like it cut out a bunch of the hd API changes, and it should hopefully be letting you launch the tasks slightly earlier as well!

I feel like if we fold the WaitAndExtractFilterTask call into some kind of datasource access call inside MaterialXSyncSceneIndex we can additionally get rid of some of the new render delegate API in hdSt/renderDelegate.h, and make the scene index more modular/reusable, which I think a bunch of folks would appreciate past just the Storm team!

Happy to expand on my thoughts in the checkin tomorrow or on slack but this is definitely a good direction.

PR #3848, opened 2025-10-09
Autodesk: Simplify materials with CM on static values

erikaharrison-adsk:
### Description of Change(s)

Reduces the number of shader compilation across materials that differ only by a static color managed value.

There is no need to give color managed static colors to MaterialX as this will require adding CM nodes to convert constant values. Far better to compute the values once and set them into the shader already in the rendering color space.

Also note that MaterialX will not know how to handle a node with two color managed static colors when they differ in the color space name.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/imaging/hdSt/materialXFilter.cpp
+++ b/pxr/imaging/hdSt/materialXFilter.cpp
@@ -20,7 +20,11 @@
 #include "pxr/imaging/hio/glslfx.h"
 #include "pxr/imaging/hgi/capabilities.h"
 
+#include "pxr/base/gf/color.h"
+#include "pxr/base/gf/colorSpace.h"
 #include "pxr/base/gf/vec2f.h"
+#include "pxr/base/gf/vec3f.h"
+#include "pxr/base/gf/vec4f.h"
 #include "pxr/base/gf/matrix3d.h"
 #include "pxr/base/gf/matrix4d.h"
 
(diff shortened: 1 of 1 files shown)


JGamache-autodesk:
@meshula A test case for this change is:
[nanocolor_names_test.zip](https://github.com/user-attachments/files/22850588/nanocolor_names_test.zip)
With expected result:
image
Note that this file does not exercise the new ColorSpaceAPI, we might have a more extensive test coming at some point in the future. It does test a mix of MaterialX and Nanocolor names though.

JGamache-autodesk:
Tried adding ColorSpaceAPI to that example file:
[nanocolor_names_test_with_CS_API.zip](https://github.com/user-attachments/files/22851420/nanocolor_names_test_with_CS_API.zip)
And the color space does not seem to be currently transported by Hydra even though a call to `ComputeColorSpaceName` returns the expected result at the USD level.

I suppose that work is yet to be done, and requires deciding how the color space gets propagated to descendants. Will we end up with a `colorSpace:__color_attr_name__` for each affected attribute, or will we get the equivalent of a `colorSpace:__full_prim__` to tell that all GfColor and SdfAsset values in that prim are affected. To be confirmed.

jesschimein:
Filed as internal issue #USD-11538

spiffmon:
Just confirming that the Hydra side of ColorSpaceAPI is planned but not yet available.

JGamache-autodesk:
Thanks @spiffmon! I see that it will be available in next release via c62c30951dbf91 by @klucknav.

PR #3853, opened 2025-10-13
Fix LNK4217/LNK4286 warnings

miguelh-nvidia:
### Description of Change(s)

We observed and increased number of LNK4217/LNK4286 warnings.

- Build configuration: This ocurrs in v25.11-alpha.
- Platform: Windows
- Approximate number of linker warnings: 18129
- Suspected commit: See https://gitlab-master.nvidia.com/omniverse/USD/-/commit/f5da4e26a598a14fa80811d080e8243cdf6617a4
- References to the code: Up to 25.08 the following code existed
```diff
-    # When building a monolithic library we want all API functions to be
-    # exported.  So add FOO_EXPORTS=1 for every library in PXR_OBJECT_LIBS,
-    # where FOO is the uppercase version of the library name, to every
-    # library in PXR_OBJECT_LIBS.
-    set(exports "")
-    foreach(lib ${PXR_OBJECT_LIBS})
-        string(TOUPPER ${lib} uppercaseName)
-        list(APPEND exports "${uppercaseName}_EXPORTS=1")
-    endforeach()
-    foreach(lib ${PXR_OBJECT_LIBS})
-        set(objects "${objects};\$<TARGET_OBJECTS:${lib}>")
-        target_compile_definitions(${lib} PRIVATE ${exports})
-    endforeach()
```
i.e. https://gitlab-master.nvidia.com/omniverse/USD/-/blob/v25.08/cmake/macros/Public.cmake#L1244-1252, it was added in 25.11 for static, but not for shared libraries.

### Checklist

### DIFF
--- a/cmake/macros/Public.cmake
+++ b/cmake/macros/Public.cmake
@@ -1202,20 +1202,26 @@ function(pxr_toplevel_epilogue)
             )
         endif()
 
-        # When building a monolithic library (static or shared) we want all API functions to be
-        # exported. So add FOO_EXPORTS=1 for every library in PXR_OBJECT_LIBS,
-        # where FOO is the uppercase version of the library name, to every
-        # library in PXR_OBJECT_LIBS.
+        # When building a monolithic library (static or shared) we want all
+        # API functions to be exported. So add FOO_EXPORTS=1 for every
+        #library in PXR_OBJECT_LIBS, where FOO is the uppercase version
+        #of the library name, to every library in PXR_OBJECT_LIBS.
         set(exports "")
         foreach(lib ${PXR_OBJECT_LIBS})
             string(TOUPPER ${lib} uppercaseName)
             list(APPEND exports "${uppercaseName}_EXPORTS=1")
         endforeach()
 
+        if (TARGET python)
+            # The boost python target uses a different export macro so
+            # add that as well.
+            list(APPEND exports "PXR_BOOST_PYTHON_SOURCE=1")
+        endif()
+
         foreach(lib ${PXR_OBJECT_LIBS})
             target_compile_definitions(${lib} PRIVATE ${exports})
         endforeach()
-    
+
         if(BUILD_SHARED_LIBS)
             target_link_libraries(usd_m
                 PUBLIC
--- a/cmake/macros/Public.cmake
+++ b/cmake/macros/Public.cmake
@@ -1202,6 +1202,20 @@ function(pxr_toplevel_epilogue)
             )
         endif()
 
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-11545

cmake/macros/Public.cmake
```diff
@@ -1202,6 +1202,20 @@ function(pxr_toplevel_epilogue)
...
+        # When building a monolithic library (static or shared) we want all API functions to be
+        # exported. So add FOO_EXPORTS=1 for every library in PXR_OBJECT_LIBS,
+        # where FOO is the uppercase version of the library name, to every
+        # library in PXR_OBJECT_LIBS.
+        set(exports "")
+        foreach(lib ${PXR_OBJECT_LIBS})
+            string(TOUPPER ${lib} uppercaseName)
+            list(APPEND exports "${uppercaseName}_EXPORTS=1")
+        endforeach()
+
```
sunyab, line 1214:
```suggestion

if (TARGET python)
    # The boost python target uses a different export macro so
    # add that as well.
    list(APPEND exports "PXR_BOOST_PYTHON_SOURCE=1")
endif()

```

When we tested this fix on our side, we noticed there were still a few thousand warnings from the Python binding code in `pxr/external/boost/python`. This library doesn't follow the same `<FOO>_EXPORTS` pattern, so we needed to add this in to get rid of those warnings.

Could you give this a try on your side to confirm this works? If so, would you mind updating your PR to include this?

cmake/macros/Public.cmake
```diff
@@ -1202,6 +1202,20 @@ function(pxr_toplevel_epilogue)
             )
         endif()
 
+        # When building a monolithic library (static or shared) we want all API functions to be
+        # exported. So add FOO_EXPORTS=1 for every library in PXR_OBJECT_LIBS,
+        # where FOO is the uppercase version of the library name, to every
+        # library in PXR_OBJECT_LIBS.
```
sunyab: Per our code conventions, would you mind line-wrapping this comment to 80 chars?

sunyab (changes requested):
Thank you for flagging this and for the fix! Passing along some notes from @tlewis.

cmake/macros/Public.cmake
miguelh-nvidia: Fixed.

cmake/macros/Public.cmake
miguelh-nvidia: Added. Thanks for checking.

PR #3855, opened 2025-10-13
[gf] Remove C4018 warnings from GfNumericCast

nvidia-jomiller:
### Description of Change(s)

Using boolean types with GfNumericCast would implicitly cast to a signed integer in GfIntegerCompareLess, causing Windows to issue C4018 warnings.

The change in this PR checks to see if either type used in `GfNumericCast` is a bool. If it is, the bool will either be explicitly cast to `0` or `1` or the the integer type will be tested for `true` or `false`. This eliminates the `warning C4018: '<': signed/unsigned mismatch` on WIndows

### Checklist

### DIFF
--- a/pxr/base/gf/numericCast.h
+++ b/pxr/base/gf/numericCast.h
@@ -30,15 +30,6 @@ template <class T, class U>
 constexpr bool
 GfIntegerCompareLess(T t, U u) noexcept
 {
-    // XXX: 
-    // On Visual Studio warning C4018 (signed/unsigned mismatch) is emitted
-    // when this function is used with boolean values. Just disable this
-    // for now.
-#if defined(ARCH_COMPILER_MSVC)
-    ARCH_PRAGMA_PUSH
-    ARCH_PRAGMA(warning(disable:4018))
-#endif    
-
     static_assert(std::is_integral_v<T> && std::is_integral_v<U>);
 
     if constexpr (std::is_signed_v<T> == std::is_signed_v<U>) {
@@ -50,10 +41,6 @@ GfIntegerCompareLess(T t, U u) noexcept
     else {
         return u >= 0 && t < std::make_unsigned_t<U>(u);
     }
-
-#if defined(ARCH_COMPILER_MSVC)
-    ARCH_PRAGMA_POP
-#endif
 }
 
 enum GfNumericCastFailureType {
--- a/pxr/base/gf/numericCast.h
+++ b/pxr/base/gf/numericCast.h
@@ -103,7 +103,7 @@ GfNumericCast(From from, GfNumericCastFailureType *failType = nullptr)
     // bool -> int
     if constexpr (std::is_same_v<From, bool> && std::is_integral_v<To>) {
         // No need to range check bool to int
-        return from ? static_cast<To>(1) : static_cast<To>(0);
+        return To{from};
     }
     // int -> bool
     else if constexpr (std::is_integral_v<From> && std::is_same_v<To, bool>) {
--- a/pxr/base/gf/numericCast.h
+++ b/pxr/base/gf/numericCast.h
@@ -100,9 +100,27 @@ GfNumericCast(From from, GfNumericCastFailureType *failType = nullptr)
         };
     };
 
+    // bool -> int
+    if constexpr (std::is_same_v<From, bool> && std::is_integral_v<To>) {
+        // No need to range check bool to int
+        return from ? static_cast<To>(1) : static_cast<To>(0);
+    }
+    // int -> bool
+    else if constexpr (std::is_integral_v<From> && std::is_same_v<To, bool>) {
+        // Range check int to bool
+        if (GfIntegerCompareLess(from, 0)) {
+            setFail(GfNumericCastNegOverflow);
+            return {};
+        }
+        if (GfIntegerCompareLess(1, from)) {
+            setFail(GfNumericCastPosOverflow);
+            return {};
+        }
+        return from != 0;
+    }
     // int -> int.
-    if constexpr (std::is_integral_v<From> &&
-                  std::is_integral_v<To>) {
+    else if constexpr (std::is_integral_v<From> &&
+                       std::is_integral_v<To>) {
         // Range check integer to integer.
         if (GfIntegerCompareLess(from, ToLimits::min())) {
             setFail(GfNumericCastNegOverflow);


nvidia-jomiller:
@nvmkuruc for vis

sunyab:
@nvidia-jomiller Just to confirm, are these warnings new to 25.11? Or is this a fix for a preexisting issue?

nvidia-jomiller:
@sunyab these warnings are new to 25.11 (I'm not seeing them in 25.08). Unfortunately, I wasn't able to easily track down what commit caused it to start issuing warnings. But if you need the specific commit, I can try and track it down

jesschimein:
Filed as internal issue #USD-11548

nvidia-jomiller:
@sunyab after some internal review we've fixed this warning by handling `bool` types in `GfNumericCast` instead of supporting `bool` in `GfIntegerCompareLess`.

But as far as `GfIntegerCompareLess` is concerned we wanted to highlight that there is a difference when comparing `GfIntegerCompareLess` with [std::cmp_less](https://en.cppreference.com/w/cpp/utility/intcmp.html). std::cmp_less will issue a compile-time error when trying to use a `bool` type. Specifically,
```
It is a compile-time error if either T or U is a non-[integer](https://en.cppreference.com/w/cpp/language/types.html#Integer_types) type, a character type, or bool.
```
I'm not sure if this difference is intentional but maybe it's something we can mention in the code docs since `GfIntegerCompareLess` is [modeled after std::cmp_less](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/gf/numericCast.h#L25-L27). Or if it's not intended, perhaps issue a similar compile-time error if a `bool` type is used with `GfIntegerCompareLess`

sunyab:
@nvmkuruc @nvidia-jomiller We have a counter-proposal for the change to `GfNumericCast` that I'll post as a review note.

But separately: I am hesitant to land a low-level behavior change like this during the alpha/rc period for 25.11. How would you feel about us just putting a pragma in place to disable this warning for 25.11 and then landing this PR for the next release?

pxr/base/gf/numericCast.h
```diff
@@ -86,9 +86,27 @@ GfNumericCast(From from, GfNumericCastFailureType *failType = nullptr)
...
+        if (GfIntegerCompareLess(from, 0)) {
+            setFail(GfNumericCastNegOverflow);
+            return {};
+        }
+        if (GfIntegerCompareLess(1, from)) {
+            setFail(GfNumericCastPosOverflow);
+            return {};
+        }
+        return from != 0;
+    }
```
sunyab, line 107: We think it could be better to have `GfNumericCast` handle bools the same way the language does:
- bool -> T gives either T{0} or T{1}
- T -> bool gives false if the T is T{0} else true, with no other range checks.

Then we wouldn't need to use `GfIntegerCompareLess` here and could disallow bools in that function.

What do you think?

sunyab (changes requested):

pxr/base/gf/numericCast.h
nvmkuruc: We had originally thought about doing that but were hesitant to change behavior since `bool` was previously undergoing range checks.  However, if someone wanted to avoid range checking the cast to bool, wouldn't they just use `static_cast`?

nvmkuruc:
> @nvmkuruc @nvidia-jomiller We have a counter-proposal for the change to `GfNumericCast` that I'll post as a review note.
> 
> But separately: I am hesitant to land a low-level behavior change like this during the alpha/rc period for 25.11. How would you feel about us just putting a pragma in place to disable this warning for 25.11 and then landing this PR for the next release?

If we approached this with a #pragma for 25.11, would it be reasonable to still merge the test cases to make sure we're not masking any issues?

pxr/base/gf/numericCast.h
nvidia-jomiller: I think both of those changes would be fine. The only reason the range checking was added to T -> bool was to maintain behavior with what was already happening inside the `int -> int` conversion.

Disabling bools in `GfIntegerCompareLess` would definitely be welcome to align with behavior of `std::cmp_less`

nvmkuruc: In case it's helpful, I verified that `boost::numeric_cast` which I believe `GfNumericCast` was designed to replace produces overflow errors when converting out of range integers to bools.

```
    std::cout<<boost::numeric_cast<bool>(2)<<std::endl;
    std::cout<<boost::numeric_cast<bool>(-1)<<std::endl;
```

Note that even in Josh's implementation, `GfIntegerCompareLess` does not need to support bools because he's passing in explicit integer values as the operands to `GfIntegerCompareLess`.

nvmkuruc: @sunyab I think I missed your first note in my first readthrough.  Are you suggesting that `bool -> T` could be more compactly implemented as
```
if constexpr (std::is_same_v<From, bool> && std::is_integral_v<To>) {
     // No need to range check bool to int
     return To{from}
}
```

If so, I agree.

sunyab: > @sunyab I think I missed your first note in my first readthrough. Are you suggesting that `bool -> T` could be more compactly implemented as 
> <snip>

We were more concerned about the second bullet point re: range checks, the first point was just there for exposition. I see @nvidia-jomiller updated the PR already with this anyway, which is fine for us.

Regarding the range checks, those are great points that I'll bring back to the team to discuss.

sunyab:
> If we approached this with a #pragma for 25.11, would it be reasonable to still merge the test cases to make sure we're not masking any issues?

We could do that, I'll run that through and see if anything falls out.

nvidia-jomiller:
@sunyab thanks for [disabling the warning](https://github.com/PixarAnimationStudios/OpenUSD/commit/8310fdb3f74301a59cde2021dd93e31c65aa8a8e) and getting that into 25.11-rc1! However, we did notice that there is still another warning being issued since a `bool` still makes it into `GfIntegerCompareLess`. We're now getting a [C4804](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4804?view=msvc-160) warning:
```
C:/Users/jomiller/projects/usd-ci/USD/pxr/base/gf/numericCast.h(51,26): warning C4804: '<': unsafe use of type 'bool' i
n operation (compiling source file C:\Users\jomiller\projects\usd-ci\USD\pxr\base\vt\value.cpp) [C:\Users\jomiller\proj
ects\usd-ci\_build\0.25.11-rc1.py312.windows-x86_64.stock.release\pxr\base\vt\vt.vcxproj]
```

Is there any chance that C4804 could also be disabled for 25.11?

PR #3886, opened 2025-11-04
APPLE: Make Accessibility Priority variable

dgovil:
### Description of Change(s)

We have been showing the USD Accessibility API to more members of the accessibility community and a request was made to make priority not be time varying. Originally, we made this uniform because we were thinking of it as a way to help prioritize accessibility purposes globally, but have realized that this would also benefit from being varying.

For example, if there was an accessibility purpose that was only high priority for a given part of an animation, like flashing red lights, that would be useful for a system to know when it changes. These would be temporary priority increases as a result.

Looking at the USD code and spec, this should be a change that doesn't necessitate versioning the schema. Coupled with the low likelihood of anything having adopted the schema yet, and the even lower likelihood of people using the priority attribute yet, I think this is a safe assumption. I've done a few tests to make sure that nothing in USD seems to get mad when the uniform token is taken away or added as well, and all seems fine.

This should only be an issue if someone authored a file after this PR was merged with time sampled information and tried to read it into an older version of USD where the attribute was uniform. But in that case it appears USD will default to reading just the static default value, which we always recommend be authored.

Therefore I'd posit that this does not require versioning since its loosening things up, and the impact is very low.

### Checklist

### DIFF
--- a/pxr/usd/usdUI/accessibilityAPI.cpp
+++ b/pxr/usd/usdUI/accessibilityAPI.cpp
@@ -231,7 +231,7 @@ UsdUIAccessibilityAPI::CreatePriorityAttr(VtValue const &defaultValue, bool writ
                            UsdUITokens->accessibility_MultipleApplyTemplate_Priority),
                        SdfValueTypeNames->Token,
                        /* custom = */ false,
-                       SdfVariabilityUniform,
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-11621

PR #3898, opened 2025-11-25
Autodesk: [hdSt] Refactor MaterialXShaderGen with generics

kohakukun:
### Description of Change(s)

- Introduce a base class for the shader generators that implements common functionality to improve code reuse and reduce code duplication.
- Guard MaterialX's Metal shader code generation with PXR_METAL_SUPPORT_ENABLED.


jesschimein:
Filed as internal issue #USD-11672

PR #3915, opened 2025-12-09
Autodesk: [hd] Allow passing HdRendererCreateArgs to CreateDelegate() so IsSupported() can work correctly

erikaharrison-adsk:
### Description of Change(s)

The main issue is that the `IsSupported()` call checks the default Hgi, which isn't necessarily what one is trying to use. If the default isn't supported, then you can't create a render delegate even if a non-default is supported. We've encountered an issue (Windows laptop) where HgiGL fails to create due to an OpenGL version problem, but HgiVulkan works. It should be possible to use HgiVulkan (non-default) even when HgiGL (default) doesn't work. This solution addresses this issue.

Note: You can set `HGI_ENABLE_VULKAN` to change the defaults, but that's not always an option, and isn't a proper solution as the environment variables are for development, not production.

Example use case: https://github.com/Autodesk/hydra-viewport-toolbox/blob/a24d055fcf0d146249afb7e6a693fd5763355ca0/source/engine/renderIndexProxy.cpp#L49-L56

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/imaging/hd/rendererCreateArgs.h
+++ b/pxr/imaging/hd/rendererCreateArgs.h
@@ -23,6 +23,16 @@ struct HdRendererCreateArgs
     bool gpuEnabled { true };
     // An Hgi instance to check backend support against.
     Hgi* hgi { nullptr };
+
+    bool operator==(const HdRendererCreateArgs& other) const
+    {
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-11704

PR #3925, opened 2025-12-12
[arch] Ignore unused-result warning from debugger.cpp

nvidia-jomiller:
### Description of Change(s)

On gcc, or clang, `write()` from unistd.h will issue a warning if the result is not checked. In `debugger.cpp`, `write()` is being called and the result is not being used (understandably, since the process is exited immediately after). This warning can not be suppressed using a void cast so the changes here will will explicitly ignore the result via `std::ignore`.

### Checklist

- [] I have added unit tests that exercise this functionality (Reference: 

### DIFF
--- a/pxr/base/arch/debugger.cpp
+++ b/pxr/base/arch/debugger.cpp
@@ -34,6 +34,7 @@
 #include <string>
 #include <string_view>
 #include <system_error>
+#include <utility>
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-11718

PR #3927, opened 2025-12-16
UsdAbc: translate the property's arrayExtent metadata into elementSize

### LINKED ISSUE #3362: Alembic importer bring in 3 float attribute as single float
### Description of Issue
Alembic importer bringing in my "rest" attribute as a single float rather than a 3 float in Houdini. I've reported the issue to Sidefx and they said to log an issue here as any fix on there end would require changes to the alembic importer.

### Steps to Reproduce
1. Use a sublayer node in Houdini to bring in the alembic file (example files attached).

Rocky 8.7
Houdini 20.5.332

[rest_attribute_wrong_format.zip](https://github.com/user-attachments/files/17365655/rest_attribute_wrong_format.zip)

cameronwhite:
### Description of Change(s)

For the example file in the referenced bug report, the Alembic property has the metadata `arrayExtent: 3, podExtent: 1, podName: float32_t`.
This was translated into a `float[]` array with too many elements since the `arrayExtent` was ignored. This metadata seems to have an equivalent meaning to USD's elementSize, describing a grouping of consecutive elements in the array.

Houdini's Alembic exporter, for example, uses this in the generic case for translating float tuple attributes (which might have an arbitrary tuple size) if they're not a standard type like positions or colors.

(Note that if the property instead had metadata `arrayExtent:1, podExtent:3`, this behaved correctly as-is and would translate into a `float3[]` array)

N/A

### Fixes Issue(s)

Fixes: #3362

### Checklist

### DIFF
--- a/pxr/usd/plugin/usdAbc/CMakeLists.txt
+++ b/pxr/usd/plugin/usdAbc/CMakeLists.txt
@@ -74,6 +74,7 @@ pxr_plugin(usdAbc
 
 pxr_test_scripts(
     testenv/testUsdAbcAlembicData.py
+    testenv/testUsdAbcArrayExtent.py
     testenv/testUsdAbcBugs.py
     testenv/testUsdAbcCamera.py
     testenv/testUsdAbcConversionBasisCurves.py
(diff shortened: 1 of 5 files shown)


jesschimein:
Filed as internal issue #USD-11727

spiffmon:
@cameronwhite , while I can't be sure, it seems like `arrayExtent` is more general than just primvars in alembic?  Last year we introduced [`arraySizeConstraint` metadata for general array types in USD](https://openusd.org/release/api/class_usd_attribute.html#Usd_AttributeArraySizeConstraintAPI), and would like to promote it as preferred over `elementSize` (since it is also more expressive).

Could we ask that when reading in Alembic, you also add `arraySizeConstraint` to the attribute spec, and when writing from USD to alembic, check for that first and translate it to arrayExtent if present?

cameronwhite:
> Could we ask that when reading in Alembic, you also add `arraySizeConstraint` to the attribute spec, and when writing from USD to alembic, check for that first and translate it to arrayExtent if present?

Thanks for the suggestion - I wasn't aware `arraySizeConstraint` existed :)
I've updated the PR with those changes

PR #3931, opened 2025-12-22
HgiVulkan: Allow Headless Rendering on X11

jeroenbakker-atmind:
This is a change to allow headless rendering on Linux systems. In this case the 
instance will have enabled `VK_KHR_surface` and `VK_KHR_xlib_surface` but no
X11 display could be created.

`XOpenDisplay` will return `nullptr` (See https://www.x.org/releases/X11R7.7/doc/man/man3/XOpenDisplay.3.xhtml)
but `vkGetPhysicalDeviceXlibPresentationSupportKHR` requires a valid pointer to
a display struct.

Other solutions would be to check this during instance creation. I didn't go for
that solution as instance creation only does extension checking and not usage
checking.

A more future proof solution would be for the client to tell the renderer in what setup
it is being used. Now it is a bit hardcoded.

This fixes an issue in Blender to be able to render usd/Hydra images headless.
See https://projects.blender.org/blender/blender/issues/149631

### Checklist

### DIFF
--- a/pxr/imaging/hgiVulkan/device.cpp
+++ b/pxr/imaging/hgiVulkan/device.cpp
@@ -62,6 +62,10 @@ _SupportsPresentation(
                     physicalDevice, familyIndex);
 #elif defined(VK_USE_PLATFORM_XLIB_KHR)
         Display* dsp = XOpenDisplay(nullptr);
+        // Allow headless rendering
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-11750

Issue #3582, opened 2025-03-27
usdcat crashes with very large files


enrac5:
Hi there, I have some `usda` files that are very large, about 185GB, or 74GB and I'd like to convert them to `usdc` files. From the documentation, the suggested tool is `usdcat`. I'm using it like this:

```
usdcat -o big_file.usdc big_file.usda
```

but for the very large files, the process just crashes. Is there any way to convert very large USD files from `usda` to `usdc`?

spiffmon:
Can you confirm whether there's a typo in your command?  I.e. you want `big_file.usdc`, not `big_file_usdc`

enrac5:
There was a typo, fixed it. Thanks for the catch!

spiffmon:
OK, thanks.  Might I also inquire what process in your pipeline is producing multi-G usda files in the first place, and is it using OpenUSD to do so, or its own text exporter?

enrac5:
Hi there, we're doing stress tests on GPUs using `usdrecord` on files with lots of assets. To generate these files, we're just using USD's API, I guess the other option would be to manually tool the files, since they're `ascii`. It's "just" a lot of spheres.

spiffmon:
OK, it sounds like the files are easily regeneratable, would it be possible for you to do so?  The usda format is **not** meant for encoding really large scenes, and the load-time performance and memory consumption will be far inferior to usdc files, and given that there are no hardcoded limits in our code relating to usda file size (I'm pretty sure), it would be great not to spend time tracking down the source of this crash for a pessimal use-case :-) 

Note that you can also create .usd files that are actually usdc format, as noted [here](https://openusd.org/release/usdfaq.html#what-file-format-is-my-usd-file).

enrac5:
Okay, so just to be clear, you're suggesting using the API to build the file (not to manually encode the file) and there's a flag to set it to USDC? Do you have an example in Python? The link you sent uses the C++ API.

spiffmon:
Sure!  As noted in the link I sent, the default format for .usd files is crate/usdc, so all you need to do is:
```
stage = Usd.Stage.CreateNew("big_file.usd")
# add all the stuff to your scene
stage.Save()

jesschimein:
Filed as internal issue #USD-10847

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

PR #3934, opened 2026-01-07
Support reading large USDA files

dgovil:
### Description of Change(s)

We were seeing failures when reading usda files great than INT_MAX on macos because the handling of larger buffers was implementation defined. The implementation here should resolve the issue instead and be more portable, allowing reads of much larger files.

This shouldn't cause any performance regressions since pread on linux also caps out at INT_MAX, but just handles the capping internally to itself.

Thanks to Nate Reid for the patch.

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/3582

### Checklist


jesschimein:
Filed as internal issue #USD-11755

PR #3937, opened 2026-01-07
cmake: modernize Boost usage

davvid:
CMake policy CMP0167 is deprecated. Newer versions of CMake do not provide FindBoost.cmake. Use the cmake configuration provided by the upstream Boost project.

Related-to: https://github.com/aousd/build-ig-initiatives/issues/13


jesschimein:
Filed as internal issue #USD-11758

pmolodo:
So... just to make sure I understand CMP0167 correctly... with it set to new, how will the `find_package(Boost)` call work if the build has configured things such that their own / custom FindBoost.cmake is somewhere on the module path, AND the config is found?

ie, with CMP0167 set to new - does it just make it so that cmake's own FindBoost.cmake effectively doesn't exist... but doesn't change the priority of old FindBoost.cmake vs boostConfig.cmake... so that if both are found, FindBoost.cmake still takes priority?

Or does it actually alter priority of FindBoost.cmake vs boostConfig.cmake, so that even if a custom FindBoost.cmake is on the CMAKE_MODULE_PATH, boostConfig.cmake is now always preferred?

davvid:
CMP0167 = NEW should only remove CMake's own built-in FindBoost.cmake from its internal module search. It should not alter the fundamental Module vs. Config mode priority in find_package().

This makes older cmake behave like the newer cmake versions where FindBoost.cmake is no longer provided.

The search order for find_package(Boost) (no CONFIG/MODULE keyword) remains:
1. Module mode first: scans CMAKE_MODULE_PATH, then CMake's own modules dir
2. Config mode fallback: scans for BoostConfig.cmake / boost-config.cmake

With CMP0167 = NEW, step 1 should no longer find anything in CMake's modules. A custom FindBoost.cmake on CMAKE_MODULE_PATH should still win as it's found in step 1 before config mode is ever attempted.

So if your build has FindBoost.cmake somewhere in CMAKE_MODULE_PATH, CMP0167 = NEW should not change anything. That custom module still takes priority over upstream's BoostConfig.cmake.

This is my understanding at least. The docs don't completely spell out the config vs. module details, but they do specifically say that the new policy makes it not find cmake's FindBoost.cmake module.

https://cmake.org/cmake/help/v3.30/policy/CMP0167.html

pmolodo:
Ok, makes sense to me...

This PR gets my vote of approval...

PR #3938, opened 2026-01-07
cmake: use the Ptex module provided by upstream Ptex

davvid:
Ptex has provided its own cmake packaging since 2018 in v2.3.1. We don't need to ship or maintain our own FindPTex.cmake these days.

Related-to: https://github.com/aousd/build-ig-initiatives/issues/13


jesschimein:
Filed as internal issue #USD-11759

PR #3971, opened 2026-02-03
crate: issue error for duplicate key in crate map

pmolodo:
### Description of Change(s)

crate: issue error for duplicate key in crate map

...if PXR_PREFER_SAFETY_OVER_SPEED is set

Currently, when the [crate file format reads maps or variant selections](https://github.com/PixarAnimationStudios/OpenUSD/blob/v25.11/pxr/usd/sdf/crateFile.cpp#L1184-L1186), it [does no checking for duplicate keys](https://github.com/PixarAnimationStudios/OpenUSD/blob/v25.11/pxr/usd/sdf/crateFile.cpp#L1140-L1151).

This came up in the AOUSD core spec working group, where it was determined that this was likely:

- unintentional (or at least, a situation the original OpenUSD authors didn't really care about)
- changing behavior by making it error would likely have little practical impact (besides performance implications), as we expect there are very few actual assets with this condition.

Therefore, we were wondering if OpenUSD would consider making the crate reader error when encountering such data, for clearer alignment with the spec.

### Checklist

### DIFF
--- a/pxr/usd/sdf/crateFile.cpp
+++ b/pxr/usd/sdf/crateFile.cpp
@@ -1145,6 +1145,13 @@ public:
             // separate because the two modifications to 'src' must be correctly
             // sequenced.
             auto key = Read<typename Map::key_type>();
+            if constexpr (SafetyOverSpeed) {
+                if map.find(key) != map.end() {
+                    TF_RUNTIME_ERROR("Corrupt asset <%s>: Duplicate map key %s",
+                                     crate->GetAssetPath().c_str(),
+                                     key.GetText().c_str());
+                }
+            }
             map[key] = Read<typename Map::mapped_type>();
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-11858

spiffmon:
@pmolodo , in fact it's impossible to get such a "corrupt" file using the OpenUSD API's - malicious intent/hacking would be required.  Given that, and that it cannot be exploited for harm (I think), is the extra check really needed?

nvmkuruc:
Hi @spiffmon, I think I would more frame it as software isn't immune from bugs. It's just something that could happen while mapping the crate format to sdf.  Emitting a warning seems reasonable to me, but I don't know that I would prioritize it.

Issue #3970, opened 2026-02-03
UsdGeomMotionAPI should have an API schema adapter


robp-sidefx:
At present, the data associated with UsdGeomMotionAPI only ends up on Hydra prims via UsdImagingGprimAdapter which means it can't be used to provide information to, for example, a Hydra Generative Procedural.

I'd like to propose an API schema adapter to serve this data translation purpose instead.

I'll submit a basic PR shortly, but I desperately need expertise & insight into what's missing and whether there are any "blockers" that would suggest it can't/shouldn't be moved out of GprimAdapter.

jesschimein:
Filed as internal issue #USD-11857

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

spiffmon:
Hey @robp-sidefx , is there more work to do here that you are hoping comes from us?

robp-sidefx:
All good ... sorry, I'd not even realised this hadn't been closed alongside the PR closing.

PR #3974, opened 2026-02-04
Adding MotionAPIAdapter

robp-sidefx:
### Description of Change(s)

Adding a MotionAPIAdapter so that the schema can be applied to arbitrary prims and the data will survive translation via Hydra (in particular we want it for GenerativeProcedural prims).

### Fixes Issue(s)

#3970 

### Checklist

### DIFF
--- a/pxr/usdImaging/usdImaging/CMakeLists.txt
+++ b/pxr/usdImaging/usdImaging/CMakeLists.txt
@@ -106,6 +106,7 @@ pxr_library(usdImaging
         materialBindingsSchema
         materialBindingsResolvingSceneIndex
         meshAdapter
+        motionAPIAdapter
         niPrototypePropagatingSceneIndex
         nurbsCurvesAdapter
         nurbsPatchAdapter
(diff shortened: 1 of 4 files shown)


spiffmon:
I am no expert, but:
 a) what you have looks right to me (thank you for not including `velocityScale`, which we really need to remove...)
 b) the thing that seems missing, but I'm not sure where it needs to get added, is logic to propagate the data down namespace, as all of the properties are "inherited" to descendant prims

jesschimein:
Filed as internal issue #USD-11864

robp-sidefx:
Thanks @spiffmon ... good catch! The latest commit should address the inheritance issue.

robp-sidefx:
Any further thoughts on this?

tgvarik:
Sorry @robp-sidefx, I'm taking this up now. Shouldn't be too long.

tgvarik:
I went ahead and also removed the handling of the MotionAPI attributes from the general Gprim datasource in usdImaging, which means MotionAPI will need to actually be applied to the prim for these two to show up (via your new adapter). Passes internally and should show up in the next set of commits.

Sorry for the delay!!

Issue #3976, opened 2026-02-09
Support reading "Ids" from UsdGeomPointInstancer in shaders


crydalch:
UsdGeomPointInstancer has an "ids" attribute, which is not available by default to shaders. This necessitates users author `primvars:id` or `primvars:ids` separately. It would be great if this were made available to shaders without requiring an additional Primvar authoring, if possible. A long time ago the idea of was discussed on the usd-interest mailing list and validated. However, it turns out I never filed an issue for it - my apologies, making up for lost time now :)

PR addressing this incoming shortly.

Thanks!

jesschimein:
Filed as internal issue #USD-11871

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

PR #3977, opened 2026-02-09
Makes the 'ids' for a UsdGeomPointInstancer available to shaders

crydalch:
### Description of Change(s)

Makes the 'ids' for a UsdGeomPointInstancer available to shaders, without requiring users author a separate Primvar for this purpose. The mesh has matching values for ids, authored as a Primvar, just to visually ensure we're getting the same results.

Storm_PointInstancer_IDs

It doesn't preclude authoring a different `primvars:ids`, as explicitly authored `primvars:ids` take precedence. We also did some testing with an internal build of Karma, and are seeing the same, expected behavior.

Storm_PointInstancer_IDs_existing

Here are the scene files we used for testing:

[PointInstancer_IDs_test_scenes.zip](https://github.com/user-attachments/files/25190536/PointInstancer_IDs_test_scenes.zip)

### Fixes Issue(s)

#3976

### Storm

In order to validate this with Storm, we had to make some changes, which are in this patch. However we aren't sure if the Storm/UsdMaterialX changes are the best way forward. We leave the patch here, for any interested parties to pick up as a starting point for having Storm to support integer primvars, as we probably aren't best situation to shepherd this into the baseline :)

[ids_Storm_MtlX_support.patch](https://github.com/user-attachments/files/25190485/ids_Storm_MtlX_support.patch)

### DIFF
--- a/pxr/imaging/hd/tokens.h
+++ b/pxr/imaging/hd/tokens.h
@@ -55,6 +55,7 @@ PXR_NAMESPACE_OPEN_SCOPE
     (geometry)                                  \
     (hermite)                                   \
     (hullIndices)                               \
+    (ids)                                       \
     (indices)                                   \
     (isFlipped)                                 \
     (itemsDrawn)                                \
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-11872

tcauchois:
Storm changes look reasonable except that we already use int primvars; need to dig a bit into why they aren't working for you!

tcauchois:
It's slightly awkward that they are 64-bit.  I think we need to add an extension pragma, and probably datatype packing into uvec2/ivec2, in order to make it portable.

PR #3992, opened 2026-02-18
APPLE: Add Polygonal Mesh Compression support from Alliance for OpenMedia VVM Group

dgovil:
### Description of Change(s)

This PR adds support for the Polygonal Mesh Coding (PMC) standard as developed by the Alliance for Open Media (AOMedia) ’s VVM Working Group (WG).

Press Release: https://aomedia.org/press%20releases/AOMedia-Releases-Reference-Software-for-Polygonal-Mesh-Coding-Standard/

The PMC standard offers significant size reductions compared to other mesh compression codecs. We will provide comparative metrics in a subsequent post to assess its performance against existing solutions. Notably, the PMC standard was designed from the outset to be fully compatible with USD features, including holes, creases, and GeomSubsets.

We add the following:

1. A Plugin for PMC encoding and decoding. The file format plugin is not limited to USDZ files.
2. Support in build_usd.py to install this plugin, along with a downloaded dependency on the codec library. This is enabled by default in this PR for easier testing.
3. A new command line tool, tentatively called usdcrush to align with a similar tool in macOS 14+. While these command line tools have different implementations, we think there’s some benefits in sharing a name + arguments to socialize use, but are open to other names. 
This will flatten stages on import, but supports reading/writing usd compatible formats.
4. Extended the USDZ spec list to include PMC files as an allowed file format

In addition to the AOMedia VVM Group, I'd like to especially credit Khaled Mammou, David Flynn, Christian Tulvan, Zhenzhen Gao and Valery Valentin on our team for their work on this plugin.

For now we have the following graph to show results.
These are assets from our [gallery](https://developer.apple.com/augmented-reality/quick-look/) with textures removed so that we can focus primarily on mesh sizes.

image (2)

### Checklist


jesschimein:
Filed as internal issue #USD-11905

e4lam:
Is there online link somewhere describing the PMC spec?

dgovil:
@e4lam We're working towards that. There was a bit of a timing mismatch but I'll update this PR once we can get more information available.

dgovil:
As a note, I removed the USDZ only limits of the `usdcrush` command line tool

pxr/bin/usdcrush/usdcrush.cpp
matthewcpp, line 1: Would it be possible to move the code for usdcrush into pxr/bin?  We have started to migrate a few of the tools there already, and going forward see this location as the home for new tools such as this.

pxr/usd/plugin/usdPmc/testenv/testUsdPmc.py
matthewcpp, line 1: The tests here look good, but perhaps would be more suited for inclusion with the usdPmc plugin?  They seem to be exercising the functionality of the plugin, as opposed to the usdCrush application itself.

If we were going to add tests for usdcrush itself, an option would be to capture and compare output with a baseline. [usdzip's CMakeLists.txt](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/bin/usdzip/CMakeLists.txt) has an example of this.

pxr/usd/bin/usdcrush/usdcrush.cpp
```diff
@@ -0,0 +1,94 @@
...
+        ->required(true);
+
+    app->add_option(
+        "--meshes", args.meshes, 
+        "Controls whether meshes are processed for optimization.")
+        ->default_val(true);
+}
+
+// Quarantine a potentially corrupt file by renaming it
+static void Quarantine(const std::string &filepath) {
```
matthewcpp: I think we might be able to remove this function, it doesn't seem to be called.

matthewcpp (commented):
Thanks for submitting this PR!  I will be working on getting it merged in, with a target of the 26.08 release.  I left a few initial high level structural comments.  I will likely follow up with some additional detailed comments next week.

dgovil:
Thanks, @matthewcpp! I'll hit those notes up this week and rebase on latest dev in the process.

dgovil:
@matthewcpp Okay I think I hit up all the notes, but let me know if you have more! Cheers

pxr/usd/plugin/usdPmc/testenv/testUsdPmc.py
dgovil: Ah sorry, I need to hit this one up quick. I thought I'd changed it to not call usdcrush directly, but I lost the change while rebasing.

dgovil: Okay restored my missing stuff. Should be good to go now

pxr/usd/plugin/usdPmc/usdPmcEncoder.cpp
```diff
@@ -0,0 +1,718 @@
...
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+/// Creates a unique temporary directory path using the system's temporary
+/// directory and a random identifier to ensure uniqueness across concurrent
+/// encoding operations. This implementation is platform-agnostic and works
+/// on macOS, Windows, and Linux.
+std::string
+_GetTempDir() {
```
matthewcpp: Do you think it would make sense to make use of [ArchMakeTmpSubdir](https://openusd.org/release/api/group__group__arch___system_functions.html#gaf16da5b9f8a5f90356e73952519215ea) along with `ArchGetTmpDir()` in lieu of this function?

pxr/usd/plugin/usdPmc/usdPmcDecoder.hpp
```diff
@@ -0,0 +1,224 @@
...
+    bool CanDecode(const char* buffer, size_t length);
+    
+    /// Decode PMC buffer into UsdGeomMesh.
+    /// \param buffer Pointer to the PMC data buffer
+    /// \param length Size of the buffer in bytes
+    /// \param decodedMesh The UsdGeomMesh to populate with decoded data
+    /// \return true if decoding was successful, false otherwise
+    bool Decode(const char* buffer, size_t length, UsdGeomMesh& decodedMesh);
+
+protected:
```
matthewcpp: I think these could be private instead of protected

matthewcpp (commented):
Thanks for making that first round of changes, they look good to me!  i looked at the plugin and found only some minor stuff. Going to see if any folks on the team here have additional thoughts which i will relay here.

pxr/usd/plugin/usdPmc/usdPmcEncoder.cpp
dgovil: Yeah I think so. Sorry missed that one on internal review when porting open. I'll swap that over

pxr/usd/plugin/usdPmc/usdPmcDecoder.hpp
dgovil: @matthewcpp addressed both notes. Thanks!

dgovil:
A press release is now available for PMC https://aomedia.org/press%20releases/AOMedia-Releases-Reference-Software-for-Polygonal-Mesh-Coding-Standard/

cmake/defaults/Packages.cmake
```diff
@@ -307,6 +307,11 @@ if (PXR_BUILD_ANIMX_TESTS)
     find_package(AnimX REQUIRED)
 endif()
 
+if ( PXR_BUILD_PMC_PLUGIN)
+    cmake_policy(SET CMP0144 NEW)
```
matthewcpp: When i tried to build the library on my workstation i got an error that this policy was not known to CMake.  This was using CMake 3.26.  It seems like this policy was added in 3.27.  When I commented it out, the build seemed to work fine.  Do you know if this is indeed needed? if we do need it here and/or in the other location, i think we will need to guard it on Cmake version.

pxr/usd/usdPmc/CMakeLists.txt
matthewcpp, line 1: So after some internal discussion, we decided that we ould like to request usdPmc be moved to `pxr/usd/usdPmc`.  The rationale behind this is that since we are adding a new tool which leverages this code, we see it more as a library than a plugin.

pxr/usd/plugin/usdPmc/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+set(PXR_PREFIX pxr/usd)
+set(PXR_PACKAGE usdPmc)
+
+cmake_policy(SET CMP0144 NEW)
```
matthewcpp: I left another detailed comment relating to this in `Packages.cmake`.  Tagging this line here for reference to the same policy

pxr/usd/plugin/usdPmc/usdPmcDecoder.hpp
```diff
@@ -0,0 +1,224 @@
...
+    /// \param length Size of the buffer in bytes
+    /// \return true if the buffer contains valid PMC data, false otherwise
+    bool CanDecode(const char* buffer, size_t length);
+    
+    /// Decode PMC buffer into UsdGeomMesh.
+    /// \param buffer Pointer to the PMC data buffer
+    /// \param length Size of the buffer in bytes
+    /// \param decodedMesh The UsdGeomMesh to populate with decoded data
+    /// \return true if decoding was successful, false otherwise
+    bool Decode(const char* buffer, size_t length, UsdGeomMesh& decodedMesh);
```
matthewcpp: The internal coding conventions we adhere to in the studio require us to use pointers in lieu of non-const references for output parameters.  Unfortunately, we do not explicitly mention this in in our coding conventions.  Would you mind fixing up the output parameters in `usdPmc` functions to use pointers?  We are going to work on getting the coding convention document updated to avoid any potential uncertainty on this in the future

pxr/usd/plugin/usdPmc/usdPmcDecoder.hpp
```diff
@@ -0,0 +1,224 @@
...
+};
+
+/// Convert USD buffer to PMC ArrayBuffer format.
+/// \param usdBuffer The USD buffer to convert
+/// \param cpv Components per vector
+/// \param dt PMC data type
+/// \return PMC ArrayBuffer structure
+template<typename T>
+inline pmc::ArrayBuffer
+ToPmcBuffer(T& usdBuffer, int cpv, pmc::DataType dt) {
```
matthewcpp: Even though this is a private header, if these functions are intended to be used by other files in the library we'd still prefix the names with `UsdPmc_` , e.g. `UsdPmc_ToPmcBuffer`.

If these aren't being used by other files, then we should move these to an anonymous namespace within the .cpp file.

pxr/usd/plugin/usdPmc/usdPmcDecoder.hpp
matthewcpp, line 1: We prefer `.h` for non third party vendored header files. Would you mind updating the file names in `usdPmc`?

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+#include <type_traits>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+namespace {
+
+/// Template utility for validating expected values during PMC encoding
+/// operations.
+template<typename T>
+struct Expect {
```
matthewcpp: For file-private stuff, we'd typically prefix function/class names with `_`.

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+{};
+
+template<typename T>
+constexpr auto oneextent_v = oneextent<T>::value;
+
+/// Analyzes the VtValue type to determine how many components each element has.
+int
+GetExtentFromType(const VtValue& vtv)
+{
+    switch (vtv.GetKnownValueTypeIndex()) {
```
matthewcpp: We should be able to replace this with [VtVisitValueType](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/vt/visitValue.h#L134) now, which should also let us get rid of the VtVtindex_v helper.

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+        std::vector<MinMax> minmax(1);
+        for (const auto v : vta) {
+            minmax[0].min = std::min(minmax[0].min, double(v));
+            minmax[0].max = std::max(minmax[0].max, double(v));
+        }
+        return minmax;
+    };
+
+    switch (vals.GetKnownValueTypeIndex()) {
+#define CASE(T, fn) VtVtindex_v<T>: return fn(vals.UncheckedGet<T>())
```
matthewcpp: We should be able to leverage [VtVisitValue](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/vt/visitValue.h#L41) here.

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+    }
+};
+/// Sequentially apply op to each elementary value in the VtArray.
+
+template<typename T, typename Op>
+void
+ScanVtArray(const VtArray<T>& vta, Op op)
+{
+    for (const auto& sv : vta) {
+        if constexpr (GfIsGfVec<T>::value)
```
matthewcpp: Conventionally use `{}` for single line bodies

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+    result.dataType = pmc::DataType::Int32;
+    return result;
+}
+/// Convert source array to destination vector<int> with optional scaling.
+
+pmc::ArrayBuffer
+ToPmc(const VtValue& src, std::vector<int>& dst, int fracbits)
+{
+    switch (src.GetKnownValueTypeIndex()) {
+#define CASE(T, op) VtVtindex_v<T>: return ToPmc(src.UncheckedGet<T>(), dst, op)
```
matthewcpp: Can we leverage `VtVisitValue` or `VtVisitValueType` here?

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+}
+
+pmc::MeshFaceType
+GetMeshFaceTypeFromFaceVertexCounts(const VtArray<int> fvcs)
+{
+    const auto minmax = std::minmax_element(fvcs.cbegin(), fvcs.cend());
+    const int min = *minmax.first;
+    const int max = *minmax.second;
+
+    if (max == min && min == 3) return pmc::MeshFaceType::TRIANGULAR;
```
matthewcpp: (here and below) code convention is to use braces and put the body on a separate line

pxr/usd/plugin/usdPmc/usdPmcEncoder.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+/// \class UsdPmcMeshEncoder
+///
+/// Encode UsdGeomMesh data into PMC compressed format.
+///
+/// This class provides functionality to encode USD geometry into PMC
+/// compressed mesh data. It handles the conversion of USD
+/// geometry, attributes, and metadata into the PMC format for efficient
+/// storage and transmission.
+class USDPMC_API UsdPmcMeshEncoder {
```
matthewcpp: Per convention we should put API exports on individual functions instead of the entire class.

pxr/usd/usdPmc/usdPmcFileFormat.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return true;
+    } catch(...) {
+        TF_RUNTIME_ERROR("Exception: Failed to read from PMC file \"%s\"",
+                         resolvedPath.c_str());
+        return false;
+    }
+}
+
+bool
+UsdPmcFileFormat::ReadFromString(SdfLayer* layer,
```
matthewcpp, line 106: This is not super well documented, but `ReadFromString `isn't intended to take a binary blob and import it into the layer. It's the counterpart to `WriteToString`, so this function should be able to read whatever that function generates and put it into the layer. Since `WriteToString` is forwarding to the usda format (as it should), `ReadToString` should do the same.

pxr/usd/plugin/usdPmc/usdPmcFileFormat.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return false;
+    }
+}
+
+// Defer to the usda file format for this.
+bool
+UsdPmcFileFormat::WriteToString(const SdfLayer& layer,
+                                std::string* str,
+                                const std::string& comment) const {
+    try {
```
matthewcpp: No need to try/catch here since the `.usda` doesn't throw exceptions (and won't, per our coding conventions)

pxr/usd/plugin/usdPmc/usdPmcFileFormat.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return false;
+    }
+}
+
+// Defer to the usda file format for this.
+bool
+UsdPmcFileFormat::WriteToStream(const SdfSpecHandle& spec,
+                                std::ostream& out,
+                                size_t indent) const {
+    try {
```
matthewcpp: No need to try/catch here since the `.usda` doesn't throw exceptions

pxr/bin/usdcrush/usdcrush.cpp
```diff
@@ -0,0 +1,76 @@
...
+    app->add_option(
+        "inputFile", args.inputFile, "The input USD file to process.")
+        ->required(true);
+
+    app->add_option(
+        "-o,--out", args.outputFile,
+        "The output USD file to write to.")
+        ->required(true);
+
+    app->add_option(
```
matthewcpp, line 43: This program seems to exit if --meshes is disabled – is there a reason we need this argument?

matthewcpp (commented):
I added comments that came in from my own testing and internal review of this change.  Appreciate you taking a look at them, and let me know if you have any additional questions!

Thanks!

pxr/bin/usdcrush/usdcrush.cpp
dgovil: Our thinking was that you might have other forms of compression in the future, so you'd have toggles for each kind. 

But we could remove it for now and add it in the future if needed.

pxr/usd/plugin/usdPmc/usdPmcEncodeSession.cpp
```diff
@@ -0,0 +1,747 @@
...
+    // Ignore constant primitives (no point encoding them)
+    if (pv.GetInterpolation() == pxr::UsdGeomTokens->constant) {
+        return;
+    }
+
+    auto pvName = pv.GetPrimvarName();
+
+    const auto vals = GetAs<VtValue>(pv);
+    const auto idxs = GetAs<VtArray<int>>(pv.GetIndicesAttr());
+    if (!vals.IsArrayValued()) {
```
matthewcpp: I was testing usdcrush on some meshes here at the studio and was hitting the exception thrown in `GetExtentFromType`.  This was due to the presence primvars of type `string[]`

I changed this line to:
```
if (!vals.IsArrayValued() || vals.IsHolding<VtArray<std::string>>() || vals.IsHolding<VtArray<TfToken>>()) {
        return;
    }
```
and everything seemed to work fine.   I am not sure if this is the best way to handle this particular situation, but thought I would include the change i made as a potential point of departure.

dgovil:
We addressed a bunch of the notes, but need to update the codec version so will have one more set of changes coming soon. After which I'll clean up the merge conflcits

dgovil:
@matthewcpp Apologies for the delay, I've updated the PR with the latest round  of notes, as well as an updated codec version from the AOMedia and some more stringent checking / failure case handling.

cmake/defaults/Packages.cmake
```diff
@@ -335,6 +335,10 @@ if(PXR_ENABLE_OSL_SUPPORT)
     add_definitions(-DPXR_OSL_SUPPORT_ENABLED)
 endif()
 
+if (PXR_BUILD_USDPMC)
+    find_package(Pmc REQUIRED)
```
nvmkuruc, line 329: @dgovil Does Pmc have a `CONFIG`?

dgovil: Unfortunately, no. I'll make a request of the team to add it for the future, but for now its just a emitting static libs and headers

nvmkuruc: Thanks @dgovil

PR #3994, opened 2026-02-23
cmake: embed VERSIONINFO resources in Windows DLLs

imcalderon:
## Summary

USD shared libraries and plugins built on Windows currently ship with no
embedded version or identity metadata. Right-clicking any `usd_*.dll` in
Windows Explorer shows a blank Version tab — no file version, no company
name, no description, no copyright. Tools like WinDbg, Dependency Walker,
and Process Explorer also show nothing useful, making it harder to identify
which USD build a DLL came from.

**Root cause:** `_pxr_library()` in `cmake/macros/Private.cmake` never
generates or links a Windows VERSIONINFO resource. Version numbers already
exist in `cmake/defaults/Version.cmake` and are wired into C++ headers via
`pxr/pxr.h.in`, but are not propagated into the DLL binary metadata.

### `cmake/macros/WindowsVersionInfo.rc.in` (new)

A per-target RC template configured by CMake with library-specific values:

| Field | Example value |
|---|---|
| `FileVersion` / `ProductVersion` | `0.26.3.0` (from `PXR_*_VERSION`) |
| `CompanyName` | `Pixar Animation Studios` |
| `FileDescription` | `Universal Scene Description: usd_gf.dll` |
| `InternalName` | `usd_gf` |
| `OriginalFilename` | `usd_gf.dll` |
| `ProductName` | `Universal Scene Description (USD)` |
| `LegalCopyright` | `Copyright 2016 Pixar` |

### `cmake/macros/Private.cmake` (modified)

After `set_target_properties()` in `_pxr_library()`, for WIN32 non-static
non-object targets, a unique `.rc` file is generated in
`CMAKE_CURRENT_BINARY_DIR` and added to the target via `target_sources()`:

```cmake
configure_file(WindowsVersionInfo.rc.in  ${NAME}_version.rc  @ONLY)
target_sources(${NAME} PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/${NAME}_version.rc)
```

Each library gets its own `.rc` to avoid collisions in parallel builds.
STATIC and OBJECT targets are excluded (no DLL produced). The block is
guarded by `if(WIN32)` so Linux and macOS builds are unaffected.

## Testing

Built with MSVC + Ninja on Windows 10. After the change, right-clicking
`usd_sdf.dll` -> Properties -> Details shows:

- File version: `0.26.3.0`
- Product version: `0.26.3.0`
- Company: `Pixar Animation Studios`
- File description: `Universal Scene Description: usd_sdf.dll`
- Original filename: `usd_sdf.dll`
- Product name: `Universal Scene Description (USD)`
- Copyright: `Copyright 2016 Pixar`


jesschimein:
Filed as internal issue #USD-11916

PR #4002, opened 2026-03-02
Physics parser rigid body collider multithreaded fix

pmolodo:
### Description of Change(s)

There is a bug in the multithreaded code for parsing UsdPhysics colliders - when multiple colliders are under the same rigid body, the multithreaded code can call `std::vector::push`, leading to possible race conditions / crash if the vector has to re-allocate.

I've included a unittest to replicate the issue.

### Fixes Issue(s)

- multithreaded crash when parsing multiple colliders under same rigid body

### Checklist

### DIFF
--- a/pxr/usd/usdPhysics/parseUtils.cpp
+++ b/pxr/usd/usdPhysics/parseUtils.cpp
@@ -2957,7 +2957,8 @@ bool LoadUsdPhysicsFromRange(const UsdStageWeakPtr stage,
     // first get the type
     std::vector<UsdPhysicsObjectType> collisionTypes;
     collisionTypes.resize(collisionPrims.size());
-    std::vector<TfToken> customTokens;
+    std::vector<TfToken> customGeomTokens;
+    customGeomTokens.resize(collisionPrims.size());
     {
         const auto workLambda = [&](const size_t beginIdx, const size_t endIdx)
         {
@@ -2967,18 +2968,18 @@ bool LoadUsdPhysicsFromRange(const UsdStageWeakPtr stage,
                 {
                     TfToken shapeToken;
                     const UsdPhysicsObjectType shapeType =
-                        _GetCollisionType(collisionPrims[i], 
-                                         &customPhysicsTokens->shapeTokens, 
+                        _GetCollisionType(collisionPrims[i],
+                                         &customPhysicsTokens->shapeTokens,
                                          &shapeToken);
                     collisionTypes[i] = shapeType;
                     if (shapeType == UsdPhysicsObjectType::CustomShape)
                     {
-                        customTokens.push_back(shapeToken);
+                        customGeomTokens[i] = shapeToken;
                     }
                 }
                 else
                 {
-                    collisionTypes[i] = _GetCollisionType(collisionPrims[i], 
+                    collisionTypes[i] = _GetCollisionType(collisionPrims[i],
                                                          nullptr, nullptr);
                 }
             }
@@ -2999,6 +3000,7 @@ bool LoadUsdPhysicsFromRange(const UsdStageWeakPtr stage,
     std::vector<UsdPrim> meshShapePrims;
     std::vector<UsdPrim> spherePointsShapePrims;
     std::vector<UsdPrim> customShapePrims;
+    std::vector<TfToken> customTokens;
     for (size_t i = 0; i < collisionTypes.size(); i++)
     {
         UsdPhysicsObjectType type = collisionTypes[i];
@@ -3052,6 +3054,7 @@ bool LoadUsdPhysicsFromRange(const UsdStageWeakPtr stage,
         case UsdPhysicsObjectType::CustomShape:
         {
             customShapePrims.push_back(collisionPrims[i]);
+            customTokens.push_back(customGeomTokens[i]);
         }
         break;
         case UsdPhysicsObjectType::SpherePointsShape:
--- a/pxr/usd/usdPhysics/parseUtils.cpp
+++ b/pxr/usd/usdPhysics/parseUtils.cpp
@@ -2045,17 +2045,20 @@ void _FinalizeCollisionDescs(
     const size_t numPrimPerBatch = 10;
     WorkParallelForN(physicsPrims.size(), workLambda, numPrimPerBatch);
 
-    // Merge into bodyDesc->collisions single-threaded. Sort by body pointer
-    // so all collisions for the same body are consecutive (better locality).
+    // Merge into bodyDesc->collisions single-threaded. Sort by body path
+    // so all collisions for the same body are consecutive (better locality)
+    // and by collision path for deterministic ordering.
     if (!bodyCollisionPairs.empty())
     {
-        std::vector<BodyCollisionPair> sorted(
-            bodyCollisionPairs.begin(), bodyCollisionPairs.end());
-        std::sort(sorted.begin(), sorted.end(),
+        std::sort(bodyCollisionPairs.begin(), bodyCollisionPairs.end(),
             [](const BodyCollisionPair& a, const BodyCollisionPair& b) {
-                return a.first < b.first;
+                if (a.first->primPath < b.first->primPath)
+                    return true;
+                if (b.first->primPath < a.first->primPath)
+                    return false;
+                return a.second < b.second;
             });
-        for (const BodyCollisionPair& pair : sorted)
+        for (const BodyCollisionPair& pair : bodyCollisionPairs)
         {
             pair.first->collisions.push_back(pair.second);
         }
--- a/pxr/usd/usdPhysics/parseUtils.cpp
+++ b/pxr/usd/usdPhysics/parseUtils.cpp
@@ -55,6 +55,9 @@
 #include "pxr/base/work/dispatcher.h"
 #include "pxr/base/work/loops.h"
 
+#include <algorithm>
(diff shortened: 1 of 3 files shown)


jesschimein:
Filed as internal issue #USD-11936

pmolodo:
Updated with an additional commit from @AlesBorovicka, which fixes a similar multi-threaded issue with customGeomToken, and adds a test.

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1984,13 +1987,19 @@ void _FinalizeCollision(UsdStageWeakPtr stage,
...
+    const std::map<SdfPath, std::unordered_set<SdfPath,
     SdfPath::Hash>>& collisionGroups)
 {
-    const auto workLambda = [physicsPrims, &physicsDesc, bodyMap, 
-        collisionGroups]
+    // Collect (body, collision path) pairs in a thread-safe container so we
+    // avoid concurrent push_back on bodyDesc->collisions (not thread-safe).
+    using BodyCollisionPair =
+        std::pair<UsdPhysicsRigidBodyDesc*, SdfPath>;
+    tbb::concurrent_vector<BodyCollisionPair> bodyCollisionPairs;
```
tallytalwar, line 1999: I updated this internally to do a reserve on physicsPrims.size().
```
bodyCollisionPairs.reserve(physicsPrims.size());
```

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1984,13 +1987,19 @@ void _FinalizeCollision(UsdStageWeakPtr stage,
...
+
+    const auto workLambda = [physicsPrims, &physicsDesc, bodyMap,
+        collisionGroups, &bodyCollisionPairs]
```
tallytalwar, line 2002: I know these cpatures were there before, but I think `physicsPrims`, `bodyMap` and `collissionGroups` can be captured by reference, or am I missing something?

Noting that all these 3 are const ref in `_FinalizeCollisionDescs`.

lmk if you folks agree and I can make this change internally when merging this PR.

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1984,13 +1987,19 @@ void _FinalizeCollision(UsdStageWeakPtr stage,
 // Finalize the collision desc, run in parallel
 template <typename DescType>
 void _FinalizeCollisionDescs(
-    UsdGeomXformCache& xfCache, const std::vector<UsdPrim>& physicsPrims, 
+    UsdGeomXformCache& xfCache, const std::vector<UsdPrim>& physicsPrims,
```
tallytalwar, line 1990: This is also from before, so apologies for not asking about it before. I am forgetting and can't seem to figure out whats the use of `xfCache` here? I do not see it being used at all.

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -2035,6 +2044,22 @@ void _FinalizeCollisionDescs(
...
+    // Merge into bodyDesc->collisions single-threaded. Sort by body pointer
+    // so all collisions for the same body are consecutive (better locality).
+    if (!bodyCollisionPairs.empty())
+    {
+        std::vector<BodyCollisionPair> sorted(
+            bodyCollisionPairs.begin(), bodyCollisionPairs.end());
+        std::sort(sorted.begin(), sorted.end(),
+            [](const BodyCollisionPair& a, const BodyCollisionPair& b) {
+                return a.first < b.first;
+            });
```
tallytalwar, line 2060: Any reason of not sorting `bodyCollisionPairs` directly? I think `tbb::concurrent_vector` does support `std::sort` and / or `tbb::parallel_sort`. Parallel sort might be overkill but main question is about why not directly sort `bodyCollisionsPairs instead of copying to another vector.

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -2035,6 +2044,22 @@ void _FinalizeCollisionDescs(
...
+
+    // Merge into bodyDesc->collisions single-threaded. Sort by body pointer
+    // so all collisions for the same body are consecutive (better locality).
+    if (!bodyCollisionPairs.empty())
+    {
+        std::vector<BodyCollisionPair> sorted(
+            bodyCollisionPairs.begin(), bodyCollisionPairs.end());
+        std::sort(sorted.begin(), sorted.end(),
+            [](const BodyCollisionPair& a, const BodyCollisionPair& b) {
+                return a.first < b.first;
```
tallytalwar: We are sorting by first, which is a pointer, do we want to have a stable sort order, or it doesn't matter in the usecase its being used right now?

pxr/usd/usdPhysics/parseUtils.cpp
tallytalwar: Irrespective of xfCache used or not here, based on the docs here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdGeom/xformCache.h#L36-L38, it seems its not safe to use it from multiple threads.

pxr/usd/usdPhysics/testenv/testUsdPhysicsParsing.py
```diff
@@ -425,6 +425,25 @@ def test_rigidbody_collision_parse(self):
         self.assertTrue(rigidbody_found)
         self.assertTrue(cube_found)
 
+    def test_rigidbody_collision_multihreading_parse(self):
```
tallytalwar: typo `multithreading`

pxr/usd/usdPhysics/testenv/testUsdPhysicsParsing.py
```diff
@@ -425,6 +425,25 @@ def test_rigidbody_collision_parse(self):
...
+
+        stage = Usd.Stage.CreateInMemory()
+        body = UsdGeom.Xform.Define(stage, "/Body")
+        UsdPhysics.RigidBodyAPI.Apply(body.GetPrim())
+
+        for k in range(NUM_COLLIDERS):
+            sphere = UsdGeom.Sphere.Define(stage, f"/Body/SphereCollider_{k}")
+            UsdPhysics.CollisionAPI.Apply(sphere.GetPrim())
+
+        UsdPhysics.LoadUsdPhysicsFromRange(stage, [Sdf.Path.absoluteRootPath])
```
tallytalwar: Might make sense to also validate the results ... maybe as simple as checking number of parsed collisions is same as NUM_COLLIDERS?

tallytalwar (commented):
This looks good overall, some minor notes, kindly let me know if I should address these internally or if you can take a stab at it?

Thanks

pxr/usd/usdPhysics/parseUtils.cpp
AlesBorovicka: Sure makes sense, thanks!

pxr/usd/usdPhysics/parseUtils.cpp
AlesBorovicka: Yes, its not used, I am aware of that. It used to be used when the code was single threaded, now _GetCollisionShapeLocalTransfrom uses the computation from the UsdGeomXformable as its multithreaded. It can be removed. Do you want me to remove it or can you internally remove it please? Thanks!

pxr/usd/usdPhysics/parseUtils.cpp
AlesBorovicka: Thats a very good point, that could be a problem, the order should be deterministic across runs, so we probably need to change the sorting. @pmolodo can you update the code please or shall I give it a try?

tallytalwar: Kindly let me know how this testing go and if any updates will be posted here? Thanks

pxr/usd/usdPhysics/parseUtils.cpp
tallytalwar: Yup, I can take it out internally.

pxr/usd/usdPhysics/testenv/testUsdPhysicsParsing.py
AlesBorovicka: Added

pxr/usd/usdPhysics/parseUtils.cpp
AlesBorovicka: Redone

pxr/usd/usdPhysics/parseUtils.cpp
AlesBorovicka: Changed to sort by the SdfPath

pxr/usd/usdPhysics/parseUtils.cpp
tallytalwar: Addressed while integrating internally.

pxr/usd/usdPhysics/parseUtils.cpp
tallytalwar: Addressed while integrating internally.

pxr/usd/usdPhysics/parseUtils.cpp
tallytalwar: Addressed while integrating internally.

PR #4022, opened 2026-03-25
APPLE: Add HTTPResolver example

dgovil:
### Description of Change(s)

This adds a new example plugin under extras/usd/examples/httpResolver/ that implements an ArResolver that can load content from Http sources. 

**NOTE:** I have this as an example project because there are possibly edge cases that I haven't handled but I think having a resolver example like this would be useful for various folks in a few domains. I think this would eventually be something that could be valuable to promote from an example to a built-in for USD, but that's perhaps a discussion for the future.

The plugin handles http:// and https:// URI schemes, resolves relative paths against HTTP anchors, and fetches assets into memory using either NSURLSession (Apple default) or libcurl (Linux/Windows/macOS). For USDZ packages, it uses HTTP Range requests to stream only the ZIP central directory and individual sub-assets on demand rather than downloading the full archive. Downloaded assets are cached in-memory (or optionally on disk) to avoid redundant network requests. There is also an optional mode where the server can redirect the resolver to a local file path instead of returning bytes directly.

This plugin uses cURL on Windows and Linux, and Foundation's NSURLSession on Apple Platforms. You can also swap to using cURL on Apple platforms too, but there are several system benefits to using the native API. It's spiritually similar to the wasmFetch resolver, and perhaps in the future they could be merged too.

A self-contained test suite (testHttpResolver) is included. It starts a local HTTP server and exercises all code paths: plain asset loading, USDZ streaming, disk caching, server-filepath hints, and concurrent downloads.

The plugin is enabled by default for the purposes of testing and can be disabled with --no-http-resolver-example.

N/A

### Fixes Issue(s)

N/A

### Checklist


jesschimein:
Filed as internal issue #USD-12026

PR #4036, opened 2026-04-10
Avoid `nonnull` warning in rvalue construction of `TfSmallVector` without local storage

nvmkuruc:
### Description of Change(s)

`TfSmallVector` supports instantiations with no local storage (the capacity template argument set to 0). These are utilized in `pxr/exec`.

The `TfSmallVector` move constructor assumes is that if the sizes of the local storage and the rvalue parameter are the same, the rvalue's data can be moved into the constructing object's local storage. However, if the local storage size is 0, the `_UninitializedMove` tries to move an empty range into non-existent local storage.  In newer versions of GCC, this causes a warning.

```
# Truncated stack trace
inlined from ‘static Iterator pxrInternal_v0_26_5__pxrReserved__::TfSmallVectorBase::_UninitializedMove(Iterator, Iterator, Iterator) [with Iterator = pxrInternal_v0_26_5__pxrReserved__::GfQuatd*]’ at pxr/base/tf/smallVector.h:84:39,
inlined from ‘pxrInternal_v0_26_5__pxrReserved__::TfSmallVector<T, N>::TfSmallVector(pxrInternal_v0_26_5__pxrReserved__::TfSmallVector<T, N>&&) [with T = pxrInternal_v0_26_5__pxrReserved__::GfQuatd; unsigned int N = 0]’ at pxr/base/tf/smallVector.h:245:35
...
error: argument 1 null where non-null expected [-Werror=nonnull]
```

This change introduces a `constexpr` guard to only apply the move if local storage is available.

With this change, the following build command should build without warnings or errors in GCC 13.
```
python3 build_scripts/build_usd.py ../inst/usd-strict --no-imaging --no-examples --no-materialx  --build-args USD,'-DPXR_STRICT_BUILD_MODE=ON -DCMAKE_CXX_FLAGS="-Wno-array-bounds -Wno-uninitialized"' -vv --tests
```

### Checklist

### DIFF
--- a/pxr/base/tf/smallVector.h
+++ b/pxr/base/tf/smallVector.h
@@ -156,7 +156,6 @@ template <typename T, uint32_t N>
 class TfSmallVector : public TfSmallVectorBase
 {
 public:
-
     /// XXX: Functionality currently missing, and which we would like to add as
     ///  needed:
     ///     - emplace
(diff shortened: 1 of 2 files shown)


sunyab:
Filed as internal issue #USD-12078

Issue #4040, opened 2026-04-16
Changing a point instancer into equivalent native instancing makes prims disappear in hydra


marktucker:
Extract the files in this zip file: 
[usd.zip](https://github.com/user-attachments/files/26786143/usd.zip)

1. Run usdview cubes_nativeinstances.usda. Four cubes as expected.
2. Run usdview cubes_pointinstances.usda. Four cubes as expected.
3. In the interpreter window run:
```
l1 = usdviewApi.stage.GetRootLayer()
l2 = Sdf.Layer.FindOrOpen("cubes_nativeinstances.usda")
l1.TransferContent(l2)
```
4. No cubes. This is bad.
5. Reload all layers, and the four cubes appear again.

This happens with Storm, karma, and Houdini's vulkan delegate, so I strongly suspect it is a problem at the hydra level.

jesschimein:
Filed as internal issue #USD-12093

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

blevin:
Fix is merged in our local dev tree and should make its way onto github in the next push.

robp-sidefx:
@blevin did this go in? scanning the may15 & may18 commits i don't see anything relevant

blevin:
Yes, https://github.com/PixarAnimationStudios/OpenUSD/commit/7385066a5847530903a1711375462c2a4f341478 should have the relevant fix.

robp-sidefx:
Oh, brilliant ... it obviously didn't click with me when I scanned right over that commit message :)

jesschimein:
Closing this issue since https://github.com/PixarAnimationStudios/OpenUSD/commit/7385066a5847530903a1711375462c2a4f341478 should address it.

PR #4042, opened 2026-04-17
Better tracking of instancer prims that get retyped via PrimsAdded

robp-sidefx:
### Description of Change(s)

Previously, new entries (in PrimsAdded) were filtered to identify instancers, and those paths were added to a local cache. But this didn't capture the event where an existing instancer was retyped. This change also captures the non-instancers from the new entries, and uses that list to remove entries that might be in the cache from an earlier representation of the scene.

I expect this new code path will get hit a lot (i.e., the vast majority of PrimsAdded prims will _not_ be instancers), so I am mildly nervous about potential performance implications, but I can't immediately come up with a lighter-weight alternative. I can verify that the change resolves the reported issue, but you may have a more elegant solution instead.

### Fixes Issue(s)

#4040 

### Checklist


jesschimein:
Filed as internal issue #USD-12098

PR #4049, opened 2026-04-20
[hgi] On shader compiling errors, output ALL source with line numbers.

### LINKED ISSUE #4048: HGI Shader errors are commonly logged with line numbers but the injected code make it impossible to know reconsile
When a shader fails to compile, the driver normally gives errors that include line numbers.

As the shader is data driven, via the code generation, it is impossible/really-hard to figure out which code the line numbers relate to. 
This is worse if the code being referred to is part of the injected code, as it is not always easy to work out what code that happens to be.

Outputting the whole shader submitted to the driver when there is an error would fix this.

DuncanHopkinsFoundry:
### Description of Change(s)

The normal hgi behaviour is to log just the reported errors. To make these errors faster to deal with, the shader compiling logs the full source, annotated with the same line numbers as the driver error messages.

This output will only be seen if there is a shader compiling error.

hgi GL, Vk, and Metal have been patched to do the same thing.

### Fixes Issue(s)

Fixes #4048

### Checklist

### DIFF
--- a/pxr/imaging/hgiGL/shaderFunction.cpp
+++ b/pxr/imaging/hgiGL/shaderFunction.cpp
@@ -12,6 +12,9 @@
 #include "pxr/imaging/hgiGL/shaderFunction.h"
 #include "pxr/imaging/hgiGL/shaderGenerator.h"
 
+#include <string>
+#include <sstream>
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 HgiGLShaderFunction::HgiGLShaderFunction(
@@ -44,9 +47,26 @@ HgiGLShaderFunction::HgiGLShaderFunction(
     if (status != GL_TRUE) {
         int logSize = 0;
         glGetShaderiv(_shaderId, GL_INFO_LOG_LENGTH, &logSize);
-        _errors.resize(logSize+1);
-        glGetShaderInfoLog(_shaderId, logSize, NULL, &_errors[0]);
+
+        std::string errors;
+        errors.resize(logSize+1);
+        glGetShaderInfoLog(_shaderId, logSize, NULL, &errors[0]);
         glDeleteShader(_shaderId);
+
+        _errors = "#### Source\n";
+        {
+            std::stringstream stream(shaderCode);
+            size_t indexLine = 1;
+
+            std::string line;
+            while(std::getline(stream, line)) {
+                _errors += std::to_string(indexLine) + ":" + line + "\n";
+                indexLine += 1;
+            }
+        }
+        _errors += "#### Errors\n";
+        _errors += errors;
+
         _shaderId = 0;
     }
 
--- a/pxr/imaging/hgiMetal/shaderFunction.mm
+++ b/pxr/imaging/hgiMetal/shaderFunction.mm
@@ -14,6 +14,8 @@
 #include "pxr/base/tf/diagnostic.h"
 
 #include <unordered_map>
+#include <string>
+#include <sstream>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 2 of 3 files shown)


jesschimein:
Filed as internal issue #USD-12106

tcauchois:
Note that we already have a set of functions and debug flags to do this in hdSt:
https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/imaging/hdSt/glslProgram.cpp#L175
(in particular, _GetCompileErrorCodeContext and HDST_DUMP_FAILING_SHADER_SOURCE).  Modulo discussions of whether this feature is more properly located in Hgi, is there a reason you need these additions and can't use the existing code?

DuncanHopkinsFoundry:
@tcauchois, these changes allow you to locate the actual code a driver reports an error about.
The user gets to see the whole shader text submitted to the API and not just small and incomplete sections.
The logged source contains line numbers in the space as the driver sees them, so any driver errors with line numbers can be easily found. 

While trying to add more complex/missing parts to HGI, across all three GL, Vk, and Metal, it was really hard to locate shader code when none of the glue logic was shown.

tcauchois:
The error string you're overwriting ends up all sorts of places in existing hydra deployments and really isn't an appropriate place to put hundreds of lines of shadertext.

We do attempt to annotate the error messages with the lines that failed + context, although it looks like our error message parsing might be specific to certain compilers; I might be able to dig up a fix for that.  As mentioned there are a few debug flags you can turn on in Storm to print the whole shadertext, either on fail or on compile, to either stdout or a tmp file.  In my experience, the line breaks in that debug print match up with compiler error lines, meaning it's fine to let an editor add the line numbers.

Can you explain your use case in more detail?  Is this just for personal debugging, or is this for an editor where you want a way to actually pull more of the context out of Storm?  Are you trying to avoid env vars or debug flags?  Are you looking for this data on a diagnostic channel, or do you want a structured accessor?  If you can tell me more about what you're trying to do I might be able to suggest alternative places to add this data.

DuncanHopkinsFoundry:
I am happy to have this hidden behind extra envvars or debug flags.

My use case is that I was/am trying to extend the compute shader functionality in the HGI, I need it for the Nuke Hydra viewport Gaussian Splat rendering. This code runs as part of Storm, but also is in an external plugin.

I am using HGI directly to create compute shaders. I am not using GLSLFX, as too much of its APIs (as used by Storm) are private to Storm classes and library. The compute shaders (at the time of writing) do not support local shared variables and a couple of other minor bit s and pieces like work group size related things (IIRC). To get these to work I was having to add in new shader glue code to the GLSL and Metal shader generation for the OpenGL, Metal, and Vulkan code paths.

When these shaders fail to compile I needed to see the whole shader, including all the Metal and GLSL glue code, submitted to the driver. And the line numbers also save me a lot of time and errors finding the correct line(s).

tcauchois:
Thanks for the discussion; talked it over internally as well.

I think to take this we'd need the new behavior hidden behind a debug flag (or at least, that's the most idiomatic way to turn something like this on/off).  Could you update the change?  I won't have the cycles for a bit.

In terms of what this could look like if we cleaned the related code up, we sketched out something like:
- for failing shaders, add HgiShaderFunction::GetContextualizedError() (add context lines to error message, copying glslProgram.cpp impl) and HgiShaderFunction::GetFailingSourceText().  For shaders that compile without a problem, don't keep these strings.
- rebase glslProgram.cpp debug codes on the new HgiShaderFunction API.

... but since most of that is cleanup, I won't ask you to do that unless you're excited about the new API surface or something.  If you added the shader source to GetFailingSourceText instead of GetCompileErrors, we wouldn't need the debug codes, of course.

PR #4054, opened 2026-04-28
Support labels in `pxr_register_test`

nvmkuruc:
### Description of Change(s)

`ctest` offers labels as a way to group tests for downstream runners.  This introduces support for test labeling and adds a single label `flaky`, implying that the test occasionally fails but usually succeeds on retry.

The initial set of flaky tests were identified by looking through recent pipeline logs for failures that appeared spurious.  The `work` tests have been known to fail occasionally for most of the repos lifetime.

Future PRs may contribute additional labels (such as those used in regex-based exclusion rules in the CI pipelines).

Example usage:
```
# Exclude flaky tests
ctest -C Release -LE flaky
# Allow multiple runs of flaky tests
ctest -C Release -L flaky --repeat until-pass 3
```

### Link to proposal ([if applicable](https://openusd.org/release/contributing_to_usd.html#step-1-get-consensus-for-major-changes))

https://github.com/aousd/build-ig-initiatives/issues/18

### Checklist

- [x] I have created this PR based on the dev branch

- [x] I have followed the [coding conventions](https://openusd.org/release/api/_page__coding__guidelines.html)

- [ ] I have added unit tests that exercise this functionality (Reference: 
  [testing guidelines](https://openusd.org/release/api/_page__testing__guidelines.html))

- [ ]  I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement (Reference: 
  [Contributor License Agreement instructions](https://openusd.org/release/contributing_to_usd.html#contributor-license-agreement))

PR #4055, opened 2026-04-28
Repeat flaky tests 3 times before failing

nvmkuruc:
### Description of Change(s)

#4054 introduced `flaky` labels to a limited set of tests.  This adds new steps to the CI pipeline which will allow flaky tests to run up to 3 times before failing.  This should improve the reliability of the GitHub CI.

https://github.com/aousd/build-ig-initiatives/issues/18

### Checklist

### DIFF
--- a/.github/workflows/buildusd.yml
+++ b/.github/workflows/buildusd.yml
@@ -95,7 +95,14 @@ jobs:
         working-directory: ./USDgen/build/OpenUSD
         run: |
           ctest --output-on-failure -j4 -C Release \
-                --exclude-regex "^(testUsdview|testUsdRecord|testUsdAppUtilsFrameRecorder|testUsdImaging|testHgiGL|testHdx|testHdSt|testExecGeomXformable_Perf_Large|testWorkThreadLimitsDefault)"
+                --exclude-regex "^(testUsdview|testUsdRecord|testUsdAppUtilsFrameRecorder|testUsdImaging|testHgiGL|testHdx|testHdSt|testExecGeomXformable_Perf_Large|testWorkThreadLimitsDefault)" \
+                -LE flaky --no-tests=error
+      - name: Run USD Headless Tests (flaky)
+        working-directory: ./USDgen/build/OpenUSD
+        run: |
+          ctest --output-on-failure -j4 -C Release \
+                --exclude-regex "^(testUsdview|testUsdRecord|testUsdAppUtilsFrameRecorder|testUsdImaging|testHgiGL|testHdx|testHdSt|testExecGeomXformable_Perf_Large|testWorkThreadLimitsDefault)" \
+                -L flaky --repeat until-pass:3
 
   GPUTests:
     needs:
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-12137

.github/workflows/buildusd.yml
```diff
@@ -144,13 +151,15 @@ jobs:
           find ${{ github.workspace }}/USDgen/build -type f -path "*/bin/*" -exec chmod a+x {} +
       - name: Run USD GPU Tests
         working-directory: ./USDgen/build/OpenUSD
+        # There's no dedicated step to run flaky GPU tests because no imaging tests are currently tagged as flaky
```
sunyab: Even though we currently have no "flaky" GPU tests, would it break anything to just add this step now? I worry that we'll have to remember to come back and add the step if someone does add that label to a GPU test in the future.

sunyab (changes requested):

.github/workflows/buildusd.yml
nvmkuruc: @sunyab Hi Sunya!  For the other steps, I have been using `-no-tests=error` to alert us to steps that might not be necessary, spurious, or improperly specified.  So if we followed that pattern here, that step would fail.  If you'd like, we could--
* Include it commented out
* Include it with just the `-no-tests=error` commented out

nvmkuruc: Also-- note that we don't exclude flaky GPU tests yet either.  So the tests would run (unrepeated) until someone adds the step.

sunyab: Oh, does the change at line 162 not already exclude "flaky" tests?

sunyab: FWIW I think it'd be ok to include it with `-no-tests=error` commented out (or omitted), so long as doing so wouldn't unnecessarily re-run tests like you mentioned.

nvmkuruc: You're right.  Apologies.

nvmkuruc: I pushed something that's a little different-- it adds the step but leaves the step disabled.  I think I'd encourage that as the pattern over commenting out part of the command, but if you'd prefer the original suggestion, I can push an update.

sunyab:
Hey @nvmkuruc , sorry I'm just getting back to this now. Thinking about this a bit more, would you object to:

1. Enabling the flaky test step for GPU tests
2. Removing `--no-tests=error` from _all_ of the flaky test steps

I think removing this would reduce maintenance burden. For example, if we fixed all the flaky tests, we'd have to remember to update this file to disable the step, but then if we added another flaky test we'd have to reenable the step, etc. I think it'd be OK to just leave the flaky step there even if it didn't find any tests, since that shouldn't cost anything anyway.

Adding `--no-tests=error` to the non-flaky test step seems good to me, since it would be very surprising if that step didn't find any tests.

nvmkuruc:
@sunyab I suspect the long term "right" answer will be to organize these opinions in a ctest `preset` akin to #4106.  I'll enable the extra step for now.

sunyab (approved):
Thanks! Will get this merged in with the next dev push.

PR #4057, opened 2026-04-29
Fix nested native instance selection

hsabri-SMS:
### Description of Change(s)

Hello, this PR fixes an issue in Hydra 2.0 code path where selecting a nested native instance could highlight geometry
from sibling instances. I should also mention that this change was partially developed with AI assistance for code exploration and initial draft. 

From what I understood the root cause is  that `UsdImagingSelectionSceneIndex::_ExpandToDescendants` could walk directly into the `UsdNiPropagatedPrototypes` helper subtree under a propagated `UsdNiPrototype` during descendant expansion. That subtree is an internal structure used to represent nested native-instancing prototypes. Visiting it directly caused nested propagated prototypes to be walked more than once, which produced incorrect flattened instance indices and leaked selection across sibling instances.

I did some digging in history and it appears the the regression was caused by this change:

PR: **refactoring to aggregate native instancing before descending into their propagated prototype**
https://github.com/PixarAnimationStudios/OpenUSD/commit/5efed7c5398235db6ddc5299764c4cb92b0b1f9d

### Fixes Issue(s)
To fix this we skip the direct `UsdNiPropagatedPrototypes` child during descendant traversal and let nested propagated prototypes be reached only through the aggregated native-instance seeds. This preserves the correct nested instance context and avoids duplicate traversal of nested prototypes.

I added a regression test for a 2x2x2x2 nested native-instancing fixture (`Tower -> Floor -> Row -> Cube`). The test
verifies that selecting `Tower_1` produces only the expected flattened cube instance ids (`8..15`) and does not
include the `Tower_0` ids (`0..7`). 

Note: I only have access to a Windows machine and won't be able to run the test on MacOS and Linux at the moment.

```
    Start 960: testUsdImagingNestedInstancingSelection
1/1 Test #960: testUsdImagingNestedInstancingSelection ...   Passed    0.39 sec
```

**Before Fix:**
image

**After Fix:**
image


jesschimein:
Filed as internal issue #USD-12142

Issue #4056, opened 2026-04-29
instanceProxyPathTranslationSceneIndex getting tripped up by "pre-expanded" paths


robp-sidefx:
In https://github.com/PixarAnimationStudios/OpenUSD/commit/8ba7b01306bbf53d29ca2c701fdc4e2d6bb9d0ed the instanceProxyPathTranslationSceneIndex was changed to follow breadcrumbs for nested instancing. While this seems to work well for nested native instancing, it can break when point instancing is included in the mix.

Let's say that after all the NI & PI SceneIndexes have run, there's a resulting hierarchy of: `/a/b/c/d/e/f/g/mat`. Further, the `instance` breadcrumbs on `/a/b` say that the replacement prototype is `/a/b/c/d` and the `instance` breadcrumbs on `/a/b/c/d/e` say that the replacement prototype is `/a/b/c/d/e/f/g`.

Now, if we have a material binding to `/a/b/e/mat`, the Scene Index works nicely:
* start with `/`
* append `a` - have `/a` - no breadcrumb, carry on
* append `b` - have `/a/b` - breadcrumb found - replace work with `/a/b/c/d`
* append `e` - have `/a/b/c/d/e` - breadcrumb found - replace work with `/a/b/c/d/e/f/g`
* append `mat` - have `/a/b/c/d/e/f/g/mat` - success!

However ......... somewhere in the mix of PI & NI Scene Indexes, `/a/b/e/mat` is being "pre-expanded" to `/a/b/c/d/e/mat`. It's still not the final path, so the instanceProxyPathTranslationSceneIndex still intercepts it:

* start with `/`
* append `a` - have `/a` - no breadcrumb, carry on
* append `b` - have `/a/b` - breadcrumb found - replace work with `/a/b/c/d`
* append `c` - have `/a/b/c/d/c` - we've now gone off the rails, prim doesn't even exist, but loop continues
* append `d` - have `/a/b/c/d/c/d` - still going, even though this prim doesn't even exist
* append `e` - have `/a/b/c/d/c/d/e` - still going, even though this prim doesn't even exist
* append `mat` - have `/a/b/c/d/c/d/e/mat` - we've reached the end of the path segments to append

Either we need to ensure there are no "pre-expanded" paths before this scene index runs, or we should work back-to-front, looking for the _longest_ "relocation", not the shortest. In that scenario we'd do something like this:

* start with `/a/b/c/d/e/mat` - not a prim, carry on
[...]

PR #4058, opened 2026-04-30
Reworking the logic of the instanceProxyPathTranslationSceneIndex

robp-sidefx:
### Description of Change(s)

Reworking the logic of the instanceProxyPathTranslationSceneIndex so that rather than working our way through the path front-to-back (matching the shortest path with an instanceSchema), we go back-to-front (matching the longest path with an instanceSchema). This has, for our tests, proven more robust in practice as it will correctly match both "fully-unexpanded" paths (as did the original implementation) and "partially-expanded" paths (which the original implementation did not - see linked issue below).

It is worth noting and considering that the new implementation is potentially more costly as it restarts the loop each time there's a match (whereas the original implementation had a fixed cost of one pass through the loop), but ultimately we need something that gives us the right results (which hopefully this one does - again, our testing suggests this is an improvement).

### Fixes Issue(s)

#4056

### Checklist


jesschimein:
Filed as internal issue #USD-12143

blevin:
PR merged today and will be on github with the next push.

PR #4068, opened 2026-05-08
tf: Align TfSingleton ABI between Python builds

erslavin:
### Description of Change(s)

Unifies the ABI of TfSingleton such that it's the same between Python and no-Python builds of OpenUSD.

### Checklist

### DIFF
--- a/pxr/base/tf/instantiateSingleton.h
+++ b/pxr/base/tf/instantiateSingleton.h
@@ -45,8 +45,15 @@ struct Tf_SingletonPyGILDropper
     TF_API
     ~Tf_SingletonPyGILDropper();
 private:
+
+    // Keep class ABI compatible between builds with python enabled or not
+    // Otherwise there will be stack corruption when loading plug-ins
+    // built between the two versions.
 #ifdef PXR_PYTHON_SUPPORT_ENABLED
     std::unique_ptr<class TfPyLock> _pyLock;
+    static_assert(sizeof(std::unique_ptr<class TfPyLock>) == sizeof(std::unique_ptr<void*>));
+#else
+    std::unique_ptr<void*> _unused;
 #endif // PXR_PYTHON_SUPPORT_ENABLED
 };
 


nvmkuruc:
Thanks to @marsupial for the initial fix.

jesschimein:
Filed as internal issue #USD-12183

gitamohr:
Hey there -- even though this "fixes" the stack layout for these functions when trying to combine Python-enabled and Python-disabled builds, it still leaves you with ODR violations, and, I'm afraid this is just the tip of the iceberg.  There are myriad reasons why we can't generally support mixing `.o` files with varying Python-enabled & -disabled build configs.

Just one example: a `VtValue` created in a Python-disabled TU and passed to a Python-enabled TU that then attempts to convert it to Python will crash.  Insert Woody & Buzz meme -- "ODR Violations ... ODR Violations Everywhere".  Python enabled/disabled compilations can't be mixed in general.

erslavin:
Yes - we have another solution that rebinds those objects if they were created on the no python side.  This was just the first step to align the ABI.

nvmkuruc:
Hey @gitamohr -- this particular PR addresses the concerns of different libraries in the same process using the same OpenUSD that don't send nopython originated values to python.

You could imagine a render delegate strictly consuming values may not need or want to provide multiple builds for python/nopython if it doesn't use python (especially when you might need to consider supporting multiple python releases).

As Ed mentioned we'd like to offer solutions for broader nopython/python intermixing problem as well, but we were hoping to start with this smaller change that solves the concrete problem.

gitamohr:
I appreciate your ambition but doing this fully generally goes _very_ deep.  Another one that comes to mind is the RefPtr/WeakPtr Python lifetime management.  Another is all the places where we do `TF_PY_ALLOW_THREADS_IN_SCOPE()` before calling virtual functions/plugin code.  If the outer frame is a Python TU that holds the GIL and it calls an intermediate no-Python TU that nominally does `TF_PY_ALLOW_THREADS_IN_SCOPE()` (which does nothing), and the ultimate callee is a Python TU that needs to wait for something that needs the GIL, you're liable to deadlock.  All the careful GIL protections, hard won by years of debugging in uniformly-Python builds are shot, essentially.

I can see how you could partly handle `VtValue` under a limited set of circumstances -- if a no-Python TU created a `VtValue` holding a type that _is known to_ the Python TU (so it could have instantiated the to-python function), then it can plausibly look up that conversion.  But that doesn't cover the case where the type is not known to the Python TU.  It's possible that there are some Python TUs that "know" the type and some that don't, so you'll get strange and confusing, possibly nondeterministic behavior.

I think it will be incredibly difficult / impossible to feel confident that you've covered all possibilities and can support fully general mixing of Python/no-Python `.o`s.  How would you even approach the test CI setup for it?  Maybe it would be more plausible if you set down very strict rules about what _exactly_ you can and can't do when mixing, but again trying to decide what those rules are, and feel confident that they fully cover all the territory seems daunting.

In the end I believe you'll still end up with ODR violations everywhere, which is UB and all bets are off.  I think the only way to do it really correctly is to make Python-enabled/optional builds work in terms of a dynamic runtime switch, supported by `dlopen()`ed Python support lib(s).  Then Python-enabled builds could run in no-Python mode if Python isn't present/desired.  No-python built modules would abort if loaded into Python-active runtimes.  Still a huge project.

gitamohr:
Hmm, after reading through our code more, maybe it's not _quite_ as dire as I initially thought.  I do think that there are real hurdles, but maybe something could be done...

Anyway I have a change that's been sitting in my workspace that stubs out `TfPyLock` to exist but do nothing in Python-disabled builds.  With that, I think we can just remove the `#ifdef PXR_PYTHON_SUPPORT_ENABLED` here and it ought to "work".  I'll see about doing that.

marsupial:
As noted there are some additional changes to get everything working well.  I think it maybe should be emphasized that if a plugin doesn’t know about Python at all (and just to be a bit more accurate, does not publish VtValues or care to publish across py/nopy boundary), OpenUsd will crash from the above.  The side-effect being that a plugin that doesn’t care at all about Python needs to be compiled and linked against a specific Python version.

nvmkuruc:
Thanks Alex for the consideration.  We're hoping that the scope of this change is symmetrical to https://github.com/PixarAnimationStudios/OpenUSD/pull/1729 which similarly aimed to make python/nopython builds ABI compatible.

gitamohr:
Sure -- I think this is a reasonable step to take, but to be clear mixing py/no-py builds is officially unsupported.

PR #4079, opened 2026-05-22
Shift headers into CMake `FILE_SET` sources

nvmkuruc:
### Description of Change(s)

Modern CMake allows for the declaration of headers as [`FILE_SET`s](https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets) and made an explicit source of a target.  The declaration of a `FILE_SET` allows the specification of a `BASE_DIR` which naturally preserves the relative paths and subdirs without any specialized `cmake` logic.

Adopting this allows us to address https://github.com/aousd/build-ig-initiatives/issues/37 and drop `_copy_headers`.

An interesting side effect of this is that the percentages in the `cmake` output are more representative.  Previously, the first 50% of the progress bar accounted for the initial copy of headers.

Another interesting observation is that plugins may declare `PUBLIC_HEADERS`. It's unclear if this is intentional or not, but it is supported.

The resulting `include` (and `share/**/include`) directories are identical.  The `compile_commands.json` is identical except for the addition of the source directory as an include.  The resulting `pxrTargets.cmake` provides a more precise list of the library's public headers.

Note if it's determined that the isolation provided by the initial copy is still desired, this change can be easily adapted to retain a simpler copy step into the build directory before constructing the `FILE_SET`.

### Fixes Issue(s)

https://github.com/aousd/build-ig-initiatives/issues/37

### Checklist

### DIFF
--- a/cmake/macros/Private.cmake
+++ b/cmake/macros/Private.cmake
@@ -6,60 +6,6 @@
 #
 include(Version)
 
-# Copy headers to the build tree.  Under pxr/ the include paths match the
-# source tree paths but elsewhere they do not. Instead we use include
-# paths like rmanArgsParser/rmanArgsParser.h.  So if /pxr/ is not in the
-# source tree path then copy the headers (public and private) into the
-# build tree under paths of the latter scheme.
-function(_copy_headers LIBRARY_NAME)
-    set(options  "")
-    set(oneValueArgs PREFIX)
-    set(multiValueArgs FILES)
-    cmake_parse_arguments(_args
-        "${options}"
-        "${oneValueArgs}"
-        "${multiValueArgs}"
-        ${ARGN}
-    )
-
-    set(files_copied "")
-    set(hpath "${_args_PREFIX}/${LIBRARY_NAME}")
-    if ("${CMAKE_CURRENT_SOURCE_DIR}" MATCHES ".*/pxr/.*")
-        # Include paths under pxr/ match the source path.
-        file(RELATIVE_PATH hpath "${PROJECT_SOURCE_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}")
-    endif()
-    set(header_dest_dir "${PROJECT_BINARY_DIR}/${PXR_INSTALL_SUBDIR}/include/${hpath}")
-    if( NOT "${_args_FILES}" STREQUAL "")
-        set(files_copied "")
-        foreach (f ${_args_FILES})
-            set(infile "${CMAKE_CURRENT_SOURCE_DIR}/${f}")
-            set(outfile "${header_dest_dir}/${f}")
-            get_filename_component(dir_to_create "${outfile}" PATH)
-            add_custom_command(
-                OUTPUT ${outfile}
-                COMMAND ${CMAKE_COMMAND} -E make_directory ${dir_to_create}
-                COMMAND ${CMAKE_COMMAND} -Dinfile=${infile} -Doutfile=${outfile} -P "${PROJECT_SOURCE_DIR}/cmake/macros/copyHeaderForBuild.cmake"
-                MAIN_DEPENDENCY ${infile}
-                COMMENT "Copying ${f} ..."
-                VERBATIM
-            )
-            list(APPEND files_copied ${outfile})
-        endforeach()
-    endif()
-
-    # Add a headers target.
-    add_custom_target(${LIBRARY_NAME}_headerfiles
-        DEPENDS ${files_copied}
-    )
-    set_target_properties(${LIBRARY_NAME}_headerfiles
-        PROPERTIES
-            FOLDER "headerfiles"
-    )
-
-    # Make sure headers are installed before building the library.
-    add_dependencies(${LIBRARY_NAME} ${LIBRARY_NAME}_headerfiles)
-endfunction() # _copy_headers
-
 # Copy doxygen files for documentation builds to the build tree. Doxygen
 # files can be .dox files, C++ source with doxygen comments, or 
 # resource files (images, etc). Files will get copied to a parallel structure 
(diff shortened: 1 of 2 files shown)


cmake/macros/Private.cmake
```diff
@@ -1363,16 +1303,43 @@ function(_pxr_library NAME)
...
+                ${headerBaseDir}
+            FILES ${args_PUBLIC_HEADERS}
+    )
+    target_sources(${NAME}
+        PRIVATE
+            FILE_SET private_headers
+            TYPE HEADERS
+            BASE_DIRS
+                ${headerBaseDir}
+            FILES ${args_PRIVATE_HEADERS}
```
dlarsson-adobe, line 1355: I think it would be nice with a comment on what this does and how it's used later when installing since I think `FILE_SET` is a quite rare feature (at least I've never seen it before)

cmake/macros/Private.cmake
```diff
@@ -1461,11 +1400,13 @@ function(_pxr_library NAME)
     else()
         # Do not include plugins libs in externally linkable targets
         if(isPlugin)
+            # Note some plugins have public headers.
             install(
```
dlarsson-adobe, line 1420: Not sure if in scope for this PR but it seems like most of these install statements are pretty much identical, can we consolidate the conditionals here?

dlarsson-adobe (commented):
Looks like an improvement to me

cmake/macros/Private.cmake
nvmkuruc: I've updated the comment.  `FILE_SET` is new as of CMake 2.23.

cmake/macros/Private.cmake
nvmkuruc: I suspect you're right, but might be nice to address in a follow up PR.

jesschimein:
Filed as internal issue #USD-12249

dlarsson-adobe:
> Note if it's determined that the isolation provided by the initial copy is still desired, this change can be easily adapted to retain a simpler copy step into the build directory before constructing the `FILE_SET`.

I was investigating a build performance problem on windows related to this copy step and in my environment the copying and forcing of the directory to exist (in practice mostly process creation) was a significant performance sink. I made a solution where everything happens inside one process and cut down this stage dramatically. We should consider doing that in case we want to keep copying

dlarsson-adobe (approved):
Looks good to me!

I have been using this myself successfully as a data point that this seems to be working

pmolodo:
Well... crap. This is why why can't have new things...

**Good news**/ **bad news**

**Good news**: my new access to the build-ig branch works - I have confirmed I can both push commits to it (pushed to update to dev) and merge PRs to it (merged https://github.com/PixarAnimationStudios/OpenUSD/pull/4223)

**Bad news**: https://github.com/PixarAnimationStudios/OpenUSD/pull/4223 was a dummy test PR I had made to test squash-and-merge, and I THOUGHT was targeting the build-ig branch in my FORK... but I forgot that the default behavior for PRs from forks is to target the upstream branch ([PixarAnimationStudios/build-ig](https://github.com/PixarAnimationStudios/OpenUSD/tree/build-ig)), not your fork's branch ([pmolodo/build-ig](https://github.com/pmolodo/OpenUSD/tree/build-ig))

So... I created a bogus commit on build-ig ([Test multi commit pr (](https://github.com/PixarAnimationStudios/OpenUSD/commit/04ca7c18df829d626ae718011106c8818e59c092)[#4223](https://github.com/PixarAnimationStudios/OpenUSD/pull/4223)[)](https://github.com/PixarAnimationStudios/OpenUSD/commit/04ca7c18df829d626ae718011106c8818e59c092)), which I then reverted one commit later ([Revert "Test multi commit pr (](https://github.com/PixarAnimationStudios/OpenUSD/commit/b0cfa1cc3992484c2ec9e9b417afe5bbfd664293)[#4223](https://github.com/PixarAnimationStudios/OpenUSD/pull/4223)[)"](https://github.com/PixarAnimationStudios/OpenUSD/commit/b0cfa1cc3992484c2ec9e9b417afe5bbfd664293))

I do not have force-push access, which seems reasonable... but I'm wondering if we want to just leave this wart in the history, or do a one-time force-push of build-ig back to [dev a3d77ff](https://github.com/PixarAnimationStudios/OpenUSD/commit/a3d77ff5e1d405b0a5080e7c9f4c6492898fb5c5)?
[@DavidLarssonAdobe](https://aousdgroup.slack.com/team/U08LPRH4GLU) [@Sunya Boonyatera](https://aousdgroup.slack.com/team/U07RJ8LD1SB) [@Shahbaz Khan](https://aousdgroup.slack.com/team/U09SRRYJWBX) [@Matt Kuruc](https://aousdgroup.slack.com/team/U07RYMZG8DR)

Apologies :disappointed:

pmolodo:
In other news - the [test PR](https://github.com/PixarAnimationStudios/OpenUSD/pull/4223) did confirm that squash-and-merge in github does behave as @dlarsson-adobe described, not how I did - it's a squash followed by a 1-commit-rebase
Apologies again, I think I was getting my wires crossed with gitlab...

sunyab:
I just tried testing this against the latest `dev` branch and got the following error:

```
CMake Error at cmake/macros/Public.cmake:688 (_copy_headers):
  Unknown CMake command "_copy_headers".
Call Stack (most recent call first):
  pxr/exec/execTest/CMakeLists.txt:4 (pxr_build_test_library)
```

The `pxr_build_test_library` function is a relatively new addition, so we'll need to adapt this function to use `FILE_SET` as well.

nvmkuruc:
@sunyab I pushed a fix and revalidated the install differences and verified that the only difference is the addition of the header sets to `pxrTargets.cmake`.  Since the `build-ig` branch is now behind `dev`, I retargeted the PR to be against `dev`.

```
# Build top of tree dev
python3.12 ./build_scripts/build_usd.py ../inst/usd --no-imaging --tests --examples -vv
mv ../inst/usd ../inst/usd-dev
# Build with PR
python3.12 ./build_scripts/build_usd.py ../inst/usd --no-imaging --tests --examples -vv
mv ../inst/usd ../inst/usd-headers
xxdiff -r --exclude "build" --exclude "__pycache__" ../inst/usd-{dev,headers}
```

PR #4097, opened 2026-06-03
[APPLE] Bugfix multiple primvars with GeomSubsets and subdiv. 

ld-kerley:
### Description of Change(s)

When a subdivided mesh has both geom subsets (materialBind family) and multiple face-varying primvars whose index buffers differ (multiple fv-topology channels), the geom-subsets path of `HdStMesh::_PopulateTopology()` was only adding the LAST channel's fvarIndices buffer to each subset draw item's topology BAR.

The local fvarIndicesSource variable in the channel-walk loop was overwritten each iteration and then passed by value to `_CreateTopologyRangeForGeomSubset()`, which only ever inserted that single source into the BAR.  Meanwhile codeGen continued to emit `HdGet_fvarIndices<N>()` calls for every channel a primvar uses, so any draw item whose primvar lived on a missing channel produced a Metal shader compile error of the form:

```
  error: use of undeclared identifier 'HdGet_fvarIndices0';
         did you mean 'HdGet_fvarIndices1'?
```

To solve this, in this PR we plumb a `HdBufferSourceSharedPtrVector` through `_CreateTopologyRangeForGeomSubset()` so that every channel's indices source (and its chained fvarPatchParam<N> buffer, which `HdSt_IndexSubsetComputation()` already preserves) ends up in the topology BAR for both the original draw item and each GeomSubset draw item.

Rendering unit tests for UsdImagingGL are disabled for Metal - so attaching locally tested unit test example that fails to render correctly - and raises the error above without the patch.

[unittest.zip](https://github.com/user-attachments/files/28568928/unittest.zip)

### Checklist

### DIFF
--- a/pxr/imaging/hdSt/mesh.cpp
+++ b/pxr/imaging/hdSt/mesh.cpp
@@ -895,7 +895,7 @@ HdStMesh::_PopulateTopology(HdSceneDelegate *sceneDelegate,
         } else {
             // Geom subsets case
             HdBufferSourceSharedPtr indicesSource;
-            HdBufferSourceSharedPtr fvarIndicesSource;
+            HdBufferSourceSharedPtrVector fvarIndicesSources;
 
             bool refined = false;
             bool quadrangulated = false;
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-12297

PR #4105, opened 2026-06-08
DRAFT: Support OneTBB via FetchContent

nvmkuruc:
### Description of Change(s)
Add FetchContent support for OneTBB. This allows `cmake` to download OneTBB if it can't be located via `find_package`.

This facilitates minimal builds of OpenUSD with `cmake` without using `build_usd.py`.  This feature is tagged as experimental and gated behind the `PXR_ALLOW_FETCH_CONTENT` flag.

Notes:
* Only OneTBB and `PXR_FIND_TBB_IN_CONFIG` are supported
* The preferred OneTBB version is identical to the one that `build_usd` would have downloaded
* OneTBB Version Range 2021.12.x-2022.y is supported

Before integrating, some consideration should be given to the ordering of the cmake options, project defaults, and packages cmake.  Global cmake variables configured in project defaults may now leak into the `FetchContent_MakeAvailable` call unless explicitly blocked.  (See `CMAKE_INCLUDE_CURRENT_DIR` which causes OneTBB to fail to build).

### Link to proposal ([if applicable]
https://github.com/aousd/build-ig-initiatives/issues/16

### Checklist

### DIFF
--- a/cmake/defaults/Packages.cmake
+++ b/cmake/defaults/Packages.cmake
@@ -13,12 +13,23 @@ set(build_shared_libs "${BUILD_SHARED_LIBS}")
 # Core USD Package Requirements 
 # ----------------------------------------------
 
+include(FetchContent)
+
 # Threads.  Save the libraries needed in PXR_THREAD_LIBS;  we may modify
 # them later.  We need the threads package because some platforms require
 # it when using C++ functions from #include <thread>.
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 find_package(Threads REQUIRED)
 set(PXR_THREAD_LIBS "${CMAKE_THREAD_LIBS_INIT}")
+
+if (PXR_ALLOW_FETCH_CONTENT)
+    FetchContent_Declare(
+        tbb
+        URL https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.12.0.zip
+        URL_HASH SHA256=fe6ca052b5bdd2c6e0616b360c9b0dcbcc46e01bbd0aa8fd0517c17fc58931db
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-12304

cmake/defaults/Packages.cmake
```diff
@@ -13,13 +13,24 @@ set(build_shared_libs "${BUILD_SHARED_LIBS}")
...
 # Threads.  Save the libraries needed in PXR_THREAD_LIBS;  we may modify
 # them later.  We need the threads package because some platforms require
 # it when using C++ functions from #include <thread>.
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 find_package(Threads REQUIRED)
 set(PXR_THREAD_LIBS "${CMAKE_THREAD_LIBS_INIT}")
 
+if (PXR_ALLOW_FETCH_CONTENT)
+    FetchContent_Declare(
+        onetbb
```
dlarsson-adobe: I think onetbb should be tbb


I tried extending this fetch approach to opensubdiv and ran into an issue

When OpenSubdiv calls find_package(TBB 2018 CONFIG COMPONENTS tbb), CMake looks for a config file named TBBConfig.cmake or tbb-config.cmake (the lowercased <name>-config.cmake form) in its search paths. The name it searches for is TBB — nothing else. A file called onetbb-config.cmake is invisible to it.

FetchContent names its generated redirect after the declared dependency name, not the package.
When you do FetchContent_Declare(<name> … FIND_PACKAGE_ARGS …) and the dependency ends up built from source, CMake writes a small stub config into ${CMAKE_FIND_PACKAGE_REDIRECTS_DIR} so that a later find_package(<name>) succeeds. That file is named <name>-config.cmake — using the FetchContent dependency name, not the NAMES in FIND_PACKAGE_ARGS.

PR #4117, opened 2026-06-19
[build_usd.py] Add wasm and native deps for webgpu support

kohakukun:
### Description of Change(s)

- Add DawnHeaders, Tint, glslang dependencies both for wasm and native builds
- Add webgpu option. Defaults to off. This option is still work in progress.


jesschimein:
Filed as internal issue #USD-12348

Issue #4074, opened 2026-05-20
`Tf.RealPath` behavior is inconsistent with symlink to missing file


nvmkuruc:
`Tf.RealPath` is unable to resolve symlinks to files that don't exist, when `allowInaccessibleSuffix` is set to `True`.  This seems curiously inconsistent with `Tf.RealPath` when run directly on the missing target of the symlink.  For a comparison, I examined what `os.path.realpath`, `pathlib.Path.resolve`, and the Linux `realpath` command.  All are able to resolve the symlink.

See the following python script.

```python
import os
import pathlib
import subprocess
import tempfile

from pxr import Tf

with tempfile.TemporaryDirectory() as temp_dir:
    temp_dir_path = pathlib.Path(temp_dir)
    does_not_exist_path = temp_dir_path / "does-not-exist.txt"
    symlink_path = temp_dir_path / "symlink"
    symlink_path.symlink_to(does_not_exist_path)
    print("SYMLINK PATH\n",
          f"Tf.RealPath: {Tf.RealPath(str(symlink_path), True)}\n",
          f"pathlib.Path.resolve: {symlink_path.resolve()}\n",
          f"os.path.realpath: {os.path.realpath(str(symlink_path))}\n",
          f"realpath: {subprocess.run(['realpath', symlink_path], capture_output=True).stdout.strip()}\n")
    print("DOES_NOT_EXIST_PATH\n",
          f"Tf.RealPath: {Tf.RealPath(str(does_not_exist_path), True)}\n",
          f"pathlib.Path.resolve: {does_not_exist_path.resolve()}\n",
          f"os.path.realpath: {os.path.realpath(str(does_not_exist_path))}\n",
          f"realpath: {subprocess.run(['realpath', does_not_exist_path], capture_output=True).stdout.strip()}\n")
```

Evaluating this script should produce the following output on Linux.
```
SYMLINK PATH
 Tf.RealPath:
 pathlib.Path.resolve: /tmp/tmpbxsz8qt8/does-not-exist.txt
 os.path.realpath: /tmp/tmpbxsz8qt8/does-not-exist.txt
 realpath: b'/tmp/tmpbxsz8qt8/does-not-exist.txt'

DOES_NOT_EXIST_PATH
 Tf.RealPath: /tmp/tmpbxsz8qt8/does-not-exist.txt
 pathlib.Path.resolve: /tmp/tmpbxsz8qt8/does-not-exist.txt
[...]

PR #4119, opened 2026-06-22
Implement `TfRealPath` via `std::filesystem` to handle broken symlinks

nvmkuruc:
### Description of Change(s)

#4074 reported that broken symlinks were not handled consistently with other implementations of `realpath` when `allowInacessibleSuffix` was set to `True`.

This provides an implementation of `TfRealPath` in terms of `std::filesystem` operations, mapping `allowInaccessibleSuffix=False` to `std::filesystem::canonical` and `allowInaccessibleSuffix=True` to `std::filesystem::weakly_canonical`.

Note that `std::filesystem::weakly_canonical` behaves similarly to `realpath -m` and other permissive forms of `realpath` EXCEPT that it does not resolve the broken symlink to its target before appending the suffix.  The broken symlink is therefore considered a part of the "inaccessible suffix".  This means `Tf.RealPath("/path/to/broken-symlink", allowInaccessibleSuffix=True)` resolves to `"/path/to/broken-symlink"` instead of `"/path/to/missing-target"`. While not providing perfect symmetry to `realpath -m`, this is preferable to the current behavior of failing to resolve anything, returning an empty string, and raising a python exception (if `raiseOnError` is enabled).

### Fixes Issue(s)

#4074 

### Checklist

### DIFF
--- a/pxr/base/tf/pathUtils.cpp
+++ b/pxr/base/tf/pathUtils.cpp
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <cctype>
 #include <errno.h>
+#include <filesystem>
 #include <limits.h>
 #include <string>
 #include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


jesschimein:
Filed as internal issue #USD-12350

PR #4126, opened 2026-06-25
Fix regression which could produce incorrect blend shape weights in Hydra

cameronwhite:
### Description of Change(s)

We observed some randomly-occurring motion blur issues with UsdSkel blend shapes (example image below) and tracked this down to starting in USD 26.03 (ea0b4de9)

The issue is creating a `TfSpan` from an rvalue `VtArray`, which is a use-after-free if the array's data isn't shared.

Because the blendshape weights data source in `dataSourceResolvedSkeletonPrim.cpp` caches the value for shutter offset 0, this issue only happens when evaluating other time samples for motion blur.

The easiest way to reliably reproduce this issue is with the following patch (which avoids the cached `_valueAtZero` and fills in sentinel values in the VtArray destructor to make the use-after-free obvious), which will then cause the `testUsdImagingGLSkelBlendShapes` unit test to fail

```
diff --git a/pxr/base/vt/array.h b/pxr/base/vt/array.h
index ab9943ad5..ff83c257a 100644
--- a/pxr/base/vt/array.h
+++ b/pxr/base/vt/array.h
@@ -1069,7 +1069,11 @@ ARCH_PRAGMA_POP
                 std::atomic_thread_fence(std::memory_order_acquire);
                 for (value_type *p = _data, *e = _data + _shapeData.totalSize;
                      p != e; ++p) {
-                    p->~value_type();
+                    if constexpr (std::is_same_v<value_type, float>) {
+                        *p = 1.0; // Set to any sentinel value
+                    } else {
+                        p->~value_type();
+                    }
                 }
                 ::operator delete(static_cast<void *>(
                                       std::addressof(_GetControlBlock(_data))));
diff --git a/pxr/usdImaging/usdSkelImaging/dataSourceResolvedSkeletonPrim.cpp b/pxr/usdImaging/usdSkelImaging/dataSourceResolvedSkeletonPrim.cpp
index f3c152e9e..e58579201 100644
--- a/pxr/usdImaging/usdSkelImaging/dataSourceResolvedSkeletonPrim.cpp
+++ b/pxr/usdImaging/usdSkelImaging/dataSourceResolvedSkeletonPrim.cpp
@@ -265,9 +265,11 @@ public:
     }
 
     VtFloatArray GetTypedValue(const Time shutterOffset) override {
+#if 0
         if (shutterOffset == 0.0f) {
             return _valueAtZero;
         }
+#endif
 
         return _Compute(shutterOffset);
     }

```

usdskel_blendshape_issue

N/A 

### Fixes Issue(s)

N/A

### Checklist

### DIFF
--- a/pxr/usdImaging/usdSkelImaging/dataSourceResolvedPointsBasedPrim.cpp
+++ b/pxr/usdImaging/usdSkelImaging/dataSourceResolvedPointsBasedPrim.cpp
@@ -597,11 +597,15 @@ public:
     GetTypedValue(const HdSampledDataSource::Time shutterOffset) override {
         TRACE_FUNCTION();
 
-        const TfSpan<const TfToken> blendShapes(  
+        const VtTokenArray blendShapesValue(
             UsdSkelImagingGetTypedValue(_blendShapes, shutterOffset));
-        const TfSpan<const float> blendShapeWeights(
+        const TfSpan<const TfToken> blendShapes(blendShapesValue);
+
+        const VtFloatArray blendShapeWeightsValue(
             UsdSkelImagingGetTypedValue(_blendShapeWeights, shutterOffset));
-        const TfSpan<const GfVec2i> blendShapeRanges(
+        const TfSpan<const float> blendShapeWeights(blendShapeWeightsValue);
+
+        const VtVec2iArray blendShapeRanges(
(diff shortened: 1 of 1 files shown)


jesschimein:
Filed as internal issue #USD-12374

PR #4128, opened 2026-06-29
Simplify transitive `imath` dependency handling

nvmkuruc:
### Description of Change(s)

This change simplifies management of `Imath` which is utilized by OSL, OpenImageIO, Alembic, and OpenEXR, all optional dependencies of OpenUSD.

`Imath` used to be a part of the `OpenEXR` project before being separated out during its `3.x` timeline.  `OpenEXR` can still provide imath as a dependency (see `OPENEXR_FORCE_INTERNAL_IMATH` used by `build_usd`).  Effort was made in the OpenUSD cmake build to be compatible with both the old OpenEXR 2.x setup and new OpenEXR 3.x setup.

Since then--
* OpenEXR 2.x hasn't been a part of the VFXPlatform since 2020.
* OpenVDB 10.x does not use `Imath` by default.  The dependency `build_usd.py` has been assuming is spurious.  Specifically, `USE_IMATH_HALF` defaults to `False`.
* Alembic 1.8.9 has been updated to include `half.h` as `imath/half.h`, avoiding the need for `ImathConfig` (https://github.com/alembic/alembic/issues/431)

To simplify the usage of `Imath` in the OpenUSD project, this PR
* Drops the OpenEXR 2.x compatibility layer
* Drops `Imath` as a dependency of `OpenVDB` (and `hioOpenVDB`) since it's not used
* Updates Alembic's patch version.

Note that `build_usd` doesn't support OSL, so that has yet to be verified

### Fixes Issue(s)

https://github.com/aousd/build-ig-initiatives/issues/10

### Checklist

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -387,11 +387,7 @@ The additional dependencies that must be supplied when invoking cmake are:
 | Dependency Name                   | Description                                       |
 | ----------------------------------|-------------------------------------------------- |
 | ALEMBIC_DIR                       | The location of [Alembic](https://https://github.com/alembic/alembic)   | 
-| OPENEXR_LOCATION                  | The location of [OpenEXR](http://www.openexr.com) |
-| Imath_DIR (If not using OpenEXR)  | Path to the CMake package config of a Imath SDK install. (With OpenEXR 3+, Imath can be used explicitly instead of OpenEXR.)|
-
-Either OpenEXR or Imath is required depending on which library is used by the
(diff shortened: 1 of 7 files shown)


sunyab:
Filed as internal issue #USD-12386

nvmkuruc:
An earlier version of this change removed the explicit `Imath` dependency, but this can't be done on all platforms until all dependencies (`Alembic`, `OSL`, and `OpenImageIO` are converted to targets.

Issue #3804, opened 2025-09-05
RUNPATHs contain absolute paths from build machine


tylerm-nv:
[INSTALL_RPATH_USE_LINKPATH](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/cmake/macros/Private.cmake#L721) puts absolute paths to the dependencies into the RUNPATH for the shared libraries. $ORIGIN relative paths are already added though, the absolute link paths seem unnecessary. Can they be removed?

jesschimein:
Filed as internal issue #USD-11410

(This is an automated message. See [here](https://openusd.org/release/contributing_to_usd.html#github-issues) for more information.)

PR #4129, opened 2026-06-29
Remove absolute RPATHs

tylerm-nv:
### Description of Change(s)

Removes absolute RPATHs from libraries. The first RPATH entries are already valid $ORIGIN relative paths, so the non-relocatable absolute paths are not necessary.

### Checklist

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -615,6 +615,17 @@ There are four ways to link USD controlled by the following options:
 | PXR_BUILD_MONOLITHIC   | `OFF`     | Build single or several libraries         |
 | PXR_MONOLITHIC_IMPORT  |           | CMake file defining usd_m import library  |
 
+By default, installed binaries and plugins use origin-relative runtime paths
+to find libraries in the USD install layout. Dependencies installed elsewhere
+must be discoverable through the platform's normal library search mechanism.
+Set `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` for compatibility with builds that
+require CMake to add external linker search directories to installed runtime
+paths. This compatibility mode may add absolute paths and make the install
+specific to the machine or dependency layout where it was built. The
+`build_usd.py` script enables this compatibility mode by default; pass
+`--no-use-link-path-rpath` to produce an install containing only the configured
+origin-relative runtime paths.
+
 ##### Shared Libraries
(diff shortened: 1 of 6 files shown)


sunyab:
Filed as internal issue #USD-12387

nvmkuruc:
@tylerm-nv This would address #3804 correct?

This seems related to https://github.com/aousd/build-ig-initiatives/issues/20 as well.

tylerm-nv:
> @tylerm-nv This would address #3804 correct?
> 
> This seems related to [aousd/build-ig-initiatives#20](https://github.com/aousd/build-ig-initiatives/issues/20) as well.

Yes.

tlewis:
Hi @tylerm-nv - I tested this change locally in our setup (both on our internal CI and locally on my workstation) and found that it breaks rpath handling of all of our plugins that link to external dependencies (such as Alembic). 

Would you be able to build via `build_usd.py` with the `--alembic` flag, run tests, and see if there's a good way to handle this? In local testing I was also seeing test failures related to .so's from Opensubdiv, draco, MaterialX, and Renderman not being found at runtime.

nvmkuruc:
@tylerm-nv Would it be reasonable to just set the RPATH cache variable in the `build_usd` script so that the pure CMake solution can move forward while not impacting users like @tlewis?

tylerm-nv:
> Hi @tylerm-nv - I tested this change locally in our setup (both on our internal CI and locally on my workstation) and found that it breaks rpath handling of all of our plugins that link to external dependencies (such as Alembic).
> 
> Would you be able to build via `build_usd.py` with the `--alembic` flag, run tests, and see if there's a good way to handle this? In local testing I was also seeing test failures related to .so's from Opensubdiv, draco, MaterialX, and Renderman not being found at runtime.

@tlewis I tested a build_usd.py config with Alembic, Embree, OpenVDB, OpenImageIO, OpenColorIO, MaterialX, and Ptex enabled. All tests passed, although ldd did find a link problem with testHdEmbree linking directly to `hdEmbree.so`.

Do you have modifications to the dependency layout? I verified that could cause the unresolved behavior, either through `--inst`,  `CMAKE_PREFIX_PATH`,  `ALEMBIC_DIR`, or similar args.

PR #4132, opened 2026-07-01
Windows CI: add missing shell: cmd to Test USD step

asluk:
The Windows `Test USD` step has no `shell:` set, so it runs under pwsh while its body uses cmd syntax (`call set PATH=...`). The adjacent `Test USD (flaky)` step already declares `shell: cmd`.

This one-line fix aligns them so the Windows headless tests run in a valid environment.


sunyab:
Thanks @asluk , we already have the same fix merged internally so I'm going to close this out.

(FYI this wasn't sufficient to get the Windows CI fully passing in my tests; I needed another change to change the line continuation character from '\' to '^')

PR #4141, opened 2026-07-07
Add Python bindings for usdValidation fixers

miguelh-nvidia:
### Description of Change(s)

Adds Python support for UsdValidationFixer, including construction from Python callbacks and optional fixer registration through the validation registry APIs. The change keeps Python callables alive across copied fixers, acquires the GIL before invoking callbacks, adds coverage for registration/lookup/invocation/error/threaded cases, and documents Python fixer usage.

This PR was prepared with AI assistance. I reviewed the implementation, tests, and documentation and take responsibility for the contribution.

### Checklist


jesschimein:
Filed as internal issue #USD-12419

pxr/usdValidation/usdValidation/README.md
```diff
@@ -746,6 +746,96 @@ registry.RegisterValidatorSuite(
 )
 ```
 
+### Adding Fixers in Python
+
+Fixers can be created in Python and passed to any registration method
+via the optional `fixers` parameter. Each fixer requires two callables:
+
+| Callable | Signature | Purpose |
+|---|---|---|
```
nvmkuruc: Is making the unrendered (text) table possible?  I understand if not.

pxr/usdValidation/usdValidation/README.md
```diff
@@ -746,6 +746,96 @@ registry.RegisterValidatorSuite(
...
+        return [
+            UsdValidation.ValidationError(
+                "MissingDocumentation",
+                UsdValidation.ValidationErrorType.Warn,
+                [UsdValidation.ValidationErrorSite(
+                    prim.GetStage(), prim.GetPath())],
+                f"Prim '{prim.GetPath()}' has no documentation.",
+            )
+        ]
+    return []
```
nvmkuruc:
```suggestion
    if prim.GetDocumentation() or prim.IsPseudoRoot():
        return []
    return [
        UsdValidation.ValidationError(
            "MissingDocumentation",
            UsdValidation.ValidationErrorType.Warn,
            [UsdValidation.ValidationErrorSite(
                prim.GetStage(), prim.GetPath())],
            f"Prim '{prim.GetPath()}' has no documentation.",
        )
    ]
```
This is hopefully more compact and has fewer branches.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
```diff
@@ -179,32 +184,63 @@ _WrapPrimTaskFn(object pyFn)
...
+    if (pyFixers.is_none()) {
+        return fixers;
+    }
+    list fixerList(pyFixers);
+    fixers.reserve(len(fixerList));
+    std::transform(
+        stl_input_iterator<object>(fixerList),
+        stl_input_iterator<object>(),
+        std::back_inserter(fixers),
+        [](const object &fixer) {
```
nvmkuruc: Can this extraction fail?

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+                os.path.dirname(__file__),
+                "testUsdValidationFixerPy",
+                cls.PLUGIN_NAME,
+            )
+        try:
+            plugins = Plug.Registry().RegisterPlugins(pluginDir)
+            assert len(plugins) == 1
+            assert plugins[0].name == cls.PLUGIN_NAME
+        except RuntimeError:
+            pass  # Plugin may already be registered.
```
nvmkuruc, line 47: Why would this happen?  Should you be registering in `setUpModule`?
https://docs.python.org/3/library/unittest.html#setupmodule-and-teardownmodule

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        )
+        self.assertEqual(fixer.name, "testFixer2")
+        self.assertEqual(fixer.errorName, "SomeError")
+        self.assertIn("studio", fixer.keywords)
+        self.assertIn("lighting", fixer.keywords)
+        self.assertTrue(fixer.HasKeyword("studio"))
+        self.assertFalse(fixer.HasKeyword("fx"))
+        self.assertTrue(fixer.IsAssociatedWithErrorName("SomeError"))
+
+    def test_ConstructionCallablesByValue(self):
```
nvmkuruc: I'm questioning how much coverage this buys you versus the other tests.  Is it redundant or testing something specific that wasn't obvious to me?

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+
+        layer = Sdf.Layer.CreateAnonymous(".usda")
+        error = self._LayerTask(layer)[0]
+        editTarget = Usd.EditTarget(layer)
+
+        with self.assertRaises(AttributeError):
+            fixer.ApplyFix(error, editTarget)
+
+    def test_ValidationFixerByValue(self):
+        """Registered validators keep copied fixer callables by value."""
```
nvmkuruc: If the fixers were copied by value, then how come their `_CanApplyFn` and `_ImplFn` show they've been called?

nvmkuruc (commented):
@miguelh-nvidia Do you think it would be reasonable to try to make some of the tests parameterized based on whether they cover layer, stage, or prim to reduce redundancy and potential for drift between cases?

pxr/usdValidation/usdValidation/wrapFixer.cpp
```diff
@@ -7,22 +7,94 @@
...
 
 using namespace pxr_boost::python;
 
+namespace
+{
+
+// ---------------------------------------------------------------------------
+// Python callable → C++ fixer function wrappers
+//
+// These follow the same GIL-safety pattern as the validator task function
```
tallytalwar: nitpick on the code comment here. Unlike the Validators which are run in a worker thread via the Context, Fixers are not run by the framework, but yes client can run these from within a worker thread. So I agree we should wrap these in TfPyObjWrapper for GIL-safety, the reasoning is different from the Validator Task Function.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
```diff
@@ -179,32 +184,63 @@ _WrapPrimTaskFn(object pyFn)
...
 }
 
+// ---------------------------------------------------------------------------
+// Fixer extraction helper
+//
+// Converts a Python list of ValidationFixer objects to a C++ vector.
+// Used by the registration shims below when the caller provides fixers.
+
+static std::vector<UsdValidationFixer>
+_ExtractFixers(object pyFixers)
```
tallytalwar: I tested this locally, I think we can simply this by removing `_ExtractFixers` because `wrapFixers.cpp` has `TfPyRegisterStlSequencesFromPython<UsdValidationFixer>();`:

- Remove _ExtractFixers
- Use `const std::vector<UsdValidationFixer> &fixers` in the static _Register* methods below, and
- use `arg("fixers") = list()` in the def below.

Can also remove a couple of headers also.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
```diff
@@ -179,32 +184,63 @@ _WrapPrimTaskFn(object pyFn)
...
 // The three _Register* functions below are thin shims that wrap the Python
 // callable and forward to the appropriate RegisterValidator() overload.
 // They are bound as Python methods on ValidationRegistry below.
 
 static void
 _RegisterLayerValidator(UsdValidationRegistry &registry,
                         const UsdValidationValidatorMetadata &metadata,
-                        object pyFn)
+                        object pyFn,
+                        object pyFixers)
```
tallytalwar: If the above sugg is taken, this will be `const std::vector<UsdValidationFixer> &fixers`

pxr/usdValidation/usdValidation/wrapRegistry.cpp
```diff
@@ -179,32 +184,63 @@ _WrapPrimTaskFn(object pyFn)
...
 {
-    registry.RegisterValidator(metadata, _WrapLayerTaskFn(pyFn));
+    registry.RegisterValidator(
+        metadata, _WrapLayerTaskFn(pyFn), _ExtractFixers(pyFixers));
```
tallytalwar: If above sugg is taken, this will be just `fixers`

pxr/usdValidation/usdValidation/wrapRegistry.cpp
```diff
@@ -323,25 +365,35 @@ void wrapUsdValidationRegistry()
              return_value_policy<TfPySequenceToList>(), (args("schemaTypes")))
         // Explicit registration -- caller provides full metadata.
         // Use when registering validators at runtime without a plugin.
+        // An optional list of ValidationFixer objects can be provided to
+        // associate fixers with the validator.
         .def("RegisterLayerValidator", &_RegisterLayerValidator,
-             (args("metadata", "layerTaskFn")))
+             (args("metadata", "layerTaskFn"),
+              arg("fixers") = object()))
```
tallytalwar: if above sugg is taken this will be `arg("fixers") = list()`

tallytalwar (commented):
I still have to look at README and test, but will check on this in my next round of review. Thanks

pxr/usdValidation/usdValidation/README.md
```diff
@@ -746,6 +746,96 @@ registry.RegisterValidatorSuite(
...
+### Adding Fixers in Python
+
+Fixers can be created in Python and passed to any registration method
+via the optional `fixers` parameter. Each fixer requires two callables:
+
+| Callable | Signature | Purpose |
+|---|---|---|
+| `fixerImplFn` | `(error: ValidationError, editTarget: Usd.EditTarget, timeCode: Usd.TimeCode) -> bool` | Apply the fix; return `True` on success |
+| `canApplyFn` | `(error: ValidationError, editTarget: Usd.EditTarget, timeCode: Usd.TimeCode) -> bool` | Return whether the fix is applicable |
+
```
tallytalwar, line 759: Maybe just refer to Adding Fixer section (line#282) here and doesn't need to mention the 2 methods again here?

pxr/usdValidation/usdValidation/README.md
```diff
@@ -746,6 +746,96 @@ registry.RegisterValidatorSuite(
...
+errors = validator.Validate(prim)
+
+for error in errors:
+    for f in error.GetFixersByErrorName():
+        editTarget = Usd.EditTarget(stage.GetRootLayer())
+        if f.CanApplyFix(error, editTarget):
+            f.ApplyFix(error, editTarget)
+```
+
+Note that `ApplyFix` calls `editTarget.GetLayer()->Save()` internally
```
tallytalwar: Thanks for adding this, but maybe move this above around line#339.. ?

pxr/usdValidation/usdValidation/README.md
```diff
@@ -746,6 +746,96 @@ registry.RegisterValidatorSuite(
...
+After validation, retrieve and apply fixers:
+
+```python
+validator = registry.GetOrLoadValidatorByName(
+    "myPackage:RequiresDocumentation"
+)
+
+# Fixers are accessible from the validator or from individual errors.
+fixers = validator.GetFixers()
+fixer = validator.GetFixerByName("AddPlaceholderDoc")
```
tallytalwar, line 822: Maybe document that such an API is useful for UI population to see what validators have fixers available.

pxr/usdValidation/usdValidation/wrapFixer.cpp
```diff
@@ -7,22 +7,94 @@
...
+        TfPyLock lock;
+        try {
+            object result = wrapper.Get()(
+                object(error),
+                object(editTarget),
+                object(timeCode));
+            return extract<bool>(result);
+        }
+        catch (error_already_set const &) {
+            TfPyConvertPythonExceptionToTfErrors();
```
tallytalwar, line 60: Maybe this is just a documentation question. From a pure python point of view its fine here, but say someone calls this from c++, in that case the python exception is flagged as a TfError in cpp. I think its worth a note about this in README maybe?

pxr/usdValidation/usdValidation/wrapFixer.cpp
```diff
@@ -7,22 +7,94 @@
...
+// Returns a heap-allocated pointer as required by make_constructor.
+UsdValidationFixer *
+_MakeValidationFixer(const TfToken &name,
+                     const std::string &description,
+                     object fixerImplFn,
+                     object canApplyFn,
+                     const TfTokenVector &keywords,
+                     const TfToken &errorName)
+{
+    return new UsdValidationFixer(
```
tallytalwar, line 83: Maybe check that both `fixerImplFn` and `canApplyFn` are callable? I think something like this might work:

```
  if (!PyCallable_Check(fixerImplFn.ptr()) ||
      !PyCallable_Check(canApplyFn.ptr())) {
      TfPyThrowTypeError("fixerImplFn and canApplyFn must be callable");
  }
```

And while you are adding this could you also add it to `_RegisterLayerValidator`, ``_RegisterStageValidator`, `_RegisterPrimValidator`, as well please in wrapRegistry.cpp?

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
tallytalwar: +1

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+
+        layer = Sdf.Layer.CreateAnonymous(".usda")
+        error = self._LayerTask(layer)[0]
+        editTarget = Usd.EditTarget(layer)
+
+        with self.assertRaises(AttributeError):
+            fixer.CanApplyFix(error, editTarget)
+
+    def test_ConstructionFixerImplRaises(self):
+        """Exceptions from fixerImplFn are restored at the Python boundary."""
```
tallytalwar, line 164: This is the thing I also mentioned in the wrapper code as well, but if a fixer is called from cpp, this exception will / should be made visible as TfError, I think we should at least document that. Unless you feel its important to also add a test for this.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(len(fixers), 1)
+        self.assertEqual(fixers[0].name, "layerFixer")
+
+        layer = Sdf.Layer.CreateAnonymous(".usda")
+        errors = validator.Validate(layer)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(layer)
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: Should these assert to true and false respectively.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(len(fixers), 1)
+        self.assertEqual(fixers[0].name, "stageFixer")
+
+        stage = Usd.Stage.CreateInMemory()
+        errors = validator.Validate(stage)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(stage.GetRootLayer())
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: assert here as well.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(fixers[0].name, "primFixer")
+
+        stage = Usd.Stage.CreateInMemory()
+        prim = stage.DefinePrim("/TestPrim")
+        errors = validator.Validate(prim)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(stage.GetRootLayer())
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: Assert here as well

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(len(fixers), 1)
+        self.assertEqual(fixers[0].name, "pluginLayerFixer")
+
+        layer = Sdf.Layer.CreateAnonymous(".usda")
+        errors = validator.Validate(layer)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(layer)
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: Assert here as well.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(len(fixers), 1)
+        self.assertEqual(fixers[0].name, "pluginStageFixer")
+
+        stage = Usd.Stage.CreateInMemory()
+        errors = validator.Validate(stage)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(stage.GetRootLayer())
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: assert here as well

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(fixers[0].name, "pluginPrimFixer")
+
+        stage = Usd.Stage.CreateInMemory()
+        prim = stage.DefinePrim("/PluginPrim")
+        errors = validator.Validate(prim)
+        self.assertEqual(len(errors), 1)
+
+        editTarget = Usd.EditTarget(stage.GetRootLayer())
+        fixers[0].CanApplyFix(errors[0], editTarget)
+        fixers[0].ApplyFix(errors[0], editTarget)
```
tallytalwar: assert here as well.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
```diff
@@ -0,0 +1,556 @@
...
+
+        layer = Sdf.Layer.CreateAnonymous(".usda")
+        errors = validator.Validate(layer)
+        self.assertEqual(len(errors), 1)
+
+        fixers = errors[0].GetFixersByErrorName()
+        self.assertEqual(len(fixers), 2)
+
+    def test_ThreadedFixerCallback(self):
+        """A Python-backed fixer callback can be invoked from threads."""
```
tallytalwar: I'm not sure this is actually testing the `TfPyLock` mechanism in the wrapper. If I'm reading it correctly, `fixer.CanApplyFix` is called from a Python thread, which already holds the GIL, so `TfPyLock` in _WrapFixerFn is just a recursive re-acquire. I believe the test passes with the TfPyLock removed, though worth confirming.

To test the path TfPyLock exists for, the callback would need to run on a thread created in C++ — a WorkDispatcher task or similar — that has never held the GIL

For validator callables something similar is exercised by running the validators through `UsdValidationContext` in `test_PythonValidatorInContext`, but `UsdValidationContext` provides that mechanism of running it via `WorkDispatcher`, unlike fixers for which the client is responsible for it.

tallytalwar (commented):
Added some more feedback. Thanks @miguelh-nvidia 

lmk if you want to talk about any of this with me. Thanks

pxr/usdValidation/usdValidation/README.md
miguelh-nvidia: Fixed (though in a latter commet, this is removed)

pxr/usdValidation/usdValidation/README.md
miguelh-nvidia: Applied

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Test removed, since it is almost similar to LayerValidation.

pxr/usdValidation/usdValidation/wrapFixer.cpp
miguelh-nvidia: Doc changed.

pxr/usdValidation/usdValidation/README.md
miguelh-nvidia: Done.

pxr/usdValidation/usdValidation/README.md
miguelh-nvidia: Moved, please check.

pxr/usdValidation/usdValidation/README.md
miguelh-nvidia: Comment added before this code block.

pxr/usdValidation/usdValidation/wrapFixer.cpp
miguelh-nvidia: Added documentation.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Done

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: You’re right, this was redundant. I checked the case I originally had in mind `_WrapFixerFn(const object &pyFn)` still passes because the returned lambda stores `TfPyObjWrapper(pyFn)` by value. Since this test does not catch that distinction and overlaps with the other construction/invocation coverage, I removed it.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Documentation added.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: This follows the same `plugInfo.json` setup pattern as the other Python tests, for example `testUsdValidationPyPlugin.py`, `testUsdValidationRegistryPy.py`, and `testUsdValidationRegistryPyRegister.py`. There isn’t a fixer-specific reason for it here. Should I remove it from this test?

pxr/usdValidation/usdValidation/wrapRegistry.cpp
miguelh-nvidia: This has been rewritten.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
miguelh-nvidia: Good point, changed.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
miguelh-nvidia: Changed.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
miguelh-nvidia: Changed.

pxr/usdValidation/usdValidation/wrapRegistry.cpp
miguelh-nvidia: Changed.

pxr/usdValidation/usdValidation/wrapFixer.cpp
miguelh-nvidia: Good point, conditions and tests added.

pxr/usdValidation/usdValidation/testenv/testUsdValidationFixerPy.py
miguelh-nvidia: Good point. This test only exercised Python-created threads, so it was not actually validating the C++-created-thread path that TfPyLock protects. I removed it rather than keep a misleading test.

miguelh-nvidia:
@nvmkuruc I gave this a try, but for these tests the parameterized version mostly moved the repetition into the case data and was harder to follow. I think the explicit layer/stage/prim tests are clearer and less fragile here, so I left them as-is.

tallytalwar (approved):
Thanks @miguelh-nvidia for addressing the notes, getting this setup to be checked in internally.

PR #4144, opened 2026-07-08
Autodesk: GfColorSpace must always initialize transfer function parameters

erikaharrison-adsk:
### Description of Change(s)

Modified the GfColorSpace initialization code in nanocolor.c, so that it always initializes the transfer function parameters K0 and phi.

Added unit test coverage to testGfColorCpp.cpp.

Details: 
Currently, if a GfColorSpace is created using a 3x3 matrix rather than chromaticity coordinates, K0 and phi are not initialized to their correct values. Leaving them both at 0. will generate either Inf or NaN in the _FromLinear or _ToLinear functions when evaluating the resulting color space.

In addition, there is a more subtle bug when using the UsdColorSpaceDefinitionAPI and CreateColorSpaceAttrsWithMatrix to create color spaces that use the sRGB transfer function. Because the test for the linear segment in _ToLinear results in NaN, that branch is never taken, and only the power function is used. This results in the chromaticity coordinates extracted from the 3x3 matrix having small errors. If ComputeColorSpaceFromDefinitionAttributes is then called to get the color space and GetRGBToXYZ is used to extract the regenerated 3x3 matrix, it no longer matches the original matrix used to initialize the color space definition.

- N/A

### Fixes Issue(s)

- N/A

### Checklist

### DIFF
--- a/pxr/base/gf/nc/nanocolor.c
+++ b/pxr/base/gf/nc/nanocolor.c
@@ -338,12 +338,15 @@ static NcM33f _M33fMultiply(NcM33f lh, NcM33f rh) {
 }
 
 static void _InitColorSpace(NcColorSpace* cs) {
-    if (!cs || cs->rgbToXYZ.m[8] != 0.0)
+    if (!cs)
         return;
 
     const float a = cs->desc.linearBias;
     const float gamma = cs->desc.gamma;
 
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-12432

PR #4145, opened 2026-07-08
Autodesk: Conversions from Data GfColorSpace should not modify RGB values

erikaharrison-adsk:
### Description of Change(s)

- Adds an isData member to NcColorSpace.
- Sets isData=true for built-in color spaces Data, Raw, Unknown, and Identity.
- Adds an isData method on GfColorSpace.
- Adds isData to Python binding in wrapColorSpace.cpp.
- Adds an NcIsDataColorSpace function in nanocolor.c.
- Color conversion functions between color spaces return as a no-op if isData=true for either the src or dst color space.
- GetRGBToRGB returns an identity matrix if isData=true for either the src or dst color space.
- NcColorSpaceEqual is updated to compare isData.
- NcColorSpaceEqual was fixed to correct the comparison on shortName. 

Extensive unit tests are provided in testGfColorCpp.cpp. Validation of isData is provided in testGfColorPy.py.

#### Details

Currently, conversions between the built-in Data GfColorSpace and other color spaces modify the RGB values. Such conversions should be a no-op.

Additionally fixes a small bug in NcColorSpaceEqual where the shortName was not being compared correctly.

#### Open Questions

- Should isData=true for all four spaces: Data, Raw, Unknown, and Identity? (My view is "yes".)
- Should comparison between two different isData spaces return equality? (I'm tempted to say "yes", but the existing comparison does test based on the name too, so the implemented behavior is "no".)
- Should it be possible to create an isData=true color space via ColorSpaceDefinitionAPI? (I don't feel it's necessary since the only difference would be the name.)

- N/A

### Fixes Issue(s)

- N/A

### Checklist


jesschimein:
Filed as internal issue #USD-12433

doug-walker:
Here's a relevant line of code from [third_party/renderman/plugin/hdPrman/renderingColorSpaceSceneIndexPlugin.cpp](https://github.com/PixarAnimationStudios/OpenUSD/blob/3e3890068e17f0e31dc34f906f7f1fa1037dd7da/third_party/renderman/plugin/hdPrman/renderingColorSpaceSceneIndexPlugin.cpp#L210) . It's working around the defect fixed by this PR by manually bypassing conversions if the color space is Raw, Identity, or Data. It should not be necessary for client code to handle this, it would be error-prone and difficult to maintain.

I notice that one of the CI checks failed, but I don't believe it is related to this PR.

PR #4146, opened 2026-07-09
[hgiWebGPU] Add pxr/imaging/hgiWebGPU with minimal Hgi/HgiGL/HgiInterop plumbing

kohakukun:
### Description of Change(s)

Add pxr/imaging/hgiWebGPU with minimal Hgi/HgiGL/HgiInterop plumbing
Brings pxr/imaging/hgiWebGPU and pxr/imaging/hgiInterop/webgpu.{cpp,h},
plus the necessary Hgi API surface the backend needs:
- New HgiShaderTextureTypeDepth and HgiDeviceCapabilitiesBits enumerators
  (GeometricStage, PushConstants, Timestamps, TriangleLineFill, Tessellation)
- WebGPU token and platform default support
- HGI_HAS_IEEE_FLOAT_SPECIAL_VALUES capability macro for shader guards

Key design: hgiWebGPU carries its own copy of the GLSL shader-section base
implementation (HgiWebGPUShaderSection with WriteDeclaration/WriteParameter/
VisitGlobal*/storage-qualifier support). This avoids refactoring hgiGL or
the shared Hgi base, keeping the WebGPU work self-contained and minimizing
upstream merge conflicts. hgiGL/shaderSection.{h,cpp} and hgi/shaderSection.{h,cpp}
remain untouched.

CMake wiring (PXR_ENABLE_WEBGPU_SUPPORT, Dawn/Tint/glslang/SPIRV-Tools find_package,
FindTint.cmake) enables native builds via build_usd.py with --webgpu flag, which
already had the dependency installers in place.

This pr is not the complete work required for webgpu but lays the foundation. This gets us the first pixels using webGPU as a backend.

There are some issues with tesselation, push constants, and some other things that will be followed by smaller prs: 

image


jesschimein:
Filed as internal issue #USD-12435

pxr/imaging/hdSt/shaders/domeLight.glslfx
```diff
@@ -108,9 +108,11 @@ vec3 GetCubemapVec(ivec3 coords)
 
 float SanitizeValue(float value)
 {
+#if HGI_HAS_IEEE_FLOAT_SPECIAL_VALUES
```
tcauchois, line 111: Is it possible to have HgiWebGPU just define local copies of isnan & isinf instead?  That seems more robust.

pxr/imaging/hgi/enums.h
```diff
@@ -63,6 +63,18 @@ using HgiBits = uint32_t;
...
+/// <li>HgiDeviceCapabilitiesBitsGeometricStage:
+///   Support for geometric shader stage</li>
+/// <li>HgiDeviceCapabilitiesBitsPushConstants:
+///   Support for push constants</li>
+/// <li>HgiDeviceCapabilitiesBitsTimestamps:
+///   Supports GPU timestamps</li>
+/// <li>HgiDeviceCapabilitiesBitsTriangleLineFill:
+///   Triangles can be rasterized as their edge lines.</li>
+/// <li>HgiDeviceCapabilitiesBitsTessellation:
+///   Supports tessellation, either Metal tessellation or
```
tcauchois: Naively it seems like you have this backwards?  MetalTessellation implies Tessellation, but Tessellation *doesn't* imply MetalTessellation?

(I think hgi capabilities for metal/GL/Vulkan also need to be updated for these bits)

pxr/imaging/hdSt/shaders/domeLight.glslfx
kohakukun: Well, the issue is that there is no deterministic polyfill for that. The closest is 
`any(lessThanEqual(val,vec4(-FLT_MAX))) || any(greaterThanEqual(val;,vec4(FLT_MAX))))` for is Inf but it is basically defined by the vendor implementation. That is why we went for the explicit definition.

If I remember correctly, in the case of the domeLight, based on how we load the textures we shoul not have to fall on the isInf or isNan path(this was a long time ago, so I might be wrong)

For the frustrum, we check against `FLT_MAX` since that is what the cpu side does and it is tighly coupled with the implementation.

Actually, what I wanted to do is remove usage of isNan and isInf altogether as it is the most deterministic

pxr/imaging/hgi/enums.h
kohakukun: Good point. Also, these capabilities are not used yet in this pr. I was planning on introducing it as a follow up, so I just removed them for now

pxr/imaging/hdSt/shaders/domeLight.glslfx
tcauchois: We've seen nans (in particular) pop up in weird places, like in driver-generated mipmaps for certain bit depths, etc.  So those checks need to stay for the vulkan/metal/gl paths at least.

The reason I don't like the define is that it makes it really hard to maintain shader code across different implementations, e.g. if someone develops a shader for vulkan and uses isnan but forgets to check the define.  It would be preferable to have HgiWebGPU define its own isnan, which either does a FLT_MAX check or a "a != a" check, or even just hardcodes return false to skip the sanitization logic entirely.

pxr/imaging/hgiWebGPU/baseGraphicsCmds.h
```diff
@@ -0,0 +1,344 @@
+//
+// Copyright 2026 Pixar
```
tcauchois, line 2: Would be great to split the implementations out into a cpp file here.

pxr/imaging/hgiWebGPU/debugCodes.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+
+    TF_REGISTRY_FUNCTION(TfDebug)
+    {
+        TF_DEBUG_ENVIRONMENT_SYMBOL(HGIWEBGPU_DEBUG_GRAPHICS_PIPELINE,
+                                    "HgiWebGPU report graphics pipeline attributes descriptors.");
+        TF_DEBUG_ENVIRONMENT_SYMBOL(HGIWEBGPU_DEBUG_SHADER_CODE,
+                                    "HgiWebGPU report graphics pipeline attributes descriptors.");
```
tcauchois, line 20: C&P error.

pxr/imaging/hgiWebGPU/README.md
```diff
@@ -0,0 +1,64 @@
...
+  instancing over a translucent material, the number of bound buffers will reach
+  the WebGPU limit, with all 10 out of 10 storage buffers being used.
+
+- After conversion to WebAssembly, the WASM package has a 90MB size. There
+are two problems that  make it challenging to reduce the size:
+  -  In WASM, we have to create a monolithic package, which includes all the libraries and
+possible plugins that the final application will use. This could be improved by using dynamic libraries
+and loading them at runtime as needed from the server.
+  -  We cannot apply `-O3` optimizations to the code, because it removes some of the symbols that
+should be kept. This can be solved by tagging the functions that should not be deleted with the `EMSCRIPTEN_KEEPALIVE`
```
tcauchois, line 46: Might be a bigger discussion than this PR but we use the <FOO>_API tags for basically this purpose, to keep symbols visible when linking on windows.  I wonder if we can fold the emscripten define into this system?  Via like an emscripten-specific addition to arch/export.h or something.

pxr/imaging/hgiWebGPU/shaderCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // convention, Hgi wants us to render upside down, which we
+        // do by negating the gl_Position.y value. Metal has the
+        // same issue, and it solves it by using a negative viewport
+        // height, which isn't supported by WebGPU, so we have to take
+        // a more complicate approach... Flipping Y also changes
+        // the handedness of the coordinate system, which reverses
+        // the winding order. For that we solve it the same way as
+        // Metal does: return the opposite in
+        // HgiWebGPUConversions::GetWinding().
+        if (ARCH_UNLIKELY(!ApplySpirvViewportFlip(*spirvOut))) {
```
tcauchois, line 146: IIRC for Vulkan we just rendered upside down and encoded a flip in the presentation code and that was less of a headache, since this has the potential to mess up existing image shaders as well.

pxr/imaging/hgiWebGPU/shaderCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    std::string warningErrors = logger.getAllMessages();
+
+    if (!warningErrors.empty()) {
+        errors->append(warningErrors);
+        return false;
+    }
+
+    // Storm generates a lot of dead code, and even dead buffer bindings that
+    // waste precious binding slots (we only have 10!). Clean this up.
```
tcauchois, line 122: Yeah totally, and this is great, but also I'm curious what dead buffer bindings we're generating?

pxr/imaging/hgiWebGPU/spirvTransforms.h
```diff
@@ -0,0 +1,31 @@
...
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+TF_DEBUG_CODES(
+    HGIWEBGPU_DEBUG_SPIRV_TRANSFORM
+);
+
+HGIWEBGPU_API
+bool ApplySpirvViewportFlip(
+    std::vector<uint32_t>& spirv);
```
tcauchois, line 23: This module is incredibly cool and was a great read, but these seem like things that are much better addressed in shader generation?

As mentioned, the viewport flip is pretty easy to do e.g. by compensating the projection matrix or adding a line to the vertex shader based on capability bit, although we've found that you need to adjust sampling code correspondingly which is much harder, and it's kinda easier to just keep everything upside down in hydra and flip on present or readback...

Alpha to one emulation similarly could just be a fragment shader mixin triggered on capability bit.

Issue #4135, opened 2026-07-02
Syntax errors in usdview's Python interpreter not handled correctly with Python 3.13+


mattyjams:
Hello!

I've run across a subtle usdview Python interpreter breakage when building OpenUSD against Python 3.13.

Between Python 3.12 and 3.13, it looks like the function signature for `InteractiveInterpreter.showsyntaxerror()` was changed such that it now accepts keyword arguments:

https://github.com/python/cpython/blob/3.12/Lib/code.py#L96

vs.

https://github.com/python/cpython/blob/3.13/Lib/code.py#L98

Seemingly as a result of this commit in Python:

https://github.com/python/cpython/commit/c8f4069ab1602a1f67239fef0e11cc3e72c0045d

In `Usdviewq`, a derived class of `InteractiveInterpreter` is created that overrides this function:

https://github.com/PixarAnimationStudios/OpenUSD/blob/3e3890068e17f0e31dc34f906f7f1fa1037dd7da/pxr/usdImaging/usdviewq/pythonInterpreter.py#L152

But the override still has the pre-3.13 function signature. A problem arises when Python itself tries to invoke the overridden `showsyntaxerror()` and pass it `source` as a keyword argument:

https://github.com/python/cpython/blob/3.13/Lib/code.py#L68

The override on the Usdviewq side does not accept keyword arguments, so a separate exception is raised.

The net result is that when using usdview when built against Python 3.13, syntax errors that result from code entered in the Python interpreter are not being handled or displayed correctly and instead raise exceptions into the terminal, for example simply entering the string "A Syntax Error" (without the quotes) into the interpreter:

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/code.py", line 65, in runsource
    code = self.compile(source, filename, symbol)
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/codeop.py", line 155, in __call__
    return _maybe_compile(self.compiler, source, filename, symbol)
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/codeop.py", line 75, in _maybe_compile
    return compiler(source, filename, symbol, incomplete_input=False)
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/codeop.py", line 118, in __call__
    codeob = compile(source, filename, symbol, flags, True)
  File "<input>", line 1
    A Syntax Error
      ^^^^^^
[...]

PR #4151, opened 2026-07-10
fix usdviewq SyntaxError handling for Python 3.13+

mattyjams:
### Description of Change(s)

Between Python 3.12 and 3.13, the function signature for `InteractiveInterpreter.showsyntaxerror()` was changed such that it now accepts keyword arguments (see https://github.com/python/cpython/commit/c8f4069ab1602a1f67239fef0e11cc3e72c0045d).

In `usdviewq`, a derived class of `InteractiveInterpreter` is created that overrides `showsyntaxerror()`, but it still had the signature that does not accept keyword arguments.

Python itself tries to invoke the overridden `showsyntaxerror()`, and beginning with 3.13 passes it `source` as a keyword argument. The override on the `usdviewq` side did not accept keyword arguments, so a separate exception was raised.

The net result is that when using `usdview` when built against Python 3.13, syntax errors that result from code entered in the Python interpreter were not being handled or displayed correctly and instead raised exceptions into the terminal.

This is addressed by updating the overridden `showsyntaxerror()` in `usdviewq` to accept keyword arguments, and to pass those along to the parent class' version. This solution works for pre-3.13 versions of Python as well.

A new test function was added to the existing `testUsdviewInterpreterNoRender.py` test to verify that `SyntaxError`s are handled correctly, and I confirmed that all tests pass when OpenUSD is built against either Python 3.11.9 or 3.13.14.

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/4135

### Checklist

- [ ] I have submitted a signed Contributor License Agreement (Reference: 

### DIFF
--- a/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInterpreterNoRender/testUsdviewInterpreterNoRender.py
+++ b/pxr/usdImaging/bin/testusdview/testenv/testUsdviewInterpreterNoRender/testUsdviewInterpreterNoRender.py
@@ -46,7 +46,31 @@ def _testInterpreterWorks(appController):
     # initializes. This is pretty brittle, but other options seem just as 
     # brittle,
     assert "usdviewApi" in appController._console.locals()
-    
+
+def _testInterpreterHandlesSyntaxError(appController):
+    #
+    # Simulate entering some invalid Python code into the interpreter and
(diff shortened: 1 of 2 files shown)


jesschimein:
Filed as internal issue #USD-12443

PR #4153, opened 2026-07-10
[26.8 alpha bug fix] Fix --no-build-apple-framework on embedded platforms

dgovil:
### Description of Change(s)

The `--no-build-apple-framework` flag was ignored on embedded platforms because buildAppleFramework was computed as:

    `args.build_apple_framework or MacOSTargetEmbedded(self)`

On embedded platforms MacOSTargetEmbedded() is True, so the 'or' short-circuited to True even when the user explicitly passed --no-build-apple-framework (which sets args.build_apple_framework to False).

Give the flag a None default so an unspecified flag can be distinguished from an explicit False, and only fall back to the embedded platform default when the flag is unspecified. Embedded platforms still default to building the framework, but can now disable it.

### Checklist

### DIFF
--- a/build_scripts/build_usd.py
+++ b/build_scripts/build_usd.py
@@ -2127,12 +2127,14 @@ if MacOS():
         "--build-apple-framework",
         dest="build_apple_framework",
         action="store_true",
+        default=None,
         help=("Build USD as an Apple Framework "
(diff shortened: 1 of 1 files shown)


sunyab:
Filed as internal issue #USD-12445

Issue #4157, opened 2026-07-15
Massive performance degradation of HdDependencyForwardingSceneIndex::_PrimsRemoved (with lots of prims)


sebastienberube-adsk:
Since these (semi-)recent OpenUSD commits:

663d3dd _[Hd] Optimize DependencyForwardingSceneIndex prim removal._
09a9d9b _[Hd] HdDependencyForwardingSceneIndex notice handler improvements._

The performance of [HdDependencyForwardingSceneIndex::_PrimsRemoved](https://github.com/PixarAnimationStudios/OpenUSD/blob/c32a44659e16a8a3c132d3205dd5c439317f6aee/pxr/imaging/hd/dependencyForwardingSceneIndex.cpp#L171-L306) is orders of magnitude worse than it was before with the same internal state and the same removed prims input (in some cases, it is >100X worse).

Subsequent commits such as:

708ab23 _[hd] DependencyForwardingSceneIndex notice optimization thread utilization._
6fafb55 _[hd] HdDependencyForwardingSceneIndex now avoids calling GetPrim()_

do not fully fix the underlying algorithmic complexity issue, and some prim removal operations that were previously instaneous not take minutes to complete.

<img width="1703" height="798" alt="Image" src="https://github.com/user-attachments/assets/70c0f82f-a286-443c-bb62-d1cebb5ec15b" />

That one use case shown above is indeed an extreme one, but we also notice a significant reduction of performance in scenes with a more moderate amount of prims.

We have a local fix for this, where we transfer the input removed entries prims into a local SdfPath HashSet, and replace the inner loop that repeatedly runs for all removed prims with queries of this new SdfPath HashSet.
See attached file.

I see in the original code a comment that suggests something like sorting the received removed prims, perhaps that could also work instead of allocating a new hashset, but would probably still be slower than a hashset in more extreme use cases.

Proposed fix based on a SdfPath HashSet:
[OpenUSD-3ff40b2-Fix HdDependencyForwardingSceneIndex_PrimsRemoved.patch](https://github.com/user-attachments/files/30062793/OpenUSD-3ff40b2-Fix.HdDependencyForwardingSceneIndex_PrimsRemoved.patch)

tcauchois:
Prior to that change, this scene index was potentially not sending correct invalidations on subtree deletion.  The notice handler is doing a lot more work now, but that work is necessary for correct behavior.

[...]

PR #4133, opened 2026-07-01
Add `flaky` label to `testWorkThreadLimitsRawTBBMax`

nvmkuruc:
### Description of Change(s)

#4054 added support for labels and tagged several `work` tests known to spuriously fail as `flaky`.  #4055 added a retry for these `flaky` tests.

This adds `testWorkThreadLimitRawTBBMax` to that list as it recently failed on a MacOS CI job.

https://github.com/PixarAnimationStudios/OpenUSD/actions/runs/28410432254/job/84182124043?pr=4130

```
1017/1022 Test  #237: testWorkThreadLimitsRawTBBMax ....................................***Failed    0.07 sec
```

### Checklist

- [x] I have created this PR based on the dev branch

- [x] I have followed the [coding conventions](https://openusd.org/release/api/_page__coding__guidelines.html)

- [ ] I have added unit tests that exercise this functionality (Reference: 
  [testing guidelines](https://openusd.org/release/api/_page__testing__guidelines.html))

- [ ] I have verified that all unit tests pass with the proposed changes

- [x] I have submitted a signed Contributor License Agreement (Reference: 
  [Contributor License Agreement instructions](https://openusd.org/release/contributing_to_usd.html#contributor-license-agreement))

PR #4160, opened 2026-07-17
Autodesk: Fix HdDependencyForwardingSceneIndex::_PrimsRemoved

ashwinbhat:
### Description of Change(s)

Improve performance of [HdDependencyForwardingSceneIndex::_PrimsRemoved](https://github.com/PixarAnimationStudios/OpenUSD/blob/c32a44659e16a8a3c132d3205dd5c439317f6aee/pxr/imaging/hd/dependencyForwardingSceneIndex.cpp#L171-L306) with large number of removed prims. For use cases where `entries` input has thousands of prims, avoid the inner loop and use a SdfPath HashSet instead.

The previous logic based on `SdfPath::HasPrefix(SdfPath other)`, which would return true if `other` is either equal or a parent of the SdfPath was replaced with the use of a new local function `hasBeenRemoved(const& SdfPath path)` which walks up through the parents of `path` (including /) and returns true on the first hashset hit.

### Fixes Issue(s)

https://github.com/PixarAnimationStudios/OpenUSD/issues/4157

### Checklist

### DIFF
--- a/pxr/imaging/hd/dependencyForwardingSceneIndex.cpp
+++ b/pxr/imaging/hd/dependencyForwardingSceneIndex.cpp
@@ -8,6 +8,7 @@
 #include "pxr/imaging/hd/dependenciesSchema.h"
 #include "pxr/imaging/hd/retainedDataSource.h"
 #include "pxr/imaging/hd/overlayContainerDataSource.h"
+#include "pxr/base/tf/hashset.h"
 #include "pxr/base/trace/trace.h"
 #include "pxr/base/work/loops.h"
 #include "pxr/base/work/utils.h"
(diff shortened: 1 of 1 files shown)


sebastienberube-adsk:
Note: I see `testWorkThreadLimitsRawTBBMax` failing on MacOS, it does not appear to be related to the current changes, though.

See this existing PR (opened by somebody else) that flags that test as flaky:
https://github.com/PixarAnimationStudios/OpenUSD/pull/4133

See also these other PRs having the same error:
https://github.com/PixarAnimationStudios/OpenUSD/pull/4129
https://github.com/PixarAnimationStudios/OpenUSD/pull/4130

pxr/imaging/hd/dependencyForwardingSceneIndex.cpp
```diff
@@ -188,9 +189,34 @@ HdDependencyForwardingSceneIndex::_PrimsRemoved(
...
+    // Create a hashset of entries
+    TfHashSet<SdfPath, SdfPath::Hash> removedEntriesHashSet;
+    removedEntriesHashSet.reserve(entries.size());
+    for (const HdSceneIndexObserver::RemovedPrimEntry& entry : entries) {
+        removedEntriesHashSet.insert(entry.primPath);
+    }
+
+    // Function to quickly check if a prim (including any parent prim) has been removed.
+    // (When a parent prim is removed, all child prims are also removed).
+    auto hasBeenRemoved = [&removedEntriesHashSet](const SdfPath& primToCheck)
```
sebastienberube-adsk, line 201: Additional suggestion: To avoid the hashmap overhead, this function could be assigned a different implementation depending on the number of prims being removed.

If the count is below a certain threshold (e.g. `size(entries) < 1000`), then we could avoid allocating and filling the hashmap and just loop through entries like the function was doing before:

```
if(size(entries) < 1000)
{
    // Lightweigth use case: simply loop through entries
    hasBeenRemoved = [&entries](const SdfPath& primToCheck)
    {
       for (const HdSceneIndexObserver::RemovedPrimEntry &entry : entries){
          if (pair.first.HasPrefix(entry.primPath)) {
               return true;
          }
       }
       return false;
    };
}
else
{
    // Heavy prim removal: use a hashset
    removedEntriesHashSet.reserve(entries.size());
    for (const HdSceneIndexObserver::RemovedPrimEntry& entry : entries) {
        removedEntriesHashSet.insert(entry.primPath);
    }
   
    hasBeenRemoved = [&removedEntriesHashSet](const SdfPath& primToCheck)
    {
        // Check if the prim has been removed, or if a parent of the prim
        for (SdfPath path = primToCheck; !path.IsEmpty(); path = path.GetParentPath())
        {
            if (removedEntriesHashSet.find(path) != removedEntriesHashSet.end())
            {
                return true;
            }
            if (path.IsAbsoluteRootPath())
            {
                break;
            }
        }
        return false;
    };
}
```

Let me know if this strategy would be worth it to balance the performance of heavy prim removal and lightingweight prim removal. I have not included this idea in the PR code because it makes the function more complicated, but I am leaving the idea here in case the reviewer of this PR has an opinion on the matter.

jesschimein:
Filed as internal issue #USD-12468

erikaharrison-adsk:
@jesschimein My understanding is Ashwin is already on the CLA list (I was out on vacation; he's happy to be removed/not on it though). Please advise if we should close and re-open or not.

jesschimein:
Hi Erika -- yes, that's right; we didn't have his github handle logged, so the automation flagged the CLA, but we did have his name as a valid contributor in our records when the PR was submitted!  (Note that he was removed from the CLA list as of yesterday)

PR #4164, opened 2026-07-22
usdPhysics: exclude disabled colliders from mass computation

andrewkaufman:
### Description of Change(s)

`UsdPhysicsRigidBodyAPI::ComputeMassProperties` gathers the colliders beneath a rigid body to compute its aggregated mass, center of mass, and inertia. The gathering loop collected every descendant prim with `PhysicsCollisionAPI` applied without checking whether that collider was actually enabled, so a collider with `physics:collisionEnabled = false` still contributed to the body's mass properties.

Per the [`UsdPhysicsCollisionAPI` documentation](https://openusd.org/release/api/class_usd_physics_collision_a_p_i.html), `physics:collisionEnabled` "Determines if the PhysicsCollisionAPI is enabled." A collider whose collision is disabled takes no part in simulation, so it should not contribute to the body's aggregated mass, center of mass, or inertia.

This change skips colliders whose resolved `physics:collisionEnabled` is `false` when gathering shapes for the mass computer. Enabled colliders (including those relying on the `true` schema fallback) are unaffected.

**Observed behavior before/after** — a rigid body with one enabled unit cube and one disabled unit cube offset along z:

| | mass | center of mass |
|---|---|---|
| before | 2000 (both cubes) | (0, 0, 2) — shifted by the disabled cube |
| after | 1000 (enabled cube only) | (0, 0, 0) |

N/A — bug fix in existing behavior, no schema or API change.

### Fixes Issue(s)

- Disabled colliders (`physics:collisionEnabled = false`) incorrectly contributing to `ComputeMassProperties` mass aggregation.

### Checklist

### DIFF
--- a/pxr/usd/usdPhysics/parseUtils.cpp
+++ b/pxr/usd/usdPhysics/parseUtils.cpp
@@ -1635,6 +1635,7 @@ bool _HasDynamicBodyParent(const UsdPrim& usdPrim, const RigidBodyMap& bodyMap,
 {
     bool physicsAPIFound = false;
     UsdPrim parent = usdPrim;
+    UsdPrim disabledBodyPrim = UsdPrim();
     while (parent != usdPrim.GetStage()->GetPseudoRoot())
     {
         if (_IsDynamicBody(parent, bodyMap, &physicsAPIFound))
@@ -1645,11 +1646,26 @@ bool _HasDynamicBodyParent(const UsdPrim& usdPrim, const RigidBodyMap& bodyMap,
 
         if (physicsAPIFound)
         {
-            *outBodyPrimPath = parent;
-            return false;
+            // A disabled rigid body takes no part in simulation, so it does
+            // not own this prim. Keep searching the ancestors: a nested
+            // disabled body may still have an enabled body above it, which
+            // this prim belongs to. Remember the nearest disabled body so it
+            // can still be reported if no enabled body is found at all.
+            if (disabledBodyPrim == UsdPrim())
+            {
+                disabledBodyPrim = parent;
+            }
         }
 
         parent = parent.GetParent();
+    }
+
+    // No enabled body above this prim. Report the nearest disabled body, if
+    // any, so it is still recognized as belonging to a body that is present
+    // but not simulating rather than as a static collision.
+    if (disabledBodyPrim != UsdPrim())
+    {
+        *outBodyPrimPath = disabledBodyPrim;
     }
     return false;
 }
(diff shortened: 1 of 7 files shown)


andrewkaufman:
@AlesBorovicka per our discussion today

andrewkaufman:
https://github.com/PixarAnimationStudios/OpenUSD/pull/4165 contains the same fix, but then goes further to add a public method to retrieve the colliders associated with the given body.

Either PR landing fixes the bug. If https://github.com/PixarAnimationStudios/OpenUSD/pull/4165 lands it would be nice to keep those 2 commits independant to make backport cherry-picks simpler for folks who aren't able to incorporate a new public method.

AlesBorovicka:
Thanks the changes look good to me

AlesBorovicka:
As for the collider get, if we agree to add these custom functions, we probably should discuss how far we want to go. If we add this, we should probably add symetric get rigid body on the colliders and very likely things like get rigid body 0/1 on joints. I am not 100% sure how far we should go here.

jesschimein:
Filed as internal issue #USD-12473

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1645,12 +1646,27 @@ bool _HasDynamicBodyParent(const UsdPrim& usdPrim, const RigidBodyMap& bodyMap,
...
         parent = parent.GetParent();
     }
+
+    // No enabled body above this prim. Report the nearest disabled body, if
+    // any, so it is still recognized as belonging to a body that is present
+    // but not simulating rather than as a static collision.
+    if (disabledBodyPrim != UsdPrim())
+    {
+        *outBodyPrimPath = disabledBodyPrim;
+    }
```
andrewkaufman, line 1669: I'm not sure about this... It matches the previous logic, but why was a collider belonging to a disabled body considered part of a body at all?

andrewkaufman:
@AlesBorovicka I updated this to account for diabled nested bodies as well. I think we should discuss in Physics WG though. Does a disabled body still own the colliders underneath it? Or do they now belong to the parent body? I can see the argument both ways...

andrewkaufman:
Superseded by https://github.com/PixarAnimationStudios/OpenUSD/pull/4187, which carries these three commits unchanged and adds the two remaining traversals.

@AlesBorovicka to answer the question I left above, about whether a disabled body still owns the colliders underneath it: the colliders belong to the nearest enabled body above it. `physics:rigidBodyEnabled = false` means the body takes no part in simulation, which is equivalent to not applying the API at all, so an ancestor search should continue past it.

Applying that consistently turned up two more traversals that stopped at a disabled body, which is why the new PR replaces this one rather than building on it. `_GetBodyPrim`, which resolves a joint body relationship, never consulted `physics:rigidBodyEnabled` at all, and `PhysicsJointChecker` reported a joint below a disabled body as bodyless even with an enabled body above it. Together with the collider attribution here, a collider under a disabled inner body under an enabled outer body resolved three different ways depending on which entry point you asked. All three now agree.

https://github.com/PixarAnimationStudios/OpenUSD/pull/4165 is unaffected as a follow-on and can rebase onto #4187.

PR #4165, opened 2026-07-22
usdPhysics: add public getters for joint/collider body resolution and refactor the joint parser onto them

andrewkaufman:
### Description of Change(s)

Adds a set of public getters that expose the body-resolution and mass-collider traversals that previously lived only inside usdPhysics internals, and refactors the joint parser to use them so the resolution logic has a single implementation.

**New API:**

- `UsdPhysicsRigidBodyAPI::GetCollisionPrims()` returns the collider prims belonging to a rigid body, the same traversal `ComputeMassProperties()` uses internally to decide which colliders' mass is accumulated into the body. Starting at the rigid body's prim the subtree is traversed (including instance proxies) and every prim with `UsdPhysicsCollisionAPI` applied is collected, with two exceptions: a collider whose `physics:collisionEnabled` resolves to `false` is skipped, and the subtree rooted at a nested prim with `UsdPhysicsRigidBodyAPI` is pruned (those colliders belong to that other body). `ComputeMassProperties()` is refactored to call it.
- `UsdPhysicsCollisionAPI::GetBody()` resolves a collider to its owning enabled rigid body via the ancestor walk, returning an invalid prim when there is none.
- `UsdPhysicsJoint::GetBody0()` / `GetBody1()` return the enabled rigid body each side of the joint attaches to, or an invalid prim.
- `UsdPhysicsJoint::GetLocalPose0()` / `GetLocalPose1()` return the joint-local anchor pose in that body's frame, with the frame scale baked into the translation (physics is scale-free) and an optional `UsdGeomXformCache` for nested hierarchies.

**Refactor:** the joint parser's private `_GetBodyPrim` / `_GetLocalPose` are deleted and `_FinalizeJoint` now calls the public getters, so joint body resolution has a single implementation shared between the parser and clients.

**Behavior change:** a joint side with no enclosing enabled body now gives an empty `body0`/`body1`, where the previous parser fell back to the nearest collider. This is called out for review in the PR discussion.

Each getter has Python wrapping and regression tests.

**Motivation:** the traversals that map a rigid body to its colliders, and a joint or collider to its owning body, previously existed only inside `ComputeMassProperties` and the joint parser. Clients that need this had to re-implement it (instance-proxy traversal and enabled-body resolution are easy to get subtly wrong). Exposing the canonical logic lets clients reuse it and lets the parser drop its private copy.

### Link to proposal

N/A, additive public API, no schema change.

### Dependency

Based on top of #4187 (treat disabled rigid bodies as unapplied in all ancestor traversals); these getters share that enabled-body filter. Restacked from #4164, which #4187 supersedes.

### AI Assistance

Developed with AI assistance (Anthropic Claude, primarily Opus 5): the code, tests, and this description were AI-drafted and then reviewed by me before submission.

### Checklist

- [x] I have followed the coding conventions

### DIFF
--- a/pxr/usd/usdPhysics/massAPI.h
+++ b/pxr/usd/usdPhysics/massAPI.h
@@ -37,16 +37,14 @@ class SdfAssetPath;
 ///
 /// Defines explicit mass properties (mass, density, inertia etc.).        
 /// MassAPI can be applied to any object that has a PhysicsCollisionAPI or
-/// a PhysicsRigidBodyAPI. MassAPI only has an effect where the underlying
-/// PhysicsCollisionAPI or PhysicsRigidBodyAPI is enabled; a collider whose
-/// physics:collisionEnabled is false, or a rigid body whose
-/// physics:rigidBodyEnabled is false, takes no part in simulation, so any
-/// MassAPI applied to it is likewise ignored. When neither an enabled
-/// collider nor an enabled rigid body remains on a prim, it is not a physics
-/// prim and its MassAPI has no effect. Note that this concerns only the prim
-/// the MassAPI is applied to: disabling a prim does not discard the mass of
-/// the enabled colliders beneath it, which continue to contribute to the
-/// nearest enabled rigid body above them.
+/// a PhysicsRigidBodyAPI, and takes effect only while that collider or rigid
+/// body is enabled. A prim whose physics:collisionEnabled or
+/// physics:rigidBodyEnabled is false is not simulated, and its MassAPI is
+/// ignored.
+///
+/// This applies only to the prim the MassAPI is applied to. Enabled colliders
+/// beneath it are unaffected and still contribute their mass to the nearest
+/// enabled rigid body above them.
 ///
 class UsdPhysicsMassAPI : public UsdAPISchemaBase
 {
--- a/pxr/usd/usdPhysics/massAPI.h
+++ b/pxr/usd/usdPhysics/massAPI.h
@@ -43,7 +43,10 @@ class SdfAssetPath;
 /// physics:rigidBodyEnabled is false, takes no part in simulation, so any
 /// MassAPI applied to it is likewise ignored. When neither an enabled
 /// collider nor an enabled rigid body remains on a prim, it is not a physics
-/// prim and its MassAPI has no effect.
+/// prim and its MassAPI has no effect. Note that this concerns only the prim
+/// the MassAPI is applied to: disabling a prim does not discard the mass of
+/// the enabled colliders beneath it, which continue to contribute to the
+/// nearest enabled rigid body above them.
 ///
 class UsdPhysicsMassAPI : public UsdAPISchemaBase
 {
--- a/pxr/usd/usdPhysics/massAPI.h
+++ b/pxr/usd/usdPhysics/massAPI.h
@@ -37,7 +37,13 @@ class SdfAssetPath;
 ///
 /// Defines explicit mass properties (mass, density, inertia etc.).        
 /// MassAPI can be applied to any object that has a PhysicsCollisionAPI or
-/// a PhysicsRigidBodyAPI.
+/// a PhysicsRigidBodyAPI. MassAPI only has an effect where the underlying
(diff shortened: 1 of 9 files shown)


jesschimein:
Filed as internal issue #USD-12474

andrewkaufman:
Restacked onto https://github.com/PixarAnimationStudios/OpenUSD/pull/4187, which supersedes #4164. Only the `GetCollisionPrims` commit is mine on top; the rest are #4187 unchanged.

@AlesBorovicka on your question about how far to take this: you are right that it invites symmetric accessors, a get rigid body on a collider and get body 0/1 on a joint. #4187 is relevant to that now, because it makes all of those the same traversal. It resolves which body owns a prim consistently across mass computation, collider attribution, joint relationship resolution and validation, so symmetric accessors would be thin wrappers over one shared walk rather than four separate ones. We would like to discuss the scope before adding them, so this PR stays limited to the collider getter.

tallytalwar:
@andrewkaufman could you rebase this on top of #4187 to remove commits which are already in 4187? Thanks

Or if its just [7faa4fb](https://github.com/PixarAnimationStudios/OpenUSD/pull/4165/commits/7faa4fbffb96fefcca2075d717511d86f4780b93) which needs to be included, I can extract that out internaly, lmk. Thanks

andrewkaufman:
@tallytalwar I've rebased on #4187 to incorporate the code review changes there. The commits still show here though... I can't seem to stack the PRs for some reason, perhaps because the source branch lives on my fork, not quite sure.... Its just the last commit that is unique here.

I have one addition planned before this is ready: alongside `GetCollisionPrims` I'll add `GetRigidBody()` on colliders and `GetBody0()`/`GetBody1()` on joints. These expose the same non-obvious resolution the parser handles (the rel can point at a collider and the parser walks the ancestor hierarchy to the nearest enabled body). I've seen multiple client side mistakes caused by this non-obvious traversal behaviour, so I think its worth exposing the helpers utils.

andrewkaufman:
@AlesBorovicka @tallytalwar, pushed the new public getters and refactored the joint parser onto them.

**New API:**
- `UsdPhysicsCollisionAPI::GetBody()` resolves a collider to its owning enabled rigid body via the ancestor walk, returning an invalid prim when there is none.
- `UsdPhysicsJoint::GetBody0()` / `GetBody1()` return the enabled rigid body each side of the joint attaches to, or an invalid prim.
- `UsdPhysicsJoint::GetLocalPose0()` / `GetLocalPose1()` return the joint-local anchor pose in that body's frame, with the frame scale baked into the translation (physics is scale-free) and an optional `UsdGeomXformCache` for nested hierarchies.

The joint parser's private `_GetBodyPrim` / `_GetLocalPose` are deleted. `_FinalizeJoint` now calls the public getters, so joint body resolution has a single implementation.

One behavior change worth a check: a joint side with no enclosing enabled body now gives an empty `body0`/`body1`, where the previous parser fell back to the nearest collider. A collider is not a body, so anchoring a joint to one looks wrong to me, but I would rather confirm than assume there was a reason for it.

@AlesBorovicka, does this match your expectation for the body-resolution semantics, and do you agree the joint parser is more correct now?

PS. `CollisionAPI::GetBody()` and the parser's internal collider resolution now use the same enabled-body walk, but I deliberately left the parser's path in place rather than routing it through `GetBody()`. The parser is scope aware (only bodies within the parsed range) while `GetBody()` is a plain stage-ancestry walk, so they can differ at a parse-scope boundary. `test_collider_get_body_matches_parser` asserts they agree in the simple and disabled-body cases.

AlesBorovicka:
> One behavior change worth a check: a joint side with no enclosing enabled body now gives an empty `body0`/`body1`, where the previous parser fell back to the nearest collider. A collider is not a body, so anchoring a joint to one looks wrong to me, but I would rather confirm than assume there was a reason for it.
> 
> @AlesBorovicka, does this match your expectation for the body-resolution semantics, and do you agree the joint parser is more correct now?
> 

Well a standalone collider is a static body, it boils down to the transformation used in the end of the day that you compute from the joint local pose. So its going to be different based on what you return. This will create different joints in PhysX for sure. So changes like this will have some consequences.

PR #4177, opened 2026-08-04
Build and publish aarch64 wheels for Linux

johnnynunez:
## Description of Change(s)

`usd-core` publishes wheels for manylinux x86_64, macOS and win_amd64 only, so `pip install usd-core` fails outright on Linux aarch64:

```
Because usd-core>=24.8 has no wheels with a matching platform tag
(e.g., manylinux_2_39_aarch64) and you require usd-core, we can
conclude that your requirements are unsatisfiable.
```

This affects every aarch64 host (Grace Hopper / GB10 workstations, Jetson, Ampere servers, Apple Silicon under Linux) and pushes downstream projects into either vendoring their own USD build or dropping the dependency. Filed as #3741 (internal `USD-11252`).

The PyPI workflow already does everything required to produce these wheels; it was pinned to one architecture in exactly two places.

**`build_scripts/pypi/docker/Dockerfile`** hardcoded the x86_64 manylinux image. It now takes an `ARCH` build arg **defaulting to `x86_64`**, so any existing caller running `docker build` with no extra arguments is unaffected.

**The `Linux` job** hardcoded `runs-on: ubuntu-22.04`. It now expands over an `ARCH` axis using the GitHub-hosted `ubuntu-22.04-arm` runners, which are free for public repositories.

Supporting changes:

- Artifact names gain the architecture (`dist-linux-<arch>-<pytag>`) so the two architectures no longer collide. `CollectPackages`' existing `dist-*-*` pattern matches these as well as the unchanged mac/windows names.
- The `Test` job gains the six matching aarch64 configurations, so the new wheels are installed and exercised rather than merely built.
- Linux `timeout-minutes` 60 → 120: the arm runners are slower than their x86_64 counterparts for a monolithic USD build.

### Note on CI cost

This doubles the Linux build jobs from 6 to 12. Given the comment at the top of this workflow about the limited number of slots, that is your call rather than mine. If the budget is the blocker, I am happy to reduce this to a subset (cp312 alone would cover most of the reported demand), or to gate the aarch64 axis behind the `workflow_dispatch` path so it only runs for releases. Say which you prefer and I will amend.

I could not exercise the workflow itself, since that requires this repository's CI. What I could verify, I did; see below.

## Verification

Run on aarch64 (Ubuntu 24.04, glibc 2.39, Python 3.12, 20 cores) against `dev` at `47154dc`, using the same `build_usd.py` invocation and packaging steps this workflow performs:

```
build_usd.py --build-monolithic --no-materialx --no-imaging \
    --no-examples --no-tutorials \
    --build-args USD,"-DPXR_PY_UNDEFINED_DYNAMIC_LOOKUP=ON ..."
setup.py bdist_wheel --python-tag cp312
auditwheel repair
```

Result: `usd_core-26.11-cp312-none-manylinux_2_38_aarch64.whl` (27.6 MB), built in ~4.5 min.

Installed into a clean venv:

- `build_scripts/pypi/test.py`: **3 passed**
- `from pxr import Usd, UsdGeom, UsdPhysics, Sdf`: OK, `Usd.GetVersion() == (0, 26, 11)`
- authored a stage with `RigidBodyAPI` + `CollisionAPI`, saved, reopened, both APIs survived
- opened a real 160-prim robot asset with a variant set, composed correctly

And the case that motivated this: `pip install simready-validate`, which declares `usd-core>=22.11` and is **currently uninstallable on this platform**, resolves and imports successfully against the built wheel, with no `--no-deps` workaround.

The static properties of the workflow change were checked by parsing it: the matrix expands to 12 jobs with unique artifact names, all of which match the existing `dist-*-*` collection pattern; the pre-existing `dist-mac-*` / `dist-windows-*` names still match; every runner label used is a valid GitHub-hosted label; and the `Dockerfile` keeps `ARG ARCH` declared before `FROM` with an `x86_64` default.

## Checklist

- [x] I have referenced the GitHub issue this PR addresses (#3741)
- [ ] I have added unit tests: n/a, this is a CI/packaging change; the existing `build_scripts/pypi/test.py` now runs on aarch64 too


jesschimein:
Filed as internal issue #USD-12512

o2b3d:
@johnnynunez, it looks like this PR is a fix for #3741, is that correct?

johnnynunez:
> @johnnynunez, it looks like this PR is a fix for #3741, is that correct?

Yes

PR #4187, opened 2026-08-18
usdPhysics: treat disabled rigid bodies as unapplied in all ancestor traversals

andrewkaufman:
### Description of Change(s)

Per the schema docs, `physics:rigidBodyEnabled = false` means the body takes no part in simulation, which is equivalent to not applying `UsdPhysicsRigidBodyAPI` at all. Every ancestor traversal that resolves which body owns a prim should therefore continue past a disabled body to the nearest enabled one.

This supersedes #4164 and fixes two more traversals that were missed there.

- `ComputeMassProperties` counted colliders with `physics:collisionEnabled = false` toward mass, center of mass and inertia, even though it skipped them as shapes.
- A disabled nested body pruned its subtree, so the colliders below it contributed mass to no body at all.
- `_HasDynamicBodyParent` stopped at a disabled body, so colliders below it were treated as static instead of belonging to the enabled body above.
- `_GetBodyPrim`, which resolves a joint's body relationship, returned the disabled body and never looked at `physics:rigidBodyEnabled` at all.
- `PhysicsJointChecker` only looked for `RigidBodyAPI` on the prim a body relationship targets. A relationship may point at any `UsdGeomXformable`, and joint parsing resolves it to the closest ancestor body, so a joint anchored to a collider under a body was flagged as bodyless while parsing resolved it to a real enabled body.

Before this change, a collider under a disabled inner body under an enabled outer body got three different answers: collider attribution said the outer body, the joint relationship resolved to the disabled inner body, and validation reported an error. All three now resolve to the outer body.

Enabled bodies and colliders relying on the `true` fallback are unaffected. `JointNoEnabledRigidBody` is evaluated per joint rather than per relationship, and is unchanged here: one of the two relationships reaching an enabled body is sufficient, either directly or through an ancestor, so an empty relationship or a non-body target opposite a body does not error.

### Fixes Issue(s)

Supersedes and includes #4164.

### AI Assistance

Developed with AI assistance (Anthropic Claude, primarily Opus 5): the
code, tests, and this description were AI-drafted and then reviewed by
me before submission.

### Checklist

### DIFF
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -319,15 +319,14 @@ a prim higher in the tree, overrides any density specified via a material
 times the locally effective density.
 
 \li Implicit mass of a rigid body is the total implicit mass of all children in
-the subtree belonging to the rigid body. Only enabled colliders are counted
-among these children: a collider whose physics:collisionEnabled is false takes
-no part in simulation and is excluded from the mass computation, so any
-UsdPhysicsMassAPI applied to it has no effect. Likewise, a UsdPhysicsMassAPI
-applied to a nested rigid body whose physics:rigidBodyEnabled is false has no
-effect. Note however that a disabled nested rigid body does not itself form the
-root of a subtree: because it takes no part in simulation it owns no colliders,
-so the enabled colliders below it still belong to the nearest enabled rigid body
-above it and still contribute their mass to it.
+the subtree belonging to the rigid body. Only enabled colliders count toward
+that total. A collider whose physics:collisionEnabled is false is not simulated,
+and any UsdPhysicsMassAPI applied to it is ignored, as is a UsdPhysicsMassAPI
+applied to a disabled nested rigid body.
+
+\li A disabled nested rigid body does not, however, exclude the colliders
+beneath it. It owns no colliders while it is disabled, so those colliders belong
+to the nearest enabled rigid body above it and contribute that body's mass.
 
 \li Density is assumed to be 1000.0 kg/m3 (approximately the density of water) 
 for volume computation when no other density is specified locally, or in bound 
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -313,7 +313,11 @@ a prim higher in the tree, overrides any density specified via a material
 times the locally effective density.
 
 \li Implicit mass of a rigid body is the total implicit mass of all children in
-the subtree belonging to the rigid body.
+the subtree belonging to the rigid body. Only enabled colliders are counted
+among these children: a collider whose physics:collisionEnabled is false, or a
+child rigid body whose physics:rigidBodyEnabled is false, takes no part in
+simulation and is excluded from the mass computation, so any UsdPhysicsMassAPI
+applied to it has no effect.
 
 \li Density is assumed to be 1000.0 kg/m3 (approximately the density of water) 
 for volume computation when no other density is specified locally, or in bound 
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -226,7 +226,8 @@ of a rigid body that are not parts of other rigid bodies. Multiple rigid bodies
 in a subtree move completely independently, even if their transforms are stored
 relative to one another.  
 
-This exception applies only while the nested UsdPhysicsRigidBodyAPI is enabled.
+This exception applies only while the prim to which UsdPhysicsRigidBodyAPI is
+applied has physics:rigidBodyEnabled enabled.
 A nested rigid body whose physics:rigidBodyEnabled is false takes no part in
 simulation and therefore does not create a separate rigid body, so it does not
 form the root of its own subtree; the prims below it remain part of the nearest
@@ -322,11 +323,12 @@ times the locally effective density.
 the subtree belonging to the rigid body. Only enabled colliders count toward
 that total. A collider whose physics:collisionEnabled is false is not simulated,
 and any UsdPhysicsMassAPI applied to it is ignored, as is a UsdPhysicsMassAPI
-applied to a disabled nested rigid body.
+applied to a rigid body whose physics:rigidBodyEnabled is false.
 
-\li A disabled nested rigid body does not, however, exclude the colliders
-beneath it. It owns no colliders while it is disabled, so those colliders belong
-to the nearest enabled rigid body above it and contribute that body's mass.
+\li A rigid body whose physics:rigidBodyEnabled is false does not, however,
+exclude the colliders beneath it. It owns no colliders while it is disabled, so
+those colliders belong to the nearest enabled rigid body above it and contribute
+that body's mass.
 
 \li Density is assumed to be 1000.0 kg/m3 (approximately the density of water) 
 for volume computation when no other density is specified locally, or in bound 
--- a/pxr/usd/usdPhysics/overview.dox
+++ b/pxr/usd/usdPhysics/overview.dox
@@ -226,6 +226,12 @@ of a rigid body that are not parts of other rigid bodies. Multiple rigid bodies
 in a subtree move completely independently, even if their transforms are stored
 relative to one another.  
 
+This exception applies only while the nested UsdPhysicsRigidBodyAPI is enabled.
+A nested rigid body whose physics:rigidBodyEnabled is false takes no part in
+simulation and therefore does not create a separate rigid body, so it does not
+form the root of its own subtree; the prims below it remain part of the nearest
+enabled rigid body above it in the hierarchy.
+
 If aggregate properties of the entire rigid body must be computed, such as
 total mass or the entirety of its collision volume, then the contents of the
 entire subtree are considered, taking care to exclude the contents of any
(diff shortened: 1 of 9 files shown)


pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1645,12 +1655,27 @@ bool _HasDynamicBodyParent(const UsdPrim& usdPrim, const RigidBodyMap& bodyMap,
...
         if (physicsAPIFound)
         {
-            *outBodyPrimPath = parent;
-            return false;
+            // A disabled rigid body takes no part in simulation, so it does
+            // not own this prim. Keep searching the ancestors: a nested
+            // disabled body may still have an enabled body above it, which
+            // this prim belongs to. Remember the nearest disabled body so it
+            // can still be reported if no enabled body is found at all.
+            if (disabledBodyPrim == UsdPrim())
```
nvmkuruc: @andrewkaufman Should this just be
```suggestion
            if (disabledBodyPrim)
```

pxr/usd/usdPhysics/parseUtils.cpp
```diff
@@ -1645,12 +1655,27 @@ bool _HasDynamicBodyParent(const UsdPrim& usdPrim, const RigidBodyMap& bodyMap,
...
+            }
         }
 
         parent = parent.GetParent();
     }
+
+    // No enabled body above this prim. Report the nearest disabled body, if
+    // any, so it is still recognized as belonging to a body that is present
+    // but not simulating rather than as a static collision.
+    if (disabledBodyPrim != UsdPrim())
```
nvmkuruc: Should this just be
```suggestion
    if (!disabledBodyPrim)
```

pxr/usdValidation/usdPhysicsValidators/validators.cpp
```diff
@@ -101,6 +101,7 @@ bool HasDynamicBodyParent(const UsdPrim& usdPrim, UsdPrim* outBodyPrimPath)
 {
     bool physicsAPIFound = false;
     UsdPrim parent = usdPrim;
+    UsdPrim disabledBodyPrim = UsdPrim();
```
nvmkuruc: Should this just be 
```suggestion
    UsdPrim disabledBodyPrim;
```

pxr/usdValidation/usdPhysicsValidators/validators.cpp
```diff
@@ -111,12 +112,27 @@ bool HasDynamicBodyParent(const UsdPrim& usdPrim, UsdPrim* outBodyPrimPath)
...
         if (physicsAPIFound)
         {
-            *outBodyPrimPath = parent;
-            return false;
+            // A disabled rigid body takes no part in simulation, so it does
+            // not own this prim. Keep searching the ancestors: a nested
+            // disabled body may still have an enabled body above it, which
+            // this prim belongs to. Remember the nearest disabled body so it
+            // can still be reported if no enabled body is found at all.
+            if (disabledBodyPrim == UsdPrim())
```
nvmkuruc: Should this be 
```suggestion
            if (disabledBodyPrim)
```

pxr/usdValidation/usdPhysicsValidators/validators.cpp
```diff
@@ -111,12 +112,27 @@ bool HasDynamicBodyParent(const UsdPrim& usdPrim, UsdPrim* outBodyPrimPath)
...
+            }
         }
 
         parent = parent.GetParent();
     }
+
+    // No enabled body above this prim. Report the nearest disabled body, if
+    // any, so it is still recognized as belonging to a body that is present
+    // but not simulating rather than as a static collision.
+    if (disabledBodyPrim != UsdPrim())
```
nvmkuruc: Should this be 
```suggestion
    if (!disabledBodyPrim)
```

andrewkaufman (commented):
Thanks @nvmkuruc, the `UsdPrim disabledBodyPrim;` default construction was right, and I've applied it to `parseUtils.cpp` as well, which the suggestion only covered in `validators.cpp`.

The other four inverted the polarity at both sites, so I've corrected them in [b33797415fc5aa5b791982f86eaa15b1a9ffa9ea](https://github.com/PixarAnimationStudios/OpenUSD/pull/4187/commits/b33797415fc5aa5b791982f86eaa15b1a9ffa9ea) rather than reverting. The recording site needs `!disabledBodyPrim`, so the nearest disabled body is remembered the first time one is seen; the reporting site needs `disabledBodyPrim`, so the out parameter is only written when one was actually found. As applied in [74907c7431d666a6eb5289b013bf379081a3c48b](https://github.com/PixarAnimationStudios/OpenUSD/pull/4187/commits/74907c7431d666a6eb5289b013bf379081a3c48b), `disabledBodyPrim` was never assigned and the reporting site wrote a null prim to `*outBodyPrimPath`.

No existing test covered a disabled body with no enabled body above it, so the entire physics and validation suite passed with the polarity inverted. That case is the one that regressed (the collider degraded to a static collision instead of being reported against the disabled body) so I've added it to `testUsdPhysicsParsing`, and it fails without this change.

pxr/usd/usdPhysics/overview.dox
```diff
@@ -226,6 +226,12 @@ of a rigid body that are not parts of other rigid bodies. Multiple rigid bodies
 in a subtree move completely independently, even if their transforms are stored
 relative to one another.  
 
+This exception applies only while the nested UsdPhysicsRigidBodyAPI is enabled.
```
tallytalwar: Minor, but can we rephrase this as, "This exception applies only while the prim to which UsdPhysicsRigidBodyAPI is applied has `physics:rigidBodyEnabled` enabled."

pxr/usd/usdPhysics/overview.dox
```diff
@@ -313,7 +319,14 @@ a prim higher in the tree, overrides any density specified via a material
 times the locally effective density.
 
 \li Implicit mass of a rigid body is the total implicit mass of all children in
-the subtree belonging to the rigid body.
+the subtree belonging to the rigid body. Only enabled colliders count toward
+that total. A collider whose physics:collisionEnabled is false is not simulated,
+and any UsdPhysicsMassAPI applied to it is ignored, as is a UsdPhysicsMassAPI
+applied to a disabled nested rigid body.
```
tallytalwar: maybe not have nested word here. It should be true for any rigid body. Also maybe mention "physic:rigidBodyEnabled" again here to be explicit.

pxr/usdValidation/usdPhysicsValidators/validators.cpp
tallytalwar: Not part of this change, but I think this should be true to begin with as thats what the schema's default is that rigidbody is enabled.

pxr/usdValidation/usdPhysicsValidators/testenv/testUsdPhysicsValidation.py
```diff
@@ -229,6 +229,151 @@ def test_physics_joint_requires_enabled_rigid_body(self):
...
+        errors = validator.Validate(physicsJoint.GetPrim())
+        self.assertTrue(len(errors) == 0)
+
+        # Disabling the only enabled ancestor body fails the check
+        rbo0 = UsdPhysics.RigidBodyAPI(body0.GetPrim())
+        rbo0.GetRigidBodyEnabledAttr().Set(False)
+        rbo1.GetRigidBodyEnabledAttr().Set(False)
+
+        errors = validator.Validate(physicsJoint.GetPrim())
+        self.assertTrue(len(errors) == 1)
```
tallytalwar: Also nitty but I think `self.assertEqual(len(errors, 1))` will be better here and other place as if it fails the diagnostics will tell what the len(errors) is!

tallytalwar (approved):
lgtm, some minor notes. @andrewkaufman could you please address these and I can get this in.

pxr/usd/usdPhysics/overview.dox
andrewkaufman: Done in 480690a7d, using your wording.

pxr/usd/usdPhysics/overview.dox
andrewkaufman: Done in 480690a7d — dropped "nested" and reworded both this line and the following one to apply to any rigid body, naming physics:rigidBodyEnabled explicitly.

pxr/usdValidation/usdPhysicsValidators/validators.cpp
andrewkaufman: Good catch, defaulted it to true in 480690a7d to match the schema fallback.

pxr/usdValidation/usdPhysicsValidators/testenv/testUsdPhysicsValidation.py
andrewkaufman: Switched all len(errors) checks in this file to assertEqual(len(errors), N) in 480690a7d.

andrewkaufman:
Pushed a follow-up commit (`aa29b9a01`) correcting the disabled-body collider case.

The earlier commits, when a collider sat below a disabled rigid body with no enabled body above it, reported the collider as belonging to that disabled body. That contradicts the schema: a rigid body with `physics:rigidBodyEnabled = false` takes no part in simulation and owns no colliders (see the MassAPI and `GetCollisionPrims` documentation). Such a collider is a static collision, and its body path should be empty.

`_HasDynamicBodyParent` no longer records a disabled body as a fallback owner; it walks past disabled bodies to the nearest enabled one and returns no body if there is none. `test_rigidbody_disabled_body_only_collision_parse` is updated to assert the collider resolves to an empty body path.

The nested case is unchanged: a collider below a disabled body that has an enabled body above it still belongs to the enabled body.

PR #4188, opened 2026-08-19
Apple: Add AuthorshipAPI for recording how a prim was created

dgovil:
### Description of Change(s)

This PR adds support the `UsdMediaAuthorshipAPI`, a multiple-apply schema that is meant to meet the needs of regulatory requirements around the world regarding Generative AI use, but also to be more generally useful for human authors as well.

The goal of this implementation is ease of adoption for regulatory compliance , and to meet the granular needs of complex OpenUSD authored scenes and workflows in a dynamic nature.

This is not a substitution for C2PA or other provenance standards, but is an implementation that is designed to work with OpenUSD's composition setup. In the future, other developers may opt to make C2PA implementations in USD, and we have talked to the C2PA leads to align this implementation such that it could gracefully and progressively flow into their future implementations. 

The schema is multiple apply so that a single prim can be considered as having been authored by multiple algorithms or humans.

In addition to the schema, we also provide:

* `UsdMediaAuthorshipAPI::GetAll()` which finds instances of the schema on a given prim
* `UsdMediaAuthorshipAPI::GetAllOnStage()` which finds instances of the schema across the stage, and optionally across all prim specs.
* `usdauthors`, a command listing records grouped by the prims they apply to
- `extras/usd/examples/usdApplyAuthorship`, applying records from the command line, including converting name=value inputs into the index-matched inputNames and inputValues arrays
- tests for the schema and the command, and toolset documentation

### Link to proposal 

A thorough proposal is available at the following link with further information and thinking.
[Authorship API Proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/106)

### Checklist


pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
 }
+
+class "AuthorshipAPI" (
+    inherits = </APISchemaBase>
+    doc = """Records who or what authored a prim.
+
+    Each record describes a single authoring step, like a generative AI run,
+    an export from a DCC, or a manual cleanup session. Because multiple
+    records can be applied, a prim can carry its full history of contributors
+    without them clobbering each other.
```
spiffmon, line 223: "...can carry its full history of contributors, provided that each step is recorded with a unique-to-the-prim schema instance name."

Do we want to suggest a numbering scheme and/or provide a method that, given a short `softwarePackage` string, will search applied schemas on the prim and return an instance name that does not collide with others of the same softwarePackage (e.g. we might go back and forth from Maya to Solaris and back to Maya).  This might be overthinking/premature, and the admonition above is sufficient.

pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
+
+    This schema provides a way to record authorship but does not verify or
+    sign it. It relies on the honor system to help with regulatory
+    compliance and artist credit. copyrightOwner records an ownership claim;
```
spiffmon, line 227: backticks on `copyrightOwner` ?

pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
+    it does not establish one.
+
```
spiffmon, line 229: In this overview, we don't need to get into _all_ the semantics, but we should at least disclose that the schema may be applied to any number of a prim's ancestors, and that authorship does *not* override, but rather aggregates through namespace?

pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
+        token apiSchemaType = "multipleApply"
+        token propertyNamespacePrefix = "authorship"
+    }
+)
+{
+    uniform string softwarePackage (
+        doc = """The tool or system that most directly wrote the USD data for
+        this prim. This is always a tool, like 'org.blender', not a person.
+        People go in 'creator'. If an AI model runs inside another tool,
+        this should name the model.
```
spiffmon, line 242: How would we recommend people refer to OpenUSD?  Just `OpenUSD` or `github.PixarAnimationStudios.OpenUSD` ?  I ask because I wonder if the following is a useful addition:

If a user created the USD data using the OpenUSD API's in an unnamed python script, the `softwarePackage` would be `OpenUSD`

pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
+        to avoid name conflicts. No network lookup or validation is performed;
+        it is just a unique identifier. Compare values case-insensitively."""
+    )
+
+    uniform string softwareVersion (
+        doc = """The version of the softwarePackage that created this record.
+        It versions the tool itself, not the asset. For hosted services
+        without a stable version, use whatever label they provide, like a
+        build hash. Rely on the 'created' timestamp to pin down which
+        iteration was used."""
```
spiffmon, line 254: ... pin down the iteration's ordering.

pxr/usd/usdMedia/schema.usda
```diff
@@ -212,3 +212,134 @@ class "AssetPreviewsAPI" (
...
+        doc = """A list of human-friendly names for the artists, studios, or 
+        services that contributed to this authoring step. This is intended 
+        for display and corresponds to 'dc:creator'."""
+    )
+
+    uniform string description (
+        doc = """A free-form description of how the prim was created. 
+        For AI content, use 'prompt:inputNames' and 'prompt:inputValues' 
+        for the specific generation inputs instead. This field is best for 
+        technique notes, reference material, or historical context."""
```
spiffmon, line 278: Do we have a recommendation for reference art, schematics/design-specs, or other assets that contributed to an artist-produced prim/asset?  Was going to make the same recommendation for those, but `prompt:` namespace does not feel appropriate...

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    //  - Close the include guard with #endif
+    // ===================================================================== //
+    // --(BEGIN CUSTOM CODE)--
+
+    /// Returns every authorship record on \p stage, sorted by prim path then
+    /// instance name. Prims beneath native instances are reached through the
+    /// stage's prototypes, so a record inside a prototype is reported against
+    /// the prototype rather than once per instance. Inactive and abstract prims
+    /// are included. Note that descendants of an inactive prim are not
+    /// composed onto the stage at all, so their records cannot be reported.
```
spiffmon, line 568: Also should note that only the current state of the stage (with respect to load-state and variant selections) is processed.

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// method's. This search covers cases the composed prim cannot report:
+    ///
+    /// - A stronger layer authoring an \em explicit \em apiSchemas list can
+    ///   shadow an application coming from a weaker layer such as a reference.
+    ///   The record's properties still compose onto the prim and remain
+    ///   readable, but the schema no longer counts as applied.
+    /// - Authorship properties authored without the schema ever having been
+    ///   applied, which some tools emit.
+    /// - The same instance name authored in several layers, where composition
+    ///   keeps only the strongest opinion for each field.
```
spiffmon, line 587: I know @nvmkuruc raised concens about some of these cases.  But I'm struggling with the position we're putting the consumer in by providing this "bare metal" option without any guidance on when it is appropriate to use.  Putting it out there with an admonition that "some DCC's may just produce bad data" makes me not trust the "expected, reasonable, easy-to-process algorithm".

I have two sugs:
1. This examination and flags (as warnings) would be *fantastic* for validation to provide, and validation would be a much appreciated addition to the PR
2. To facilitate validation, and manual debugging, we might indeed want to provide this logic, but in a *separate, debugging utility method*

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    ///
+    /// Nothing is merged or resolved. Records from different prims stay
+    /// distinct even when they share an instance name, since they describe
+    /// different authoring steps.
+    ///
+    /// \p searchPrimStack behaves as it does for GetAllOnStage().
+    USDMEDIA_API
+    static std::vector<UsdMediaAuthorshipAPI>
+    ComputeAccumulatedRecords(const UsdPrim &prim,
+                              bool searchPrimStack = false);
```
spiffmon, line 618: Ah, so it's actually **this method's** `searchPrimStack` logic that wants to be a separate utility method.  No need to package up `GetAllOnStage`s, I think?

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// Unselected variants are not part of a prim's spec stack, so records
+    /// inside them are not found either way. This search visits every layer
+    /// contributing to every prim and is considerably slower.
+    USDMEDIA_API
+    static std::vector<UsdMediaAuthorshipAPI>
+    GetAllOnStage(const UsdStagePtr &stage, bool searchPrimStack = false);
+
+    /// Returns the records that apply to \p prim, meaning its own plus those it
+    /// accumulates from its ancestors, sorted by prim path so ancestors come
+    /// first. Compare a record's prim against \p prim to tell an inherited
```
spiffmon, line 601: ".. to distinguish an inherited..."

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// record from one the prim declared itself.
+    ///
+    /// This proposal's convention is that a prim's children accumulate the
```
spiffmon, line 604: No longer a proposal - this is what is

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    ///
+    /// \p prim itself is included, matching UsdPrimRange. Prims beneath a
+    /// native instance are not visited, since their records live on the
+    /// prototype; use GetAllOnStage() to reach those, or query the prototype
+    /// directly.
+    ///
+    /// \p searchPrimStack behaves as it does for GetAllOnStage().
+    USDMEDIA_API
+    static std::vector<UsdMediaAuthorshipAPI>
+    GetAllUnder(const UsdPrim &prim, bool searchPrimStack = false);
```
spiffmon, line 631: Don't think this method needs a bare metal version either?

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    ///   readable, but the schema no longer counts as applied.
+    /// - Authorship properties authored without the schema ever having been
+    ///   applied, which some tools emit.
+    /// - The same instance name authored in several layers, where composition
+    ///   keeps only the strongest opinion for each field.
+    ///
+    /// A returned schema object always reads composed property values, so
+    /// repeated entries read alike. They record that several specs contribute,
+    /// not what each one said. UsdPrim::HasAPI() may report \c false for them.
+    /// Unselected variants are not part of a prim's spec stack, so records
```
spiffmon, line 592: Maybe use "prim stack" with or without reference to UsdPrim::GetPrimStack() ?

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+
+    /// Returns the path of every authorship record authored in \p layer, sorted,
+    /// in the <tt><primPath>.authorship:instanceName</tt> form that
+    /// Get(const UsdStagePtr&, const SdfPath&) accepts and
+    /// IsAuthorshipAPIPath() parses.
+    ///
+    /// This inspects one layer's scene description directly and composes
+    /// nothing, so records in its sublayers and reference targets are not
+    /// included. Use it to ask what a particular layer contributes, for example
+    /// to check a DCC's export. Because it does not compose, it reaches specs
```
spiffmon, line 641: "For example" is presuming a DCC exports a single layer, and modifies no others.

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// that no stage would report, including those inside variants that are not
+    /// selected.
+    ///
+    /// Paths are returned rather than schema objects because a record found
+    /// this way may have no composed prim to attach one to. Values can be read
+    /// from the layer's specs, or the paths can be passed to
+    /// Get(const UsdStagePtr&, const SdfPath&) once a stage is at hand.
```
spiffmon, line 648: Suggesting someone can access the records on a composed stage is problematic.  They may be beneath a deactivated prim, or behind a complex set of variant selections.  And the layer may contain references to other layers with records, so there may be **more** than just the records reported here.

At a minimum I'd take that sentence out.  More broadly, is this a utility we want to promote?  If so, I'd recommend we back off on the claims that it should be used to validate a DCC's export, and state there may not be a correspondence between the records this method reports, and the records we discover on a composed stage using this layer.

pxr/usd/usdMedia/authorshipAPI.cpp
```diff
@@ -0,0 +1,801 @@
...
+    if (tokens.size() < 3 || tokens.front() != UsdMediaTokens->authorship) {
+        return TfToken();
+    }
+
+    // A property base name can itself be namespaced (e.g. prompt:inputNames),
+    // so try the longest possible base name first: whichever trailing run of
+    // tokens matches a known property, the tokens between the namespace
+    // prefix and that run are the instance name, which may itself be
+    // namespaced too.
+    for (size_t baseNameLen = tokens.size() - 2; ; --baseNameLen) {
```
spiffmon, line 590: It is fully legal to have instance names that contain colons (see [this thread](https://forum.aousd.org/t/guidance-for-api-schema-instance-names/1498/5)).  If you're trying to intuit a property's instance name,  you're going to need to actually query the instance name. I don't think the introspection operation being performed here is actually possible to do robustly, without knowing the schema(s) applied to the prim.

pxr/usd/usdMedia/authorshipAPI.cpp
```diff
@@ -0,0 +1,801 @@
...
+        }
+    }
+
+    // Properties authored without the schema ever having been applied.
+    for (const SdfPropertySpecHandle &propertySpec : spec->GetProperties()) {
+        if (!propertySpec) {
+            continue;
+        }
+        const TfToken instanceName =
+            _GetInstanceNameFromPropertyName(propertySpec->GetName());
```
spiffmon, line 647: reiterating, I don't think this can be done robustly unless we restrict possible candidates to those with exactly three tokens, which is unsatisfying.

spiffmon (commented):
Hey @dgovil ,
   Once again breaking up review into two parts.  Had planned to do the tooling parts in a second review, but then I also paused review of authorshipAPI.cpp, as I think notes from the first review may refactor it enough to wait.

dgovil:
Thanks for the notes @spiffmon , I'll start tackling those

pxr/usd/usdMedia/schema.usda
dgovil: I added your caveat. I agree that I think a helper is premature, and I wonder if it makes sense to make a generic API schema helper? 
I can imagine there's some value in a "nextUniqueName" helper for multiple apply schemes in general?

pxr/usd/usdMedia/schema.usda
dgovil: Done

pxr/usd/usdMedia/schema.usda
dgovil: I added a short paragraph to address this

pxr/usd/usdMedia/schema.usda
dgovil: Ah that's a great point, I would go with your suggestion of just OpenUSD for that. I don't think we need to be over prescriptive on it, but maybe there's benefits to storing the actual fork of USD used to author?

pxr/usd/usdMedia/schema.usda
dgovil: Added

pxr/usd/usdMedia/schema.usda
dgovil: I feel like maybe I over indexed on prompt in the proposal here. I think just ditching the `prompt` prefix would do the trick.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Ah good catch. Added.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: I added a new validator. Naming might be a bit verbose (I took inspiration from what some other linters call similar things).
I also split out the `searchPrimStack` methods into their own methods.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Yeah, so the latest iteration I split it out which I think makes it a little less ambiguous.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Fixed

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Ah yeah good catch.  Copy paste bugs :-D

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Also split out

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Fair, also fixed when I split things out

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Fixed

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Fair. I took the sentence out and added the caveat as you suggested.

pxr/usd/usdMedia/authorshipAPI.cpp
dgovil: Ah yeah totally fair. Okay I simplified this down.

pxr/usd/usdMedia/authorshipAPI.cpp
dgovil: Resolved with the above one

pxr/usd/usdMedia/schema.usda
spiffmon: Let's just leave it with the caveat, for now.

docs/toolset.help
```diff
@@ -404,6 +404,30 @@ Options:
...
+layer authors.
+
+Usage: usdauthors [OPTIONS] inputPath
+
+Positionals:
+  inputPath TEXT REQUIRED     The input file to process
+
+Options:
+  -h,--help                   Print this help message and exit
+  -d,--deep                   Gathers records from every prim spec instead of the composed prim. Finds records shadowed by an explicit apiSchemas list in a stronger layer, and records clobbered by the same instance name in another layer. Slower.
```
spiffmon: Should we call this `--debug` ? "deep" is a _little_ ambiguous, because we have namespace depth as well as compositional depth, and "debug" helps cement that it's not the flavor you want to base your primary tooling on?

pxr/bin/usdauthors/usdauthors.cpp
```diff
@@ -0,0 +1,473 @@
...
+    bool summary = false;
+    bool layer = false;
+};
+
+void Configure(CLI::App *app, Args &args) {
+    app->add_option(
+        "inputPath", args.inputPath, "The input file to process")
+        ->required();
+    app->add_flag(
+        "-d,--deep", args.deep,
```
spiffmon, line 50: Should we call this -`--debug` ? "deep" is a little ambiguous, because we have namespace depth as well as compositional depth, and "debug" helps cement that it's not the flavor you want to base your primary tooling on?

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    //
+    // Just remember to: 
+    //  - Close the class declaration with }; 
+    //  - Close the namespace with PXR_NAMESPACE_CLOSE_SCOPE
+    //  - Close the include guard with #endif
+    // ===================================================================== //
+    // --(BEGIN CUSTOM CODE)--
+
+    /// Returns every authorship record on \p stage, sorted by prim path then
+    /// instance name. Records inside a native instance's prototype are
```
spiffmon, line 566: Records on descendant prims of a native instance's prototype are

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// composition dropped.
+    ///
+    /// Only apiSchemas metadata is scanned, not authored properties, so a
+    /// record whose schema was never applied anywhere is not found.
+    /// Instance names and base names can both be namespaced, so a property
+    /// name alone can't be split back into the two. Repeated entries read
+    /// the same composed values; HasAPI() may report false for them.
+    /// Unselected variants aren't in a prim's prim stack
+    /// (UsdPrim::GetPrimStack()) and aren't found either; use
+    /// GetAllInLayer() for those.
```
spiffmon, line 622: Rather than directing people at GetAllInLayer() as if it's something you need to call and stitch in the results, maybe move this to a "See also GetAllInLayer() for a per-layer view that includes unselected variants and inactive primSpecs" at the end.

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    ///
+    /// Considerably slower than GetAllOnStage(), ComputeAccumulatedRecords(),
+    /// and GetAllUnder(). Intended for debugging and validation, not
+    /// general use.
+    USDMEDIA_API
+    static std::vector<UsdMediaAuthorshipAPI>
+    GetAllInPrimStacks(const UsdStagePtr &stage);
+
+    /// Returns the path of every authorship record authored in \p layer, sorted,
+    /// in the <tt><primPath>.authorship:instanceName</tt> form that
```
spiffmon, line 632: For records in variants, it won't be a `primPath` - it'll more generally be a specPath that may include variant selections.  You discuss this in general below, but perhaps acknowledge it here, also?

pxr/usd/usdMedia/authorshipAPI.h
```diff
@@ -0,0 +1,656 @@
...
+    /// the same composed values; HasAPI() may report false for them.
+    /// Unselected variants aren't in a prim's prim stack
+    /// (UsdPrim::GetPrimStack()) and aren't found either; use
+    /// GetAllInLayer() for those.
+    ///
+    /// Considerably slower than GetAllOnStage(), ComputeAccumulatedRecords(),
+    /// and GetAllUnder(). Intended for debugging and validation, not
+    /// general use.
+    USDMEDIA_API
+    static std::vector<UsdMediaAuthorshipAPI>
```
spiffmon, line 628: Especially if we are returning schema objects, here, I'm thinking that the stated purpose of this diagnostic would be much better served by returning **only** duplicates.  So, called on a stage where there is no shadowing, you'd get an empty result.  `GetAllDuplicates()` . What do you think?

docs/toolset.rst
```diff
@@ -207,7 +252,7 @@ usdfixbrokenpixarschemas
 ************************
 
 :program:`usdfixbrokenpixarschemas` attempts to fix usd(a|c|z) layers for any
-updates introduced by newer Pixar schema revisions. Note that this does not
```
spiffmon, line 210: There are a ton of whitespace changes introduced here... if it's immediately obvious and fixable for you, feel free to do so, but also please feel free to remove `toolset.help` and `toolset.rst` from the PR, as we regenerate them as part of our release process, anyways.

pxr/usd/usdMedia/authorshipAPI.cpp
```diff
@@ -0,0 +1,769 @@
...
+    // A token could appear in more than one list-op field, so dedup with a
+    // set.
+    std::set<TfToken> found;
+    const SdfTokenListOp &listOp = value.UncheckedGet<SdfTokenListOp>();
+
+    // Scan every op, not just the applied result: shadowed opinions are
+    // exactly what this search is for.
+    for (const SdfListOpType listOpType :
+            {SdfListOpTypeExplicit, SdfListOpTypeAdded,
+             SdfListOpTypeOrdered, SdfListOpTypePrepended,
```
spiffmon, line 594: Can remove SdfListOpTypeOrdered... it does not indicate "presence" of the schema

pxr/usd/usdMedia/authorshipAPI.cpp
```diff
@@ -0,0 +1,769 @@
...
+        result->push_back(record);
+    }
+}
+
+static void
+_AppendPrimStackRecordsForPrim(const UsdPrim &prim,
+                               std::vector<UsdMediaAuthorshipAPI> *result)
+{
+    // Strongest to weakest, not deduplicated: a record in several layers
+    // is reported once per layer, in that order.
```
spiffmon, line 623: Per my comment in the header file... it may be that we are adding records "in layer order" but we don't tell you what layer/spec it actually is, which seems frustrating as a consumer of the information.  So either we additionally return that information, or just tell you which applications appeared in more than one spec.

pxr/usdValidation/usdMediaValidators/validators.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+// Identifies authorship records unreachable via composed views, including
+// applications shadowed by explicit apiSchemas list-ops or instance names
+// overridden across multiple layers. See
+// UsdMediaAuthorshipAPI::GetAllInPrimStacks().
+static UsdValidationErrorVector
+_ShadowedOrClobberedAuthorshipValidator(
+    const UsdStagePtr &usdStage,
```
spiffmon, line 30: I can't see any reason why this needs to be a stage validator, and formulating it as one incurs a big, single-threaded stage traversal.  Can we change it to a prim validator, which means we can be checking prims in parallel?

pxr/bin/usdauthors/usdauthors.cpp
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        const std::string value = GetDisplayValue(attr);
+        if (value.empty()) {
+            continue;
+        }
+        fields.emplace_back(baseName.GetString(), value);
+    }
+
+    if (record.GetInputNamesAttr().HasAuthoredValue() ||
```
spiffmon, line 180: Don't you want `&&` here instead of `||` ? Also reminds me that in the schema doc you state these two properties must always be authored together, so should we add a validator for that?

pxr/usd/usdMedia/authorshipAPI.cpp
```diff
@@ -0,0 +1,769 @@
...
+    SdfPathVector result;
+
+    if (!layer) {
+        TF_CODING_ERROR("Invalid layer");
+        return result;
+    }
+
+    layer->Traverse(SdfPath::AbsoluteRootPath(),
+                    [&layer, &result](const SdfPath &path) {
+        const SdfPrimSpecHandle spec = layer->GetPrimAtPath(path);
```
spiffmon, line 751: I think you want to call GetObjectAtPathHere(), or else you will not find data authored directly on VariantSpecs, since a VariantSpec is not a PrimSpec.  Probably want a test case for that?

spiffmon (commented):
Thanks for the changes, @dgovil ; have re-reviewed them and also reviewed the remaining parts.

pxr/bin/usdauthors/usdauthors.cpp
dgovil: Ah yeah, I like that. Renamed, and simplified the output.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Done

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Done, as a closing \sa on the stage-level methods."

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Ah fair, done and added an example too.

pxr/usd/usdMedia/authorshipAPI.h
dgovil: Yeah that makes sense. I did that.

docs/toolset.rst
dgovil: Ah yeah, sorry git whitespace stripping hook.
I removed both files as you suggested.

pxr/usd/usdMedia/authorshipAPI.cpp
dgovil: Removed, and added a test as well

pxr/usd/usdMedia/authorshipAPI.cpp
dgovil: I opted to go for the second. The new methods just tell you which apps are duplicated or shadowed.

pxr/usdValidation/usdMediaValidators/validators.cpp
dgovil: Fair, I just default to stage validation. But yes, switched to a prim validator

pxr/usd/usdMedia/authorshipAPI.cpp
dgovil: Ah good call. I switched over

pxr/bin/usdauthors/usdauthors.cpp
dgovil: I used `||` so a record with only one of the two authored still shows up, but flagged as malformed. Otherwise it would silently be showing no inputs. Also, just added validation for it.

dgovil:
Thanks for the second round of notes, @spiffmon .
I took a pass

PR #4207, opened 2026-09-01
Shared GPU Buffers for Hydra

AdamFelt:
# Shared GPU Buffers for Hydra

## What it does

Lets a producer that already holds geometry on the GPU hand Hydra a handle to that buffer instead of a CPU `VtArray`, so Storm can bind or blit it on-device. Removes the per-frame GPU→CPU readback + CPU→GPU re-upload round trip that the CPU primvar path forces.

## Design

Two new generated schemas as optional children under a primvar — additive and ignorable, so a delegate that doesn't implement them just reads the CPU value as before:

```
primvars/points:
  primvarValue, interpolation, role
  extGpuBuffer:            # HdExtGpuBufferSchema
    backendApi, numElements, elementType, byteOffset, byteStride, directBindable
    rawHandle, rawHandleByteSize                    ← adopt route
    externalMemoryHandle, externalHandleType,
      memoryBlockSize, memoryOffset, dedicated      ← import route
    deviceUuid, logicalDeviceId
    sync:                  # HdExtGpuSyncSchema (optional)
      write/readSemaphore, externalWrite/ReadSemaphore, handleType, kind
```

Two orthogonal axes:

- **Route** — `adopt` (backendApi + deviceUuid + logicalDeviceId all match → wrap `rawHandle` non-owningly) / `import` (physical device matches but logical device doesn't, or different API → `CreateBufferFromExternalMemory` over the producer's allocation) / fall back to the CPU value, bumping `extGpuBufferFallbackCount`. Publishing both clusters is the recommended default.
- **Consumption** — `directBindable=true` → zero-copy alias BAR (`HdStExtGpuBufferArrayRange`), not aggregatable, own draw batch; `directBindable=false` → aggregation strategy's `CopyData` detects the external source and does a GPU→GPU blit into Storm's VBO, keeping the prim in indirect-draw batches. Route-agnostic: the blit reads whichever `HgiBuffer` the route resolved.

### Device identity

Two identity fields, deliberately separate:

- `deviceUuid` answers **which GPU** holds the memory (mismatch kills both routes, since opaque handles only import on the exporting device).
- `logicalDeviceId` (from the new `Hgi::GetLogicalDeviceId()`, a process-local monotonic counter) answers **which device object** minted `rawHandle` (mismatch kills adopt only).

Two logical Vulkan devices on one GPU report the same UUID but hand out unrelated `VkBuffer` values — without the second field that case is indistinguishable from a same-device producer and the consumer binds an arbitrary object. `0`/absent = "unknown" = match, for backward compat.

### Sync

`HdExtGpuSyncSchema` (nested inside `extGpuBuffer` so degradation works at both levels) covers both ordering edges on the GPU timeline:

- **RAW** via a producer-signalled write semaphore the consumer waits on.
- **WAR** via a consumer-signalled read semaphore the producer waits on before overwriting.

Binary, not timeline — `GL_EXT_semaphore` can only wait/signal binary and a GL producer is the motivating case; the `kind` field reserves timeline for later. `QueueWait/SignalExternalSemaphore` attach to the next submission (enqueued during Sync, executed at draw submit), so neither side CPU-stalls.

### Lifetime

A separate mechanism, not a schema member: the consumer retains the data source carrying the bound handle and parks it on `HgiBuffer` as an opaque `std::shared_ptr<void>` keepalive. Because backends destroy buffers through a GC that defers until in-flight command buffers retire, one release signal answers both "consumer let go" and "no submission still names it" — no fence, no pumped frame.

### Cross-API constraint that inverts ownership

OpenGL cannot export memory (`GL_EXT_memory_object` is import-only), so the Vulkan side must allocate exportable memory and GL imports it — even when GL is conceptually the producer. Imported buffers are refcounted-cached on `HdStResourceRegistry` (key includes the handle value, since two dedicated allocations both report `memoryOffset 0`); importing per-Sync would allocate a `VkDeviceMemory` every frame.

## Instancing

Both axes covered:

- **Prototype primvars** need no special handling — one buffer bound once, drawn N times (only cost is losing batch aggregation on the direct-bind path).
- **Instancer primvars** are the higher-value share: per-instance transforms published on the instancer prim with `numElements = instance count`, and one buffer can back several streams via `byteOffset`/`byteStride` (AoS and SoA both work). Nested instancing falls out for free.

**Caveat:** a shared transform buffer must already be in the renderer's matrix layout/precision, since the GPU path skips the CPU conversion.

## Tests

`testHdStExtGpuBuffer` renders the same cube (and an instanced grid) through the CPU path and the shared path and compares pixels:

- **default / `--instancing`** — Storm's own Hgi allocates; adopt route
- **`--vulkanSync`** — Vulkan producer on the consumer's device, own submission; adopt + native RAW/WAR semaphores
- **`--vulkanInterop`** — a second `HgiVulkan` stands in for a foreign-device producer; import route + imported semaphores + zero-copy alias (the acceptance case — single-device topology can't catch UUID-negotiation bugs by construction)
- **`--vulkanInterop --copy`** — same two-device topology, `directBindable=false`; import + GPU→GPU blit

## Status / gaps

**Landed:** core schema, both routes, both consumption strategies, binary semaphores (native + imported), refcounted imported-buffer cache, keepalive lifetime, full `HgiVulkan` interop surface, `logicalDeviceId` gating.

**Open:**

- `HgiGL` is adopt-only.
- `HgiMetal` is not just missing — its generic wrapper would misinterpret a foreign handle, so consumers should reject `backendApi=Metal`.
- Timeline semaphores unimplemented.
- `opaqueFd`/Linux allowed by the schema but only Win32 exercised.
- The imported-semaphore cache is session-lifetime (refcounting needs the queue's pending lists to hold a reference first).
- No test yet publishes a `rawHandle` from a foreign logical device — the exact case the adopt gate exists for, where the failure mode is UB rather than a clean error.

## Implementation stages (if the proposal is approved)

1. Add `HdExtGpuBufferSchema` to Hydra and consumption to Storm (GL only), no cross-backend interop and synchronization in Hydra or Storm.
2. Add `HdExtGpuSyncSchema` to Hydra and consumption to Storm (GL + Vulkan).

## Noteworthy

This feature was developed with help from AI.

### Checklist


pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            Describes an externally-owned GPU buffer that a producer publishes
+            on a primvar so a renderer can consume it directly instead of the
+            CPU staging round trip.
+
+            A producer names the buffer in one of two ways, and may use both:
+            "rawHandle" for a consumer sharing its context or logical device,
+            and the foreign-memory import cluster ("externalMemoryHandle" and
+            friends) for everyone else.
+
+            Lifetime. A consumer retains the data source carrying whichever
```
tcauchois, line 92: This doesn't feel like a robust system of lifetime management.  Based on experience working with lots of other developers writing filters, this feels like we'd get lots of bugs in both directions; accidentally holding onto buffers forever, and also deleting them prematurely.

Some alternatives I can think of off the top of my head (possibly switched via an enum in this schema):
1. Storm adopts the buffer on read and is responsible for freeing it when done with it.
2. The host app guarantees that the buffer is valid for the current frame.  Hydra leaves that primvar dirty, so it will re-pull that primvar next frame.  (Note, with both this and above, there's a complication if we don't pull these primvars e.g. because of primvar pruning or their owning prims are invisible. Hydra can't take ownership until it actually pulls the primvars...)
3. The host app guarantees that the buffer is valid until either the owning hydra prim is removed, or the primvar owner of the buffer is dirtied.
4. The host app guarantees that the buffer is valid until it calls resourceRegistry->InvalidateExtGpuBuffer(rawHandle), which will synchronously drop that buffer's reference from the render delegate and flag the owning primvar as dirty.
5. Storm provides a resourceRegistry->IsExtGpuBufferActive(rawHandle) if it's still actively using "rawHandle" and the host can't delete the handle until this returns false.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            ('deviceUuid',           T_TOKEN,
+             dict(DOC = '''
+                Physical device that owns the memory, as a 32-character
+                lowercase hex encoding of the 16-byte device UUID. Opaque
+                handles are only importable on the same physical device, so a
+                consumer compares this against its own device before adopting
+                or importing. Absent means "unknown" and is treated as a match
+                for backward compatibility.
+                ''')),
+            ('logicalDeviceId',      T_UINT64,
```
tcauchois, line 213: Note that for GL, the equivalent would be a GL context.  contexts are a bit more interesting with sharing, but we can say that if a context uses a sharing context, the sharing context is what we should use for the logical device.

With that in mind, can we change this to be a "rawDevice" which is a uint64 cast of the device/context*?  That seems more solid than a sequential count maintained by Hgi...

With "rawDevice", I'm not sure you need deviceUuid at least for adopt, although you might still need it for import.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            ('directBindable',    T_BOOL,
+             dict(DOC = '''
+                True if the consumer may bind the buffer directly (zero-copy).
+                ''')),
+            # Foreign-memory import cluster: describes the OS-shareable memory
+            # allocation the buffer is bound into, so a consumer on a different
+            # device (or API) can import it instead of adopting rawHandle.
+            ('externalMemoryHandle', T_UINT64,
+             dict(DOC = '''
+                OS-shareable handle (Win32 NT handle / fd, cast to uint64)
```
tcauchois, line 171: This is underspecified...  How is it produced/how is it intended to be consumed?

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+                imports the memory; see the schema-level note.
+                ''')),
+            ('externalHandleType',   T_TOKEN,
+             dict(DOC = '''
+                How to interpret externalMemoryHandle
+                ("opaqueWin32"/"opaqueFd"). Required whenever
+                externalMemoryHandle is set; the handle kind is never inferred
+                from the value, which can collide with a native rawHandle.
+                ''')),
+            ('memoryBlockSize',      T_SIZET,
```
tcauchois, line 188: For clarity: externalMemoryBlockSize...

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+             dict(DOC = '''
+                Total size of the memory block externalMemoryHandle refers to.
+                The importer must allocate the full block, not just byteSize.
+                ''')),
+            ('memoryOffset',         T_SIZET,
```
tcauchois, line 193: For clarity: externalMemoryOffset...

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+             dict(DOC = '''
+                The buffer's offset WITHIN that memory block. Distinct from
+                byteOffset, which is this stream's offset within the buffer;
+                both may be non-zero at once.
+                ''')),
+            ('dedicated',            T_BOOL,
```
tcauchois, line 199: Underspecified: what does this mean?

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            writes before its own reads (RAW) and signal when its reads
+            complete (WAR). Nested as the "sync" child of
+            HdExtGpuBufferSchema. Optional: absent means the consumer falls
+            back to whatever implicit coherence the producer provides.
+            ''',
+        MEMBERS = [
+            ('backendApi', T_TOKEN,
+             dict(DOC = '''
+                GPU API that created the semaphores (matches
+                Hgi::GetAPIName()). A consumer ignores the sync objects if
```
tcauchois, line 260: Not strong enough... awkwardly, if the backendApi doesn't match, the only way this works is if the application issues a flush through backendApi before hydra runs, but that puts the burden on the app developer of making sure ahead of time that it's passing the right backendApi.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+                this does not match its active backend.
+                ''')),
+            ('writeSemaphore', T_UINT64,
```
tcauchois, line 263: Amusingly, in my notes I think I had these swapped (since I'm on the hydra side of the fence :p).  I do think a name clarification would help here, e.g. "beforeHydraRead"/"afterHydraRead" (probably something better than that though).

Semantics of "writeSemaphore" particularly are that hydra can't touch the buffer until it's signalled.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+             dict(DOC = '''
+                Native handle of the semaphore the producer signals after
+                writing the buffer; the consumer waits on it before reading
+                (RAW). 0/absent = none. Only usable when the producer and
+                consumer share a device -- see externalWriteSemaphore for the
+                cross-device form.
+                ''')),
+            ('readSemaphore', T_UINT64,
```
tcauchois, line 271: Underspecified: for import and adopt/copy, we can ping this when we've finished copying, but what responsibility do we have for the adopt/direct bind case?  What if we don't pull this primvar right away (e.g. because of primvar pruning); will that cause a deadlock?

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+                semaphore, so a consumer on a different device can import it.
+                Independent of writeSemaphore: a producer may publish both,
+                letting a same-device consumer use the native handle and
+                everyone else import.
+                ''')),
+            ('externalReadSemaphore', T_UINT64,
+             dict(DOC = '''
+                OS-shareable handle for the read semaphore (WAR), for import.
+                ''')),
+            ('handleType', T_TOKEN,
```
tcauchois, line 289: externalHandleType.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+             dict(DOC = '''
+                How to interpret the external semaphore handles ("opaqueWin32"
+                / "opaqueFd"). Required whenever either external handle is
+                set.
+                ''')),
+            ('kind', T_TOKEN,
```
tcauchois, line 295: Possibly also event/MTLFence?  Dunno how general we need to make it.

As with the note in backendApi, we can't *really* safely ignore the sync container, because to do so requires the app to insert a flush in its own API for the "writeSemaphore" bit, and to also know that it shouldn't expect "readSemaphore" to be signalled.  If the app & hydra have a negotiation process, they can collectively decide whether a certain sync format will work or if they should just insert flushes on either end, but we can't determine that at frametime.

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            ('deviceUuid', T_TOKEN,
+             dict(DOC = '''
+                Physical device that owns the semaphores, as a 32-character
+                lowercase hex encoding of the 16-byte device UUID. Absent
+                means "unknown" and is treated as a match. Compared the same
+                way as the buffer schema's deviceUuid: equal means the native
+                handles are usable directly, unequal means the external
+                handles must be imported.
+                ''')),
+            ('logicalDeviceId', T_UINT64,
```
tcauchois, line 310: Same note as above re logicalDeviceId.  Also, I'm *hoping* that logicalDeviceId/deviceUuid match what was passed in for buffer ownership, which would make it redundant to also specify them in the sync container.

pxr/imaging/hd/types.h
```diff
@@ -363,6 +364,13 @@ TfHashAppend(HashState &h, HdTupleType const &tt)
     h.Append(tt.type, tt.count);
 }
 
+/// Stream output for HdTupleType, e.g. "HdTupleType(HdTypeFloatVec3, 1)".
```
tcauchois, line 367: Helpful claude comment appreciated for WIP code but should be stripped from the final PR.

pxr/imaging/hdSt/extGpuBuffer.cpp
```diff
@@ -0,0 +1,69 @@
...
+    size_t byteSize)
+    : HgiBuffer(HgiBufferDesc())
+    , _rawHandle(rawHandle)
+    , _byteSize(byteSize)
+{
+    _descriptor.byteSize = byteSize;
+    // Match HdStExtGpuBufferArrayRange::GetUsageHint(): external buffers may be
+    // bound as vertex attributes or read as storage (e.g. by GPU computations),
+    // so advertise both rather than vertex-only.
+    _descriptor.usage = HgiBufferUsageVertex | HgiBufferUsageStorage;
```
tcauchois, line 36: Smaller note: we don't want to flag "Vertex" if we use this for other primvar rates in the future.  Could wire a bool through, or a usage flag.

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+        _numElements = hdDesc.numElements;
+    }
+
+    IncrementVersion();
+    return true;
+}
+
+bool
+HdStExtGpuBufferArrayRange::MergeExternalResources(
```
tcauchois, line 203: What's the difference between merge & update?  At a quick scan, they seem like duplicated code.

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+bool
+HdStExtGpuBufferArrayRange::IsValid() const
+{
+    return _valid;
+}
+
+bool
+HdStExtGpuBufferArrayRange::IsAssigned() const
```
tcauchois, line 271: Not sure this is correct...

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+{
+    return _valid;
+}
+
+bool
+HdStExtGpuBufferArrayRange::IsImmutable() const
```
tcauchois, line 277: Seems desirable for it to be either usually immutable or always immutable? But I don't know if the flag here affects shader bindings, or just buffer aggregation (which doesn't matter for us).

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+bool
+HdStExtGpuBufferArrayRange::RequiresStaging() const
+{
+    return false;
+}
+
+bool
+HdStExtGpuBufferArrayRange::Resize(int numElements)
+{
+    // Alias ranges own no real storage: element-count and allocation-size
```
tcauchois, line 291: This should only be called if we're trying to aggregate external sources with non-external sources in the same BAR, which is something we could potentially do but certainly not in this class.  I don't know how we guarantee that we won't do that, but anyway this seems like a coding error?

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+    // as reusable capacity.
+    _numElements = numElements;
+    return false;
+}
+
+void
+HdStExtGpuBufferArrayRange::CopyData(
+    HdBufferSourceSharedPtr const &bufferSource)
+{
+    // No-op for alias ranges — the external buffer IS the data.
```
tcauchois, line 305: Likewise, coding error?

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+VtValue
+HdStExtGpuBufferArrayRange::ReadData(TfToken const &name) const
+{
+    TF_CODING_ERROR("ReadData not supported on external GPU buffer array range");
```
tcauchois, line 311: Fine for now, but no reason we couldn't support this...

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
```diff
@@ -0,0 +1,448 @@
...
+HdStExtGpuBufferArrayRange::GetUsageHint() const
+{
+    return HdBufferArrayUsageHintBitsVertex |
+           HdBufferArrayUsageHintBitsStorage;
+}
+
+void
+HdStExtGpuBufferArrayRange::SetBufferArray(HdBufferArray *)
+{
+    // Alias ranges are not associated with a real buffer array.
```
tcauchois, line 367: Coding error, as you probably don't want the buffer array registry trying to manage allocations in this bar type.

pxr/imaging/hdSt/extGpuBufferSource.cpp
```diff
@@ -0,0 +1,123 @@
...
+size_t
+HdStExtGpuBufferSource::ComputeHash() const
+{
+    // The default HdBufferSource::ComputeHash hashes GetData() bytes, which
+    // dereferences nullptr for external GPU sources (no CPU payload).  Hash
+    // the GPU-side identity instead.  This is intentionally stable across
+    // animation frames -- HdStMesh::_PopulateVertexPrimvars transitions the
+    // BAR to mutable on the second dirty frame and resets the immutable
+    // share key, so this hash only gates first-time cross-rprim sharing.
+    return TfHash::Combine(
```
tcauchois, line 78: Minor note: is there an advantage to only hashing the inputs that contributed to this specific buffer depending on adopt/import type?

pxr/imaging/hdSt/extGpuImportedBuffer.h
```diff
@@ -0,0 +1,87 @@
...
+    HdSt_ImportedExtGpuBuffer(HdSt_ImportedExtGpuBuffer const &) = delete;
+    HdSt_ImportedExtGpuBuffer &operator=(
+        HdSt_ImportedExtGpuBuffer const &) = delete;
+
+    /// The buffer aliasing the producer's memory.
+    HgiBufferHandle const &GetHandle() const { return _handle; }
+
+    /// Stop reporting to the registry, which is being destroyed.  Only reached
+    /// when a holder outlives the renderer.
+    void DetachFromRegistry() { _registry = nullptr; }
```
tcauchois, line 73: When would you need this?

pxr/imaging/hdSt/extGpuImportedBuffer.h
```diff
@@ -0,0 +1,87 @@
...
+/// device-memory reference that only destruction returns.  So the registry
+/// caches them on Key below, but holds only weak references; every holder along
+/// the chain from routing to draw -- the HdStExtGpuBufferDesc, the
+/// HdStExtGpuBufferSource built from it, and the HdStExtGpuBufferArrayRange it
+/// is committed to -- holds a strong one.  The last holder to let go destroys
+/// the buffer and removes it from the cache.  That way nothing is pinned for
+/// the life of the renderer, and no holder can be left pointing at a buffer
+/// somebody else decided to free.
+///
+class HdSt_ImportedExtGpuBuffer
```
tcauchois, line 42: This whole thing seems to be a refcount object, but the resource instance registry already has refcounting?  Unclear why you need this...

pxr/imaging/hdSt/extGpuBufferConsumer.cpp
```diff
@@ -0,0 +1,296 @@
...
+    // semaphores belong to the active backend.
+    if (HdExtGpuSyncSchema sync =
+            HdExtGpuSyncSchema::GetFromParent(schema.GetContainer())) {
+        const HdTokenDataSourceHandle apiDs = sync.GetBackendApi();
+        if (apiDs && apiDs->GetTypedValue(0.0f) == hgi->GetAPIName()) {
+            const uint64_t writeSem = _ResolveSemaphore(
+                sync, registry, hgi,
+                sync.GetWriteSemaphore(), sync.GetExternalWriteSemaphore());
+            if (writeSem) {
+                hgi->QueueWaitExternalSemaphore(writeSem);
```
tcauchois, line 227: We want to queue this before either:
(a) Importing
(b) Copying into a destination bar
(c) Drawing a batch using an alias bar

To simplify things, if we make a list of them we can just queue them all before jumping into the buffer resolve code in CommitResources.

pxr/imaging/hdSt/extGpuBufferConsumer.cpp
```diff
@@ -0,0 +1,296 @@
...
+            }
+            // WAR: signal the read semaphore after this consumer's reads
+            // complete, so the producer may wait before overwriting the buffer.
+            const uint64_t readSem = _ResolveSemaphore(
```
tcauchois, line 231: We want to queue this after:
(a) Importing
(b) Copying into a destination bar
(c) Drawing a batch using an alias bar

Depending on how latency sensitive this is, we could just make a list to queue in Hgi::EndFrame.  If you need more granularity, we could queue (a) and (b) after the buffer resolve code in CommitResources...  Queueing them here is absolutely incorrect, though.

pxr/imaging/hdSt/extGpuBufferConsumer.cpp
```diff
@@ -0,0 +1,296 @@
...
+        //    retried (uncached, since failures are not cached) every Sync.
+        //    Both consumption strategies are served from the imported buffer:
+        //    a directBindable stream is aliased and bound zero-copy, while a
+        //    non-directBindable stream is GPU-to-GPU blitted into Storm's
+        //    aggregated VBO. The blit still avoids the CPU round trip this path
+        //    exists to remove -- for a foreign-device producer, which has no
+        //    cheap CPU copy of its own, that is cheaper than the CPU fallback,
+        //    not more expensive.
+        hdDesc.importedBuffer =
+            registry->GetOrCreateImportedExtGpuBuffer(hdDesc);
```
tcauchois, line 203: Where does the actual import happen?  Hopefully not here.  We should be deferring it to CommitResources to get better thread scheduling...

pxr/imaging/hdSt/interleavedMemoryManager.cpp
```diff
@@ -733,6 +734,37 @@ HdStInterleavedMemoryManager::_StripedInterleavedBufferRange::CopyData(
         HdDataSizeOfTupleType(bufferSource->GetTupleType());
     size_t const elementStride = _stripedBuffer->GetElementStride();
 
+    if (auto const *extSrc = dynamic_cast<HdStExtGpuBufferSource const *>(
```
tcauchois, line 737: The dynamic cast as a structural support here is a bit sketch.  We could add two functions to HdBufferSource instead:
bool IsGpuBacked() const;
VtValue GetGpuResource() const; (which would return desc.cachedHgiHandle).

All of the other fields of desc I think are already exposed on HdBufferSource...

pxr/imaging/hdSt/extGpuBuffer.cpp
tcauchois: Ahh yeah, we're already using this code for stuff like face-varying.  Different APIs are more or less strict or will add more or less performance nonsense around whether Vertex is incorrectly set here, so seems better to pass it in from the prim handling.

pxr/imaging/hdSt/vboMemoryManager.cpp
```diff
@@ -668,6 +669,51 @@ HdStVBOMemoryManager::_StripedBufferArrayRange::CopyData(
     }
     size_t vboOffset = bytesPerElement * _elementOffset;
 
+    if (auto const *extSrc = dynamic_cast<HdStExtGpuBufferSource const *>(
```
tcauchois, line 672: See comment in interleavedMemoryManager.cpp.

pxr/imaging/hdSt/vboSimpleMemoryManager.cpp
```diff
@@ -458,6 +460,50 @@ HdStVBOSimpleMemoryManager::_SimpleBufferArrayRange::CopyData(
         srcSize = dstSize;
     }
 
+    if (auto const *extSrc = dynamic_cast<HdStExtGpuBufferSource const *>(
```
tcauchois, line 463: See comment in interleavedMemoryManager.cpp.

pxr/imaging/hgi/buffer.cpp
```diff
@@ -21,6 +21,18 @@ HgiBuffer::GetDescriptor() const
     return _descriptor;
 }
 
+void
```
tcauchois, line 24: This is an implementation detail of extGpuBuffer, not sure why you need to modify the base class. Please revert.

pxr/imaging/hgi/buffer.h
```diff
@@ -119,12 +120,41 @@ class HgiBuffer
     HGI_API
     virtual void* GetCPUStagingAddress() = 0;
 
+    /// Attach an opaque reference that this buffer holds for as long as it
```
tcauchois, line 123: See comment in buffer.cpp.

pxr/imaging/hgi/hgi.h
```diff
@@ -37,6 +37,72 @@ class HgiIndirectCommandEncoder;
 
 using HgiUniquePtr = std::unique_ptr<class Hgi>;
 
+/// \enum HgiExternalHandleType
```
tcauchois, line 40: Would be nice to move some of this (the structs specifically) to buffer.* and semaphore.* (or if we want a more generic name, sync.*).

pxr/imaging/hgi/hgi.h
```diff
@@ -37,6 +37,72 @@ class HgiIndirectCommandEncoder;
...
+    HgiSemaphoreKindBinary = 0,
+    HgiSemaphoreKindTimeline,
+};
+
+/// \struct HgiInteropBufferInfo
+///
+/// Describes an interop-allocated buffer's external memory so another GPU API
+/// (e.g. OpenGL via GL_EXT_memory_object) can import and alias the same memory.
+/// Returned by Hgi::CreateInteropBuffer.
+struct HgiInteropBufferInfo
```
tcauchois, line 66: Naming suggestion: HgiExportableMemoryBufferDesc.

pxr/imaging/hgi/hgi.h
```diff
@@ -37,6 +37,72 @@ class HgiIndirectCommandEncoder;
...
+    // Whether the allocation is a dedicated memory object.
+    bool     dedicated = false;
+};
+
+/// \struct HgiExternalMemoryBufferDesc
+///
+/// Describes a foreign memory allocation to import and wrap in a buffer of this
+/// backend, so the backend can read memory another device allocated. The
+/// producer-side counterpart is HgiInteropBufferInfo.
+struct HgiExternalMemoryBufferDesc
```
tcauchois, line 85: Naming suggestion: HgiImportableMemoryBufferDesc.

pxr/imaging/hgi/hgi.h
```diff
@@ -37,6 +37,72 @@ class HgiIndirectCommandEncoder;
...
+    // Size of the whole memory block the handle refers to. The import allocates
+    // the full block even when only byteSize of it is used here.
+    size_t memoryBlockSize = 0;
+    // Offset of the buffer within that block.
+    size_t memoryOffset = 0;
+    // Size of the buffer to create at memoryOffset.
+    size_t byteSize = 0;
+    // Must match how the producer allocated the memory or the import fails.
+    bool dedicated = false;
+    HgiBufferUsage usage = 0;
```
tcauchois, line 102: Not sure "Usage" belongs in this struct.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
     HgiBufferHandle CreateBuffer(HgiBufferDesc const & desc);
 
+    /// Wrap an externally-owned native GPU buffer in a NON-OWNING HgiBuffer so
+    /// this backend can bind it without copying. \p rawHandle is interpreted per
+    /// backend (a GL buffer name, a VkBuffer, ...). The returned handle must NOT
+    /// free the underlying native resource when destroyed via DestroyBuffer.
+    /// Backends that cannot adopt a foreign handle return an empty handle; the
+    /// caller then falls back to its own generic wrapper. Default: empty handle.
+    HGI_API
+    virtual HgiBufferHandle CreateExternalBuffer(
```
tcauchois, line 304: Name suggestion: CreateBufferFromRawHandle.

Is there a way to make it safe to call DeleteBuffer on this? e.g. if it dynamic casts a certain way it's a no-op?  I'd rather say that "Hgi buffers returned from this function aren't owned by this Hgi instance, and Hgi->DestroyBuffer will be a no-op", rather than "please don't destroy this".  Or we could make it a coding error/etc.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+        uint64_t rawHandle, size_t byteSize, HgiBufferUsage usage);
+
+    /// Allocate a buffer whose memory is EXPORTABLE to other GPU APIs (interop).
+    /// Returns a real, owning HgiBuffer and, in \p outInfo, the external-memory
+    /// description another API can import to alias the same memory. Backends
+    /// without interop support return an empty handle and leave \p outInfo at
+    /// its defaults. Default: unsupported.
+    HGI_API
+    virtual HgiBufferHandle CreateInteropBuffer(
```
tcauchois, line 313: Name suggestion: CreateBufferForExport.

Wondering whether we want a capability for buffer interop, so we can say that this function returns null if it's not supported?  I think it failing if interop isn't supported is better than making an allocation, as (e.g.) if we fail to interop a buffer we will probably unwind a bunch and go to the CPU fallback.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+
+    /// IMPORT memory another device (or API) allocated as exportable, and wrap
+    /// it in a real, OWNING buffer of this backend: the returned handle owns the
+    /// buffer object and the imported memory reference, but not the underlying
+    /// allocation, which stays alive as long as any importer holds it. Use this
+    /// when the producer's memory does not live on this backend's device, so
+    /// CreateExternalBuffer's adopt path is not available. Backends that cannot
+    /// import return an empty handle. Default: unsupported.
+    HGI_API
+    virtual HgiBufferHandle CreateBufferFromExternalMemory(
```
tcauchois, line 324: Name suggestion: CreateBufferFromImport?

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+        HgiExternalMemoryBufferDesc const& desc);
+
+    /// IMPORT a semaphore created by another device (or API) from its OS handle,
+    /// returning a backend-native handle usable with QueueWait/QueueSignal and
+    /// DestroyExternalSemaphore, or 0 if unsupported. This is the cross-device
+    /// counterpart of CreateExternalSemaphore, which creates one locally.
+    /// \p externalHandle ownership follows the platform convention (fd: taken;
+    /// Win32: retained by the caller). Default: unsupported.
```
tcauchois, line 332: To match the rest of Hgi, I feel like we want an HgiSemaphoreHandle/HgiSyncHandle and paired Create/Destroy calls, rather than just passing u64s around.

I also think just like we have import/export structs for buffers, we should have them for semaphores as well.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+    /// Returns an id for the LOGICAL device this backend renders through --
+    /// unique within this process, 0 when the backend has no such object. Where
+    /// GetDeviceUuid names the GPU, this names the device object whose handle
+    /// namespace a native resource handle belongs to; the two differ when two
+    /// logical devices drive one GPU, and that is exactly when adopting a
+    /// foreign native handle would bind an unrelated object. Backends without a
+    /// logical device (GL, whose namespace is the context share group) return 0,
+    /// which callers treat as "unknown". Default: 0.
+    HGI_API
+    virtual uint64_t GetLogicalDeviceId() const;
```
tcauchois, line 356: I feel like a pointer to the device would be better than a sequential int here, and also think a pointer to the current GL context (or the share context if there is one) would be safer than just disabling this check for GL.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+
+    /// Create a binary semaphore whose signal state is EXPORTABLE to other GPU
+    /// APIs. Returns a backend-native handle (0 if unsupported) and, in
+    /// \p outExternalHandle, an OS handle (Win32 NT handle / fd) the other API
+    /// imports. Used to order a producer's writes in another API against this
+    /// backend's reads of a shared buffer. Default: unsupported.
+    HGI_API
+    virtual uint64_t CreateExternalSemaphore(uint64_t* outExternalHandle);
```
tcauchois, line 364: Group with ImportExternalSemaphore and see notes there.

pxr/imaging/hgi/hgi.h
```diff
@@ -228,6 +294,92 @@ class Hgi
...
+
+    /// Destroy a semaphore returned by CreateExternalSemaphore.
+    HGI_API
+    virtual void DestroyExternalSemaphore(uint64_t semaphore);
```
tcauchois, line 368: If semaphores have an actual type in Hgi, we just need a "void DestroySemaphore(HgiSemaphore *semaphore)" which would also work for native semaphores in the future, and the wait/signal functions could move to the semaphore class instead of living on Hgi directly.

tcauchois (changes requested):
I think there are still some major issues with the schema to iron out.  The broad strokes of the HdSt/Hgi implementation look reasonable but I left a bunch of comments there as well, about cleaning up the object model a bit so that it's maintainable.  I haven't read through HgiVulkan yet but can do a readthrough of that at some point.  Curious to get David's take as well.

pxr/imaging/hd/hdSchemaDefs.py
lilike-adsk: Thanks for comments, just clarify a bit aroud the current ownership and lifetime management model: producer/host app is owning the allocation/deallocation of the external buffer, and the ref count mechnism (wrap the rawHandle via a custom HdUInt64DataSourceHandle with a buffer deleter) is leveraged to manage the lifetime of the external buffer. This model did requre both producer and consumer (including Hydra SI filters) to respect some implicit contracts, e.g., the producer need to hold the HdUInt64DataSourceHandle in a frame, and the consumer must not extract rawHandle and re-create a new HdUInt64DataSourceHandle to clone it and pass the new HdUInt64DataSourceHandle without deleter to the downstream, the renderer need to increase/decrease ref count explicitly (hgiBuffer's keepAlive case) when it can't hold HdUInt64DataSourceHandle directly. 

Is your concern about developer's filters might not respect the contract? For your proposal around the IsActive/Invalidate explicit contract on resourceRegistry, meaning new APIs on HdResourceRegistry for ExtGpuBuffer?

pxr/imaging/hd/hdSchemaDefs.py
lilike-adsk: Will polish the comment

pxr/imaging/hgi/hgi.h
lilike-adsk: Good point!

pxr/imaging/hd/types.h
lilike-adsk: Need to respect the Format (Return XXX), right?

pxr/imaging/hdSt/extGpuBufferConsumer.cpp
lilike-adsk: Good catch! Will fix.

pxr/imaging/hdSt/extGpuImportedBuffer.h
lilike-adsk: It's for cross-device import, we could use the instance registry probably as you said.

pxr/imaging/hgi/buffer.cpp
lilike-adsk: As mentioned, this was added with the ref count mechnism, we can remove if we use other mechnisms

pxr/imaging/hd/hdSchemaDefs.py
```diff
@@ -74,6 +74,260 @@
...
+            Because retention follows the published value, a filter that
+            substitutes a different handle releases the allocation the old one
+            named, which is the correct outcome -- the substituted buffer is
+            what gets bound. The converse is the limit worth knowing: a filter
+            that copies the value into a data source of its own breaks the
+            chain, so this is a convention a producer opts into with its
+            consumers, not a guarantee the scene description enforces.
+            ''',
+        MEMBERS = [
+            ('backendApi',        T_TOKEN,
```
tcauchois, line 110: One thing that might be future friendly is to create an HgiExternalResource/ExternalBuffer abstraction that provides the accessors we need from Storm, and then transport that opaquely here.  That would let us create subclasses of ExternalResource for things like HANDLE/FD based interop and hide some of the complexity rather than put it in the schema here.

tcauchois: That also lets us tie lifetime (via a strong or weak ptr check) to the external resource itself, rather than the datasource.

pxr/imaging/hd/hdSchemaDefs.py
tcauchois: From the call: I'm nervous about tying lifetime to a datasource, and had proposed instead tying it to something like "HgiExternalResource" which is something that (a) holds a strong or weak reference to the buffer; (b) provides a base class API for Storm to access the hardware resource and abstracts away stuff like the FD-based interop.  I'll come up with a concrete code proposal for what this would look like.

pxr/imaging/hd/hdSchemaDefs.py
tcauchois: As discussed on the call, a stronger wording here helps, e.g.:
If the consumer cannot interpret the semaphores because of a differing backendApi, the external resources are unsafe to access and the consumer should fall back to the CPU path.

pxr/imaging/hd/hdSchemaDefs.py
tcauchois: As mentioned above, if we decide that the renderer should just skip the extGpuBuffer if it doesn't understand the sync setup, and document that, I think this is fine.

pxr/imaging/hdSt/extGpuBufferArrayRange.cpp
tcauchois: I guess they are immutable in that you want to bind them const, but the binding itself might count as mutable if it's updating, and we want to flag it as such in the instance registry/batching code...

pxr/imaging/hd/hdSchemaDefs.py
lilike-adsk: @tcauchois Have you had a chance to make a code proposal?

pxr/imaging/hgi/externalBufferArena.h
```diff
@@ -0,0 +1,384 @@
...
+/// application-allocated GPU memory, and owns the single semaphore pair that
+/// orders that application's access against Hgi's.
+///
+/// Obtain one from Hgi::GetExternalBufferArena(), which creates it on demand
+/// and returns null when this Hgi cannot interop with the requested arena
+/// type.  That null *is* the negotiation step: the application and Hgi settle
+/// on an interop format once, when the arena is created, rather than
+/// rediscovering per frame whether a given buffer can be shared.
+///
+/// There is one arena per type per Hgi, which makes ONE PRODUCER PER ARENA a
```
tcauchois: (Still reading the rest of the PR but): if this is an issue, we could conceivably let you register multiple VK->VK arenas, one for each producer?  If this isn't a current headache for you, that's also a change we could make down the road when it becomes a requirement.

lilike-adsk: Agree with starting from simple case, I used one arena per device/context at the begining, then realized it could be added later if it's really needed.

PR #4210, opened 2026-09-09
Handle blocked mass attributes in ComputeMassProperties

### LINKED ISSUE #4155: UsdPhysicsRigidBodyAPI::ComputeMassProperties reads blocked MassAPI attributes into uninitialized locals
### Description of Issue

`UsdPhysicsRigidBodyAPI::ComputeMassProperties` returns nondeterministic garbage mass properties when any `UsdPhysicsMassAPI` attribute on the rigid-body prim (or a collider prim in its subtree) carries an attribute value block.

The cause is in `pxr/usd/usdPhysics/rigidBodyAPI.cpp`. `_ParseMassApi` reads attributes into uninitialized locals and only validates the value afterwards:

```cpp
float m;
massAttribute.Get(&m);
if (m > 0.0f)
{
    result.mass = m;
}
```

For an *unauthored* attribute, `Get()` succeeds and returns the schema fallback (`0.0`), which correctly fails the `> 0` check. For a *blocked* attribute, `Get()` **fails** and leaves `m` holding indeterminate stack memory. Any positive garbage value is then taken as an authored mass, `accumulateMass` becomes false, and the garbage is returned as the body mass. The same pattern affects `GfVec3f dg` (diagonalInertia), `GfQuatf pa` (principalAxes) — both left uninitialized by their default constructors — and `GfVec3f v` in `_GetCoM` (centerOfMass), where garbage that happens to be finite is accepted as an authored center of mass.

Because this is uninitialized-memory UB, results vary by platform and process state: in our CI (newton-physics/newton#3418) the same stage produced masses of `2.67e-39` (ubuntu), `1.4e-45` (macOS), and `2.8e-45` (windows) on GitHub Actions runners, while passing with the correct value `8.0` on other machines where the stack memory happened to be benign.

Expected behavior: a blocked attribute resolves to no value, so it should be treated exactly like an unauthored one (mass computed from colliders/density).

Suggested fix: check the return value of each `Get()` call (treating failure as unauthored), or pre-initialize the locals to the schema fallback sentinels (`0.0`, `(0,0,0)`, `(0,0,0,0)` quaternion, non-finite center of mass).

### Steps to Reproduce

```python
from pxr import Gf, Usd, UsdGeom, UsdPhysics

stage = Usd.Stage.CreateInMemory()
UsdGeom.SetStageMetersPerUnit(stage, 1.0)
body = UsdGeom.Xform.Define(stage, "/Body")
UsdPhysics.RigidBodyAPI.Apply(body.GetPrim())
mass_api = UsdPhysics.MassAPI.Apply(body.GetPrim())
mass_api.CreateMassAttr().Set(5.0)
mass_api.GetMassAttr().Block()  # attribute now resolves to no value

cube = UsdGeom.Cube.Define(stage, "/Body/Collider")
cube.CreateSizeAttr().Set(0.2)
UsdPhysics.CollisionAPI.Apply(cube.GetPrim())

def mass_info(prim):
    mi = UsdPhysics.RigidBodyAPI.MassInformation()
    mi.volume = 0.008
    i = 0.008 * (0.2**2 + 0.2**2) / 12.0
    mi.inertia = Gf.Matrix3f(i, 0, 0, 0, i, 0, 0, 0, i)
    mi.centerOfMass = Gf.Vec3f(0.0)
    mi.localPos = Gf.Vec3f(0.0)
    mi.localRot = Gf.Quatf(1.0)
    return mi

rb = UsdPhysics.RigidBodyAPI(body.GetPrim())
mass, diag, com, axes = rb.ComputeMassProperties(mass_info)
print(mass)  # expected 8.0 (volume 0.008 * default density 1000); observed: nondeterministic
```

Whether the wrong value appears depends on stack contents; running under valgrind flags the uninitialized reads deterministically.

Reproduced on Linux x86_64, macOS (arm64), Windows x86_64 (GitHub Actions runners).

### Package Versions

usd-core 26.3 (PyPI); the affected code is unchanged on the current `dev` branch (only instance-proxy traversal was added).

### Build Flags

Stock PyPI `usd-core` wheels, no custom build.

Miko997:
### Description of Change(s)

Guard failed mass-attribute reads.

Check `Get()` before using mass, diagonal inertia, principal axes and center of mass in `ComputeMassProperties`. Failed reads could leave uninitialized values in the calculations. The four guards preserve the existing validity checks and fallback behavior.

The added regression test exercises 18 cases covering body/collider blocks, unauthored controls, authored overrides, transforms and density precedence. Saved local results:

- Baseline cube mass was approximately `1.33e-43`; patched mass was `8.0`.
- Focused Valgrind Memcheck with origin tracking: 2,078 errors on the base, zero with the patch; no suppressions.
- Patched usdPhysics suite: 50 tests passed with zero skips. The 18 regression cases are subtests within that suite, not 18 additional test methods.

Full-repository and cross-platform testing were not performed.

Pixar has confirmed CLA clearance.

Codex helped with implementation.

Not applicable.

### Fixes Issue(s)

Fixes #4155.

### Checklist

- [ ] I have verified that all unit tests pass with the proposed changes — only the usdPhysics suite was run.

### DIFF
--- a/pxr/usd/usdPhysics/rigidBodyAPI.cpp
+++ b/pxr/usd/usdPhysics/rigidBodyAPI.cpp
@@ -277,15 +277,14 @@ _MassApiData _ParseMassApi(const UsdPrim& usdPrim)
         densityAttribute.Get(&result.density);
 
         float m;
-        massAttribute.Get(&m);
-        if (m > 0.0f)
+        if (massAttribute.Get(&m) && m > 0.0f)
         {
             result.mass = m;
         }
 
         GfVec3f dg;
-        diagonalInertia.Get(&dg);        
-        if (!GfIsClose(dg, GfVec3f(0.0f), COMPARE_TOLERANCE))
+        if (diagonalInertia.Get(&dg) &&
+            !GfIsClose(dg, GfVec3f(0.0f), COMPARE_TOLERANCE))
         {
             result.hasInertia = true;
             result.diagonalInertia = dg;
@@ -293,8 +292,9 @@ _MassApiData _ParseMassApi(const UsdPrim& usdPrim)
         
         // 0 0 0 0 is the sentinel value
         GfQuatf pa;
-        principalAxes.Get(&pa);
-        if (!GfIsClose(pa.GetImaginary(), GfVec3f(0.0f), COMPARE_TOLERANCE) || fabsf(pa.GetReal()) > COMPARE_TOLERANCE)
+        if (principalAxes.Get(&pa) &&
+            (!GfIsClose(pa.GetImaginary(), GfVec3f(0.0f), COMPARE_TOLERANCE) ||
+             fabsf(pa.GetReal()) > COMPARE_TOLERANCE))
         {
             result.hasPa = true;
             result.principalAxes = pa;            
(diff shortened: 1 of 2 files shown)


Miko997:
Hi, the CLA is confirmed and this is ready for review. Could a maintainer approve the pending BuildUSD workflow and take a look when convenient? Thanks!

PR #4222, opened 2026-09-16
Add initial implementation for native capsules

miguelh-nvidia:
### Description of Change(s)

Adds an initial implementation for rendering native capsules in Storm, based (very closely) on the existing native sphere implementation. The main differences are the capsule-specific rprim parameters and the shader’s bounds, lateral-surface, and spherical-cap intersection calculations. The feature is guarded by HDST_ENABLE_NATIVE_CAPSULES and includes image tests.

AI assistance was used during the development of this PR. I reviewed the implementation, tests, and documentation and take full responsibility for the contribution.

### Checklist


jesschimein:
Filed as internal issue #USD-12640

pxr/imaging/hdSt/shaders/capsule.glslfx
```diff
@@ -0,0 +1,348 @@
...
+        vec3(maxRadius, maxRadius, 0.5 * (boundsMax - boundsMin)),
+        HdGet_axis());
+
+    return ComputeCapsuleBboxEye(center, halfExtent, instanceModelView);
+}
+
+--- --------------------------------------------------------------------------
+-- glsl ImplicitSurface.Intersection.Capsule
+
+void AddCapsuleHit(CapsuleHit hit, inout CapsuleHits hits)
```
roggiezhang-nv: **[P1] Use portable reference and struct initialization syntax for Metal**

With `HDST_ENABLE_NATIVE_CAPSULES=1`, rendering this PR's `simpleCapsule.usda` on HgiMetal produces an empty image and exits with status 1. The fragment compiler reports `unknown type name 'inout'` here. This parameter needs Storm's reference abstraction:

```suggestion
void AddCapsuleHit(CapsuleHit hit, REF(thread, CapsuleHits) hits)
```

There is a second compilation issue in the same shader: GLSL-style struct construction such as `CapsuleHit(0.0, CAPSULE_HIT_NONE)` and `CapsuleHits(...)` fails with `no matching constructor for initialization`. Changing only `inout` is insufficient. Portable factory functions that explicitly assign `dist`/`kind` and `nearHit`/`farHit` resolve those errors.

I reproduced both failures through the actual Storm renderer, then confirmed that the reference-macro change plus factory functions makes the scene render. Please cover this with a Metal rendering test; a successful C++ build does not compile these runtime shaders.

pxr/imaging/hdSt/shaders/capsule.glslfx
```diff
@@ -0,0 +1,348 @@
...
+}
+
+--- --------------------------------------------------------------------------
+-- glsl ImplicitSurface.Fragment.Capsule
+
+HitPoints ComputeHitPointsObj(Ray rayObj)
+{
+    const int axis = HdGet_axis();
+    const vec3 rayOriginCapsule = ToCapsuleSpace(rayObj.origin, axis);
+    const vec3 rayDirCapsule = ToCapsuleSpace(rayObj.dir, axis);
```
roggiezhang-nv: **[P2] Normalize the object-space ray before applying fixed intersection thresholds**

`rayObj.dir` can have arbitrary magnitude after projection and the inverse instance transform. `IntersectLateral` then tests `abs(a) < 1e-8`, so changing an instance's scale can incorrectly select the linear branch of a nondegenerate quadratic.

A controlled GPU reproduction uses height 2, both radii 1, origin `(0, -2, 0)`, and direction `(0, 0.00001, 0)`. The current code returns near/far distances of 75,000/75,000 instead of 100,000/300,000, placing the near hit at y=-1.25 instead of y=-1.

This is also visible in Storm: after the Metal portability fixes, scale `simpleCapsule.usda`'s world transform, camera translation, and camera clipping distances together by 1,000,000. The intended view is unchanged, but the sidewall becomes badly distorted. Normalizing the ray direction here fixes that reproduction:

```suggestion
    const vec3 rayDirCapsule = normalize(ToCapsuleSpace(rayObj.dir, axis));
```

With this change, the scaled render differs from its equivalent unscaled render by at most one 8-bit RGB value. A regression scene can reuse the existing `simpleCapsule` baseline. This addresses the reproduced scale dependence; further tests for degenerate geometry and near-tangent rays would still be useful.

roggiezhang-nv (commented):
Tested this PR on an Apple M4 Pro with Storm/HgiMetal, using the [matching macOS CI artifact](https://github.com/PixarAnimationStudios/OpenUSD/actions/runs/35132188879) and verifying that its capsule shader matched this PR head. Two reproducible correctness findings are inline: Metal shader compilation and scale-dependent intersections.

After local shader-only fixes, all six image-test scenes added by this PR rendered successfully. Manual RGB comparisons against the supplied baselines were within the configured error thresholds. The added scale reproduction also passed after ray normalization. This was manual renderer execution and RGB comparison, not a full CTest/idiff run; OpenGL/Vulkan execution and performance were not tested.

The use of `HdStImplicitSurface` fits the recent sphere refactor. As follow-ups, it would help to run these scenes in Metal GPU CI (the current `GPUTests` job was skipped) and share the duplicated lateral-intersection/hit-accumulation helpers between the cylinder and capsule implementations.

pxr/imaging/hdSt/shaders/capsule.glslfx
miguelh-nvidia: Fixed, thanks!

pxr/imaging/hdSt/shaders/capsule.glslfx
miguelh-nvidia: Fixed.

PR #4225, opened 2026-09-19
build_usd.py: build OpenColorIO whenever OpenImageIO is requested

### LINKED ISSUE #4219: `build_usd.py` does not express `OpenImageIO` requirement of `OpenColorIO` in development branch of 26.11
https://github.com/PixarAnimationStudios/OpenUSD/commit/ca10fc609f013513134769695289641d127a368e updated the default dependencies OpenUSD supports.

OpenImageIO 3.x requires OpenColorIO but `build_usd.py` does not express it in its upstream dependencies.

https://github.com/PixarAnimationStudios/OpenUSD/blob/a3d77ff5e1d405b0a5080e7c9f4c6492898fb5c5/build_scripts/build_usd.py#L2621

`build_usd.py` will now fail if `OpenImageIO` is required such as when imaging tests are built.

hikmetba-bit:
## Summary

Fixes #4219. `requiredDependencies` in `build_scripts/build_usd.py`'s `buildImaging` block only adds `OPENCOLORIO` when `--opencolorio` is passed explicitly:

```python
if context.buildOCIO:
    requiredDependencies += [ZLIB, OPENCOLORIO]

if context.buildOIIO:
    requiredDependencies += [ZLIB, JPEG, TIFF, PNG, OPENEXR, OPENIMAGEIO]
```

But the default `OPENIMAGEIO` dependency was bumped to v3.1.8.0 (https://github.com/PixarAnimationStudios/OpenUSD/commit/ca10fc609f013513134769695289641d127a368e), and OpenImageIO 3.x links against OpenColorIO regardless of whether USD's own `PXR_BUILD_OPENCOLORIO_PLUGIN` is enabled. So `--openimageio` alone (without `--opencolorio`) now builds OpenImageIO without ever building its OpenColorIO dependency first, and the OpenImageIO build fails — exactly the repro in the issue.

The comment directly above this code already states the intended behavior ("When OCIO is required, we need to make sure it's built before OIIO, since OIIO is dependent on OCIO") — the condition just didn't cover the `buildOIIO`-without-`buildOCIO` case. Fix: `if context.buildOCIO or context.buildOIIO:`.

## Verification

`build_usd.py` exits early ("C++ compiler not found") before reaching the dependency-summary step in this sandboxed environment (no C++ toolchain available to me), so I could not run the full script end to end. To still get real, executed verification of the actual logic change:

- Added `build_scripts/test_oiio_ocio_dependency.py`, which re-implements the exact touched predicate in isolation (it can't import `build_usd.py` directly — that module runs argparse and the compiler check at import time) and asserts `OPENCOLORIO` is present, and ordered before `OPENIMAGEIO`, whenever `OpenImageIO` is requested — with or without an explicit `--opencolorio`.
- Ran this test against the pre-fix predicate (`if context.buildOCIO:`) as a mutation check: 2 of 4 assertions fail/error, confirming the test actually catches this regression.
- Ran it against the fixed predicate: all 4 pass.
- `python3 -m py_compile build_scripts/build_usd.py` confirms the edited file still parses.

I'd appreciate a maintainer or CI double-checking this against an actual compiler-enabled build, since I could not run `build_usd.py` itself here.

## Test plan
- [ ] CI / maintainer: `python3 build_scripts/build_usd.py --dry_run --imaging --openimageio --no-opencolorio <fresh_dir>` now lists OpenColorIO among the dependencies to build (it did not before this fix)
- [x] `python3 -m unittest build_scripts/test_oiio_ocio_dependency.py` passes


jesschimein:
Filed as internal issue #USD-12655

sunyab:
Thank you for this contribution, but we've already merged a fix for this internally and will push that out shortly.

PR #4226, opened 2026-09-19
usdview: fix embedded interpreter for Windows PYTHONSTARTUP paths and syntax errors

### LINKED ISSUE #4209: usdview Python interpreter fails on Windows with Python 3.14
## Description

Opening the embedded Python interpreter in `usdview` on Windows with Python 3.14 fails.

There appear to be two separate issues in `pxr/Usdviewq/pythonInterpreter.py`:

1. `PYTHONSTARTUP` paths are inserted directly into generated Python source code without escaping Windows backslashes.
2. The custom `Interpreter.showsyntaxerror()` override is incompatible with Python 3.14, which now passes a `source` keyword argument.

## Environment

* OS: Windows
* Python: 3.14.7
* OpenUSD: built from source
* usdview: PySide6
* OpenUSD install path:

  ```text
  D:\Bibliotheken\OpenUSD\install_py314
  ```

## Steps to reproduce

1. Build OpenUSD with Python 3.14 on Windows.
2. Start `usdview`.
3. Open the embedded Python interpreter.
4. Have `PYTHONSTARTUP` set to a Windows path, for example:

   ```text
   C:\Users\Jonas\AppData\Roaming\Code\User\workspaceStorage\...\ms-python.python\pythonrc.py
   ```

In my case `PYTHONSTARTUP` is set by the VS Code Python extension.

## Actual result

The interpreter fails while executing the startup file:

```text
  File "C:\Users\Jonas\AppData\Local\Python\pythoncore-3.14-64\Lib\code.py", line 64, in runsource
    code = self.compile(source, filename, symbol)
  File "C:\Users\Jonas\AppData\Local\Python\pythoncore-3.14-64\Lib\codeop.py", line 154, in __call__
    return _maybe_compile(self.compiler, source, filename, symbol, flags=self.compiler.flags)
  ...
  File "<input>", line 1
    g = dict(globals());g["__file__"] = "C:\Users\Jonas\AppData\Roaming\Code\..."
                                        ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

While handling that syntax error, a second exception occurs:

```text
  File "...pxr\Usdviewq\appController.py", line 2706, in _showInterpreter
    self._console = Myconsole(self._interpreter, self._usdviewApi)
  ...
  File "...pxr\Usdviewq\pythonInterpreter.py", line 298, in ExecStartupFile
    self.interpreter.runsource(...)
  File "...\Python\pythoncore-3.14-64\Lib\code.py", line 67, in runsource
    self.showsyntaxerror(filename, source=source)

TypeError: Interpreter.showsyntaxerror() got an unexpected keyword argument 'source'
```

## Expected result

The usdview Python interpreter should open normally and execute the configured `PYTHONSTARTUP` file.

## Issue 1: Windows path escaping

`ExecStartupFile()` currently generates Python source containing the startup path:

```python
'g["__file__"] = "{0}";'
'f = open("{0}", "rb");'
'exec(compile(f.read(), "{0}", "exec"), g);'
```

For a Windows path such as:

```text
C:\Users\Jonas\...
```

this produces a Python string containing `\U`, which is interpreted as the start of a Unicode escape sequence.

Using the `repr()` formatting conversion should avoid this:

```python
self.interpreter.runsource(
    'g = dict(globals());'
    'g["__file__"] = {0!r};'
    'f = open({0!r}, "rb");'
    'exec(compile(f.read(), {0!r}, "exec"), g);'
    'f.close();'
    'del g["__file__"];'
    'globals().update(g);'.format(path))
```

## Issue 2: Python 3.14 `showsyntaxerror()` API change

Python 3.14's `InteractiveInterpreter.runsource()` calls:

```python
self.showsyntaxerror(filename, source=source)
```

The usdview interpreter overrides `showsyntaxerror()` without accepting this argument.

For example, changing:

```python
def showsyntaxerror(self, filename=None):
    self._outputBrush = QtGui.QBrush(QtGui.QColor('#ffcc63'))
    try:
        InteractiveInterpreter.showsyntaxerror(self, filename)
    finally:
        self._outputBrush = None
```

to something like:

```python
def showsyntaxerror(self, filename=None, **kwargs):
    self._outputBrush = QtGui.QBrush(QtGui.QColor('#ffcc63'))
    try:
        InteractiveInterpreter.showsyntaxerror(self, filename, **kwargs)
    finally:
        self._outputBrush = None
```

appears to make the override compatible with Python 3.14 while remaining backwards compatible.

## Workaround

Clearing `PYTHONSTARTUP` before launching usdview avoids the first problem:

```bat
set "PYTHONSTARTUP="
python "%USD_ROOT%\bin\usdview" %*
```

However, the `showsyntaxerror()` incompatibility can still occur whenever a syntax error is processed by the interpreter.

## Additional notes

The two problems seem independent:

* The Windows path issue is exposed when `PYTHONSTARTUP` contains a normal Windows path.
* The `showsyntaxerror()` issue is caused by an API change in Python 3.14.

I can provide a PR with the changes above if this approach is considered appropriate.

hikmetba-bit:
## Summary

Fixes #4209. Two independent bugs in `pxr/usdImaging/usdviewq/pythonInterpreter.py`, both in the embedded Python interpreter (`usdview`'s built-in console):

**1. `Controller.ExecStartupFile` breaks on Windows `$PYTHONSTARTUP` paths**

It builds the source handed to `interpreter.runsource()` by splicing the startup file's path into a plain double-quoted Python string literal:

```python
'g["__file__"] = "{0}";'
'f = open("{0}", "rb");'
'exec(compile(f.read(), "{0}", "exec"), g);'
...).format(path)
```

Since `path` is embedded raw inside `"..."`, a Windows path's backslashes are read as escape sequences by the *generated* source's own parser (e.g. `\U` starts an 8-hex-digit unicode escape) — raising a `SyntaxError` for an ordinary path like `C:\Users\Jonas\AppData\Roaming\Code\...\pythonrc.py`, exactly the traceback in the issue. Fixed by using `{0!r}` (i.e. `repr(path)`) instead of manual quoting, which produces a properly escaped, safely embeddable string literal.

**2. `Interpreter.showsyntaxerror` doesn't accept the `source` keyword `runsource()` passes**

```python
def showsyntaxerror(self, filename = None):
    ...
    InteractiveInterpreter.showsyntaxerror(self, filename)
```

`code.InteractiveInterpreter.runsource` calls `self.showsyntaxerror(filename, source=source)` on a `SyntaxError`. Since the override only accepts `filename`, this raises `TypeError` instead of displaying the error — this isn't actually 3.14-specific, the currently-installed Python already does this (see Verification). Fixed by adding `**kwargs` and forwarding them to the base implementation.

## Verification

`pythonInterpreter.py` imports `pxr.Tf` and `PySide6` at module scope, so it can't be imported directly in this sandboxed environment (no built USD, no PySide6). To still get real, *executed* verification rather than just static reasoning:

- Added `pxr/usdImaging/usdviewq/testenv/testUsdviewqPythonInterpreterCompat.py`, which re-implements each fixed snippet in isolation using only the stdlib (`code.InteractiveInterpreter` + `compile()`) and actually runs it.
- `test_windows_path_compiles_after_fix` / `test_windows_path_with_backslash_u_breaks_naive_quoting`: confirms the pre-fix `"{0}"` quoting genuinely raises `SyntaxError` on a Windows-style path, and the fixed `{0!r}` quoting compiles cleanly.
- `test_fixed_interpreter_handles_syntax_error` / `test_prefix_interpreter_raises_typeerror`: confirms the *original* `showsyntaxerror(self, filename=None)` override raises `TypeError` against the presently-installed Python (3.13.12) when `runsource()` hits a syntax error, and the fixed `**kwargs`-forwarding override doesn't. (This also confirms the bug isn't 3.14-exclusive — `code.InteractiveInterpreter.runsource` already passes `source=` on 3.13.)
- `python3 -m py_compile pythonInterpreter.py` confirms the edited production file still parses.

This test script is standalone — it is **not** wired into `CMakeLists.txt`/`ctest`; there was no build environment available here to add and verify that wiring. A maintainer may want to register it as a proper usdviewq test if useful.

## Test plan
- [x] `python3 -m unittest testUsdviewqPythonInterpreterCompat` — 4/4 pass (standalone, stdlib-only)
- [ ] Maintainer/CI: build usdview and confirm the embedded interpreter's `$PYTHONSTARTUP` handling and syntax-error display now work on Windows


jesschimein:
Filed as internal issue #USD-12656

PR #4227, opened 2026-09-19
build_usd.py: fail fast when --tools is requested without usdValidation

### LINKED ISSUE #4100: `build_usd.py` fails with `--tools --no-usdValidation`
`usdzip` depends on `usdValidation`.  This means when `tools` are enabled, `usdValidation` must be enabled too.
```
python3 build_scripts/build_usd.py <OUTPUT_DIR> --no-usdview --no-examples --no-materialx --no-imaging --tests --no-python --no-usdValidation --no-tutorials -v
...
pxr/bin/usdzip/usdzip.cpp:20:10: fatal error: pxr/usdValidation/usdValidation/context.h: No such file or directory
   20 | #include "pxr/usdValidation/usdValidation/context.h"
...
```

Should `--no-usdValidation` should be deprecated or phased out if core tools depend on validation?

hikmetba-bit:
## Summary

Fixes #4100. `usdzip` (built as part of `--tools`) includes `pxr/usdValidation/usdValidation/context.h`, so requesting `--tools --no-usdValidation` currently fails deep into the C++ build with a confusing error:

```
python3 build_scripts/build_usd.py <dir> --tools --no-usdValidation ...
...
pxr/bin/usdzip/usdzip.cpp:20:10: fatal error: pxr/usdValidation/usdValidation/context.h: No such file or directory
```

This adds an early, explicit check right next to the existing "`--usdview` requires `usdImaging`/`python`" guard a few lines above (same file, same established idiom already used throughout `build_usd.py` for erroring out on an incompatible flag combination *before* doing any work), so the user gets an immediate, clear Python-level error instead of a confusing C++ compile failure potentially much later in a long build.

This only addresses the reported symptom (the confusing failure mode). The maintainer's own comment on the issue raises a separate, larger design question — whether `--no-usdValidation` should be deprecated/phased out entirely now that core tools depend on it — which I've deliberately left alone as a project decision, not something for this PR to presume.

## Verification

This is the one of my three OpenUSD PRs today I could verify **end-to-end**, since this particular code path runs and exits before the script's later "C++ compiler not found" check (which blocks fuller execution in this sandboxed environment — no C++ toolchain available to me):

- Ran `python3 build_scripts/build_usd.py --dry_run` with the **exact repro command from the issue** (`--tools --no-usdValidation ...`) — confirmed it now exits immediately with `ERROR: Cannot build tools when usdValidation is disabled.` instead of proceeding toward the C++ build.
- Ran `--dry_run` with `--tools --usdValidation` (explicit) and with `--tools` + usdValidation left at its default (on) — both proceed past this check to the next step, as expected (`ERROR: C++ compiler not found`, the environment's own separate, pre-existing limitation — confirms no false positive).
- Ran `--dry_run` with `--no-tools --no-usdValidation` — also proceeds past this check unaffected.
- `python3 -m py_compile build_scripts/build_usd.py` confirms the file still parses.

## Test plan
- [x] `python3 build_scripts/build_usd.py --dry_run --tools --no-usdValidation <dir>` now fails fast with a clear message (verified above)
- [x] Compatible flag combinations (`--tools --usdValidation`, `--tools` with default usdValidation, `--no-tools --no-usdValidation`) unaffected (verified above)
- [ ] Maintainer/CI: confirm on a compiler-enabled machine that a full build with a compatible flag combination still succeeds


jesschimein:
Filed as internal issue #USD-12657

PR #4230, opened 2026-09-21
pypi: Preserve Windows wheel DLL search paths

### LINKED ISSUE #4198: usd-core overwrites PXR_USD_WINDOWS_DLL_PATH on Windows
### Description of Issue

We ran into a Windows failure importing `pxr.Tf` from the `usd-core 26.8` PyPI wheel:

```text
ImportError: DLL load failed while importing _tf:
A dynamic link library (DLL) initialization routine failed.
```

Python's fault handler also reported an access violation while `_tf` was loading.

The problem appears to be in the Windows-specific code added by the [`usd-core` packaging script](https://github.com/PixarAnimationStudios/OpenUSD/blob/6792941c5d5b42778e6875c5ed5a3279e6936981/build_scripts/pypi/package_files/setup.py):

```python
dllPath = os.path.split(os.path.realpath(__file__))[0]
if sys.version_info >= (3, 8, 0):
    os.environ["PXR_USD_WINDOWS_DLL_PATH"] = dllPath
```

This overwrites any value set by the caller before importing `pxr`.

The [`Tf.WindowsImportWrapper`](https://github.com/PixarAnimationStudios/OpenUSD/blob/6792941c5d5b42778e6875c5ed5a3279e6936981/pxr/base/tf/__init__.py) then does the following:

```python
import_paths = os.getenv("PXR_USD_WINDOWS_DLL_PATH")
if import_paths is None:
    import_paths = os.getenv("PATH", "")
```

Since the wheel always sets `PXR_USD_WINDOWS_DLL_PATH`, the wrapper never registers the directories from `PATH`. It registers only the installed `pxr` directory.

This also conflicts with the [Windows instructions in `BUILDING.md`](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/BUILDING.md), which say that users can override the search paths with `PXR_USD_WINDOWS_DLL_PATH`.

On the machine where we reproduced the failure, the relevant versions were:

```text
usd_ms.dll linker version              14.44
C:\Windows\System32\MSVCP140.dll       14.00.24215.1
<environment>\bin\MSVCP140.dll         14.44.35211.0
```

The Python environment had a compatible runtime in its `bin` directory, but that directory was no longer in the search path registered by the `Tf` import wrapper. Windows selected the old System32 copy instead.

Microsoft requires the deployed Visual C++ Redistributable to be at least as new as the newest toolset used to build any application component. See [C++ binary compatibility between Visual Studio versions](https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170).

We ran the following cases twice in fresh processes:

| Case | Result |
|---|---|
| Import `pxr.Tf` directly | Failed 2/2 |
| Preload only `<environment>\bin\MSVCP140.dll` | Passed 2/2 |
| Register `<environment>\bin` with `os.add_dll_directory()` | Passed 2/2 |
| Import `pxr`, then add `<environment>\bin` to `PXR_USD_WINDOWS_DLL_PATH` | Passed 2/2 |

The direct case used a minimal environment with `usd-core` and Blosc only. Warp, Torch, CUDA, and NumPy were not loaded.

Importing Torch before `pxr.Tf` also made the import succeed because Torch caused a compatible `MSVCP140.dll` to be loaded first. That made the failure look intermittent and dependent on application state.

The same import succeeds on a second Windows machine where the System32 copy of `MSVCP140.dll` is version 14.51.36247.0. The wheel still overwrites `PXR_USD_WINDOWS_DLL_PATH` there, but the newer system runtime masks the problem.

The experiment and logs are available in [NVIDIA/warp#1838](https://github.com/NVIDIA/warp/pull/1838#issuecomment-5402489432).

### Steps to Reproduce

```powershell
py -3.12 -m venv .venv
.\.venv\Scripts\python.exe -m pip install usd-core==26.8 blosc==1.11.4

$env:PXR_USD_WINDOWS_DLL_PATH = "$PWD\.venv\bin"

.\.venv\Scripts\python.exe -c "import os, pxr; print(os.environ['PXR_USD_WINDOWS_DLL_PATH']); from pxr import Tf"
```

`blosc 1.11.4` is installed here only because it supplies a known compatible MSVC 14.44 runtime under the environment's `bin` directory.

After `import pxr`, the command prints:

```text
...\.venv\Lib\site-packages\pxr
```

The value supplied by the caller has been discarded.

On a machine with an old system Redistributable, the following `Tf` import fails. A machine with a current system Redistributable may import `Tf` successfully, but it will still reproduce the environment-variable overwrite.

### Expected Behavior

Importing `pxr` should preserve the paths supplied through `PXR_USD_WINDOWS_DLL_PATH`.

The wheel can add its own directory first without discarding the existing value:

```python
existing = os.environ.get("PXR_USD_WINDOWS_DLL_PATH", "")
paths = [dllPath, existing]
os.environ["PXR_USD_WINDOWS_DLL_PATH"] = os.pathsep.join(
    path for path in paths if path
)
```

It may also make sense for the wheel to register common environment runtime locations such as `sys.prefix\bin` and `sys.prefix\Library\bin`, or to bundle the runtime needed by the wheel.

Whichever packaging approach is chosen, importing `pxr.Tf` should not depend on another Python package loading a compatible `MSVCP140.dll` first.

```text
Windows Server 2025 Standard
OS build 26100
NVIDIA RTX PRO 6000
Python 3.12.13
```

### Package Versions

```text
usd-core 26.8
blosc 1.11.4
```

### Build Flags

Official `usd-core 26.8` Windows PyPI wheel.

andrewkaufman:
### Description of Change(s)

The Windows `usd-core` wheel overwrites `PXR_USD_WINDOWS_DLL_PATH` when importing `pxr`, discarding caller-supplied directories and preventing `Tf.WindowsImportWrapper` from using its normal `PATH` fallback.

Prepend the wheel's DLL directory while preserving an explicit caller override. When the override is absent, preserve the original `PATH` fallback; an explicitly empty override continues to exclude `PATH`. Normalize comparisons to remove existing spellings of the wheel directory before prepending it once to both variables. Preserve unrelated entries and their order, tolerate missing or empty `PATH`, and keep repeated initialization stable.

Add installed-wheel regression tests for caller overrides, actual Windows DLL-directory registration, unset versus empty variables, normalized duplicates, and repeated initialization. These run in the existing PyPI packaging test suite.

Local validation on Windows x64 with Python 3.12.14 and MSVC 19.43:

- Built and installed monolithic USD core, built the Windows wheel, and installed it into a separate virtual environment.
- `python -B -m unittest -v build_scripts.pypi.test`: all five tests pass, including 12 path configurations.
- Verified the new regressions fail with the original wheel initialization.
- Verified actual external dependency loading through both the override and `PATH` fallback, followed by `Tf` import and USD stage creation. The local fixture used a unique TBB DLL name to prevent an installed system copy from masking the result; the built wheel was left unchanged. An empty override correctly excludes the `PATH`-only dependency.

N/A — Windows wheel packaging bug fix.

### Fixes Issue(s)

Fixes #4198.

### AI Assistance

Developed with AI assistance (OpenAI Codex): the code, tests, and this description were AI-drafted and validated locally before submission.

### Checklist

- [x] I have verified that all unit tests pass with the proposed changes (the affected wheel installation suite)


build_scripts/pypi/package_files/setup.py
```diff
@@ -88,8 +88,19 @@ def windows():
...
 dllPath = os.path.split(os.path.realpath(__file__))[0]
+
+def _PrependDllPath(paths):
+    normalizedPath = os.path.normcase(os.path.normpath(dllPath))
+    return os.pathsep.join([dllPath] + [
+        path for path in paths.split(os.pathsep)
+        if path and os.path.normcase(os.path.normpath(path)) != normalizedPath
+    ])
+
 if sys.version_info >= (3, 8, 0):
```
nvmkuruc: The minimum version for OpenUSD is now `(3, 11)` so the version guard is likely not needed.

jesschimein:
Filed as internal issue #USD-12676

PR #4232, opened 2026-09-21
usdPhysics: park the disabled-body collider-ownership behavior pending resolution

andrewkaufman:
### Description of Change(s)

997b46e (the squash of #4187) treats a disabled rigid body as unapplied in every ancestor traversal, so a collider owned by a disabled body either propagates to the nearest enabled ancestor or, when none exists, is attributed to the disabled body itself. Whether that is the right ownership model is still under discussion (see the disabled-body / static-collision thread on #4187 and #4165). This reverts only that behavior so the ownership question can be settled before it is baked in, while keeping the bug fixes from that work that are independent of it.

Reverted (the contested collider / joint-body ownership only):
- `parseUtils.cpp` `_GetBodyPrim`: no longer skips a disabled body when resolving a joint body relationship.
- `parseUtils.cpp` `_HasDynamicBodyParent`: no longer walks past a disabled body or reports a nearest-disabled-body fallback.
- `validators.cpp` `HasDynamicBodyParent`: same revert.

Every ancestor walk again stops at the first rigid body regardless of `physics:rigidBodyEnabled`, matching pre-#4187 attribution. The three parsing tests and the one validation test that assert the disabled-body skip are dropped.

Kept (bug fixes independent of the ownership question):
- `rigidBodyAPI.cpp`: exclude `physics:collisionEnabled = false` colliders from mass, and do not prune a disabled nested body's subtree so its enabled colliders' mass is not lost.
- `validators.cpp`: `isAPISchemaEnabled` defaults to true, matching the `rigidBodyEnabled` fallback.
- `validators.cpp` `HasEnabledRigidBody`: searches ancestors so a joint anchored to a collider below a body is not flagged bodyless.

This introduces a deliberate discrepancy while parked. On dev every path is consistent. With this revert the mass computer still walks past a disabled nested body (that fix is kept), while collider attribution, joint-body resolution and validation stop at it, so a collider under a disabled nested body contributes its mass to the enabled outer body but is attributed to the disabled inner body. This is the cost of parking the ownership question and is expected to be resolved in one direction or the other once the discussion settles.

Opening as a draft: this depends on the direction of that discussion and should not merge until it is resolved.

N/A, reverts a behavior change; no schema or API change.

### Fixes Issue(s)

N/A

### AI Assistance

Developed with AI assistance (Anthropic Claude, primarily Opus 5): the code, tests, and this description were AI-drafted and then reviewed by me before submission.

### Checklist

- [x] I have followed the coding conventions


jesschimein:
Filed as internal issue #USD-12678

andrewkaufman:
From a meeting between several AOUSD Physics WG representatives, including myself, Adam, @eberled & @tallytalwar, the decision was to move forward with #4187 in its entirety, including the final aa29b9a0127597ebb1b5765e39da0e1f054496e9 commit (which is not on `dev` as of yet).

This PR is no longer needed, as it was only an attempt to walk back some of #4187 incase there was no consensus reached today.

